Issue #6, opened 2017-12-05
Gatekeeper app


cameel:
The purpose of this task is to create a small Django app that will analyze file upload/download requests coming to `nginx-storage` and decide whether they can be allowed. It does not have to have the actual authorization logic yet (this will be another task). At this point the main requirement is to have the app and its views set up correctly.

- Create a new app inside `concent_api`
    - NOTE: This is meant to be a new app in the existing Django project, not an entirely new project. It should be placed in `concent_api/gatekeeer/`.
- Add a new setting called `CONCENT_FEATURES` that controls which features are enabled. It should be a list with the following possible values: `"concent-api"`, `"gatekeeper"`, `"admin-panel"`.
    - If `"concent-api"` is not on the list, the views from the `core` app should not be included in `concent_api/concent_api/urls.py`.
    - If `"gatekeeper"` is not on the list, the views from the `gatekeeper` app should not be included in `concent_api/concent_api/urls.py`.
    - If `"admin-panel"` is not on the list, the admin panel should not be enabled in `concent_api/concent_api/urls.py`.
- Add [system checks](https://docs.djangoproject.com/en/1.11/topics/checks/) to validate the new setting
    - The setting must always be defined
    - It must always be a list
    - Only the values listed above are allowed
    - There must be no duplicates on the list
    - The corresponding app (`core` for `concent-api`, `gatekeeper` for `gatekeeper`, `django.contrib.admin` for `admin`) must be present in `INSTALLED_APPS`.
- Create `GET gatekeeper/download-auth/` and `POST gatekeeper/upload-auth/` views in the `gatekeeper` app that expect a base64-encoded golem message in the `Authorization` HTTP header on the request.
    - An example header could look like this (in curl notation): `-H "Authorization: Golem zdXegB12152DDSDBB..."`
    - The view should send an empty response with `HTTP 200 OK` status if it grants access. For now let's assume that we grant authorization if we can successfully decode the message.
    - The view should respond with a JSON response with `HTTP 401 UNAUTHORIZED` if it denies access. The response should contain the `WWW-Authenticate` header with the following value: `Golem realm="Concent Storage"`.

cameel:
Update: `file-transfer-auth/` changed to `gatekeeper/file-transfer-auth/`. Having a prefix for each app will let us handle them in nginx in a more reliable way.

cameel:
[...]

PR #8, opened 2017-12-12
Gatekeeper app

Jakub89:
#6


concent_api/concent_api/system_check.py
```diff
@@ -0,0 +1,74 @@
+from django.core.checks     import Error, register
```
rwrzesien: So as we agreed in previous review, lets at least try to keep alphabetic order, and make multi line imports, like this:
```
from from django.core.checks     import (
    Error, 
    register,
)
```

When importing  this way, please add , to last object, so next person will not need to modify this line.

concent_api/concent_api/system_check.py
```diff
@@ -0,0 +1,74 @@
...
+from django.conf            import settings
+from django.core.validators import URLValidator
+from django.core.exceptions import ValidationError
+
+
+@register()
+def check_settings_concent_features(app_configs, **kwargs):  # pylint: disable=unused-argument
+    errors = []
+
+    concent_features = settings.CONCENT_FEATURES
```
rwrzesien: There is no need to create this extra variable, just use settings.CONCENT_FEATURES directly in code below.

concent_api/concent_api/system_check.py
```diff
@@ -0,0 +1,74 @@
...
+                    Error('CONCENT_FEATURES error:', hint = 'App "gatekeeper" is not defined in INSTALLED_APPS')
+                )
+                return errors
+
+        if application == "admin-panel":
+            if "django.contrib.admin" not in settings.INSTALLED_APPS:
+                errors.append(
+                    Error('CONCENT_FEATURES error:', hint = 'App "django.contrib.admin" is not defined in INSTALLED_APPS')
+                )
+                return errors
```
rwrzesien: Each time an error appears, it is appended to errors list and returned, so actually errors will always be either empty list or one element list. So, instead of appending to errors and returning it, you can return one element list directly, like this:
```
return [Error('CONCENT_FEATURES error:', hint = 'App "django.contrib.admin" is not defined in INSTALLED_APPS')]
```
, remove the errors variable at all, and at the end return empty list. But maybe there are some other plans for this code, to gather all the errors and then return them at the end of this method, then it is ok.

concent_api/concent_api/system_check.py
```diff
@@ -0,0 +1,74 @@
...
+
+@register()
+def check_settings_storage_cluster_address(app_configs, **kwargs):  # pylint: disable=unused-argument
+    errors = []
+    storage_cluster_address = URLValidator(schemes=['http', 'https'])
+    if not hasattr(settings, 'STORAGE_CLUSTER_ADDRESS'):
+        errors.append(
+            Error('STORAGE_CLUSTER_ADDRESS error:', hint = 'STORAGE_CLUSTER_ADDRESS variable must be defined in django settings.')
+        )
+        return errors
```
rwrzesien: same here

concent_api/concent_api/system_check.py
```diff
@@ -0,0 +1,74 @@
...
+
+        if application in check_for_duplicate:
+            errors.append(
+                Error('CONCENT_FEATURES error:', hint = 'Duplicate {} in CONCENT_FEATURES'.format(application), id = 'settings_check_003')
+            )
+            return errors
+        check_for_duplicate.append(application)
+
+        if application == "concent-api":
+            if "core" not in settings.INSTALLED_APPS:
```
rwrzesien: If it is the final version of those ifs, you can combine them into one, same for the next two.

concent_api/concent_api/urls.py
```diff
@@ -13,12 +13,28 @@
...
 
-urlpatterns = [
-    url(r'^api/v1/', include(core.urls, namespace = 'core')),
-    url(r'^admin/',  admin.site.urls),
-]
+urlpatterns = []
+
+assert hasattr(settings, 'CONCENT_FEATURES') is True, "CONCENT_FEATURES variable must be defined in django settings."
+
+if "concent-api" in settings.CONCENT_FEATURES:
```
rwrzesien: Maybe we should add some kind of dictionary that would work as registry and auto load all urls from it ? What I am trying to point out is that at some point we might have a lot of those ifs. But maybe that can be added as separate feature, what do you think @cameel ?

concent_api/concent_api/system_check.py
cameel: OK, but let's stick to the style I suggested in the other pull request:

``` python
from django.core.checks import Error
from django.core.checks import register
```

concent_api/concent_api/urls.py
```diff
@@ -13,12 +13,28 @@
...
 import core.urls
+import gatekeeper.urls
 
-urlpatterns = [
-    url(r'^api/v1/', include(core.urls, namespace = 'core')),
-    url(r'^admin/',  admin.site.urls),
-]
+urlpatterns = []
+
+assert hasattr(settings, 'CONCENT_FEATURES') is True, "CONCENT_FEATURES variable must be defined in django settings."
```
cameel: `is True` is redundant here. `hasattr(settings, 'CONCENT_FEATURES')` returns a boolean value so it's sufficient.

Jakub89:
system_check.py was added in previous pull request, so I don't know if I should make changes to it in the gatekepper feature/branch.

concent_api/concent_api/urls.py
cameel: Sure. Sounds good. I'm fine with it being simple `if`s for now but having a dict that maps these names to URLs, required modules and other values is a good idea as well.

We could do something like this in some simple module that contains only definitions:

``` python
AVAILABLE_CONCENT_FEATURES = {
    "admin": {
        "required_django_apps": [
            "django.contrib.admin",
        ],
        "url_patterns":         [
            url(r'^admin/', admin.site.urls),
        ],
    },
    "concent-api": {
        "required_django_apps": [
            "core",
        ],
        "url_patterns":         [
            url(r'^api/v1/', include(core.urls, namespace = 'core')),
        ],
    },
    "gatekeeper": {
        "required_django_apps": [
            "gatekeeper",
        ],
        "url_patterns":         [
            url(r'^gatekeeper/', include(gatekeeper.urls, namespace = 'gatekeeper')),
        ],
    },
}
```

cameel:
> system_check.py was added in previous pull request, so I don't know if I should make changes to it in the gatekepper feature/branch.

No, it's in a commit right here. I don't see any separate pull request just for system check.

concent_api/concent_api/system_check.py
```diff
@@ -0,0 +1,74 @@
...
+                errors.append(
+                    Error('CONCENT_FEATURES error:', hint = 'App "core" is not defined in INSTALLED_APPS')
```
cameel: Please add `id`s for all errors.

concent_api/concent_api/system_check.py
Jakub89: Done

concent_api/concent_api/system_check.py
Jakub89: Done

concent_api/concent_api/system_check.py
Jakub89: Done

concent_api/concent_api/system_check.py
Jakub89: Done

concent_api/concent_api/urls.py
```diff
@@ -13,14 +13,28 @@
...
 import gatekeeper.urls
 
-urlpatterns = [
-    url(r'^',        include(gatekeeper.urls, namespace = 'gatekeeper')),
-    url(r'^api/v1/', include(core.urls, namespace = 'core')),
-    url(r'^admin/',  admin.site.urls),
-]
+urlpatterns = []
+
+assert hasattr(settings, 'CONCENT_FEATURES'), "CONCENT_FEATURES variable must be defined in django settings."
```
cameel: This needs to be changed to a normal `if` that raises `ImproperlyConfigured` exception. `urls.py` is loaded before our Django checks run so we can't assume that missing `CONCENT_FEATURES` is bug. It can be a configuration error on user's part.

cameel:
@Jakub89 Please always include a link to the original issue in the description when you create a pull request. Otherwise there's no direct connection between them. I've added it to the description here.

maaktweluit:
@cameel @Jakub89 There is another trick in linking, when you add "Resolves #..." it will auto close the issue when the pull gets merged.

concent_api/concent_api/system_check.py
```diff
@@ -0,0 +1,46 @@
...
+
+@register()
+def check_settings_concent_features(app_configs, **kwargs):  # pylint: disable=unused-argument
+
+    if not isinstance(settings.CONCENT_FEATURES, list):
+        return [Error('CONCENT_FEATURES error:', hint = 'CONCENT_FEATURES should be a list.', id = 'settings_check_101')]
+
+    check_for_duplicate = []
+    for application in settings.CONCENT_FEATURES:
+        if application not in ["concent-api", "gatekeeper", "admin-panel"]:
```
rwrzesien: How about adding a single line comment above each if which describes what is currently checked, that can be copy-pasted from documentation, like
```
# The setting must always be defined
```

concent_api/gatekeeper/views.py
```diff
@@ -0,0 +1,41 @@
...
+        decoded_header_content = (b64decode(header_content))
+    except ValueError:
+        return JsonResponse({"WWW-Authenticate": "Golem realm='Concent Storage'"}, status = 401)
+
+    try:
+        check_golem_msg = load(decoded_header_content, settings.CONCENT_PRIVATE_KEY, settings.CONCENT_PUBLIC_KEY)
+    except AttributeError:
+        return JsonResponse({"WWW-Authenticate": "Golem realm='Concent Storage'"}, status = 401)
+
+    if isinstance(check_golem_msg, Message):
```
rwrzesien: How about switching it to 
```
if not isinstance(check_golem_msg, Message):
    return JsonResponse({"WWW-Authenticate": "Golem realm='Concent Storage'"}, status = 401)

return HttpResponse("", status=200)
```
which fits into code pattern above.

rwrzesien:
@Jakub89 @cameel I have reviewed this against documentation https://github.com/golemfactory/concent/issues/6 and for me it looks like everything should work as described :+1:

concent_api/gatekeeper/views.py
cameel: I agree.

And actually I'd make this check more explicit. The result here can be only `None` or `Message`. `None` indicates a failure to decode the message. Anything else would be a bug. This assumption should be documented with `assert`:

``` python
if decoded_golem_message is None:
    return JsonResponse({"WWW-Authenticate": "Golem realm='Concent Storage'"}, status = 401)

assert isinstance(decoded_golem_message, Message)
return HttpResponse("", status = 200)
```

BTW, @Jakub89 please avoid naming variables like this. `check_golem_msg` sounds like a function name.

concent_api/concent_api/urls.py
```diff
@@ -13,14 +13,17 @@
...
-    url(r'^',        include(gatekeeper.urls, namespace = 'gatekeeper')),
-    url(r'^api/v1/', include(core.urls, namespace = 'core')),
-    url(r'^admin/',  admin.site.urls),
-]
+urlpatterns = []
+
+if not hasattr(settings, 'CONCENT_FEATURES'):
+    raise ImproperlyConfigured("CONCENT_FEATURES variable must be defined in django settings.")
+
+assert set(settings.CONCENT_FEATURES) - set(AVAILABLE_CONCENT_FEATURES.keys()) == set()
```
rwrzesien: A comment above this line would be nice.

Jakub89: I've added quick fixup.

rwrzesien: I have meant the line with assert ;)

Jakub89: Okey, after talk to @cameel  we decide to switch asset with if statement, and there was also a bug in for loop to fix ;)

concent_api/concent_api/urls.py
```diff
@@ -1,29 +1,16 @@
...
-    1. Add an import:  from my_app import views
-    2. Add a URL to urlpatterns:  url(r'^$', views.home, name='home')
-Class-based views
-    1. Add an import:  from other_app.views import Home
-    2. Add a URL to urlpatterns:  url(r'^$', Home.as_view(), name='home')
-Including another URLconf
-    1. Import the include() function: from django.conf.urls import url, include
-    2. Add a URL to urlpatterns:  url(r'^blog/', include('blog.urls'))
-"""
 from django.conf            import settings
```
cameel: Why did you remove this? In a fixup?

concent_api/concent_api/urls.py
```diff
@@ -1,29 +1,16 @@
...
 
 urlpatterns = []
 
+# Check for CONCENT_FEATURES variable existence
 if not hasattr(settings, 'CONCENT_FEATURES'):
     raise ImproperlyConfigured("CONCENT_FEATURES variable must be defined in django settings.")
 
 assert set(settings.CONCENT_FEATURES) - set(AVAILABLE_CONCENT_FEATURES.keys()) == set()
 
+# Run project only with apps specified in CONCENT_FEATURES variable
```
cameel: Those are not apps. Please don't call them that to avoid confusion with `INSTALLED_APPS`.

concent_api/gatekeeper/views.py
```diff
@@ -1,11 +1,41 @@
...
 
+from golem_messages.shortcuts       import load
+from golem_messages.message         import Message, init_messages
+
 
 @csrf_exempt
 @require_POST
 def file_transfer_auth(request):
-    return HttpResponse("", status=200)
+    init_messages()
```
cameel: Why do you call `init_messages` here? `golem-messages` already calls this at module level. This should not be necessary.

concent_api/gatekeeper/views.py
cameel: @Jakub89 Still waiting for this change.

cameel:
One more thing. Please also include commits that implement the changes to the original issue: https://github.com/golemfactory/concent/issues/6#issuecomment-351388736

cameel:
Another one: pylint reports a warning:

```
./concent_api/concent_api/__init__.py:1:1: F401 '.system_check' imported but unused
```

concent_api/concent_api/system_check.py
```diff
@@ -0,0 +1,29 @@
...
+
+@register()
+def check_settings_concent_features(app_configs, **kwargs):  # pylint: disable=unused-argument
+
+    if not isinstance(settings.CONCENT_FEATURES, list):
+        return [Error('The value of CONCENT_FEATURES setting is not a list', id = 'concent.E001')]
+
+    uknown_features = set(settings.CONCENT_FEATURES) - {"concent-api", "gatekeeper", "admin-panel"}
+    if len(uknown_features) > 0:
+        return [Error('Unknown features specified CONCENT_FEATURES setting',               hint = 'Did you make a typo in the name?', id = 'concent.E002')]
```
rwrzesien: Something went wrong with code formatting, there is a lot of extra spaces in this line and following ifs.

concent_api/concent_api/system_check.py
```diff
@@ -0,0 +1,29 @@
...
+from django.conf            import settings
+
+
+@register()
+def check_settings_concent_features(app_configs, **kwargs):  # pylint: disable=unused-argument
+
+    if not isinstance(settings.CONCENT_FEATURES, list):
+        return [Error('The value of CONCENT_FEATURES setting is not a list', id = 'concent.E001')]
+
+    uknown_features = set(settings.CONCENT_FEATURES) - {"concent-api", "gatekeeper", "admin-panel"}
```
rwrzesien: `{"concent-api", "gatekeeper", "admin-panel"}` This can be created from AVAILABLE_CONCENT_FEATURES.keys(), and also used directly in if below (but a comment that you check for unknown features above if would be nice).

concent_api/concent_api/urls.py
```diff
@@ -1,24 +1,16 @@
...
-Including another URLconf
-    1. Import the include() function: from django.conf.urls import url, include
-    2. Add a URL to urlpatterns:  url(r'^blog/', include('blog.urls'))
-"""
-from django.conf.urls import url, include
-from django.contrib   import admin
+from concent_api.constants  import AVAILABLE_CONCENT_FEATURES
 
-import core.urls
+urlpatterns = []
```
rwrzesien: It would be more readable if you move this to line 14, just before the first usage of this variable.

concent_api/gatekeeper/urls.py
```diff
@@ -0,0 +1,8 @@
+from django.conf.urls import url
+
+from .views import upload, download
```
rwrzesien: Please split to two lines.
```
from .views import download
from .views import upload
```

concent_api/gatekeeper/views.py
```diff
@@ -0,0 +1,72 @@
+from base64                         import b64decode
+
+from django.conf                    import settings
+from django.http                    import JsonResponse, HttpResponse
+from django.views.decorators.http   import require_POST, require_GET
+from django.views.decorators.csrf   import csrf_exempt
```
rwrzesien: Please split into multiple lines and also try to keep alphabetical.

concent_api/concent_api/system_check.py
Jakub89: Spaces where added by @cameel  for better readability.

concent_api/concent_api/system_check.py
Jakub89: Can't do that if I try:
from concent_api.constants  import AVAILABLE_CONCENT_FEATURES
I get:
django.core.exceptions.AppRegistryNotReady: Apps aren't loaded yet.

concent_api/gatekeeper/views.py
```diff
@@ -0,0 +1,72 @@
...
+    except AttributeError:
+        return json_response('Authorization header is not a golem message.')
+
+    if loaded_golem_msg is None:
+        return json_response('Undefined golem message.')
+    assert isinstance(loaded_golem_msg, Message)
+
+
+
+def json_response(message):
```
rwrzesien: Please move this function to utils or something similar.

concent_api/concent_api/system_check.py
rwrzesien: Splitting to multiple lines should do the job but as you guys want.

concent_api/concent_api/urls.py
Jakub89: done

concent_api/gatekeeper/urls.py
Jakub89: done

concent_api/gatekeeper/views.py
Jakub89: done

concent_api/gatekeeper/views.py
Jakub89: I've made gatekeeper_view.py in utils and put it there

concent_api/concent_api/system_check.py
cameel: That's because we're using `url()` and `include()` in `constants.py`. This could be avoided by using plain strings, though `AVAILABLE_CONCENT_FEATURES` would become more complex and less flexible...

rwrzesien: @cameel Fix for this is what I have described yesterday on slack, we discussed it with Jakub and as agreed with you lets turn concent_api into django app and load system_check in AppConfig ready() method.

concent_api/gatekeeper/views.py
cameel: I don't think it's generic enough to warrant moving it to `utils`. It should have much more specific name. `json_response()` does not give any hint that it's hard-coded to return HTTP 401 and add `WWW-Authenticate` header. By looking at the call you can't even tell that it will return an error. It should be something like `gatekeeper_access_denied_response()`. And in my opinion this is something very specific to gatekeeper, not a generic utility that could be used by other apps.

concent_api/concent_api/system_check.py
cameel: I know. I just wanted to explain why this happens because making it not use these functions is a more straightforward solution in my opinion. But making `concent_api` an app works too so let's continue with that.

concent_api/gatekeeper/views.py
rwrzesien: I agree but I still don't think it should be in views.py. How about putting it into gatekeeper/utils.py or gatekeeper/helpers.py ?

cameel: Personally, I'd keep it in views (maybe as a nested function in the view?) because it's been made into a function mostly to save repetition. It's not really reusable outside of that single view.

But well, as long as it's inside the gatekeeper app it does not matter that much so if you stronly prefer to put it in some kind of `utils` file, I'm fine with that.

rwrzesien: From my experience when there is permission to put objects that are not strictly elements that should be in that files (like views in views, models in models etc.) sooner or later there is more and more of those in that files and it is getting more spaghetti. I suggest putting it into separate file in that app, or, if it is reusable, use ClassBasedViews with Base class with that function as a method, and inherit it in other views.

Jakub89: Done, switched to `gatekeeper_access_denied_response()` in  gatekeeper/utils.py

concent_api/concent_api/system_check.py
Jakub89: Done, concent_api is now django app included in INSTALLED_APPS, witch allow us to use
`from concent_api.constants  import AVAILABLE_CONCENT_FEATURES`.

concent_api/gatekeeper/views.py
```diff
@@ -0,0 +1,68 @@
...
+def check_headers(request):
+    # Decode and check if request header contains golem message:
+    if 'HTTP_AUTHORIZATION' not in request.META:
+        return gatekeeper_access_denied_response('Missing Authorization header.')
+
+    authorization = request.META['HTTP_AUTHORIZATION']
+
+    if len(authorization.split(" ")) != 2:
+        return gatekeeper_access_denied_response('Wrong construction of Authorization header.')
+    (golem_header, auth_header_content) = authorization.split()
```
rwrzesien: () is not needed here.

Jakub89: Right! Fixed.

cameel: About these parentheses. They're optional but I encourage using them because they make it actually look like a tuple. With them it's easier to see that you're assigning multiple values there.

Same with return values. `return 1, 2` is valid but for me `return (1, 2)` is more readable because you immediately see that you're returning a tuple of two values.

concent_api/gatekeeper/views.py
cameel: @Jakub89 What about this one?

cameel:
I have just re-reviewed it all, squashed to fixups and added small tweaks. There are still a few API issues. I'm going to add them as comments now. This should be the last round of review.

concent_api/gatekeeper/views.py
```diff
@@ -1,11 +1,77 @@
...
+    if response is not None:
+        return response
+
+    return HttpResponse("", status = 200)
+
+
+@csrf_exempt
+@require_GET
+def download(request):
+    if request.content_type == 'application/x-www-form-urlencoded':
```
cameel: Download is not supposed to get any data from the client so why is it checking the `Content-Type` header here? The download request should not have this header at all.

concent_api/gatekeeper/views.py
```diff
@@ -1,11 +1,77 @@
...
+        return gatekeeper_access_denied_response("Missing token in the 'Authorization' header.")
+
+    (scheme, token) = authorization_scheme_and_token
+
+    if scheme != 'Golem':
+        return gatekeeper_access_denied_response("Unrecognized scheme in the 'Authorization' header.")
+
+    try:
+        decoded_auth_header_content = b64decode(token)
+    except ValueError:
```
cameel: 1) Pass `validate=False` to `b64decode()`

2) Why are you expecting `ValueError` here? The [docs for b64decode()](https://docs.python.org/3.6/library/base64.html#base64.b64decode) say that you will get `binascii.Error` if there are problems with padding or invalid characters in the input.

concent_api/gatekeeper/views.py
Jakub89: When I'm sending curl message with GET it always have:
'CONTENT_TYPE': 'text/plain'
That's why I wanted to have that check.

concent_api/concent_api/urls.py
```diff
@@ -13,14 +13,21 @@
...
+from concent_api.constants  import AVAILABLE_CONCENT_FEATURES
+
+
+if not hasattr(settings, 'CONCENT_FEATURES'):
+    raise ImproperlyConfigured("CONCENT_FEATURES variable must be defined in settings")
+
+if not set(settings.CONCENT_FEATURES) - set(AVAILABLE_CONCENT_FEATURES.keys()) == set():
+    raise ImproperlyConfigured("Unrecognized features in CONCENT_FEATURES")
+
+urlpatterns = []
```
cameel: @rwrzesien 
I'm getting an error from `mypy` here:

```
concent_api/concent_api/urls.py:13: error: Need type annotation for variable)
```

We should probably add a type annotation here, like `urlpatterns: List[...] = []`. But what would be the type of items on the list? How would you handle this?

cameel:
Here's the list of remaining issues to fix:

- [Why do we need `init_messages()`](https://github.com/golemfactory/concent/pull/8/commits/4a2a5e204e3fc7f0fcc9038859772dff70445fb0#r157444950)? golem-messages runs it at module load time so it should not be needed.
- [Download requests should have no content](https://github.com/golemfactory/concent/pull/8/commits/4a2a5e204e3fc7f0fcc9038859772dff70445fb0#r159426629)
- [Incorrect error handling for `b64decode()`](https://github.com/golemfactory/concent/pull/8/commits/4a2a5e204e3fc7f0fcc9038859772dff70445fb0#r159427515)
- [Type error from `mypy`](https://github.com/golemfactory/concent/pull/8/commits/2a15c8220832ae9b4140058373a91447920c39f3#r159427933)

concent_api/gatekeeper/views.py
cameel: Could you give me a curl command that sends this request? Are you sure it's a download request? The client should not be doing that if there is no body in the request.

For example try this:
``` bash
curl http://google.com -v
```

This is a typical GET request. You can see that there is no `Content-Type` header:
```
> GET / HTTP/1.1
> Host: google.com
> User-Agent: curl/7.57.0
> Accept: */*
```

Jakub89: For example:
```
curl -v -H "Authorization: Golem D6UABX47Mo+sAAGmPhp5+S5/iGSvxfrxE+2kCOblP35IcFIGIW4q7es0I1cXWOYs3Tq1SFm5DEPVZj5N5PFWpLqq3d4sgnMRtaGOAQTNrdDEW84chuDsicYhHDfVVGah9Ao33Km1BUhBECZMB38pWTsTnMnu9AvbQ8w+85YX8JnzmUiNuZFkDfnk3ra2klg3WNSZJhpZg+HhmMn5bJqBmtxON+tZ0vJKlws64+W35w7GKYROJ5VLQOp62cQKJamiE+Us8KzulZlLjFobc9o4cKLUQe5FVMuBhNSpk0xVMpoRJgrgXrTPwQjqch+ADyHT88X/0GOkaohfYB2nJCDvrZQmaoCBg1uzQqUktImBwtpiYqzS15HFav8Dk03gCo4dSHTSoqiJtRFCVdFAe1uqUHMcJ9tO8jg2hoXfBVf6YJ5rxULM/52AVnUuDJX4TfrJ8t2Fh0ElHhYY+cTjyC9BRsfWQeLWshyUl3whCRRhM6nxXIHJ69JrwO+gPILri/nI1JArMMRBma9QOtrbF634SYg/2oqn0fRVLTPTi84pQUKiGNlJ8+4gnKD5vPvyWFHPLEd9wiqokkih9veI2MU87Fg8zcwogSSgUd1fhFpDAejqkNtraxVumUIWI2l/pApnkdFPuq7qaMWr8r6dpsb9uYCP7au/UiWzaGFHurGgazBoVTeat78JXkeO52Hhw3QdZJ3z6nxH/HmUhHxW4Nh5iPEPX9Pid8shKaso" -H "Concent-Client-Public-Key: 85cZzVjahnRpUBwm0zlNnqTdYom1LF1P1WNShLg17cmhN2UssnPrCjHKTi5susO3wrr/q07eswumbL82b4HgOw==" -X GET http://127.0.0.1:8000/gatekeeper/download-auth/blender/benchmark/test_task/scene-Helicopter-27-cycles.blend
```
always gives:
```
*   Trying 127.0.0.1...
* Connected to 127.0.0.1 (127.0.0.1) port 8000 (#0)
> GET /gatekeeper/download-auth/blender/benchmark/test_task/scene-Helicopter-27-cycles.blend HTTP/1.1
> Host: 127.0.0.1:8000
> User-Agent: curl/7.47.0
> Accept: */*
> Authorization: Golem D6UABX47Mo+sAAGmPhp5+S5/iGSvxfrxE+2kCOblP35IcFIGIW4q7es0I1cXWOYs3Tq1SFm5DEPVZj5N5PFWpLqq3d4sgnMRtaGOAQTNrdDEW84chuDsicYhHDfVVGah9Ao33Km1BUhBECZMB38pWTsTnMnu9AvbQ8w+85YX8JnzmUiNuZFkDfnk3ra2klg3WNSZJhpZg+HhmMn5bJqBmtxON+tZ0vJKlws64+W35w7GKYROJ5VLQOp62cQKJamiE+Us8KzulZlLjFobc9o4cKLUQe5FVMuBhNSpk0xVMpoRJgrgXrTPwQjqch+ADyHT88X/0GOkaohfYB2nJCDvrZQmaoCBg1uzQqUktImBwtpiYqzS15HFav8Dk03gCo4dSHTSoqiJtRFCVdFAe1uqUHMcJ9tO8jg2hoXfBVf6YJ5rxULM/52AVnUuDJX4TfrJ8t2Fh0ElHhYY+cTjyC9BRsfWQeLWshyUl3whCRRhM6nxXIHJ69JrwO+gPILri/nI1JArMMRBma9QOtrbF634SYg/2oqn0fRVLTPTi84pQUKiGNlJ8+4gnKD5vPvyWFHPLEd9wiqokkih9veI2MU87Fg8zcwogSSgUd1fhFpDAejqkNtraxVumUIWI2l/pApnkdFPuq7qaMWr8r6dpsb9uYCP7au/UiWzaGFHurGgazBoVTeat78JXkeO52Hhw3QdZJ3z6nxH/HmUhHxW4Nh5iPEPX9Pid8shKaso
> Concent-Client-Public-Key: 85cZzVjahnRpUBwm0zlNnqTdYom1LF1P1WNShLg17cmhN2UssnPrCjHKTi5susO3wrr/q07eswumbL82b4HgOw==
> 
* HTTP 1.0, assume close after body
< HTTP/1.0 500 Internal Server Error
< Date: Fri, 05 Jan 2018 14:19:56 GMT
< Server: WSGIServer/0.2 CPython/3.5.2
< Content-Type: text/html
< Content-Length: 121508
< Vary: Cookie
< X-Frame-Options: SAMEORIGIN
```

concent_api/concent_api/urls.py
rwrzesien: @cameel I would do `# type: ignore` here as this is django stuff.

concent_api/gatekeeper/views.py
cameel: > < HTTP/1.0 500 Internal Server Error
> < Date: Fri, 05 Jan 2018 14:19:56 GMT
> < Server: WSGIServer/0.2 CPython/3.5.2
> < Content-Type: text/html
> < Content-Length: 121508

This is a response header, not a request header.

But after looking into this, looks like you're really getting `text/plain` in `request.content_type` in Django. Even though curl does not send the `Content-Type` header, Django returns `'text.plain'` from `request.content_type`. The WSGI sever must be putting it there as a default. I see that this happens only with `manage.py runserver.` It does not if you run it on `gunicorn` for example. So looks like `runserver` is broken in that regard.

Well, not much we can do about this except hacking around this bug. The whole check is actually more or a sanity check than anything else so it does not have to be perfect. Let's add an exception for `text/plain`. Please make the check accept this one content type and add a comment explaining that this is because of this weird behavior of `runserver`. And that adding this exception does not compromise security because we ignore the body anyway. The check is there just so that the client knows it's sending an incorrect request.

Jakub89: Done and pushed on branch gatekeeper-app.

concent_api/concent_api/urls.py
cameel: Well, OK for now but I think that we should eventually come up with a better solution. I hope we won't have to have these ignores everywhere.

concent_api/gatekeeper/views.py
cameel: Oops, I meant `validate=True` :) Sorry for the confusion.

PR #9, opened 2017-12-13
Golem messages in api view

kbeker:
Golem messages branch ready for review


cameel:
This branch still needs some cleanup before it can be submitted as a pull request. There are for example unsquashed fixups and temporary commits.

concent_api/api-test.py
```diff
@@ -1,72 +1,152 @@
 #!/usr/bin/env python3
 
 import sys
-import json
 import datetime
 import random
+import http.client
+from base64                             import b64encode
+
+from golem_messages.message             import Message, MessageForceReportComputedTask, MessageRejectReportComputedTask, MessageCannotComputeTask, MessageTaskToCompute, MessageAckReportComputedTask
```
rwrzesien: We want to introduce some changes in imports. Generally two rules:
1. Try to keep alphabetical order in imports group.
2. When doing many imports from single module, split each one into separate line, like this:
```
from golem_messages.message import Message
from golem_messages.message import MessageForceReportComputedTask
from golem_messages.message import MessageRejectReportComputedTask
....
```

concent_api/api-test.py
```diff
@@ -1,72 +1,152 @@
...
+            message_task_to_compute = message_task_to_compute_dumped
+        )
+
+    elif message_type == 'MessageAckReportComputedTask':
+        return MessageAckReportComputedTask(
+            timestamp = current_time,
+            message_task_to_compute = message_task_to_compute_dumped,
+        )
+    else:
+        assert False, 'Invalid message type'
```
rwrzesien: How about rising exception here instead of assert ?

concent_api/api-test.py
```diff
@@ -1,72 +1,152 @@
...
 DEFAULT_HEADERS = {
-    'Content-Type':              'application/json',
-    'concent-client-public-key': '85cZzVjahnRpUBwm0zlNnqTdYom1LF1P1WNShLg17cmhN2UssnPrCjHKTi5susO3wrr/q07eswumbL82b4HgOw==',
+    'Content-Type':              'application/octet-stream',
+    'concent-client-public-key': b64encode(CONCENT_PUBLIC_KEY).decode('ascii'),
 }
 
 
 def create_data(message_type, task_id):
+    assert task_id > 0
```
rwrzesien: How about switching this one to if and rising exception instead of assert ?
```
if not task_id:
    raise Exception('No task_id provided.')
```

concent_api/api-test.py
```diff
@@ -1,72 +1,152 @@
...
+
+    elif message_type == 'MessageAckReportComputedTask':
+        return MessageAckReportComputedTask(
+            timestamp = current_time,
+            message_task_to_compute = message_task_to_compute_dumped,
+        )
+    else:
+        assert False, 'Invalid message type'
+
+    return None
```
rwrzesien: return None is not needed, this is what function returns by default when no return is specified.

concent_api/api-test.py
```diff
@@ -1,72 +1,152 @@
...
+
+
+def create_reject_data(task_id):
+    assert task_id > 0
```
rwrzesien: Same here.

concent_api/api-test.py
```diff
@@ -1,72 +1,152 @@
...
+        timestamp = current_time,
+        reason = 'provider-quit',
+        task_id = task_id,
+    )
+
+    message_cannot_compute_task_dumped = dump(
+        message_cannot_compute_task,
+        CONCENT_PRIVATE_KEY,
+        CONCENT_PUBLIC_KEY,
+
```
rwrzesien: Please remove blank line.

concent_api/api-test.py
```diff
@@ -1,72 +1,152 @@
...
+    return MessageRejectReportComputedTask(
+        timestamp = current_time,
+        reason = 'cannot-compute-task',
+        message_cannot_compute_task = message_cannot_compute_task_dumped,
+    )
+
+
+def print_golem_message(message, indent = 4):
+    assert isinstance(message, Message)
```
rwrzesien: I see you are a great fan of asserts :) Asserts are ok in general, but I think we want more controlled flow here, with rising and catching exceptions. @cameel what do you think ?

concent_api/api-test.py
```diff
@@ -1,72 +1,152 @@
...
+        else:
+            print('{}{:30} = {}'.format(' ' * indent, field, value))
 
 
 def api_request(host, endpoint, data = None, headers = None):
-    if headers is None:
-        headers = DEFAULT_HEADERS
+    assert endpoint not in ['', None]
+    assert host     not in ['', None]
+    assert headers  not in ['', None]
```
rwrzesien: It can be changed to 
```
if not all([endpoint, host, headers]):
    raise Exception()
```
but also please remove `= None ` from function definition if it is not accepted.

concent_api/api-test.py
```diff
@@ -1,72 +1,152 @@
...
-        response = requests.post("{}/api/v1/{}/".format(host, endpoint), headers = headers, data = json.dumps(data))
-
-    try:
-        decoded_response = response.json()
-    except json.JSONDecodeError:
-        decoded_response = response.text
-        print(response, decoded_response)
+        response = requests.post("{}".format(url), headers = headers, data = data)
+
+    if len(response.content) != 0:
```
rwrzesien: ```if response.content:```
should be enough.

concent_api/api-test.py
```diff
@@ -1,72 +1,152 @@
...
+            CONCENT_PRIVATE_KEY,
+            CONCENT_PUBLIC_KEY,
+        )
+        print('STATUS: {} {}'.format(response.status_code, http.client.responses[response.status_code]))
+        print('MESSAGE:')
+        print_golem_message(decoded_response)
     else:
-        print(response, json.dumps(decoded_response, indent = 4))
+        print('STATUS: {} {}'.format(response.status_code, http.client.responses[response.status_code]))
+        if response.text not in ['', None]:
```
rwrzesien: ```if response.text```

concent_api/core/tests/test_core_views.py
```diff
@@ -1,39 +1,59 @@
...
-from django.test    import TestCase, Client
-from django.urls    import reverse
-from django.http    import JsonResponse, HttpResponse
-from django.utils   import timezone
+from freezegun                      import freeze_time
+from django.test                    import TestCase, Client, override_settings
+from django.urls                    import reverse
+from django.http                    import JsonResponse, HttpResponse
+from django.conf                    import settings
+from django.utils                   import timezone
```
rwrzesien: Please apply rules described above.

concent_api/api-test.py
cameel: See my comments above.

Yes, we use `assert` a lot because it's a great tool to ensure that the assumptions you make are valid. We never use them to control the program flow though. If there's any chance that an assertion will be false even though the code is pefectly valid (e.g. due to user's input, network conditions, hardware failure, etc.) it should indeed be replaced with a standard `if` and raising an exception.

This is not the case here. The function is deliberately written to handle only messages that derive from the `Message` type. Any other case is not even worth coding for and it's the responsibility of the caller to ensure that this precondition holds. E.g. if one wanted to use this function with data where the type is not known, it's his job to check the type and call the function only when he knows it's a `Message`.

This is perfectly appropriate in internal code where you have total control over what arguments the function is called with. It lets you not worry about cases that will never happen but also catch them at the implementation stage if they do happen due to your own mistake.

concent_api/api-test.py
cameel: 1) Yes, but `assert` is fine and let's be explicit about comparisons:

``` python
assert all(value not in ['', None] for value in [endpoint, host, headers]), "Some description."
```

2) (I think I talked about it with @Jakub89 before) I think that we should avoid having too many non-obvious defaults. It often causes hours lost in debugging down the line. And the most obvious default here is to simply have no headers.

concent_api/api-test.py
cameel: We're usually explicit about checks like that.

concent_api/api-test.py
cameel: See my remark about comparisons above.

concent_api/utils/api_view.py
```diff
@@ -35,9 +36,10 @@ def wrapper(request, *args, **kwargs):
         else:
             if request.content_type == '':
                 raise Http400('Content-Type is missing.')
-            if request.content_type == 'application/json':
+            if request.content_type == "application/json":
                 message = json.loads(request.body.decode('ascii'))
```
cameel: @kbeker Please avoid committing random changes like this one.

concent_api/api-test.py
```diff
@@ -1,72 +1,152 @@
...
+import http.client
+from base64                             import b64encode
+
+from golem_messages.message             import Message, MessageForceReportComputedTask, MessageRejectReportComputedTask, MessageCannotComputeTask, MessageTaskToCompute, MessageAckReportComputedTask
+from golem_messages.shortcuts           import dump, load
 
 import requests
 
+CONCENT_PUBLIC_KEY  = b'\xf3\x97\x19\xcdX\xda\x86tiP\x1c&\xd39M\x9e\xa4\xddb\x89\xb5,]O\xd5cR\x84\xb85\xed\xc9\xa17e,\xb2s\xeb\n1\xcaN.l\xba\xc3\xb7\xc2\xba\xff\xabN\xde\xb3\x0b\xa6l\xbf6o\x81\xe0;'
+CONCENT_PRIVATE_KEY = b'l\xcdh\x19\xeb$>\xbcG\xa1\xc7v\xe8\xd7o\x0c\xbf\x0e\x0fM\x89lw\x1e\xd7K\xd6Hnv$\xa2'
```
cameel: You should import keys from settings.

concent_api/core/views.py
```diff
@@ -1,207 +1,300 @@
...
-        if data['message_task_to_compute']['deadline'] < current_time:
-            data['type']   = "MessageRejectReportComputedTask"
-            data['reason'] = "deadline-exceeded"
-            return data
+        if loaded_message.deadline < current_time:
+            reject_message = MessageRejectReportComputedTask(
+                reason = "deadline-exceeded",
+                message_task_to_compute = message.message_task_to_compute,
+            )
+            return reject_message
```
rwrzesien: You can return the message directly, without creating extra variable.

concent_api/core/views.py
```diff
@@ -1,207 +1,300 @@
...
-                decoded_message_to_compute['timestamp'] = current_time
-                return decoded_message_to_compute
-            return decoded_message_data
+    if isinstance(decoded_message_data, MessageRejectReportComputedTask):
+        message_cannot_compute_task = load(
+            decoded_message_data.message_cannot_compute_task,
+            settings.CONCENT_PRIVATE_KEY,
+            client_public_key
+        )
+        message_to_compute = Message.objects.get(type = 'MessageForceReportComputedTask', task_id = message_cannot_compute_task.task_id)
```
rwrzesien: This line is too long, please split into multiple lines.

concent_api/core/views.py
```diff
@@ -1,207 +1,300 @@
...
+        settings.CONCENT_PRIVATE_KEY,
+        client_public_key
+    )
+    if message_task_to_compute.deadline + settings.CONCENT_MESSAGING_TIME <= current_time <= message_task_to_compute.deadline + 2 * settings.CONCENT_MESSAGING_TIME:
+        new_message_data = MessageAckReportComputedTask(
+            task_id                 = decoded_message_data.task_id,
+            message_task_to_compute = decoded_message_data.message_task_to_compute,
+            timestamp               = current_time,
+        )
+        return new_message_data
```
rwrzesien: You can return directly without creating extra variable.

concent_api/core/views.py
```diff
@@ -1,207 +1,300 @@
...
         return decoded_message_data
 
     return HttpResponse("", status = 204)
 
 
 @api_view
 @require_POST
-def receive_out_of_band(_request, _message):
+def receive_out_of_band(request, _message):
+    client_public_key = b64decode(request.META['HTTP_CONCENT_CLIENT_PUBLIC_KEY'].encode('ascii'))
```
rwrzesien: I see this one is repeated several times, how about creating an util function from it ?

concent_api/api-test.py
```diff
@@ -1,72 +1,158 @@
...
-    'Content-Type':              'application/json',
-    'concent-client-public-key': '85cZzVjahnRpUBwm0zlNnqTdYom1LF1P1WNShLg17cmhN2UssnPrCjHKTi5susO3wrr/q07eswumbL82b4HgOw==',
+    'Content-Type':              'application/octet-stream',
+    'concent-client-public-key': b64encode(CONCENT_PUBLIC_KEY).decode('ascii'),
 }
 
 
 def create_data(message_type, task_id):
+    assert task_id >= 0
+    assert task_id is not int
```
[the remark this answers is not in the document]
cameel: I recommend using `assert` a lot but I have mixed feelings about doing type assertions like this. I think that we should start using type annotations instead.

@rwrzesien Have you used type annotations before? Maybe you could add some type checker to `requirements-development` and extend `full-check` to run it?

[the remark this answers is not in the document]
rwrzesien: I have used them but not in Python, but I am familiar with the concept. I have done quick research and this one looks fine http://mypy-lang.org/ .

concent_api/concent_api/settings/development.py
```diff
@@ -10,5 +10,5 @@
 DATABASES['default']['USER']     = 'postgres'
 DATABASES['default']['PASSWORD'] = ''
 
-CONCENT_PUBLIC_KEY  = b'\xf3\x97\x19\xcdX\xda\x86tiP\x1c&\xd39M\x9e\xa4\xddb\x89\xb5,]O\xd5cR\x84\xb85\xed\xc9\xa17e,\xb2s\xeb\n1\xcaN.l\xba\xc3\xb7\xc2\xba\xff\xabN\xde\xb3\x0b\xa6l\xbf6o\x81\xe0;'
-CONCENT_PRIVATE_KEY = b'l\xcdh\x19\xeb$>\xbcG\xa1\xc7v\xe8\xd7o\x0c\xbf\x0e\x0fM\x89lw\x1e\xd7K\xd6Hnv$\xa2'
```
cameel: Why does this key need update? What was wrong with the old one?

concent_api/concent_api/settings/production.py
```diff
@@ -2,3 +2,6 @@
 from .base import *  # NOQA  # pylint: disable=wildcard-import
 
 DEBUG = False
+
+CONCENT_PUBLIC_KEY  = b'D\x1e\x9d\xfe\x0b3\x1a\x9e\xaf\xe8\x1d^\xbb\xf1~\xb1\x15A\xce\xda\xf5\x8f\xe4\x166\x9b\xa9\x00\xe1+]\xcd\xe7\x7f\x93\x17t\xa6\xbd\xbfY\xcf|]\xa5\x9e;|lJ\xbf(q\xb7`\x12S\xcd\n5\xdey\x86\xa5'
+CONCENT_PRIVATE_KEY = b'[\x10!\xca\rv\xb1\x8d7\x84\xd2\x83\xb1\xa5\xab\xf2\x1c\xe0\xb0\x04Pj>hE|tr~,^\xb5'
```
cameel: We absolutely do not want a default key in production.

In `concent-deployment` we will have one but it should be in secrets and not in the repository.

concent_api/core/views.py
```diff
@@ -1,207 +1,308 @@
...
+    client_public_key = decode_client_public_key(request)
+    if isinstance(message, MessageForceReportComputedTask):
+        loaded_message = load(
+            message.message_task_to_compute,
+            settings.CONCENT_PRIVATE_KEY,
+            client_public_key
+        )
+        validate_golem_message_task_to_compute(loaded_message)
+        if Message.objects.filter(task_id = loaded_message.task_id).exists():
+            raise Http400(message.__class__.__name__, " is already being processed for this task.")
```
cameel: This will not construct the message correctly. You'll get something like
`('MessageForceReportComputedTask', ' is already being processed for this task.')`,
not
`MessageForceReportComputedTask is already being processed for this task.`

concent_api/core/views.py
```diff
@@ -1,207 +1,308 @@
...
+            client_public_key,
+        )
+        message_task_to_compute = load(
+            decoded_message_from_database.message_task_to_compute,
+            settings.CONCENT_PRIVATE_KEY,
+            client_public_key,
+        )
+        if current_time <= message_task_to_compute.deadline + settings.CONCENT_MESSAGING_TIME:
 
-        if current_time <= decoded_message_data['message_task_to_compute']['deadline'] + settings.CONCENT_MESSAGING_TIME:
```
cameel: You're missing an `else` here.

concent_api/core/views.py
```diff
@@ -1,207 +1,308 @@
...
-    data = message
-    validate_message(data)
-
-    if data['type'] == "MessageForceReportComputedTask":
-        validate_message_task_to_compute(data['message_task_to_compute'])
-
-        if Message.objects.filter(task_id = data['message_task_to_compute']['task_id']).exists():
-            raise Http400("'ForceReportComputedTask' is already being processed for this task.")
+    client_public_key = decode_client_public_key(request)
+    if isinstance(message, MessageForceReportComputedTask):
```
cameel: This `if` has no final `else` block. Please make sure that there's an explicit `return` for every case in the `if`.

concent_api/core/tests/test_core_views.py
```diff
@@ -46,97 +85,176 @@ def test_send_should_return_http_200_if_message_timeout(self):
...
+        response_400 = self.client.post(
+            reverse('core:send'),
+            data = dump(self.message_want_to_compute, settings.CONCENT_PRIVATE_KEY, settings.CONCENT_PUBLIC_KEY),
+            content_type                   = 'application/octet-stream',
+            HTTP_CONCENT_CLIENT_PUBLIC_KEY = b64encode(CONCENT_PUBLIC_KEY).decode('ascii'),
+        )
+
+        self.assertEqual(response_400.status_code, 400)
+        self.assertIn('error', response_400.json().keys())
+        self.assertIn(response_400.json()['error'], 'Invalid golem message type')
```
cameel: Never test messages like this. You can check that the message is present but the message itself might be different for various reasons like translations. We also don't want the test to fail if someone simply rewords the message.

concent_api/core/views.py
```diff
@@ -1,223 +1,321 @@
...
+            return dump(
+                MessageAckReportComputedTask(
+                    message_task_to_compute = decoded_message_data.message_task_to_compute
+                ),
+                settings.CONCENT_PRIVATE_KEY,
+                client_public_key,
+            )
+    else:
+        last_undelivered_message_status.delivered = True
+        last_undelivered_message_status.save()
```
cameel: `return` missing here.

concent_api/core/views.py
```diff
@@ -1,223 +1,321 @@
...
-    if decoded_message_data['type'] == "MessageAckReportComputedTask":
-        if current_time <= decoded_message_data['message_task_to_compute']['deadline'] + 2 * settings.CONCENT_MESSAGING_TIME:
-            return decoded_message_data
+    # Mark message as delivered
+    if last_undelivered_message_status.message.type == "MessageForceReportComputedTask":
+        message_task_to_compute_from_force = load(
+            decoded_message_data.message_task_to_compute,
+            settings.CONCENT_PRIVATE_KEY,
+            client_public_key)
+        if message_task_to_compute_from_force.deadline + settings.CONCENT_MESSAGING_TIME < current_time:
```
cameel: Missing `else` and `return`.

concent_api/core/views.py
```diff
@@ -1,223 +1,321 @@
...
+        raw_message_data,
+        settings.CONCENT_PRIVATE_KEY,
+        client_public_key,
+    )
 
-    if decoded_message_data['type'] == "MessageAckReportComputedTask":
-        if current_time <= decoded_message_data['message_task_to_compute']['deadline'] + 2 * settings.CONCENT_MESSAGING_TIME:
-            return decoded_message_data
+    # Mark message as delivered
+    if last_undelivered_message_status.message.type == "MessageForceReportComputedTask":
```
cameel: Sometimes you're checking `message.type` and sometimes `isinstance(decoded_message_data, ...)`. Please choose one and use it consistently. I think you should stick to `isinstance()` here.

concent_api/core/views.py
```diff
@@ -1,223 +1,321 @@
...
+                    message_task_to_compute = decoded_message_data.message_task_to_compute
+                ),
+                settings.CONCENT_PRIVATE_KEY,
+                client_public_key,
+            )
+    else:
+        last_undelivered_message_status.delivered = True
+        last_undelivered_message_status.save()
+
+    if isinstance(decoded_message_data, MessageForceReportComputedTask):
```
cameel: This will be always false. You have already processed `MessageForceReportComputedTask` above.

concent_api/core/tests/test_core_views.py
```diff
@@ -46,97 +85,176 @@ def test_send_should_return_http_200_if_message_timeout(self):
...
+
+    @freeze_time("2017-11-17 10:00:00")
+    def test_send_should_return_http_400_if_task_to_compute_deadline_exceeded(self):
+        message_task_to_compute = MessageTaskToCompute(
+            timestamp = self.message_timestamp-10000,
```
cameel: ./concent_api/core/tests/test_core_views.py:193:47: E226 missing whitespace around arithmetic operator

concent_api/core/tests/test_core_views.py
```diff
@@ -46,97 +85,176 @@ def test_send_should_return_http_200_if_message_timeout(self):
...
+            task_id = 8,
+            deadline = self.message_timestamp-9000,
```
cameel: ./concent_api/core/tests/test_core_views.py:195:46: E226 missing whitespace around arithmetic operator

concent_api/core/views.py
```diff
@@ -1,223 +1,321 @@
...
+        return raw_message_data
+    elif isinstance(decoded_message_data, MessageAckReportComputedTask):
+        message_task_to_compute = load(
+            decoded_message_data.message_task_to_compute,
+            settings.CONCENT_PRIVATE_KEY,
+            client_public_key
+        )
+        if current_time <= message_task_to_compute.deadline + 2 * settings.CONCENT_MESSAGING_TIME:
+            return raw_message_data
```
cameel: This is wrong. If the message is already in queue, you should return it. If the client runs into this, he will get 204 and information that there are no messages even if there are other messages already queued behind it.

concent_api/core/views.py
```diff
@@ -1,223 +1,321 @@
...
+        )
+        if current_time <= message_task_to_compute.deadline + 2 * settings.CONCENT_MESSAGING_TIME:
+            if decoded_message_data.reason == "deadline-exceeded":
+                message_ack_report_computed_task = MessageAckReportComputedTask(
+                    timestamp = current_time,
+                    message_task_to_compute = decoded_message_from_database.message_task_to_compute,
+                )
+                return message_ack_report_computed_task
+            return raw_message_data
         return HttpResponse("", status = 204)
```
cameel: This is wrong. You cannot just return HTTP 204 if the last message in the queue was discarded. You must process the next message in the queue.

concent_api/core/views.py
```diff
@@ -1,223 +1,321 @@
...
+            message_ack_report_computed_task.message_task_to_compute    = force_report_computed_task.message_task_to_compute
+            dumped_message_ack_report_computed_task = dump(
+                message_ack_report_computed_task,
+                settings.CONCENT_PRIVATE_KEY,
+                client_public_key
+            )
+            message_verdict.message_ack_report_computed_task = dumped_message_ack_report_computed_task
             return message_verdict
 
     return HttpResponse("", status = 204)
```
cameel: Same here. You cannot simply return if the current message was not returned. You have to check if there are any other messages waiting to be delivered. If so, process them until you find one that can be delivered or you run out of messages.

concent_api/core/views.py
```diff
@@ -1,223 +1,321 @@
...
+        return HttpResponse("", status = 204)
 
 
 @api_view
 @require_POST
-def receive_out_of_band(_request, _message):
-    last_task_message = Message.objects.order_by('id').last()
+def receive_out_of_band(request, _message):
+    client_public_key = decode_client_public_key(request)
+    last_task_message = Message.objects.order_by('timestamp').last()
```
cameel: You have a big race condition here. You process messages meant to be delivered by `receive()`. So the client may or may not get his out-of-band message depending on whether the other client has retrieved his message or not.

This is completely wrong. There should be two separate queues, one for normal messages and the other for out-of-band messages. Receiving a message from one of them should not affect the other.

cameel:
@kbeker When you're closing a pull request or an issue you should always explain why,

kbeker:
@cameel Ok, I havent known about it. Next time I will of course write explanation

cameel:
Well, it's not like you have to write a full page report :)

Just drop a line saying that you closed it because there were big changes needed and we decided that you're going to rewrite it and sumit it as a new pull request for review from scratch.

Also, it's a good idea to link to to the old request from the new one so that if someone later wants to figure out what happened, he's not confused seeing two seemingly indentical pull requests.

PR #11, opened 2017-12-18
Add mypy static type checker

rwrzesien:


requirements-development.txt
```diff
@@ -7,6 +7,7 @@ pipdeptree
 flake8
 pylint-django
 requests
+mypy==0.560
```
cameel: This should not have a version hard-coded.

Our convention is to have two dependency files: `requirements.txt` and `requirements.lock`.
- `requirements.txt` contains only top-level dependencies and does not state any specific version unless there's a good reason for it. For example it can use `!= x.y.z` if version x.y.z is known to be broken. Or can use `>= a.b.c` if versions older than a.b.c are too old for us.
- `requirements.lock` contains all the dependencies, including dependencies of dependencies, each one pinned to the exact version we used to develop and test the code. This is so that on the server we use only versions known to be good. `requirements.lock` is basically the output of `pip freeze` after installing stuff from `requirements.txt`. This is similar to how `Gemfile`/`Gemfile.lock` work in Ruby. Or `package.json`/`npm-shrinkwrap.json` in node.js.

The file you've modified (`requirements-development.txt`) follows similar rules, but it does not have a corresponding `.lock` file because these packages never go on the server. So It's ok to put something like `mypy >= 0.560` here but you should not hard-coded it to only that specific version unless you know for a fact that no other version will work for us.

rwrzesien: @cameel Sounds very reasonable, I have pushed fix.

cameel: OK. But could you change it to a fixup to the original commit?

cameel:
Merged into master.

BTW, @rwrzesien Our convention is to use dashes rather than underscores in branch names. I've renamed it to match our style before merge.

Issue #18, opened 2017-12-22
Response header with version of golem-messages


cameel:
In all HTTP responses from concent-api and gatekeeper include a header with version of golem-message being used on the server.

If we're using a tagged version, include that tag. Otherwise include the commit ID.

Requested by @jiivan to help in debugging Concent-related issues on the Golem side.

rwrzesien:
@cameel I would like to use GitPython for this http://gitpython.readthedocs.io . Now, it has attribute name_rev on commit object, which have value like:
'ca6f01c15451fbf877ef6e3043bd55e607b7d935 tags/v1.2.1' when commit is tag, and 
'f94d72024f8fc4d496978915c31256d6a45fc55f remotes/origin/concent-messages' when commit is not a tag. 

Would that be ok to use this as version header ? It contains all the info plus extra branch name and origin for non-tag commits.

If you want to have like in the original description, it is not a problem, it will just require one or two more ifs.

cameel:
I think that we should have just the tag (i.e. `v1.2.1`) when it's a tagged commit. I think that most of the time we'll be running a tagged version and the commit ID is just noise in that case. It's still helpful if there is no tag though to so it's fine to add it on untagged commits. Branch name is completely unnecessary.

Another thing is that I'd prefer if we did not rely entirely on git. The code does not have to come from git. It can be installed from a python package.

cameel:
I see that @jiivan has added `__version__` to golem-messages a few days ago: https://github.com/golemfactory/golem-messages/commit/eed9f9340f97513e949f64c8ec8be96334003dbf#diff-ae716e1448ba977101f358c0355724abR8. We should use it.

It looks like it also can handle installing from a package (via a `RELEASE-VERSION` file that must be placed in the package during build). See [version.py](https://github.com/golemfactory/golem-messages/blob/v1.3.1/version.py).

rwrzesien:
@cameel Pull request created https://github.com/golemfactory/concent/pull/33

cameel:
Merged into master: 83cd12b8f021e692e68460b91752fa5174a9b2c5

PR #22, opened 2017-12-28
Add version of golem_message to every response header

rwrzesien:
https://github.com/golemfactory/concent/issues/18

I have started from golem-messages-in-api-view as contains necessary stuff for tests.

Please note that tests are put into concet_api/tests directory, as concent_api will be django app.


concent_api/concent_api/settings/base.py
```diff
@@ -1,5 +1,12 @@
 import os
 
+from git import (
+    InvalidGitRepositoryError,
+    Repo,
+)
```
cameel: You forgot to add this library to `requirements.txt` and to regenerate `requirements.lock`.

concent_api/concent_api/settings/base.py
```diff
@@ -196,3 +204,10 @@
 # A global constant defining the length of the time window within which a requestor or a provider is supposed to
 # contact concent and send or receive a message as defined in the protocol.
 CONCENT_MESSAGING_TIME = 3600  # seconds
+
+# Loads git tag or commit id of the golem_messages package currently used.
+try:
+    repo = Repo(os.path.dirname(golem_messages.__file__), search_parent_directories=True)
+    GOLEM_MESSAGES_VERSION = repo.commit().name_rev
+except InvalidGitRepositoryError:
+    GOLEM_MESSAGES_VERSION = None
```
cameel: 1) I think that adding such a heavy piece of code to settings is not a good idea. Now even simply importing settings would require lots of dependencies: gitpython and git, golem-messages, all its cryptographic libraries and even OpenSSL. Sometimes we may want to import settings in scripts that should not have that much overhead.

    Also note that this information is not universally needed. We want to return the header from `concent-api` but there is also `gatekeeper` and later other features sharing the same code base but running somewhere else on the cluster (verification agent and eth wallet agent) that will never ever need it.

    I think this does not really work as a setting. I'd make it a function in `utils`. And then call that function during middleware initialization. Possibly, caching the result somewhere - but definitely not in settings.

2) I'm not sure if relying entirely on git is a good idea here. What if golem-messages is installed from a python package? Installing it directly from github is just a temporary arrangement.

concent_api/concent_api/tests/test_middleware.py
```diff
@@ -0,0 +1,51 @@
...
+(PROVIDER_PRIVATE_KEY,  PROVIDER_PUBLIC_KEY)  = generate_ecc_key_pair()
+
+
+@override_settings(
+    CONCENT_PRIVATE_KEY    = CONCENT_PRIVATE_KEY,
+    CONCENT_PUBLIC_KEY     = CONCENT_PUBLIC_KEY,
+)
+class GolemMessagesVersionMiddlewareTest(TestCase):
+
+    def test_golem_messages_version_middleware_attach_http_header_to_response(self):
```
cameel: Just FYI: our convention with naming tests methods is `test_X_should_Y` where X is what you want to test and Y is what you expect it to do. You're missing the `_should_` here.

concent_api/concent_api/tests/test_middleware.py
```diff
@@ -0,0 +1,51 @@
...
+        serialized_ping_message = dump(ping_message, PROVIDER_PRIVATE_KEY, CONCENT_PUBLIC_KEY)
+
+        response = self.client.post(
+            reverse('core:send'),
+            data                           = serialized_ping_message,
+            content_type                   = 'application/octet-stream',
+            HTTP_CONCENT_CLIENT_PUBLIC_KEY = b64encode(PROVIDER_PUBLIC_KEY).decode('ascii'),
+        )
+
+        self.assertFalse(str(response.status_code).startswith('5'))
```
cameel: What would you say about
``` python
self.assertTrue(500 <= response.status_code < 600)
```

or

```
self.assertGreaterEqual(response.status_code, 500)
self.assertLess(response.status_code, 600)
```
which is, admittedly, longer and a bit less readable but gives more meaningful error messages if a test fails.

concent_api/concent_api/settings/base.py
rwrzesien: This will be not removed if I use golem_messages.__version__

concent_api/concent_api/settings/base.py
rwrzesien: This will be not removed if I use golem_messages.__version__

concent_api/concent_api/tests/test_middleware.py
rwrzesien: :+1:

concent_api/concent_api/tests/test_middleware.py
rwrzesien: :+1:

cameel: One more thing. Now that I think about it, this check will never fail. That's because in tests Django does not catch view exceptions to replace them with HTTP 500. If there's an exception in the view, it simply escapes and interrupts the test.

In general, we never test for 5xx errors (or their absence) in tests.

rwrzesien: It was just added to confirm that we get expected response from Concent.

rwrzesien:
@cameel Rebased on latest master and valid-golem-messages, tests passed so it looks like it is working correctly with golem_messages version with __version__.

PR #24, opened 2018-01-03
Add coverage report for django tests

### LINKED ISSUE #23: Test coverage report
Add a tool that checks test coverage. Might be integrated with test runner or a separate tool.

Always run it in `full_check.sh`.

rwrzesien:
Resolves #23
https://www.pivotaltracker.com/story/show/153992502

### DIFF
--- a/coverage-config
+++ b/coverage-config
@@ -1,6 +1,18 @@
 # .coveragerc to control coverage.py
 [run]
-omit=*/migrations/*,*/mocks/*,*/tests*,*/management/*,*/wsgi.py,*/admin.py,manage.py,*/settings/*,*/apps.py,*/__init__.py
+omit=
+    api-integration-test.py
+    api-test.py
+    manage.py
+    */__init__.py
+    */admin.py
+    */apps.py
+    */wsgi.py
+    */management/*
+    */migrations/*
+    */mocks/*
+    */tests*
+    */settings/*
 
 [html]
 directory = ../htmlcov
--- a/full-check.sh
+++ b/full-check.sh
@@ -17,6 +17,5 @@ printf "========================= UNIT TESTS WITH COVERAGE =================\n"
 rm -rf contrib/coverage/coverage_html
 cd concent_api/
 coverage run --rcfile=../coverage-config --source='.' manage.py test
-coverage html --rcfile=../coverage-config
 coverage report --show-missing
 cd ..


full-check.sh
```diff
@@ -12,8 +12,11 @@ printf "========================= MYPY STATIC TYPE CHECKER =================\n"
...
 
-printf "============================ UNIT TESTS ============================\n"
+printf "========================= UNIT TESTS WITH COVERAGE =================\n"
 # NOTE: 'manage.py test' does not find all tests unless we run it from within the app directory.
+rm -rf contrib/coverage/coverage_html
 cd concent_api/
-python3 manage.py test
+coverage run --rcfile=../coverage-config --source='.' manage.py test
+coverage html --rcfile=../coverage-config
+coverage report -m
```
cameel: Please use long form of options whenever possible. It makes the calls easier to understand for someone who is not familiar with the command and we're not running it by hand to the command does not need to be short.

So, please replace `-m` with `--show-missing`.

rwrzesien: :+1:

full-check.sh
```diff
@@ -12,8 +12,11 @@ printf "========================= MYPY STATIC TYPE CHECKER =================\n"
...
 printf "\n"
 
-printf "============================ UNIT TESTS ============================\n"
+printf "========================= UNIT TESTS WITH COVERAGE =================\n"
 # NOTE: 'manage.py test' does not find all tests unless we run it from within the app directory.
+rm -rf contrib/coverage/coverage_html
 cd concent_api/
-python3 manage.py test
+coverage run --rcfile=../coverage-config --source='.' manage.py test
+coverage html --rcfile=../coverage-config
```
cameel: I'd remove the html report from this script. There's no point in generating it every time we run checks of the code. The report in the console is enough. I think that most of the time the programmer will not even look at the html report.

It may be useful to generate it from time to time and look at it but I think there should be a separate script for that. Or just run the command manually.

coverage-config
```diff
@@ -0,0 +1,6 @@
+# .coveragerc to control coverage.py
+[run]
+omit=*/migrations/*,*/mocks/*,*/tests*,*/management/*,*/wsgi.py,*/admin.py,manage.py,*/settings/*,*/apps.py,*/__init__.py
```
cameel: Please add our integration tests (`api-test.py` and `api-integration-test.py`) here. They're at 0% coverage and I think they'll stay at that because, after all, we won't be writing tests for tests :)

Also, is it possible to wrap this into multiple lines? Just like with imports, it would make future diffs easier to read.

full-check.sh
rwrzesien: :+1:

coverage-config
rwrzesien: :+1:

full-check.sh
rwrzesien: Depends on what we are trying to achieve, if we want to push only 100% tested code then it might be useful, if not then it is ok to remove imho.

cameel: But the summary in the console already shows coverage? HTML report is just a nicely formatted version of it, right? So it's not necessary to generate it every time. Only if you want to look at it.

rwrzesien: No, HTML report shows a list of html files, where each other is a separate file that was examined for coverage, and it marks not covered lines in red, covered lines in green. It also show some extra information, like for example if you have
`
if something:
    do()
`
if both paths (True and False) were taken. Go ahead and generate one and take a look. It is very useful if you want to check which lines exactly are not covered. So, please let me know if you want to keep it or not. It could take a while to generate if the project gets really big, but how often does developer runs full_check ? If he is stuck on test or lint, he can run only a single command.

So please

rwrzesien:
@cameel I have added fixes for all issues, also removed html report for now, but there is no problem in bringing it back off course.

full-check.sh
cameel: OK. I haven't looked at the details of the report. I like it.

I just don't think `full_check.sh` is a good place for it. It's meant to be comprehensive but still pretty cheap to run during development. If use it when I want to run all the tests because extra checks do not take all that long and are useful. Only when I start debugging a specific problem I start running `lint.sh` or specific unit test. I don't think we'll be looking at the report just as often.

That being said, I think looking at this detailed report is pretty useful. I suggest making a separate script for generating it. I was also thinking about serving the coverage report from the development cluster though I'm not sure how useful it would be. Maybe not that useful because we should check coverage before we even let the code run on the cluster. Anyway, if you have any other idea on when to run it, we can discuss it.

PR #25, opened 2018-01-03
Golem messages in api view

kbeker:
Ready for another pull request. All fixups squashed, all tests are ok.

### DIFF
--- a/concent_api/utils/api_view.py
+++ b/concent_api/utils/api_view.py
@@ -1,3 +1,4 @@
+import binascii
 from functools                      import wraps
 from base64                         import b64decode
 
@@ -19,16 +20,19 @@ def api_view(view):
     @wraps(view)
     @csrf_exempt
     def wrapper(request, *args, **kwargs):
-        if 'HTTP_CONCENT_CLIENT_PUBLIC_KEY' not in request.META:
+        if 'HTTP_CONCENT_CLIENT_PUBLIC_KEY' not in request.META or request.META['HTTP_CONCENT_CLIENT_PUBLIC_KEY'] == '':
             return JsonResponse({'error': 'Concent-Client-Public-Key HTTP header is missing on the request.'}, status = 400)
 
         try:
-            client_public_key = b64decode(request.META['HTTP_CONCENT_CLIENT_PUBLIC_KEY'].encode('ascii'))
-        except TypeError:
-            # From b64decode() docs:
-            # A TypeError is raised if s is incorrectly padded.
-            # Characters that are neither in the normal base-64 alphabet nor the alternative alphabet are discarded prior to the padding check.
+            client_public_key = b64decode(request.META['HTTP_CONCENT_CLIENT_PUBLIC_KEY'].encode('ascii'), validate=True)
+        except binascii.Error:
             return JsonResponse({'error': 'The value in the Concent-Client-Public-Key HTTP is not a valid base64-encoded value.'}, status = 400)
+
+        if len(client_public_key) != 64:
+            return JsonResponse(
+                {'error': 'The length in the Concent-Client-Public-Key HTTP is wrong.'},
+                status = 400
+            )
 
         if request.content_type not in ['application/octet-stream', '', 'application/json']:
             return JsonResponse({'error': 'Concent supports only application/octet-stream and application/json.'}, status = 415)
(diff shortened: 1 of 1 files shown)


concent_api/core/tests/test_integration_report_computed_task.py
```diff
@@ -429,3 +429,950 @@ def test_provider_forces_computed_task_report_and_requestor_does_not_respond(sel
...
+        with freeze_time("2017-12-01 10:59:00"):
+            response_2 = self.client.post(
+                reverse('core:send'),
+                data                           = serialized_force_report_computed_task,
+                content_type                   = 'application/octet-stream',
+                HTTP_CONCENT_CLIENT_PUBLIC_KEY = b64encode(PROVIDER_PUBLIC_KEY).decode('ascii'),
+            )
+
+        self.assertEqual(response_2.status_code, 400)
+        self.assertEqual(response_2.json(), {"error": "MessageForceReportComputedTask is already being processed for this task."})
```
cameel: @rwrzesien Please don't test the exact text of error messages. This makes for very brittle tests. It's enough to check i the `error` field is present.

rwrzesien: @cameel I think it makes a decent test. If someone is changing the implementation, he should also change the test. Also, checking only for `error` field present may be not enough as we are returning different errors for different cases. Also, all strings, including error messages, should be defined as constants and used in both implementation and tests without concern about text change.

cameel: From my experience, this produces a lot of false positives. I can't remember any case where a test failing due to a tweaked message actually found any breakage. But I remember a lot of cases where I had to tediously go over the tests and find and fix each and every changed piece of text. It does not add any value but forces you to write the text twice.

Also it's hard to test this reliably. Messages often go through translation and you have to expend extra effort to make sure you're comparing with the right version. And I'm not sure if extracting all the strings into a translation table is the best way to combat this. What I've seen more often is something like: `_("This is an error message.")`, where `_()` is a `gettext` call.

And it's the same for the actual client - no one should compare these messages on the client side. They're only a hint for the user. What we really should have are machine-readable error codes.

concent_api/utils/api_view.py
```diff
@@ -19,17 +20,23 @@ def api_view(view):
...
     @csrf_exempt
     def wrapper(request, *args, **kwargs):
-        if 'HTTP_CONCENT_CLIENT_PUBLIC_KEY' not in request.META:
+        if 'HTTP_CONCENT_CLIENT_PUBLIC_KEY' not in request.META or request.META['HTTP_CONCENT_CLIENT_PUBLIC_KEY'] == '':
             return JsonResponse({'error': 'Concent-Client-Public-Key HTTP header is missing on the request.'}, status = 400)
 
         try:
             client_public_key = b64decode(request.META['HTTP_CONCENT_CLIENT_PUBLIC_KEY'].encode('ascii'))
-        except TypeError:
+        except (AttributeError, BinAsciiError):
```
cameel: In what situation can an `AttributeError` be thrown here? And why `BinAsciiError` instead of `binascii.Error`? The [docs for `b64decode()`](https://docs.python.org/3.6/library/base64.html#base64.b64decode) mention only `binascii.Error`.

Also, please use `verify=True`.

rwrzesien: @cameel When I was implementing it `AttributeError` was already there, so this diff must be somehow malformed along the way, I will change it, and also remove comment and add `verify=True`

About BinAsciiError, please check the import part, I am importing `binascii.Error` as `BinAsciiError` because `Error` is to common name.

concent_api/core/tests/test_integration_report_computed_task.py
rwrzesien: I agree that error codes are better, but currently we don't have those. Also there are custom exception classes which are also cool. And as you described a situation when you go over the tests and find and fix each and every changed piece of text, that is a perfect reason to use constants.

And you can use translation on both constant, and in tests, as they are run with django context with a specific settings which determine language. It is ok to do:
```
ERROR_1 = _("This is an error message.")
```
If you want I can extract strings into constants from whole application, which I think should free us from maintaining them on both sides in the future, but maybe lets do this as separate task.

concent_api/utils/api_view.py
cameel: Oh, OK. Maybe it would be better to do `import binascii` and then catch `binascii.Error`? I think it would be a bit less confusing. Also, it's easier to grep for stuff when it's not renamed like that.

rwrzesien: Right, I will update now.

concent_api/core/tests/test_integration_report_computed_task.py
cameel: Generally I think we should remove these checks for now. We can reintroduce them when they can be done right. Leaving them as is the worst of both approaches.

I've been through various approaches to this problem (gettext, costants, global message tables, etc.) in various languages and all of them have significant drawbacks. It's even more evident in applications where you have UI and other people (but not necessarily translators) want to be able to edit those strings. My opinion is simply that text is too volatile and it's not a good idea to test it. It's not worth the hassle.

If you want to refactor the code to add constants for all messages, that would be a bit less bad so I'm fine with it, but please let's not leave the text hard-coded like this.

rwrzesien: @cameel Ok, as we discussed on daily they will be removed by @kbeker .

cameel:
Merged into master.

There were lots of small changes. Finally the branch is good enough to be merged.

PR #29, opened 2018-01-08
Feature authenticating golem clients based on public key

### LINKED ISSUE #16: Identifying clients based on public keys passed in headers
We don't have an authorization mechanism yet and we also don't even check who is sending/receiving a message. All we currently do is to ensure that a message signature matches the key sent in the `Concent-Client-Public-Key` HTTP header.

The authorization mechanism is still to be designed but for now we should at least start to identify the clients.
- Every message to be delivered from `receive/` and `receive-out-of-band/` endpoints should only be delivered to the client it's meant for. So you should store receiver's public key with the message and only return that message if the key is the same as the one in the `Concent-Client-Public-Key` HTTP header.
- In the `send/` endpoint:
    - The public key or the client who sent `ForceReportComputedTask` should be remembered as provider's key for that task. All the subsequent messages from the provider regarding this task must use the same key.
    - When sending `ForceReportComputedTask`, the client must send requestor's public key in the `Concent-Requestor-Public-Key` header. The key must match the signature on the `TaskToCompute` message. All subsequent messages from the requestor regarding this task must use the same key.
    - `AckReportComputedTask` and `RejectReportComputedTask` must be submitted and signed by the requestor.
    - `CannotComputeTask` and `TaskFailure` messages sent inside other messages must be signed by the provider.

rwrzesien:
https://www.pivotaltracker.com/story/show/153696350
Resolves #16


rwrzesien:
@cameel This is rebased on `fix-issues-valid-golem-messages` and ready for review.

rwrzesien:
@cameel Done.

concent_api/core/views.py
```diff
@@ -437,5 +552,17 @@ def store_receive_out_of_band(golem_message, message_timestamp):
...
-    return b64decode(request.META['HTTP_CONCENT_CLIENT_PUBLIC_KEY'].encode('ascii'), validate = True)
+    return decode_key(request.META['HTTP_CONCENT_CLIENT_PUBLIC_KEY'])
+
+
+def decode_requestor_public_key(request):
+    if 'HTTP_CONCENT_REQUESTOR_PUBLIC_KEY' in request.META:
+        try:
+            return decode_key(request.META['HTTP_CONCENT_REQUESTOR_PUBLIC_KEY'])
+        except TypeError:
+            return JsonResponse({'error': 'The value in the Requestor-Public-Key HTTP is not a valid base64-encoded value.'}, status = 400)
```
cameel: You're missing the final `return`. If your intention was to to return `None`, please do it explicitly. Otherwise it looks as if you forgot about `else`.

concent_api/core/views.py
```diff
@@ -419,6 +525,15 @@ def store_message(golem_message_type, task_id, raw_golem_message):
     return (golem_message, message_timestamp)
 
 
+def store_message_auth(golem_message, provider_public_key, requestor_public_key):
```
cameel: The function tries to insert keys into fields that are non-nullable so it should state as a precondition that `provider_public_key` and `requestor_public_key` are not `None` (using `assert`).

concent_api/core/views.py
```diff
@@ -43,15 +46,33 @@ def send(_request, client_message):
             golem_message,
             message_timestamp,
         )
+        store_message_auth(
+            golem_message,
+            provider_public_key  = client_public_key,
+            requestor_public_key = requestor_public_key,
```
cameel: You're inserting `requestor_public_key` into a non-nullable column but it can be `None` (`decode_requestor_public_key()` may return `None`) and you do not have any checks against this condition here.

concent_api/utils/fields.py
```diff
@@ -0,0 +1,24 @@
...
+
+    def contribute_to_class(self, cls, name, private_only=False, virtual_only=models.fields.NOT_PROVIDED):
+        if self.db_column is None:
+            self.db_column = name
+        self.field_name = name + '_base64'
+        super().contribute_to_class(cls, self.field_name, private_only=private_only, virtual_only=virtual_only)
+        setattr(cls, name, property(self.get_data, self.set_data))
+
+    def get_data(self, obj):
+        return base64.b64decode(getattr(obj, self.field_name))
```
cameel: Please use `validate = True` when decoding base64 data.

concent_api/utils/fields.py
```diff
@@ -0,0 +1,24 @@
...
+
+    def set_data(self, obj, data):
+        setattr(obj, self.field_name, base64.b64encode(data))
```
cameel: Is this decoding/encoding really necessary here? In the views you assign values that are already base64-encoded (because the client sends them in an encoded form) so aren't you encoding them twice?

concent_api/utils/fields.py
```diff
@@ -0,0 +1,24 @@
+import base64
+
+from django.db import models
+
+
+class Base64Field(models.TextField):
```
cameel: Does this field expect the user to assign strings that are already base-64 encoded or unencoded binary data? Or can it handle both? This should be documented here.

concent_api/utils/fields.py
```diff
@@ -0,0 +1,24 @@
...
+    """ Base64 encoding field for storing binary data in Django TextFields. """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.field_name = None
+
+    def contribute_to_class(self, cls, name, private_only=False, virtual_only=models.fields.NOT_PROVIDED):
+        if self.db_column is None:
+            self.db_column = name
+        self.field_name = name + '_base64'
```
cameel: What's the purpose of this `_base64` suffix?

Looks to me like something analogous to `something` and `something_id` with foreign keys where the normal field stores the object (or makes a query) and the `_id` field has the raw `_id`. But I don't see it here. You're not storing both encoded and decoded version of the data here so this is not the same.

concent_api/core/views.py
```diff
@@ -17,6 +21,7 @@
 @api_view
 @require_POST
 def send(_request, client_message):
+    client_public_key = decode_client_public_key(_request)
```
cameel: This change belongs in the next commit.

concent_api/core/models.py
```diff
@@ -44,11 +61,11 @@ class Meta:
         verbose_name_plural = ('ReceiveOutOfBand statuses')
 
 
-class MessageAuth(models.Model):
+class MessageAuth(Model):
```
cameel: This should be a fixup to the commit that adds the model. It does not belong in this commit.

concent_api/core/models.py
```diff
@@ -44,11 +61,11 @@ class Meta:
...
     """
     This class is used to store provider and requestor keys for message exchange related to given initial message.
     """
 
-    message              = models.OneToOneField('Message', related_name='auth')
+    message              = OneToOneField('Message', related_name='auth')
```
cameel: Same here. This change should be in the same fixup with the change above.

EDIT: To be clear, I don't mean a standalone commit marked as fixup. I mean that the original commit that adds the models should be amended to already have this change.

concent_api/core/migrations/0004_messageauth.py
```diff
@@ -0,0 +1,26 @@
...
+
+from django.db import migrations, models
+import django.db.models.deletion
+import utils.fields
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('core', '0002_auto_20180105_0945'),
```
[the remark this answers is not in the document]
rwrzesien: This is automated change, I think there was some migration there before that was removed later, I will regenerate it. But generally merging migration would fix it too.

concent_api/core/migrations/0004_messageauth.py
```diff
@@ -0,0 +1,26 @@
...
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='MessageAuth',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('provider_public_key_base64', utils.fields.Base64Field(db_column='provider_public_key', max_length=64)),
+                ('requestor_public_key_base64', utils.fields.Base64Field(db_column='requestor_public_key', max_length=64)),
```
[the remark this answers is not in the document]
rwrzesien: It will be regenerated.

concent_api/core/migrations/0004_messageauth.py
```diff
@@ -0,0 +1,26 @@
...
+    dependencies = [
+        ('core', '0002_auto_20180105_0945'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='MessageAuth',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('provider_public_key_base64', utils.fields.Base64Field(db_column='provider_public_key', max_length=64)),
```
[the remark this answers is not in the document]
rwrzesien: It will be regenerated.

concent_api/core/models.py
```diff
@@ -17,11 +22,22 @@ def __str__(self):
         return 'Message #{}, type:{}, {}'.format(self.id, self.type, self.timestamp)
 
 
+class ReceiveStatusManager(Manager):
+
+    def get_for_public_key(self, client_public_key):
```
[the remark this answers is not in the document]
rwrzesien: :+1:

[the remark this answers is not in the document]
rwrzesien: :+1:

concent_api/core/models.py
```diff
@@ -35,9 +51,21 @@ class ReceiveOutOfBandStatus(Model):
...
         verbose_name        = ('ReceiveOutOfBand status')
         verbose_name_plural = ('ReceiveOutOfBand statuses')
+
+
+class MessageAuth(Model):
+    """
+    This class is used to store provider and requestor keys for message exchange related to given initial message.
+    """
+
+    message              = OneToOneField('Message', related_name='auth')
```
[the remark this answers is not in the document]
rwrzesien: When I was creating it all others were defined this way, so I wanted to follow up the convention, but I see they have changed now so I will update.

concent_api/core/tests/test_core_views.py
```diff
@@ -325,6 +335,11 @@ def test_receive_should_accept_valid_message(self):
         )
         new_message_status.full_clean()
         new_message_status.save()
+        MessageAuth.objects.create(
+            message              = new_message,
+            provider_public_key  = PROVIDER_PUBLIC_KEY,
+            requestor_public_key = REQUESTOR_PUBLIC_KEY,
+        )
```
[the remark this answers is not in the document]
rwrzesien: Why should I do this in tests ?

concent_api/core/views.py
rwrzesien: How this can be moved without resetting changes of all following commits ?

concent_api/core/views.py
```diff
@@ -1,29 +1,31 @@
...
 from .models                        import ReceiveOutOfBandStatus
 
 
 @api_view
 @require_POST
 def send(_request, client_message):
+    client_public_key = decode_client_public_key(_request)
     current_time = int(datetime.datetime.now().timestamp())
     if isinstance(client_message, message.ForceReportComputedTask):
+        requestor_public_key = decode_requestor_public_key(_request)
```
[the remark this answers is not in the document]
rwrzesien: Then it must have changed during rebase conflicts, it should raise Http400 instead of returning JSON,

concent_api/core/views.py
rwrzesien: I have changed it to catch KeyError and raise Http400 when requestor key is missing in header but expected to be there.

concent_api/core/views.py
```diff
@@ -70,6 +91,11 @@ def send(_request, client_message):
                 golem_message,
                 message_timestamp,
             )
+            store_message_auth(
+                golem_message,
+                provider_public_key  = force_task_to_compute.last().auth.provider_public_key,
+                requestor_public_key = client_public_key,
+            )
```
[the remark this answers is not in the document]
rwrzesien: Why would it fail if `force_task_to_compute` query ever returns more than one object ?

concent_api/core/views.py
```diff
@@ -78,7 +104,11 @@ def send(_request, client_message):
     elif isinstance(client_message, message.RejectReportComputedTask):
         validate_golem_message_reject(client_message.cannot_compute_task)
 
-        force_report_computed_task_from_database = Message.objects.filter(task_id = client_message.cannot_compute_task.task_to_compute.compute_task_def['task_id'], type = message.ForceReportComputedTask.TYPE)
+        force_report_computed_task_from_database = Message.objects.filter(
+            task_id                           = client_message.cannot_compute_task.task_to_compute.compute_task_def['task_id'],
+            type                              = message.ForceReportComputedTask.TYPE,
+            auth__requestor_public_key_base64 = _request.META['HTTP_CONCENT_CLIENT_PUBLIC_KEY']
```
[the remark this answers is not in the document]

concent_api/core/views.py
```diff
@@ -93,20 +123,39 @@ def send(_request, client_message):
...
             )
+            store_receive_message_status(
+                golem_message,
+                message_timestamp,
+            )
+            store_message_auth(
+                golem_message,
+                provider_public_key  = force_report_computed_task_from_database.last().auth.provider_public_key,
+                requestor_public_key = client_public_key,
+            )
```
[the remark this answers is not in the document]
rwrzesien: Then it should be retrieved using `Message.objects.get()` method and DoesNotExists exception might be handled by using 
```        if not force_report_computed_task_from_database.exists():
            raise Http400("'ForceReportComputedTask' for this task has not been initiated yet. Can't accept your 'RejectReportComputedTask'.")
``` 
this code, but how would you propose to handle MultipleObjectsReturned error ?

concent_api/core/views.py
```diff
@@ -128,15 +182,22 @@ def send(_request, client_message):
 
 @api_view
 @require_POST
-def receive(request, _message):
-    client_public_key               = decode_client_public_key(request)
-    last_undelivered_message_status = ReceiveStatus.objects.filter(delivered = False).order_by('timestamp').last()
+def receive(_request, _message):
+    client_public_key               = decode_client_public_key(_request)
```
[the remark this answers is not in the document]

concent_api/core/views.py
```diff
@@ -241,11 +305,13 @@ def receive(request, _message):
 
 @api_view
 @require_POST
-def receive_out_of_band(request, _message):
+def receive_out_of_band(_request, _message):
```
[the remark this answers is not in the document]

concent_api/core/views.py

concent_api/core/views.py
```diff
@@ -419,6 +525,15 @@ def store_message(golem_message_type, task_id, raw_golem_message):
     return (golem_message, message_timestamp)
 
 
+def store_message_auth(golem_message, provider_public_key, requestor_public_key):
+    """ Store both keys related to this operation for future messages. """
+    MessageAuth.objects.create(
+        message=golem_message,
+        provider_public_key=provider_public_key,
+        requestor_public_key=requestor_public_key,
+    )
```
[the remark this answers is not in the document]

concent_api/core/views.py
```diff
@@ -437,5 +552,17 @@ def store_receive_out_of_band(golem_message, message_timestamp):
...
 def decode_client_public_key(request):
-    return b64decode(request.META['HTTP_CONCENT_CLIENT_PUBLIC_KEY'].encode('ascii'), validate = True)
+    return decode_key(request.META['HTTP_CONCENT_CLIENT_PUBLIC_KEY'])
+
+
+def decode_requestor_public_key(request):
+    if 'HTTP_CONCENT_REQUESTOR_PUBLIC_KEY' in request.META:
+        try:
+            return decode_key(request.META['HTTP_CONCENT_REQUESTOR_PUBLIC_KEY'])
+        except TypeError:
```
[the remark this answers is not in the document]

concent_api/core/views.py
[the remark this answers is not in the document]

concent_api/utils/fields.py
rwrzesien: :+1:

concent_api/utils/fields.py
rwrzesien: Yes, without it it is throwing:
`django.utils.encoding.DjangoUnicodeDecodeError: 'utf-8' codec can't decode byte 0xed in position 1:`

concent_api/utils/fields.py
rwrzesien: It rather works like MoneyField, where you have general value stored with currency, and __amount which shows only float value.
In example:
provider_public_key gives b'\xe7M\xc8\x9d\xc2\x1di\x93\xdd\x19\xd1d\xea#i\xcb^\xfb\xc4\x89\xb7\xf8\xff\xc9_p\xae+\xa2t3\xfc\xf01\x05r\x8f\x96\x0b\x97\x16\xdc*SD^\x17\t(\xfc\xb8\x80\xb0\xc5=1\x99\x01\x92Z\xcf8z.'
provider_public_key_base64 gives '503IncIdaZPdGdFk6iNpy177xIm3+P/JX3CuK6J0M/zwMQVyj5YLlxbcKlNEXhcJKPy4gLDFPTGZAZJazzh6Lg=='

Purpose for this is that you can't query on binary data.

concent_api/core/tests/test_auth_report_computed_task.py
```diff
@@ -0,0 +1,898 @@
...
+(DIFFERENT_PROVIDER_PRIVATE_KEY, DIFFERENT_PROVIDER_PUBLIC_KEY) = generate_ecc_key_pair()
+(DIFFERENT_REQUESTOR_PRIVATE_KEY, DIFFERENT_REQUESTOR_PUBLIC_KEY) = generate_ecc_key_pair()
+
+
+@override_settings(
+    CONCENT_PRIVATE_KEY = CONCENT_PRIVATE_KEY,
+    CONCENT_PUBLIC_KEY  = CONCENT_PUBLIC_KEY,
+    CONCENT_MESSAGING_TIME = 10,  # seconds
+)
+class ReportComputedTaskIntegrationTest(TestCase):
```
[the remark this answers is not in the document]
rwrzesien: :+1:

concent_api/core/tests/test_integration_report_computed_task.py
```diff
@@ -55,6 +55,7 @@ def test_provider_forces_computed_task_report_and_concent_immediately_sends_reje
                 data                           = serialized_force_report_computed_task,
                 content_type                   = 'application/octet-stream',
                 HTTP_CONCENT_CLIENT_PUBLIC_KEY = b64encode(PROVIDER_PUBLIC_KEY).decode('ascii'),
+                HTTP_CONCENT_REQUESTOR_PUBLIC_KEY=b64encode(REQUESTOR_PUBLIC_KEY).decode('ascii'),
```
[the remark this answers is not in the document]
rwrzesien: :+1:

concent_api/utils/fields.py
rwrzesien: :+1:

concent_api/core/migrations/0004_messageauth.py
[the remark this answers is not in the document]
cameel: It's the same as with merge. You can have multiple separate lines of events and you have put them together in your head when you want to understand what happened. Or you can just rebase/regenerate the migration and have linear, easy to understand changes.

concent_api/core/migrations/0004_messageauth.py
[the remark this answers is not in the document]
cameel: I did regenerate it and it does not fix this error. I think it's an issue with the custom field.

concent_api/core/models.py
```diff
@@ -35,9 +51,21 @@ class ReceiveOutOfBandStatus(Model):
...
+    provider_public_key  = Base64Field()
+    requestor_public_key = Base64Field()
```
cameel: We should have validations for these fields. The fields should ensure that they get base64-encoded data. Unless you coded it so that you assign binary data to them - in that case the fields should only accept `bytes`.

rwrzesien:
@cameel Updated.

concent_api/core/models.py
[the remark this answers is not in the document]
cameel: Yeah. I asked Karol to refactor it at one point.

concent_api/core/tests/test_core_views.py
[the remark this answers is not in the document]
cameel: Why not? If your model does not pass validations there's no point starting the test.

concent_api/core/views.py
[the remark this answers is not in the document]
cameel: OK. But please always review your code if you have conflicts after rebasing. The thing that would make the reviews much, much easier would be if you could spot problems like this yourself even before you submit the code.

concent_api/core/views.py
cameel: Please don't catch `KeyError`. Check it with an `if`.

concent_api/core/views.py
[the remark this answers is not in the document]
cameel: It would not fail and that's the point. We want it to fail if it ever happens. It's a `filter()` so if there are two objects in the database it will return two.

Assertions aren't there for perfect code. They're there to catch bugs. And such a situation would be a bug that we want to detect.

concent_api/core/views.py
cameel: Do interactive rebase, mark the commit as `edit`, fix it and fix conflicts with other commits.

concent_api/core/tests/test_core_views.py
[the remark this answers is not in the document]

concent_api/core/views.py

concent_api/core/migrations/0004_messageauth.py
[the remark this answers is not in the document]
rwrzesien: I got it fixed, but it changes field related logic in views, I will update my comments for other issues too.

concent_api/core/models.py
rwrzesien: Please check updated code and decide if this comment is still valid.

concent_api/core/views.py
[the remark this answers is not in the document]
rwrzesien: Changed to use Message.objects.get and proper exceptions handling.

concent_api/core/views.py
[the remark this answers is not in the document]
rwrzesien: Changed to use Message.objects.get and proper exceptions handling.

concent_api/core/views.py
```diff
@@ -116,6 +165,11 @@ def send(_request, client_message):
                 golem_message,
                 message_timestamp,
             )
+            store_message_auth(
+                golem_message,
+                provider_public_key  = force_report_computed_task_from_database.last().auth.provider_public_key,
+                requestor_public_key = client_public_key,
+            )
```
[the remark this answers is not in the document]
rwrzesien: Changed to use Message.objects.get and proper exceptions handling.

rwrzesien:
@cameel I have updated branch and fixed all the issues I could.

concent_api/core/views.py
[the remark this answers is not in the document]
cameel: You don't need `get()` here because we want to first check if the object exists. An assertion would be enough. Please use them more.

concent_api/core/views.py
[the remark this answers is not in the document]
cameel: Not really. We don't use `get()` because it's not certain that there's exactly one object. There might be none.

> but how would you propose to handle MultipleObjectsReturned error

You should not handle it. It will never happen. But I want this to be stated explicitly and in a way that Python can check and enforce. That's why I want an `assert` and not an exception here. This is just a sanity check that runs only in development. It's like saying ''I know that this instruction can technically return two objects but I assert that a correctly written send() will prevent this from ever happening.".

By writing this explicitly with `assert` you let Python call you out on this if you're wrong. And it helps anyone analyzing the code know that we don't have to worry about this case.

concent_api/core/views.py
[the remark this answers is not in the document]
cameel: See my comment above. We don't need exception handling here. Just an `assert`.

rwrzesien:
@cameel Updated.

concent_api/core/views.py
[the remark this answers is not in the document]

concent_api/core/views.py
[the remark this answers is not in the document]

concent_api/core/views.py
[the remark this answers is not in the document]

concent_api/core/models.py
```diff
@@ -17,11 +24,23 @@ def __str__(self):
...
+        return self.filter(
+            Q(message__auth__provider_public_key  = client_public_key) |
+            Q(message__auth__requestor_public_key = client_public_key)
+        )
+
+
 class ReceiveStatus(Model):
     message     = ForeignKey(Message)
     timestamp   = DateTimeField()
     delivered   = BooleanField(default = False)
```
Gr33ny: test

rwrzesien:
@kbeker @Jakub89 Please do first round of review before I will pass it to @cameel .

rwrzesien:
@cameel Review done and approved by Karol and Kuba, so it is ready for you, rebased on latest master.

concent_api/core/models.py
```diff
@@ -43,3 +63,13 @@ def __str__(self):
...
+
+
+class MessageAuth(Model):
+    """
+    This class is used to store provider and requestor keys for message exchange related to given initial message.
+    """
+
+    message              = OneToOneField(Message, related_name = 'auth')
+    provider_public_key  = Base64Field()
+    requestor_public_key = Base64Field()
```
cameel: Please set max length for the text fields. We don't want anyone to feed us gigabytes of data in the key field. We do have validations against that in the views but having database constraints as the last line of defense is always a good idea.

concent_api/utils/fields.py
cameel: How will it behave if you assign data that can't be decoded with `b64decode()` to the `_base64` field? Are there any validations or checks against that?

concent_api/core/views.py
```diff
@@ -586,3 +719,20 @@ def get_file_status(file_transfer_token_from_database: message.concents.FileTran
         return False
     else:
         raise exceptions.UnexpectedResponse()
+
+
+def decode_key(key: str) -> bytes:
+    return b64decode(key.encode('ascii'), validate = True)
```
cameel: This could be moved to `utils`. We're decoding keys in a lot of places, e.g. in gatekeeper. Would be convenient to have it as a shortcut.

concent_api/core/views.py
```diff
@@ -586,3 +719,20 @@ def get_file_status(file_transfer_token_from_database: message.concents.FileTran
...
+
+def decode_client_public_key(request):
+    return decode_key(request.META['HTTP_CONCENT_CLIENT_PUBLIC_KEY'])
+
+
+def decode_other_party_public_key(request):
+    if 'HTTP_CONCENT_OTHER_PARTY_PUBLIC_KEY' not in request.META:
+        raise Http400('Missing Other-Party-Public-Key HTTP when expected.')
+    try:
+        return decode_key(request.META['HTTP_CONCENT_OTHER_PARTY_PUBLIC_KEY'])
```
cameel: This was supposed to be called `Concent-Requestor-Public-Key`.

concent_api/core/views.py
```diff
@@ -586,3 +719,20 @@ def get_file_status(file_transfer_token_from_database: message.concents.FileTran
...
+    return b64decode(key.encode('ascii'), validate = True)
+
+
+def decode_client_public_key(request):
+    return decode_key(request.META['HTTP_CONCENT_CLIENT_PUBLIC_KEY'])
+
+
+def decode_other_party_public_key(request):
+    if 'HTTP_CONCENT_OTHER_PARTY_PUBLIC_KEY' not in request.META:
+        raise Http400('Missing Other-Party-Public-Key HTTP when expected.')
```
cameel: `Concent-` is part of the header name.

concent_api/core/views.py
```diff
@@ -586,3 +719,20 @@ def get_file_status(file_transfer_token_from_database: message.concents.FileTran
...
+    try:
+        return decode_key(request.META['HTTP_CONCENT_OTHER_PARTY_PUBLIC_KEY'])
+    except binascii.Error:
+        raise Http400('The value in the Other-Party-Public-Key HTTP is not a valid base64-encoded value.')
```
cameel: `Concent-` is part of the header name.

concent_api/core/views.py
```diff
@@ -45,53 +47,86 @@ def send(_request, client_message):
...
             if file_uploaded:
-                return handle_receive_force_get_task_result_upload_for_requestor(decoded_message_data)
+                return handle_receive_force_get_task_result_upload_for_requestor(
+                    request, decoded_message_data, last_delivered_message_status
+                )
             else:
-                return handle_receive_force_get_task_result_failed(decoded_message_data)
+                return handle_receive_force_get_task_result_failed(
+                    request, decoded_message_data, last_delivered_message_status
+                )
```
cameel: Please fix your style here. One parameter per line.

concent_api/core/views.py
```diff
@@ -45,53 +47,86 @@ def send(_request, client_message):
...
+        if (
+            last_delivered_message_status.message.type                            == message.concents.ForceGetTaskResultUpload.TYPE and
+            last_delivered_message_status.message.auth.requestor_public_key_bytes == client_public_key
+        ):
             decoded_message_data = deserialize_message(last_delivered_message_status.message.data.tobytes())
             file_uploaded        = get_file_status(decoded_message_data.file_transfer_token)
             if file_uploaded:
-                return handle_receive_force_get_task_result_upload_for_requestor(decoded_message_data)
+                return handle_receive_force_get_task_result_upload_for_requestor(
+                    request, decoded_message_data, last_delivered_message_status
```
cameel: Please fix your style here. One parameter per line.

concent_api/core/views.py
```diff
@@ -45,53 +47,86 @@ def send(_request, client_message):
...
             return decoded_message_data
         return None
 
     if isinstance(decoded_message_data, message.concents.ForceGetTaskResult):
+        if client_public_key != last_undelivered_message_status.message.auth.provider_public_key_bytes:
+            return None
         set_message_as_delivered(last_undelivered_message_status)
-        return handle_receive_force_get_task_result_upload_for_provider(request, decoded_message_data)
+        return handle_receive_force_get_task_result_upload_for_provider(
+            request, decoded_message_data, last_undelivered_message_status
```
cameel: Please fix your style here. One parameter per line.

concent_api/core/views.py
```diff
@@ -586,3 +719,20 @@ def get_file_status(file_transfer_token_from_database: message.concents.FileTran
...
     else:
         raise exceptions.UnexpectedResponse()
+
+
+def decode_key(key: str) -> bytes:
+    return b64decode(key.encode('ascii'), validate = True)
+
+
+def decode_client_public_key(request):
+    return decode_key(request.META['HTTP_CONCENT_CLIENT_PUBLIC_KEY'])
```
cameel: You should add a comment or an `assert` to explain why you assume that the header is always present because from the declaration alone it's not obvious that you only use it in views that use `api_view` decorator.

concent_api/core/views.py
```diff
@@ -524,9 +639,19 @@ def validate_task_id(task_id):
...
 
 
-def store_message_and_message_status(golem_message_type: int, task_id: str, raw_golem_message: bytes, status = None, delivered: bool = False):
+def store_message_and_message_status(
+    golem_message_type:     int,
+    task_id,
+    raw_golem_message:      bytes,
+    provider_public_key:    str,
+    requestor_public_key:   str,
+    status                          = None,
```
cameel: Wouldn't using `type` as the type here work?

concent_api/core/views.py
```diff
@@ -524,9 +639,19 @@ def validate_task_id(task_id):
         raise Http400("task_id cannot be longer than {} chars.".format(MESSAGE_TASK_ID_MAX_LENGTH))
 
 
-def store_message_and_message_status(golem_message_type: int, task_id: str, raw_golem_message: bytes, status = None, delivered: bool = False):
+def store_message_and_message_status(
+    golem_message_type:     int,
+    task_id,
```
cameel: `task_id` is a string now.

concent_api/core/views.py
```diff
@@ -586,3 +719,20 @@ def get_file_status(file_transfer_token_from_database: message.concents.FileTran
...
+def decode_key(key: str) -> bytes:
+    return b64decode(key.encode('ascii'), validate = True)
+
+
+def decode_client_public_key(request):
+    return decode_key(request.META['HTTP_CONCENT_CLIENT_PUBLIC_KEY'])
+
+
+def decode_other_party_public_key(request):
+    if 'HTTP_CONCENT_OTHER_PARTY_PUBLIC_KEY' not in request.META:
```
cameel: This function should return `None` explicitly if the condition here is false.

concent_api/core/views.py
```diff
@@ -160,19 +199,34 @@ def handle_send_force_report_computed_task(client_message):
         client_message.TYPE,
         client_message.task_to_compute.compute_task_def['task_id'],
         client_message.serialize(),
-        status = ReceiveStatus
+        provider_public_key  = client_public_key,
+        requestor_public_key = other_party_public_key,
```
cameel: You're never checking if `other_party_public_key` is not `None`. There is an `assert` inside `store_message_and_message_status()` but assertions are not for data validation. It won't protect you in production where assertion are disabled. The client can always send you a message without this header.

concent_api/core/views.py
```diff
@@ -214,14 +280,23 @@ def handle_send_reject_report_computed_task(client_message):
...
+            provider_public_key  = force_report_computed_task_from_database.last().auth.provider_public_key_bytes,
+            requestor_public_key = client_public_key,
+            status               = ReceiveStatus
         )
-
         return HttpResponse("", status = 202)
 
     if current_time <= force_report_computed_task.task_to_compute.compute_task_def['deadline'] + settings.CONCENT_MESSAGING_TIME:
-        ack_message             = Message.objects.filter(task_id = client_message.cannot_compute_task.task_to_compute.compute_task_def['task_id'], type = message.AckReportComputedTask.TYPE)
-        previous_reject_message = Message.objects.filter(task_id = client_message.cannot_compute_task.task_to_compute.compute_task_def['task_id'], type = message.RejectReportComputedTask.TYPE)
```
cameel: This change does not make sense on its own. It should be in the same commit that adds the lines below.

concent_api/core/views.py
```diff
@@ -231,22 +306,27 @@ def handle_send_reject_report_computed_task(client_message):
...
+            requestor_public_key = client_public_key,
+            status               = ReceiveStatus
         )
         return HttpResponse("", status = 202)
     else:
         raise Http400("Time to acknowledge this task is already over.")
 
 
-def handle_send_force_get_task_result(client_message: message.concents.ForceGetTaskResult) -> message.concents:
+def handle_send_force_get_task_result(request, client_message: message.concents.ForceGetTaskResult) -> message.concents:
```
cameel: Please fix the type annotation while you're at it. I think I reported it at some point but it's still broken.

concent_api/core/views.py
```diff
@@ -273,7 +353,9 @@ def handle_send_force_get_task_result(client_message: message.concents.ForceGetT
             client_message.TYPE,
             client_message.report_computed_task.task_to_compute.compute_task_def['task_id'],
             client_message.serialize(),
-            status = ReceiveStatus,
+            provider_public_key  = other_party_public_key,
```
cameel: You're not checking if `other_party_public_key` is not `None`.

concent_api/core/views.py
```diff
@@ -311,10 +396,15 @@ def handle_receive_ack_from_force_report_computed_task(decoded_message):
     return ack_report_computed_task
 
 
-def handle_receive_force_get_task_result_upload_for_provider(request, decoded_message: message.concents.ForceGetTaskResult) -> message.concents.ForceGetTaskResult:
+def handle_receive_force_get_task_result_upload_for_provider(
+    request,
+    decoded_message:                message.concents.ForceGetTaskResult,
+    previous_message_from_database: Message
```
cameel: It's not a `Message`. It's a message status.

Also, this looks really brittle to me. It's easy to give the wrong message here. You should add some asserts. Is ` previous_message_from_database` the same message the `decoded_message` was decoded from? You should check that.

I think it would make more sense to just pass the key here, not the whole status object. Less things that can go out of sync that way. It's not really clear by just looking at the name what `previous_message_from_database` is supposed to be.

concent_api/core/views.py
```diff
@@ -342,33 +432,47 @@ def handle_receive_force_get_task_result_upload_for_provider(request, decoded_me
...
     )
     force_get_task_result_upload.sig = None
     return force_get_task_result_upload
 
 
-def handle_receive_force_get_task_result_failed(decoded_message: message.concents.ForceGetTaskResultUpload) -> message.concents.ForceGetTaskResultUpload:
+def handle_receive_force_get_task_result_failed(
+    request,
+    decoded_message:                message.concents.ForceGetTaskResultUpload,
+    previous_message_from_database: Message
```
cameel: It's not a `Message`. It's message status.

concent_api/core/views.py
```diff
@@ -342,33 +432,47 @@ def handle_receive_force_get_task_result_upload_for_provider(request, decoded_me
...
     )
     force_get_task_result_failed.sig = None
     return force_get_task_result_failed
 
 
-def handle_receive_force_get_task_result_upload_for_requestor(decoded_message: message.concents.ForceGetTaskResultUpload) -> message.concents.ForceGetTaskResultUpload:
+def handle_receive_force_get_task_result_upload_for_requestor(
+    request,
+    decoded_message:                message.concents.ForceGetTaskResultUpload,
+    previous_message_from_database: Message
```
cameel: It's not a `Message`. It's message status.

concent_api/core/views.py
```diff
@@ -380,16 +484,18 @@ def handle_receive_force_get_task_result_upload_for_requestor(decoded_message: m
...
 
     store_message_and_message_status(
-        force_get_task_result_upload.TYPE,
-        force_get_task_result_upload.force_get_task_result.report_computed_task.task_to_compute.compute_task_def['task_id'],
-        force_get_task_result_upload.serialize(),
-        status    = ReceiveStatus,
-        delivered = True,
+        decoded_message.TYPE,
+        decoded_message.force_get_task_result.report_computed_task.task_to_compute.compute_task_def['task_id'],
+        decoded_message.serialize(),
```
cameel: This looks like a bug. `decoded_message` already comes from the database and you're storing it again. Why? The code before this change was using `force_get_task_result_upload` here.

concent_api/core/views.py
```diff
@@ -538,6 +663,14 @@ def store_message_and_message_status(golem_message_type: int, task_id: str, raw_
     golem_message.full_clean()
     golem_message.save()
 
+    message_auth = MessageAuth(
+        message                    = golem_message,
+        provider_public_key_bytes  = provider_public_key,
+        requestor_public_key_bytes = requestor_public_key,
```
cameel: You're creating `MessageAuth` for every message. Why create a separate table then? Either reuse same requestor/provider key object for all messages in a single task or just move these fields to `Message`.

rwrzesien:
@cameel Please finish the review. We agreed on last retrospection that we will not split review into parts. There are probably a lot of things in last two commits with tests that can be done better, but we are planning a general refactor there. Current comments are not really a difficult to fix. Another thing is that most of the issues that you have added now are for the code that didn't changed since you last reviewed it, and they could be already done, so please do full review.

concent_api/core/models.py
rwrzesien: :+1:

concent_api/core/views.py
rwrzesien: Why ? There is nothing about it in our coding style guidelines.

concent_api/core/views.py
rwrzesien: I have asked you about this before I have started working on this branch. You have said two things - store separate object for each message and make it a separate model. Why do you want to change it now, at the end of work ?

concent_api/core/views.py

concent_api/core/views.py
rwrzesien: But it is not always requestor key, as for example in second use case.

concent_api/core/views.py

concent_api/core/views.py
rwrzesien: If condition here is false it will either return decoded key or raise exception, there is no other exit from this function.

concent_api/core/views.py

concent_api/core/views.py

concent_api/core/views.py

concent_api/core/views.py

concent_api/core/views.py
rwrzesien: I can't remember now, probably it was related with some failing test, but it works now on the old code so it looks like not necessary, I will remove it.

concent_api/core/views.py

concent_api/core/views.py

concent_api/core/views.py
rwrzesien: The request is used in some of the function for other things, so it is better to keep their signature consistent and pass request to every one that is using it.

concent_api/core/views.py
rwrzesien: base64.b64decode can't return None.

concent_api/core/views.py
rwrzesien: base64.b64decode doesn't return None.

concent_api/core/views.py

rwrzesien:
@cameel Most of the fixes added, rest is answered.

cameel:
> Please finish the review. We agreed on last retrospection that we will not split review into parts. There are probably a lot of things in last two commits with tests that can be done better, but we are planning a general refactor there. Current comments are not really a difficult to fix. Another thing is that most of the issues that you have added now are for the code that didn't changed since you last reviewed it, and they could be already done, so please do full review.

Sorry, I did not manage to review it all yesterday. It's a lot of changes (well over 2000 lines added). We agreed to clearly indicate if it's not a full review and I did. Sometimes it's not easy to do it all at once. I had the choice of giving you half and letting you fix what I found or keeping you waiting until it's all done.

Anyway, some of the issues that were there from the beginning are ones indicated in my previous review and they were not or not fully fixed.

concent_api/utils/fields.py
cameel: So what about this one?

concent_api/core/views.py
cameel: I'm not talking about `request`. I'm talking about `previous_message_from_database` which is only used to get the key out of it.

concent_api/core/views.py
cameel: But it does not come directly from `b64decode()`. It comes from `decode_other_party_public_key()`.

concent_api/core/views.py
cameel: But it does not come directly from `b64decode()`. It comes from `decode_other_party_public_key()`.

concent_api/core/views.py
cameel: OK. Sorry I missed that. It's OK then.

concent_api/core/views.py
cameel: I see that I misread the conditions in `decode_other_party_public_key()`. In that case it's OK.

concent_api/core/views.py
cameel: I see that I misread the conditions in `decode_other_party_public_key()`. In that case it's OK.

concent_api/core/views.py
cameel: Because that's more readable. It's easy to miss that this is more than one parameter. I actually thought it's a single on until I later went back to this line to check something and looked at it closer.

There are examples in our coding style that show each parameter on a separate line. There isn't one showing this case specifically but I honestly did not think that someone would do that :) I'll add it as an example.

concent_api/core/views.py
cameel: OK.

concent_api/utils/fields.py
```diff
@@ -0,0 +1,31 @@
+import base64
+
+from django.db import models
+
+
+class Base64Field(models.TextField):
```
cameel: Adding it here as a separate remark because I see that you missed my remark above:

> How will it behave if you assign data that can't be decoded with b64decode() to the _base64 field? Are there any validations or checks against that?

concent_api/core/views.py
rwrzesien: :+1: I thought that if it is not exceeding a line it is ok, as they are other examples of this in our code, but off course I will change it.

concent_api/core/views.py

concent_api/core/views.py

concent_api/utils/fields.py
rwrzesien: Yes, it will fail with TypeError

```
In [2]: from core.models import MessageAuth, Message

In [3]: m = Message.objects.first()

In [4]: m
Out[4]: <Message: Message #55, type:4009, 2018-02-06 17:36:05.778571+00:00>

In [5]: MessageAuth(message=m, provider_public_key_bytes='aa', requestor_public_key_bytes='bb')
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-5-3bc30003f61c> in <module>()
----> 1 MessageAuth(message=m, provider_public_key_bytes='aa', requestor_public_key_bytes='bb')

~/work/codepoets/golem/venv-concent/lib/python3.6/site-packages/django/db/models/base.py in __init__(self, *args, **kwargs)
    566                     if prop in property_names or opts.get_field(prop):
    567                         if kwargs[prop] is not _DEFERRED:
--> 568                             _setattr(self, prop, kwargs[prop])
    569                         del kwargs[prop]
    570                 except (AttributeError, FieldDoesNotExist):

~/work/codepoets/golem/concent/concent_api/utils/fields.py in set_data(self, obj, data)
     29 
     30     def set_data(self, obj, data):
---> 31         setattr(obj, self.name, base64.b64encode(data))

~/work/codepoets/golem/venv-concent/lib/python3.6/base64.py in b64encode(s, altchars)
     56     application to e.g. generate url or filesystem safe Base64 strings.
     57     """
---> 58     encoded = binascii.b2a_base64(s, newline=False)
     59     if altchars is not None:
     60         assert len(altchars) == 2, repr(altchars)

TypeError: a bytes-like object is required, not 'str'

In [6]: MessageAuth(message=m, provider_public_key_bytes=22, requestor_public_key_bytes=11)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-6-1e278134efb7> in <module>()
----> 1 MessageAuth(message=m, provider_public_key_bytes=22, requestor_public_key_bytes=11)

~/work/codepoets/golem/venv-concent/lib/python3.6/site-packages/django/db/models/base.py in __init__(self, *args, **kwargs)
    566                     if prop in property_names or opts.get_field(prop):
    567                         if kwargs[prop] is not _DEFERRED:
--> 568                             _setattr(self, prop, kwargs[prop])
    569                         del kwargs[prop]
    570                 except (AttributeError, FieldDoesNotExist):

~/work/codepoets/golem/concent/concent_api/utils/fields.py in set_data(self, obj, data)
     29 
     30     def set_data(self, obj, data):
---> 31         setattr(obj, self.name, base64.b64encode(data))

~/work/codepoets/golem/venv-concent/lib/python3.6/base64.py in b64encode(s, altchars)
     56     application to e.g. generate url or filesystem safe Base64 strings.
     57     """
---> 58     encoded = binascii.b2a_base64(s, newline=False)
     59     if altchars is not None:
     60         assert len(altchars) == 2, repr(altchars)

TypeError: a bytes-like object is required, not 'int'
```

concent_api/core/views.py
rwrzesien: I would like to keep it as a reminder that it require further investigation. It still mysterious to me why it doesn't create warning, even if I have added `return 11` at the end of this function.

concent_api/utils/fields.py
rwrzesien: It will fail, good catch. I will have to think about solving this tomorrow.

rwrzesien: :+1: I have added validation in field for this scenario.

concent_api/core/tests/test_core_views.py
```diff
@@ -295,11 +305,11 @@ def test_send_reject_message_save_as_receive_out_of_band_status(self):
             reverse('core:send'),
             data = dump(
                 self.reject_report_computed_task,
-                PROVIDER_PRIVATE_KEY,
+                REQUESTOR_PRIVATE_KEY,
```
cameel: Why is the key changing here?

concent_api/core/tests/test_auth_get_task_result.py
```diff
@@ -0,0 +1,964 @@
...
+from utils.testing_helpers  import generate_ecc_key_pair
+
+
+(CONCENT_PRIVATE_KEY, CONCENT_PUBLIC_KEY)                         = generate_ecc_key_pair()
+(DIFFERENT_PROVIDER_PRIVATE_KEY, DIFFERENT_PROVIDER_PUBLIC_KEY)   = generate_ecc_key_pair()
+(DIFFERENT_REQUESTOR_PRIVATE_KEY, DIFFERENT_REQUESTOR_PUBLIC_KEY) = generate_ecc_key_pair()
+
+
+def get_file_status_true_mock(_):
+    return True
```
cameel: Please name the parameter.

concent_api/core/tests/test_auth_get_task_result.py
```diff
@@ -0,0 +1,964 @@
...
+    return False
+
+
+@override_settings(
+    CONCENT_PRIVATE_KEY    = CONCENT_PRIVATE_KEY,
+    CONCENT_PUBLIC_KEY     = CONCENT_PUBLIC_KEY,
+    CONCENT_MESSAGING_TIME = 10,  # seconds
+    FORCE_ACCEPTANCE_TIME  = 10,  # seconds
+)
+class GetTaskResultIntegrationTest(ConcentIntegrationTestCase):
```
cameel: This test name is already used elsewhere.

concent_api/core/tests/test_auth_get_task_result.py
```diff
@@ -0,0 +1,964 @@
...
+        with freeze_time("2017-12-01 11:00:08"):
+            response = self.client.post(
+                reverse('core:send'),
+                data                                = original_serialized_force_get_task_result,
+                content_type                        = 'application/octet-stream',
+                HTTP_CONCENT_CLIENT_PUBLIC_KEY      = self._get_encoded_requestor_public_key(),
+                HTTP_CONCENT_OTHER_PARTY_PUBLIC_KEY = self._get_encoded_provider_public_key(),
+            )
+
+        assert response.status_code == 200
```
cameel: This is part of the test so it should be `assertEqual()`. If it's not (i.e. if you consider it a step that prepares data for the test and not the actual test) then checks below should use `assert` too.

concent_api/core/tests/test_auth_report_computed_task.py
```diff
@@ -0,0 +1,892 @@
...
+
+from core.models            import MessageAuth
+from utils.testing_helpers  import generate_ecc_key_pair
+
+
+(CONCENT_PRIVATE_KEY, CONCENT_PUBLIC_KEY)                         = generate_ecc_key_pair()
+(PROVIDER_PRIVATE_KEY, PROVIDER_PUBLIC_KEY)                       = generate_ecc_key_pair()
+(REQUESTOR_PRIVATE_KEY, REQUESTOR_PUBLIC_KEY)                     = generate_ecc_key_pair()
+(DIFFERENT_PROVIDER_PRIVATE_KEY, DIFFERENT_PROVIDER_PUBLIC_KEY)   = generate_ecc_key_pair()
+(DIFFERENT_REQUESTOR_PRIVATE_KEY, DIFFERENT_REQUESTOR_PUBLIC_KEY) = generate_ecc_key_pair()
```
cameel: This is not something to fix now but a remark to keep in mind in the future:

`CONCENT_PRIVATE_KEY` and `CONCENT_PUBLIC_KEY` were originally defined as global constants at the top of the file only because we needed to use them in `override_settings` on multiple test classes. The other keys should not be created here. I think I added them at the top in my original integration tests and that's why they continue to be added here but we really should start defing them inside `setUp()` in new tests (they should be named in lower case of course when they're moved).

concent_api/core/views.py
cameel: Please just create an issue. I don't think this is a good way to set a reminder :)

concent_api/core/views.py
cameel: I'm not sure how exactly I worded it but it looks like a misunderstanding. The whole purpose of creating a new model is to be able to create a relationship that's not 1:1. Otherwise we would not need a separate model at all.

We can leave it out of this pull request but it's really something that can't stay that way in the long term. I'm creating another issue for it.

concent_api/core/views.py
cameel: If there's something like this somewhere in our code I must have missed it.

concent_api/core/views.py
cameel: One more thing. If something like has to change during implementation please always drop a comment in the original issue so that we can discuss it. The pull request should implement exactly what the issue says.

cameel:
One last thing. I can't find the response to my question about what happens when you assign data that can't be decoded as `base64` field. But I see some responses in the e-mail notifications. Could you summarize what the status is on this?

concent_api/core/views.py
```diff
@@ -181,28 +241,40 @@ def handle_send_ack_report_computed_task(client_message):
...
 
-    force_report_computed_task_from_database = Message.objects.filter(task_id = client_message.cannot_compute_task.task_to_compute.compute_task_def['task_id'], type = message.ForceReportComputedTask.TYPE)
+    force_report_computed_task_from_database = Message.objects.filter(
+        task_id                    = client_message.cannot_compute_task.task_to_compute.compute_task_def['task_id'],
+        type                       = message.ForceReportComputedTask.TYPE,
+        auth__requestor_public_key = request.META['HTTP_CONCENT_CLIENT_PUBLIC_KEY']
+    )
 
     if not force_report_computed_task_from_database.exists():
         raise Http400("'ForceReportComputedTask' for this task has not been initiated yet. Can't accept your 'RejectReportComputedTask'.")
```
cameel: This message is now misleading. It can also mean that someone has sent a reject for someone else's task.

concent_api/core/views.py

concent_api/core/views.py
rwrzesien: :+1: Lets see how next level of authentication will be described.

concent_api/core/tests/test_auth_get_task_result.py
rwrzesien: :+1:

concent_api/core/tests/test_auth_get_task_result.py
rwrzesien: :+1:

concent_api/core/tests/test_auth_report_computed_task.py
rwrzesien: :+1: I generally looked for a way to move generating CONCENT_PRIVATE_KEY and CONCENT_PUBLIC_KEY to `setUpClass` to override settings there but it didn't worked out as I have expected. Please note that in `test_auth_get_task_result.py` this  provider and requestor keys are generated in setup, so I will move `DIFFERENT_` keys in that file there too.

cameel: Yeah, I saw that. But I'm only taking about keys other than `CONCENT_...` ones. I think that `CONCENT_...` keys should stay where the are. It's only the `PROVIDER_...` keys, `REQUESTOR_...` keys and others that should be moved.

concent_api/core/tests/test_auth_get_task_result.py
rwrzesien: :+1:

concent_api/core/tests/test_core_views.py
rwrzesien: Because it should be send as requestor. In previous step of this test provider is sending ForceReportComputedTask, it is requestor who must send RejectReportComputedTask.

cameel: So it was a bug in the test? Please put bugfixes like this in separate commits. It's independent from the feature so it could have gone to master even before we finished the review here.

concent_api/core/views.py
rwrzesien: :+1: Updated in both places.

rwrzesien:
@cameel Updated all current issues. Yes, there is something weird with how responses to comments are handled here on github.

About base64, there was a bug as you have spotted it, that when I wanted to create an object with non-decodable string passed to `provider_public_key` or `requestor_public_key`, I was able to, but then when I wanted to read value of `provider_public_key_bytes` I get binascii.Error. So I have added validation in that field which validates if it is decodable and raises ValidationError if not.

concent_api/core/tests/test_core_views.py
rwrzesien: Not really a bug, because with previous implementation it was ok to send it as anyone, now it is not, so it is related with current changes.

rwrzesien:
@cameel All should be updated by now.

PR #32, opened 2018-01-09
Chore store message types as ints

### LINKED ISSUE #20: Store message types as ints in the database
Currently the `Message` class stores message class name in its `type` field. It was meant to be only a temporary solution while we were working with JSON. Now that we're using Golem messages, the field should be changed to an `int` and store `Message.TYPE` value.

kbeker:
Finished chore store message types in database as ints and ready for review
Resolved #20 issue


kbeker:
Changes submitted

concent_api/core/views.py
```diff
@@ -56,9 +55,9 @@ def send(request, message):
         current_time = int(datetime.datetime.now().timestamp())
         if current_time <= message.task_to_compute.compute_task_def['deadline'] + settings.CONCENT_MESSAGING_TIME:
 
-            force_task_to_compute   = Message.objects.filter(task_id = message.task_to_compute.compute_task_def['task_id'], type = "ForceReportComputedTask")
-            previous_ack_message    = Message.objects.filter(task_id = message.task_to_compute.compute_task_def['task_id'], type = "AckReportComputedTask")
-            reject_message          = Message.objects.filter(task_id = message.task_to_compute.compute_task_def['task_id'], type = "RejectReportComputedTask")
+            force_task_to_compute   = Message.objects.filter(task_id = message.task_to_compute.compute_task_def['task_id'], type = ForceReportComputedTask().TYPE)
```
rwrzesien: @kbeker One more thing, you don't have to create instance of a class here, it will work if you do
`ForceReportComputedTask.TYPE` instead of `ForceReportComputedTask().TYPE`.

kbeker:
I've added data migrations and changes which were requested

concent_api/core/migrations/0003_auto_20180109_1436.py
```diff
@@ -0,0 +1,37 @@
...
+        ('core', '0002_auto_20180105_0945'),
+    ]
+
+    operations = [
+        migrations.RunPython(migrate_golem_messages_types_from_strings_to_ints),
+        migrations.AlterField(
+            model_name='message',
+            name='type',
+            field=models.PositiveSmallIntegerField(),
+        ),
```
cameel: Does this even work? Please test this migration.

Your migration reads a string from a database column then stores a number in the same column and saves. This cannot work. The column would have to magically change its type during the operation.

This cannot be done in less than three separate operations, four if the new column must have the same name as the old one:
- Add an integer column.
- Fill the new column with values based on the string column.
- Remove the string column.
- Rename the integer column so that has the name string column used to have.

concent_api/core/tests/test_core_views.py
```diff
@@ -63,37 +62,44 @@ def setUp(self):
...
+                REQUESTOR_PRIVATE_KEY,
+                CONCENT_PUBLIC_KEY),
+            content_type = 'application/octet-stream',
+            HTTP_CONCENT_CLIENT_PUBLIC_KEY = b64encode(REQUESTOR_PUBLIC_KEY).decode('ascii')
         )
 
         self.assertEqual(response.status_code, 202)
         self.assertEqual(len(Message.objects.all()),       1)
-        self.assertEqual(Message.objects.last().type,      "MessageForceReportComputedTask")
+        self.assertEqual(Message.objects.last().type,      ForceReportComputedTask().TYPE)
```
cameel: You missed this. It still creates an instance.

concent_api/core/tests/test_core_views.py
```diff
@@ -237,23 +255,23 @@ def test_send_reject_message_save_as_receive_out_of_band_status(self):
...
             reverse('core:send'),
-            data                           = dump(self.message_force_report_computed_task, PROVIDER_PRIVATE_KEY, settings.CONCENT_PUBLIC_KEY),
+            data                           = dump(self.correct_golem_data, PROVIDER_PRIVATE_KEY, CONCENT_PUBLIC_KEY),
             content_type                   = 'application/octet-stream',
             HTTP_CONCENT_CLIENT_PUBLIC_KEY = b64encode(PROVIDER_PUBLIC_KEY).decode('ascii')
         )
 
         self.assertEqual(force_response.status_code, 202)
-        self.assertEqual(Message.objects.last().type, 'MessageForceReportComputedTask')
+        self.assertEqual(Message.objects.last().type, ForceReportComputedTask().TYPE)
```
cameel: You missed this. It still creates an instance.

concent_api/core/tests/test_core_views.py
```diff
@@ -237,23 +255,23 @@ def test_send_reject_message_save_as_receive_out_of_band_status(self):
...
             reverse('core:send'),
-            data                           = dump(self.message_reject_report_computed_task, PROVIDER_PRIVATE_KEY, settings.CONCENT_PUBLIC_KEY),
+            data                           = dump(self.reject_report_computed_task, PROVIDER_PRIVATE_KEY, CONCENT_PUBLIC_KEY),
             content_type                   = 'application/octet-stream',
             HTTP_CONCENT_CLIENT_PUBLIC_KEY = b64encode(PROVIDER_PUBLIC_KEY).decode('ascii')
         )
 
         self.assertEqual(reject_response.status_code, 202)
-        self.assertEqual(Message.objects.last().type, 'MessageRejectReportComputedTask')
+        self.assertEqual(Message.objects.last().type, RejectReportComputedTask().TYPE)
```
cameel: You missed this. It still creates an instance.

concent_api/core/migrations/0003_auto_20180109_1436.py
```diff
@@ -0,0 +1,45 @@
...
+    Message = apps.get_model('core', 'Message')
+    for message in Message.objects.all():
+        if message.type == 'ForceReportComputedTask':
+            message.tmp_type = 4001
+        if message.type == 'AckReportComputedTask':
+            message.tmp_type = 4002
+        if message.type == 'RejectReportComputedTask':
+            message.tmp_type = 4003
+        if message.type == 'VerdictReportComputedTask':
+            message.tmp_type = 4004
```
cameel: You should use `elif` if the `if` does not terminate the loop/function. If you don't Python always has to check all the conditions for every item instead of stopping immediately after finding one that matches.

kbeker: If statements changed to elif. All fixUps squashed.

kbeker:
Commits `ae4d993` and `e1d4c5a` squashed

Issue #39, opened 2018-01-12
Force report computed task use case: ForceReportComputedTask handled instead of VerdictReportComputedTask in receive-out-of-band/


rwrzesien:
In `receive_out_of_band` there is no condition for case
```
if last_undelivered_receive_out_of_band_status is None:
    if last_undelivered_receive_status.type == 'RejectReportComputedTask':
```
which should return VerdictReportComputedTask. But there is for 
```
if last_undelivered_receive_status.type == 'ForceReportComputedTask':
```
which is not covered at the moment and looks like it is unnecessary.

But it looks like right code, which at this moment is also uncovered, is below after condition
```
if isinstance(decoded_last_task_message, RejectReportComputedTask):
    if decoded_last_task_message.reason == "deadline-exceeded":
```
So the best solution for this would be to get code after `decoded_last_task_message` check and replace with it check for `last_undelivered_receive_status.type == 'ForceReportComputedTask':` and replace in that check `ForceReportComputedTask` to `RejectReportComputedTask` and do some minor fixes for message creation.

Issue #38, opened 2018-01-12
Force report computed task use case: AckReportComputedTask created but not added to the receive queue


rwrzesien:
In `receive` when provider sends request and following conditions are met
```
if current_time <= decoded_message_from_database.task_to_compute.compute_task_def['deadline'] + 2 * settings.CONCENT_MESSAGING_TIME:
    if decoded_message_data.reason is not None and decoded_message_data.reason.value == "TASK_TIME_LIMIT_EXCEEDED":
```
he should receive AckReportComputedTask. However, he gets it but from different line. Probably the reason for it is missing store of ReceiveStatus in `send` after condition
```
if message.cannot_compute_task.reason == CannotComputeTask.REASON.WrongCTD:
```

Issue #37, opened 2018-01-12
Force report computed task use case: Unnecessary code at the end of receive() view


rwrzesien:
There is a a code in `receive` function that looks like it is not needed for anything in current flow. If yes, then it should be removed.
```
    if decoded_message_data.task_to_compute.compute_task_def['deadline'] + settings.CONCENT_MESSAGING_TIME <= current_time <= decoded_message_data.task_to_compute.deadline + 2 * settings.CONCENT_MESSAGING_TIME: 
        ack_report_computed_task = AckReportComputedTask( 
            task_id                 = decoded_message_data.task_id, 
            timestamp               = current_time, 
        ) 
        ack_report_computed_task.task_to_compute = decoded_message_data.task_to_compute 
        return ack_report_computed_task 
 
    if current_time <= decoded_message_data.message_task_to_compute.compute_task_def['deadline'] + settings.CONCENT_MESSAGING_TIME: 
        return decoded_message_data 
```

PR #42, opened 2018-01-16
Fix issues valid golem messages

### LINKED ISSUE #36: Force report computed task use case: Missing test for AckReportComputedTask response
There is an piece of code in `if` in `receive` function which is not currently tested. There should be a test which will cover it.
```
if last_undelivered_message_status.message.type == "ForceReportComputedTask":
    if decoded_message_data.task_to_compute.compute_task_def['deadline'] + settings.CONCENT_MESSAGING_TIME < current_time:
    ...
```

kbeker:
Resolved #36, #37, #38, #39, #43 issues. Ready for review.


concent_api/core/tests/test_core_views.py
```diff
@@ -424,3 +470,67 @@ def test_view_receive_out_of_band_return_http_204_if_no_messages_in_database(sel
 
         self.assertEqual(response.status_code, 204)
         self.assertEqual(response.content.decode(), '')
+
+    def test_two_receive_out_of_band_in_row(self):
```
rwrzesien: I am not sure if this test expectations are correct. It means that each time somebody (for now without key checking) hits `\receive_out_of_band` he will get VerdictReportComputedTask. @kbeker @cameel is that expected behavior ?

concent_api/core/views.py
```diff
@@ -261,123 +243,96 @@ def receive_out_of_band(request, _message):
...
+            raw_last_message_verdict,
+            None,
+            check_time = False,
+        )
+
+        last_undelivered_receive_out_of_band_status.delivered = True
+        last_undelivered_receive_out_of_band_status.full_clean()
+        last_undelivered_receive_out_of_band_status.save()
 
-    return None
```
rwrzesien: I think it should still `return None` at the very end of this function.

concent_api/core/views.py
```diff
@@ -261,123 +243,96 @@ def receive_out_of_band(request, _message):
...
             )
-            store_receive_out_of_band(golem_message, message_timestamp)
+            store_receive_out_of_band(
+                golem_message,
+                message_timestamp
+            )
             message_verdict.sig = None
             return message_verdict
+        return None
+    else:
```
rwrzesien: According to my previous comment about test I think this `else:` part can should be removed too.

concent_api/core/tests/test_core_views.py
cameel: The test is wrong and needs to be changed. `VerdictReportComputedTask` should be returned only once.

By the way the name of the test function does not follow our convention. The pattern is `def test_<what you're testing>_should_<what you expect it to do>():`

PR #48, opened 2018-01-18
Chore gatekeeper logging

### LINKED ISSUE #35: More logging in Gatekeeper
`gatekeeper` should log more messages so that we can better understand what it's doing when deployed to the cluster. Add logging in the following situations:
- DEBUG: Upload or download request has been received. Include the path, client's public key, subtask ID, operation. Maybe also more information from the token.
- INFO: 4xx errors. Include the path, client's key, subtask ID, error message.

Jakub89:
Resolves #35

### DIFF
--- a/concent_api/gatekeeper/utils.py
+++ b/concent_api/gatekeeper/utils.py
@@ -1,8 +1,13 @@
 from django.http import JsonResponse
 
 
-def gatekeeper_access_denied_response(message):
-    data = {'message': message}
+def gatekeeper_access_denied_response(message, path = None, subtask_id = None, client_key = None):
+    data = {
+        'message':         message,
+        'path_to_file':    path,
+        'subtask_id':      subtask_id,
+        'client_key':      client_key,
+    }
     response = JsonResponse(data, status = 401)
     response.WWW_Authenticate = 'Golem realm="Concent Storage"'
     return response
--- a/concent_api/gatekeeper/views.py
+++ b/concent_api/gatekeeper/views.py
@@ -25,10 +25,12 @@ logger = logging.getLogger(__name__)
 @csrf_exempt
 @require_POST
 def upload(request):
+    logger.debug("Upload request received.")
     if request.content_type != 'application/x-www-form-urlencoded':
         return gatekeeper_access_denied_response('Unsupported content type.')
 
-    response = parse_headers(request)
+    path_to_file = request.get_full_path().partition(reverse('gatekeeper:upload'))[2]
+    response = parse_headers(request, path_to_file)
     if response is not None:
         logger.warning(response.content.decode())
         return response
@@ -40,6 +42,7 @@ def upload(request):
 @csrf_exempt
 @require_safe
 def download(request):
+    logger.debug("Download request received.")
     # The client should not sent Content-Type header with GET requests.
     # FIXME: When running on `manage.py runserver` in development, empty or missing Concent-Type gets replaced
     # with text/plain. gunicorn does not do this. Looks like a bug to me. We'll let it pass for now sice we ignore
@@ -47,7 +50,8 @@ def download(request):
     if request.content_type != 'text/plain' and request.content_type != '':
         return gatekeeper_access_denied_response('Download request cannot have data in the body.')
 
-    response = parse_headers(request)
+    path_to_file = request.get_full_path().partition(reverse('gatekeeper:download'))[2]
+    response = parse_headers(request, path_to_file)
     if response is not None:
         logger.warning(response.content.decode())
         return response
@@ -56,7 +60,7 @@ def download(request):
     return JsonResponse({"message": "Request passed all download validations."}, status = 200)
 
 
-def parse_headers(request):
+def parse_headers(request, path_to_file):
     # Decode and check if request header contains a golem message:
     if 'HTTP_AUTHORIZATION' not in request.META:
         return gatekeeper_access_denied_response("Missing 'Authorization' header.")
@@ -90,7 +94,17 @@ def parse_headers(request):
 
     # Check if request header contains Concent-Client-Public-Key:
     if 'HTTP_CONCENT_CLIENT_PUBLIC_KEY' not in request.META:
-        return gatekeeper_access_denied_response('Missing Concent-Client-Public-Key header.')
+        return gatekeeper_access_denied_response('Missing Concent-Client-Public-Key header.', path_to_file, loaded_golem_message.subtask_id)
+    client_public_key = request.META['HTTP_CONCENT_CLIENT_PUBLIC_KEY']
+
+    logger.debug(
+        "Client wants to {} file '{}', with subtask_id '{}'. Client public key: '{}'.".format(
+            loaded_golem_message.operation,
+            path_to_file,
+            loaded_golem_message.subtask_id,
+            client_public_key
+        )
+    )
 
     # Check ConcentFileTransferToken each field:
     # -DEADLINE
@@ -128,10 +142,6 @@ def parse_headers(request):
     # -FILES
     if not all(isinstance(file, dict) for file in loaded_golem_message.files):
         return gatekeeper_access_denied_response('Wrong type of files variable.')
-    if request.method == 'POST':
-        path_to_file = request.get_full_path().partition(reverse('gatekeeper:upload'))[2]
-    if request.method == 'GET':
-        path_to_file = request.get_full_path().partition(reverse('gatekeeper:download'))[2]
     transfer_token_paths_to_files = []
     for file in loaded_golem_message.files:
         transfer_token_paths_to_files.append(file['path'])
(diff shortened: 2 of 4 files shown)


concent_api/gatekeeper/utils.py
```diff
@@ -1,8 +1,8 @@
 from django.http import JsonResponse
 
 
-def gatekeeper_access_denied_response(message):
-    data = {'message': message}
+def gatekeeper_access_denied_response(message, path = None, subtask_id = None, client_key = None):
+    data = {'message': message, 'path to file': path, 'subtask id': subtask_id, "client's key": client_key}
```
cameel: 1) Don't use spaces or apostrophes in key names in JSON responses.
2) If you have more than one key in a dict it's better to wrap it into multiple lines. Otherwise it's hard to read.

concent_api/gatekeeper/views.py
```diff
@@ -25,12 +25,14 @@
...
+    logger.debug("Upload request received.")
     if request.content_type != 'application/x-www-form-urlencoded':
         return gatekeeper_access_denied_response('Unsupported content type.')
 
-    response = parse_headers(request)
+    path_to_file = request.get_full_path().partition(reverse('gatekeeper:upload'))[2]
+    response = parse_headers(request, path_to_file)
     if response is not None:
-        logger.warning(response.content.decode())
+        logger.info(response.content.decode())
```
cameel: This change should be in a separate commit. The description of your commit does not mention it. Please don't mix in random changes to unrelated commits.

concent_api/gatekeeper/views.py
```diff
@@ -40,102 +42,107 @@ def upload(request):
...
-        return gatekeeper_access_denied_response("Token in the 'Authorization' header is not a valid Golem message.")
+        return gatekeeper_access_denied_response("Token in the 'Authorization' header is not a valid Golem message.", path_to_file)
 
     if loaded_golem_message is None:
-        return gatekeeper_access_denied_response('Undefined golem message.')
+        return gatekeeper_access_denied_response('Undefined golem message.', path_to_file)
     assert isinstance(loaded_golem_message, Message)
+    logger.debug(
+        "Message operation: {}, subtask_id: {}, path to file: '{}'".format(
+            loaded_golem_message.operation, loaded_golem_message.subtask_id, path_to_file
```
cameel: Each parameter should be on a separate line.

concent_api/gatekeeper/views.py
```diff
@@ -40,102 +42,107 @@ def upload(request):
...
+        )
+    )
 
     # Check if request header contains Concent-Client-Public-Key:
     if 'HTTP_CONCENT_CLIENT_PUBLIC_KEY' not in request.META:
-        return gatekeeper_access_denied_response('Missing Concent-Client-Public-Key header.')
+        return gatekeeper_access_denied_response('Missing Concent-Client-Public-Key header.', path_to_file, loaded_golem_message.subtask_id)
+    client_pub_key = request.META['HTTP_CONCENT_CLIENT_PUBLIC_KEY']
```
cameel: Use full words in names. `client_pub_key` -> `client_public_key`.

concent_api/gatekeeper/views.py
```diff
@@ -40,102 +42,107 @@ def upload(request):
...
+        return gatekeeper_access_denied_response('Wrong operation variable for this request method.', path_to_file, loaded_golem_message.subtask_id, client_pub_key)
 
     # -FILES
     if not all(isinstance(file, dict) for file in loaded_golem_message.files):
-        return gatekeeper_access_denied_response('Wrong type of files variable.')
-    if request.method == 'POST':
-        path_to_file = request.get_full_path().partition(reverse('gatekeeper:upload'))[2]
-    if request.method == 'GET':
-        path_to_file = request.get_full_path().partition(reverse('gatekeeper:download'))[2]
+        return gatekeeper_access_denied_response('Wrong type of files variable.', path_to_file, loaded_golem_message.subtask_id, client_pub_key)
```
cameel: I see that this code has been removed but there was a bug here. What if the method is neither `GET` nor `POST`. You should have either an `assert` stating that it's impossible or have code to handle other methods (possibly by returning an error).

concent_api/gatekeeper/utils.py
rwrzesien: Also, you are mixing "" with ''. Please choose one and use it consistently.

concent_api/gatekeeper/views.py
Jakub89: I separated this change to commit "Switch logger.warning with access_denied_response to legger.info"

Jakub89:
I've push all the fixups.

concent_api/gatekeeper/utils.py
cameel: I'm all for consistency but please keep in mind one exception: always use the double quotes when there are single quotes inside. So for example `"That's an error. Let's fix this 'issue'."` is better than `'That\'s an error. Let\'s fix this \'issue\'.` even if all the other strings around use single quotes. In general I think that it's a good idea to always use double quotes for messages because they often contain apostrophes.

Jakub89: Okey, I've changed names of all keys, so they don't have any spaces and apostrophes. They can all stay like this: '<key>'.

concent_api/requirements.lock
```diff
@@ -6,7 +6,7 @@ coincurve==6.0.0
 Django==1.11.9
 ethereum==1.6.1
 freezegun==0.3.9
--e git://github.com/golemfactory/golem-messages.git@5e86af055a9a1ff6358238377d1050e69d78e41c#egg=Golem_Messages
+-e git://github.com/golemfactory/golem-messages.git@217489bcc057feafc325ae6a401a836ddbf5f314#egg=Golem_Messages
 pbkdf2==1.3
```
Jakub89: wtf?

concent_api/gatekeeper/views.py
cameel: You should log only the error message here.
If you want to log the whole response, it should be at the debug level, in addition to the message. Normally we don't want  this level of detail.

concent_api/gatekeeper/views.py
```diff
@@ -40,102 +42,109 @@ def upload(request):
...
     # the body anyway and the check is mostly to inform the client about its mistake.
     if request.content_type != 'text/plain' and request.content_type != '':
         return gatekeeper_access_denied_response('Download request cannot have data in the body.')
 
-    response = parse_headers(request)
+    path_to_file = request.get_full_path().partition(reverse('gatekeeper:download'))[2]
+    response = parse_headers(request, path_to_file)
     if response is not None:
-        logger.warning(response.content.decode())
+        logger.info(response.content.decode())
```
cameel: You should log only the error message here.

concent_api/gatekeeper/views.py
cameel: This was supposed to be logged at DEBUG level.

concent_api/gatekeeper/views.py
cameel: This was supposed to be logged at DEBUG level.

concent_api/gatekeeper/views.py
```diff
@@ -40,102 +42,109 @@ def upload(request):
...
+            path_to_file
+        )
+    )
 
     # Check if request header contains Concent-Client-Public-Key:
     if 'HTTP_CONCENT_CLIENT_PUBLIC_KEY' not in request.META:
-        return gatekeeper_access_denied_response('Missing Concent-Client-Public-Key header.')
+        return gatekeeper_access_denied_response('Missing Concent-Client-Public-Key header.', path_to_file, loaded_golem_message.subtask_id)
+    client_public_key = request.META['HTTP_CONCENT_CLIENT_PUBLIC_KEY']
+    logger.debug("Concent client public key: {}".format(client_public_key))
```
cameel: The public key was supposed to be included in the messages about upload/download request, not separately. We want to be able to pull it out of message in Stackdriver and filter by it.

concent_api/gatekeeper/views.py
```diff
@@ -40,102 +42,109 @@ def upload(request):
...
     except AttributeError:
-        return gatekeeper_access_denied_response("Token in the 'Authorization' header is not a valid Golem message.")
+        return gatekeeper_access_denied_response("Token in the 'Authorization' header is not a valid Golem message.", path_to_file)
 
     if loaded_golem_message is None:
-        return gatekeeper_access_denied_response('Undefined golem message.')
+        return gatekeeper_access_denied_response('Undefined golem message.', path_to_file)
     assert isinstance(loaded_golem_message, Message)
+    logger.debug(
+        "Message operation: {}, subtask_id: {}, path to file: '{}'".format(
```
cameel: This information should be in the message about upload/download operation. Not in a separate message.

Jakub89:
Okey. I've refactor logger.info to contain only message about what happened. And I've merged debug loggers in to one readable message.

PR #49, opened 2018-01-18
Add missing test for AckReportComputedTask

### LINKED ISSUE #36: Force report computed task use case: Missing test for AckReportComputedTask response

kbeker:
Resolves #36 Added one missed test for AckReportComputedTask.

### DIFF
--- a/concent_api/core/tests/test_core_views.py
+++ b/concent_api/core/tests/test_core_views.py
@@ -355,6 +355,62 @@ class CoreViewReceiveTest(TestCase):
         self.assertEqual(response.content.decode(), '')
         assert len(ReceiveStatus.objects.filter(delivered=False)) == 0
 
+    def test_receive_should_return_ack_if_the_receive_queue_contains_only_force_report_and_its_past_deadline(self):
+        self.compute_task_def               = message.ComputeTaskDef()
+        self.compute_task_def['task_id']    = 2
+        self.compute_task_def['deadline']   = int(dateutil.parser.parse("2017-11-17 10:00:00").timestamp())
+
+        self.task_to_compute = message.TaskToCompute(
+            timestamp           = int(dateutil.parser.parse("2017-11-17 10:00:00").timestamp()),
+            compute_task_def    = self.compute_task_def,
+        )
+        self.force_golem_data = message.ForceReportComputedTask(
+            timestamp       = int(dateutil.parser.parse("2017-11-17 10:00:00").timestamp()),
+            task_to_compute = self.task_to_compute
+        )
+        message_timestamp = datetime.datetime.now(timezone.utc)
+        new_message = Message(
+            type        = self.force_golem_data.TYPE,
+            timestamp   = message_timestamp,
+            data        = self.force_golem_data.serialize(),
+            task_id     = self.task_to_compute.compute_task_def['task_id']  # pylint: disable=no-member
+        )
+        new_message.full_clean()
+        new_message.save()
+        new_message_status = ReceiveStatus(
+            message   = new_message,
+            timestamp = message_timestamp,
+            delivered = False
+        )
+        new_message_status.full_clean()
+        new_message_status.save()
+
+        with freeze_time("2017-11-17 12:00:00"):
+            response = self.client.post(
+                reverse('core:receive'),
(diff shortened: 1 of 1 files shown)


concent_api/core/tests/test_core_views.py
```diff
@@ -355,6 +355,56 @@ def test_receive_return_http_204_if_no_messages_in_database(self):
...
+            timestamp           = int(dateutil.parser.parse("2017-11-17 10:00:00").timestamp()),
+            compute_task_def    = self.compute_task_def,
+        )
+        self.force_golem_data = message.ForceReportComputedTask(
+            timestamp       = int(dateutil.parser.parse("2017-11-17 10:00:00").timestamp()),
+            task_to_compute = self.task_to_compute
+        )
+        message_timestamp = datetime.datetime.now(timezone.utc)
+        new_message = Message(
+            type        = self.force_golem_data.__class__.__name__,
```
cameel: #20 has been merged. Please change it to numeric `TYPE`.

concent_api/core/tests/test_core_views.py
```diff
@@ -355,6 +355,56 @@ def test_receive_return_http_204_if_no_messages_in_database(self):
         self.assertEqual(response.content.decode(), '')
         assert len(ReceiveStatus.objects.filter(delivered=False)) == 0
 
+    @freeze_time("2017-11-17 12:00:00")
+    def test_receive_should_get_ack_after_task_to_compute_is_not_after_deadline(self):
```
cameel: Please give it a more sensible name: `test_receive_should_return_ack_if_the_receive_queue_contains_only_force_report_and_its_past_deadline`.

The current name makes no sense because `receive()` does not get anything. It only returns a message.

concent_api/core/tests/test_core_views.py
```diff
@@ -355,6 +355,56 @@ def test_receive_return_http_204_if_no_messages_in_database(self):
         self.assertEqual(response.content.decode(), '')
         assert len(ReceiveStatus.objects.filter(delivered=False)) == 0
 
+    @freeze_time("2017-11-17 12:00:00")
```
cameel: Please move this inside the function. You only need to freeze time around the `client.post()` call and I think this makes the test easier to undrstand because you can read the timestamps in order.

concent_api/core/tests/test_core_views.py
```diff
@@ -355,6 +355,56 @@ def test_receive_return_http_204_if_no_messages_in_database(self):
...
+            response.content,
+            REQUESTOR_PRIVATE_KEY,
+            CONCENT_PUBLIC_KEY,
+            check_time = False,
+        )
+
+        self.assertEqual(response.status_code, 200)
+        self.assertIsInstance(decoded_ack_response, message.AckReportComputedTask)
+        self.assertEqual(decoded_ack_response.task_to_compute.compute_task_def['task_id'],  self.task_to_compute.compute_task_def['task_id'])   # pylint: disable=no-member
+        self.assertEqual(decoded_ack_response.task_to_compute.compute_task_def['deadline'], self.task_to_compute.compute_task_def['deadline'])  # pylint: disable=no-member
```
cameel: 1) You can compare `computed_task_def` as a whole. No need to compare the individual fields one by one.
2) You should actually check that the whole `TaskToCompute` message inside is unchanged and has the same signature.
3) You should compare `subtask_id` too.
4) You should check that the message in queue has been marked as delivered.
5) You should check that the timestamp of the ack message is within the allowed range. It should not be for example too far in the past because the client won't be able to decode it and will discard it instead.

kbeker:
Changes added. Ready for another review.

cameel:
Looks like the breakage reported by pylint was not from this branch but from #51 merged earlier. I have extracted the fix you added here into master.

I did not notice this. If you see that it's master and not your branch that is broken, please tell me. master should always pass tests and have no warnings.

PR #50, opened 2018-01-18
Add missing case for RejectReportComputedTask

### LINKED ISSUE #38: Force report computed task use case: AckReportComputedTask created but not added to the receive queue

### LINKED ISSUE #39: Force report computed task use case: ForceReportComputedTask handled instead of VerdictReportComputedTask in receive-out-of-band/

kbeker:
Resolves #38 Added missing store message in receive queue.
Resolves #39 Added missing handling message for RejectReportComputedTask in receive_out_of_band() view. 

These two issues have to be fix together becuase them depend on each other. All tests pass.

### DIFF
--- a/concent_api/core/views.py
+++ b/concent_api/core/views.py
@@ -257,7 +257,7 @@ def receive_out_of_band(_request, _message):
 
             message_verdict.force_report_computed_task = force_report_computed_task
             message_verdict.ack_report_computed_task   = decoded_ack_report_computed_task
-            message_verdict.sig = None
+
             (golem_message, message_timestamp) = store_message(
                 message_verdict.TYPE,
                 decoded_ack_report_computed_task.task_to_compute.compute_task_def['task_id'],
@@ -281,7 +281,6 @@ def receive_out_of_band(_request, _message):
             message_verdict.ack_report_computed_task    = ack_report_computed_task
             message_verdict.force_report_computed_task  = decoded_force_report_computed_task
 
-            message_verdict.sig = None
             (golem_message, message_timestamp) = store_message(
                 message_verdict.TYPE,
                 decoded_force_report_computed_task.task_to_compute.compute_task_def['task_id'],
@@ -302,7 +301,6 @@ def receive_out_of_band(_request, _message):
             message_verdict.ack_report_computed_task                 = message.AckReportComputedTask()
             message_verdict.ack_report_computed_task.task_to_compute = decoded_reject_report_computed_task.cannot_compute_task.task_to_compute
 
-            message_verdict.sig = None
             store_message(
                 message_verdict.TYPE,
                 decoded_reject_report_computed_task.cannot_compute_task.task_to_compute.compute_task_def['task_id'],
@@ -332,7 +330,6 @@ def receive_out_of_band(_request, _message):
             last_undelivered_receive_out_of_band_status.delivered = True
             last_undelivered_receive_out_of_band_status.full_clean()
             last_undelivered_receive_out_of_band_status.save()
-            message_verdict.sig = None
 
             (golem_message, message_timestamp) = store_message(
                 message_verdict.TYPE,
@@ -362,7 +359,6 @@ def receive_out_of_band(_request, _message):
             last_undelivered_receive_out_of_band_status.full_clean()
             last_undelivered_receive_out_of_band_status.save()
 
-            message_verdict.sig = None
             (golem_message, message_timestamp) = store_message(
                 message_verdict.TYPE,
                 decoded_last_task_message.message_cannot_compute_task.task_to_compute.compute_task_def['task_id'],


concent_api/core/views.py
```diff
@@ -310,6 +315,30 @@ def receive_out_of_band(request, _message):
             )
             message_verdict.sig = None
             return message_verdict
+        if last_undelivered_receive_status.type == 'RejectReportComputedTask':
```
cameel: #20 has been merged. Please update message types.

concent_api/core/views.py
```diff
@@ -94,11 +94,16 @@ def send(_request, client_message):
 
         assert force_report_computed_task.task_to_compute.compute_task_def['task_id'] == client_message.cannot_compute_task.task_to_compute.compute_task_def['task_id']
         if client_message.cannot_compute_task.reason == message.CannotComputeTask.REASON.WrongCTD:
-            store_message(
+            golem_message, message_timestamp = store_message(
```
cameel: Use parentheses around tuples:
``` python
(golem_message, message_timestamp) = store_message(
```

concent_api/core/views.py
```diff
@@ -310,6 +315,30 @@ def receive_out_of_band(request, _message):
             )
             message_verdict.sig = None
             return message_verdict
+        if last_undelivered_receive_status.type == 'RejectReportComputedTask':
+            serialized_reject_report_computed_task = last_undelivered_receive_status.data.tobytes()
+            try:
+                decoded_reject_report_computed_task = load(
```
cameel: Shouldn't this be `deserialize()` now?

concent_api/core/views.py
```diff
@@ -310,6 +315,30 @@ def receive_out_of_band(request, _message):
...
+            ack_report_computed_task.task_to_compute    = decoded_reject_report_computed_task.task_to_compute
+            message_verdict.ack_report_computed_task    = ack_report_computed_task
+
+            message_verdict.sig = None
+            store_message(
+                type(message_verdict).__name__,
+                decoded_reject_report_computed_task.cannot_compute_task.task_to_compute.compute_task_def['task_id'],
+                message_verdict.serialize()
+            )
+            message_verdict.sig = None
```
cameel: You're doing `message_verdict.sig = None` twice.

concent_api/core/views.py
```diff
@@ -310,6 +315,30 @@ def receive_out_of_band(request, _message):
...
+                    settings.CONCENT_PRIVATE_KEY,
+                    client_public_key,
+                    check_time = False
+                )
+            except InvalidSignature as exception:
+                return JsonResponse({'error': "Failed to decode ForceReportComputedTask. {}".format(exception)}, status = 400)
+
+            ack_report_computed_task                    = message.AckReportComputedTask()
+            ack_report_computed_task.task_to_compute    = decoded_reject_report_computed_task.task_to_compute
+            message_verdict.ack_report_computed_task    = ack_report_computed_task
```
cameel: You could simplify these three lines to:
``` python
message_verdict.ack_report_computed_task                 = message.AckReportComputedTask()
message_verdict.ack_report_computed_task.task_to_compute = decoded_reject_report_computed_task.task_to_compute
```

concent_api/core/views.py
```diff
@@ -310,6 +315,30 @@ def receive_out_of_band(request, _message):
...
+
+            message_verdict.sig = None
+            store_message(
```
cameel: You should not only store the verdict but also mark it as delivered.

concent_api/core/views.py
```diff
@@ -310,6 +315,30 @@ def receive_out_of_band(request, _message):
...
+                    check_time = False
+                )
+            except InvalidSignature as exception:
+                return JsonResponse({'error': "Failed to decode ForceReportComputedTask. {}".format(exception)}, status = 400)
+
+            ack_report_computed_task                    = message.AckReportComputedTask()
+            ack_report_computed_task.task_to_compute    = decoded_reject_report_computed_task.task_to_compute
+            message_verdict.ack_report_computed_task    = ack_report_computed_task
+
+            message_verdict.sig = None
```
cameel: Why are you removing the signature?

concent_api/core/views.py
kbeker: All changes about deserialization are in fix-incorrect-key-changes branch.

cameel: `fix-incorrect-key-changes` is now merged. So this needs to be changed now.

concent_api/core/views.py
```diff
@@ -347,13 +369,12 @@ def receive_out_of_band(_request, _message):
...
 
             message_verdict.sig = None
-            golem_message, message_timestamp = store_message(
+            (golem_message, message_timestamp) = store_message(
                 message_verdict.TYPE,
                 decoded_last_task_message.message_cannot_compute_task.task_to_compute.compute_task_def['task_id'],
                 message_verdict.serialize()
             )
             store_receive_out_of_band(golem_message, message_timestamp)
-            message_verdict.sig = None
```
cameel: OK. Now why are you **not** removing the sig here? I mean, I was just asking because it's a change out of nowhere. You should have such changes in separate commits and properly described so that we know exactly why we're changing it instead of changing the code randomly.

concent_api/core/views.py
```diff
@@ -413,5 +434,6 @@ def store_receive_out_of_band(golem_message, message_timestamp):
         message     = golem_message,
         timestamp   = message_timestamp
     )
+    receive_out_of_band_status.delivered = True
```
cameel: You could put it in the constructor above.

concent_api/core/views.py
cameel: This is still there and you did not answer my question.

concent_api/core/views.py
cameel: This still there.

concent_api/core/views.py
```diff
@@ -291,18 +296,35 @@ def receive_out_of_band(_request, _message):
...
+
+            message_verdict.ack_report_computed_task                 = message.AckReportComputedTask()
+            message_verdict.ack_report_computed_task.task_to_compute = decoded_reject_report_computed_task.task_to_compute
+
+            message_verdict.sig = None
+            store_message(
+                message_verdict.TYPE,
+                decoded_reject_report_computed_task.cannot_compute_task.task_to_compute.compute_task_def['task_id'],
+                message_verdict.serialize()
+            )
```
cameel: You're not marking the message as delivered.

cameel:
You don't have any tests for the code you've added. The tests pass before your fix so this means that our tests are incomplete. You should add at least one for each problem you're fixing. The tests should fail on master and pass on your branch.

Issue #44, opened 2018-01-17
receive/ and receive-out-of-band/ sometimes return HTTP 204 even if there's a message queued


cameel:
The current implementation of `receive/` and `receive-out-of-band/` views processes only one message at a time and returns `None` (which results in HTTP 204) if that message does not need to be returned to the client. This is not the correct behavior. If there's another message waiting in queue, that message should be returned instead. The views should only ever return HTTP 204 if there are no pending messages for the client.

Consider for example this piece of code: https://github.com/golemfactory/concent/blob/d5404685b4c8ac02924e62b1a3c54dbaf972e595/concent_api/core/views.py#L220-L227

If the client asks about the message outside of the right time window, the view returns `None`. It does not even check if there's another message waiting.

Another thing is that there should be no timeout checks in these views. If the message has been queued, it must be returned even it has already expired. The messages should never just disappear. Such behavior will make the views less predictable and harder to test reliably.

cameel:
@rwrzesien This one was pretty important. Could you make sure it's done and close it?

rwrzesien:
@cameel Yes, done in https://github.com/golemfactory/concent/pull/162.

PR #56, opened 2018-01-18
[fix] receive/ and receive-out-of-band/ sometimes return HTTP 204 e…

rwrzesien:
…ven if there's a message queued

Resolves https://github.com/golemfactory/concent/issues/44
https://www.pivotaltracker.com/story/show/154401776

### DIFF
--- a/concent_api/core/tests/test_core_views.py
+++ b/concent_api/core/tests/test_core_views.py
@@ -595,6 +595,8 @@ class CoreViewReceiveOutOfBandTest(TestCase):
         self.force_golem_data.task_to_compute = self.task_to_compute
         assert self.force_golem_data.timestamp == 1510912800
 
+    @freeze_time("2017-11-17 11:40:00")
+    def test_view_receive_out_of_band_should_accept_valid_message(self):
         message_timestamp   = datetime.datetime.now(timezone.utc)
         new_message         = Message(
             type        = self.force_golem_data.TYPE,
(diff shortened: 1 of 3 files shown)


concent_api/core/tests/test_core_views.py
```diff
@@ -379,6 +379,8 @@ def setUp(self):
         self.force_golem_data.task_to_compute = self.task_to_compute
         assert self.force_golem_data.timestamp == 1510912800
 
+    @freeze_time("2017-11-17 11:40:00")
+    def test_view_receive_out_of_band_should_accept_valid_message(self):
```
cameel: Where does this come from? Was the test broken? Please rebase the branch on master.

If you're fixing something unrelated to the issue then this should be a separate commit that explains what and why you're fixing.

rwrzesien: Please check the whole class, which have two tests of which one was broken, but it is related to what I am doing in current issue. There was message creation in setUp(), and second test name is `test_view_receive_out_of_band_return_http_204_if_no_messages_in_database`, which was wrong because test was passing and there was a message in database, so I have moved creation of it from setup to a first test which only needed it.

rwrzesien:
@cameel Updated.

rwrzesien:
@cameel Rebased on master, please also consider this PR as little refactoring.

rwrzesien:
@cameel I have squashed fixups, there were no conflicts.

rwrzesien:
Closing as discussed, general issue related with this case will be covered in different implementation.

PR #58, opened 2018-01-19
api_handling_helpers: Add handling for status code diffrent than 200

### LINKED ISSUE #46: api-test.py and api-integration-test.py do not handle errors properly
1) The script crashes if the response is not a serialized `Message` (`load()` returns `None` but the script tries to use it as if it were a `Message`).
2) JSON messages should be decoded and their content printed to the console.

kbeker:
Resolves #46


concent_api/api_testing_helpers.py
```diff
@@ -73,7 +74,10 @@ def api_request(host, endpoint, private_key, public_key, data = None, headers =
         )
         print('STATUS: {} {}'.format(response.status_code, http.client.responses[response.status_code]))
         print('MESSAGE:')
-        print_golem_message(decoded_response, private_key, public_key)
+        if response.status_code != 200:
```
cameel: You should be checking `Content-Type`.

concent_api/api_testing_helpers.py
```diff
@@ -73,7 +74,10 @@ def api_request(host, endpoint, private_key, public_key, data = None, headers =
         )
         print('STATUS: {} {}'.format(response.status_code, http.client.responses[response.status_code]))
         print('MESSAGE:')
-        print_golem_message(decoded_response, private_key, public_key)
+        if response.headers['Content-Type'] != 'application/octet-stream':
+            print(json.loads(response.content.decode())['error'])
```
cameel: This is still wrong. If the content type is HTML, it will try to decode it as if it were JSON.

kbeker: Change statement to check if this is application/json

concent_api/api_testing_helpers.py
```diff
@@ -73,7 +74,10 @@ def api_request(host, endpoint, private_key, public_key, data = None, headers =
         )
         print('STATUS: {} {}'.format(response.status_code, http.client.responses[response.status_code]))
         print('MESSAGE:')
-        print_golem_message(decoded_response, private_key, public_key)
+        if response.headers['Content-Type'] == 'application/json':
+            print(json.loads(response.content.decode())['error'])
```
cameel: OK. Now you're not handling `application/octet-stream` again :) We're going in circles here.

Please take a step back and think about what we want to achieve here: you're getting a response from the server. It might come from any endpoint: `send/` `receive/`, maybe another one someone adds in the future. The operation might fail (2xx), return a client error (4xx) or a server error (5xx). In case of a success you may get an empty response or a Golem message. You may also get something else if the server is broken and returns 200 with something unexpected. In case of a failure you're likely to get JSON but you can also easily get HTML from nginx if the server crashes completely. You can also get something completely different if the server is misconfigured. This is a test script after all so you cannot assume that the server is working properly.

How to efficiently handle all of this? 
- First assume that the server is giving you a valid response and try to decode it according to what it says. So if it says the content type is JSON, decode it as JSON. If it says `application/octet-steam`, assume a Golem message. If you support any other type, handle it too. If there's no content type, assume that the body is empty. In any other case don't decode and just print what you got, possibly saying that you got an unexpected content type.
- Second, if the decoding method that matches the content type fails or if you can't handle the decoded data (e.g. there's no content type but body is non-empty), catch the error and print an error message saying that you could not decode the response. Then just print the response as is so that the user of the script can try to debug the problem.

concent_api/api_testing_helpers.py
```diff
@@ -73,7 +74,15 @@ def api_request(host, endpoint, private_key, public_key, data = None, headers =
         )
         print('STATUS: {} {}'.format(response.status_code, http.client.responses[response.status_code]))
         print('MESSAGE:')
-        print_golem_message(decoded_response, private_key, public_key)
+        if response.headers['Content-Type'] == 'application/octet-stream':
+            print_golem_message(decoded_response, private_key, public_key)
```
cameel: You're not handling the case where `decoded_response` is `None`.

You should actually move the `load()` call here.

concent_api/api_testing_helpers.py
```diff
@@ -73,7 +74,15 @@ def api_request(host, endpoint, private_key, public_key, data = None, headers =
         )
```
cameel: You're not handling exceptions from `load()`.

concent_api/api_testing_helpers.py
```diff
@@ -73,7 +74,15 @@ def api_request(host, endpoint, private_key, public_key, data = None, headers =
...
-        print_golem_message(decoded_response, private_key, public_key)
+        if response.headers['Content-Type'] == 'application/octet-stream':
+            print_golem_message(decoded_response, private_key, public_key)
+        elif response.headers['Content-Type'] == 'application/json':
+            try:
+                print(json.loads(response.content.decode())['error'])
+            except json.decoder.JSONDecodeError:
+                print('RAW RESPONSE: {}'.format(response.text), "\nFail to decode response content")
+        else:
+            print('RAW RESPONSE: {}'.format(response.text), "\nUnexptected content-type of response message")
```
cameel: `Unexptected` -> `Unexpected`

concent_api/api_testing_helpers.py
```diff
@@ -73,7 +74,15 @@ def api_request(host, endpoint, private_key, public_key, data = None, headers =
         )
         print('STATUS: {} {}'.format(response.status_code, http.client.responses[response.status_code]))
         print('MESSAGE:')
-        print_golem_message(decoded_response, private_key, public_key)
+        if response.headers['Content-Type'] == 'application/octet-stream':
+            print_golem_message(decoded_response, private_key, public_key)
+        elif response.headers['Content-Type'] == 'application/json':
+            try:
+                print(json.loads(response.content.decode())['error'])
```
cameel: What happens if the message does not contain a field called `error`?

cameel: Also, you're running `decode()` on the content. What happens if `content` contains unicode characters?

Issue #55, opened 2018-01-18
Non-numeric task IDs crash Concent API


jiivan:
I've received undefined 500 response. Test transcript follows:
```
Prepared message: ForceReportComputedTask(timestamp=1516288707, encrypted=False, sig=None, slots=[['task_to_compute', TaskToCompute(timestamp=1516288707, encrypted=False, sig=None, slots=[['compute_task_def', {'task_id': 'a1', 'subtask_id': 'a1/1', 'deadline': 1516288717}]])], ['result_hash', 'sha1:6597eb6497f74f425c7ebacd712562cc9234697e']])
Sending to concent...
DEBUG    [golem.network.concent.client       ] send_to_concent(): Encrypting msg ForceReportComputedTask(timestamp=1516288707, encrypted=False, sig=None, slots=[['task_to_compute', TaskToCompute(timestamp=1516288707, encrypted=False, sig=None, slots=[['compute_task_def', {'task_id': 'a1', 'subtask_id': 'a1/1', 'deadline': 1516288717}]])], ['result_hash', 'sha1:6597eb6497f74f425c7ebacd712562cc9234697e']])
[...]

PR #59, opened 2018-01-22
[fix] non numeric task ids crash concent api

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/55
https://www.pivotaltracker.com/story/show/154444926


concent_api/core/migrations/0005_auto_20180122_0044.py
```diff
@@ -0,0 +1,20 @@
...
+    dependencies = [
+        ('core', '0004_auto_20180119_1055'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='message',
+            name='task_id',
+            field=models.CharField(max_length=255, null=True),
+        ),
```
cameel: This change requires a data migration to preserve existing data.

concent_api/core/models.py
```diff
@@ -11,7 +11,7 @@ class Message(Model):
     type        = PositiveSmallIntegerField()
     timestamp   = DateTimeField()
     data        = BinaryField()
-    task_id     = IntegerField(null = True)
+    task_id     = CharField(max_length = 255, null = True)
```
cameel: Could you check if Golem code has any limit on the length of task IDs?

I think that if we impose a limit here, we should have some kind of validation of that limit in golem-messages. And perhaps a constant that defined the maximum ID length.

concent_api/core/tests/test_core_views.py
```diff
@@ -129,6 +129,41 @@ def test_send_should_return_http_200_if_message_timeout(self):
         )
         self.assertIsInstance(response_message, GolemMessage)
 
+    @freeze_time("2017-11-17 10:00:00")
+    def test_send_should_accept_valid_message_with_non_numeric_task_id(self):
```
cameel: Could you extend this test to have a list of valid and invalid ids and check them in a loop?

concent_api/core/migrations/0005_auto_20180122_0044.py
rwrzesien: It doesn't, int will be casted to string without problems.

concent_api/core/models.py
rwrzesien: I have checked in docs and code and couldn't find any.

concent_api/core/tests/test_core_views.py
rwrzesien: What would be invalid id ?

concent_api/core/models.py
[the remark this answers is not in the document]
rwrzesien: How this can break support for non-numeric IDs. Have you meant numeric IDs ? Wasn't that the purpose of this change ?

rwrzesien:
> It seems to me that the the code is going to fail and return a 5xx error if you try to use a non-numeric ID. This is not acceptable. You should add a validation to the validating function in views. The view should return a 4xx error if a message is invalid.

There is a test verifying that, could you please provide an example of id that will break ?

cameel:
Try `1`.

rwrzesien:
> Try 1.
We have a lot of tests which use exactly this ID and they all pass.

concent_api/core/migrations/0005_auto_20180122_0044.py
cameel: Interesting. So PostgreSQL can do int <-> string conversion in simple cases. That's cool.

concent_api/core/models.py
cameel: OK. But we still need a validation.

concent_api/core/migrations/0005_auto_20180122_0044.py
rwrzesien: Int => string yes
String => int probably will ask about custom handling while migrating, if you don't provide any and there will be non-numeric string then it will fail.

concent_api/core/models.py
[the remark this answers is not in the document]
cameel: >  Have you meant numeric IDs ?

Yes, sorry for the confusion.

> Wasn't that the purpose of this change ?

The purpose was to add support for non-numeric IDs. Your change does this but at the same time breaks numeric IDs in a way.

Looks like Golem lets you use anything as ID. And I did not see any string conversions there so I suspect that to Golem ids like `666` and `'666'` are different things since that's how Python works.

After your change we "support" both of these in the sense that there's no exception. But Django converts anything you assign to a `CharField` to string and `filter()` will find `'666'` if you give it `666`. So we don't really support numeric IDs any longer. We just convert them to strings (which Golem does not do).

cameel:
> We have a lot of tests which use exactly this ID and they all pass.

OK, so it does not raise an exception. That's even worse because it's still broken - at least as long as we support both numeric and non-numeric IDs - because I think that to Golem `1` and `'1'` are different IDs.

I did not know that Django will convert anything you assign to a `CharField` to a string. My natural assumption was that you'd get a validation error or a type error on save. That's troubling and I think we'll have a lot of problems with silent conversions like this. Golem does not do these conversions and we can end up with slightly different data in messages and in the database. That can lead to bugs that are hard to find unless you're very careful.

I think that the only way out of this is to agree with Golem on a more strict schema for messages. At least that IDs must always be strings and nothing else. Then on our side validate this assumption (rather than blindly convert them). There should be a validation for this in golem-messages too.

concent_api/core/models.py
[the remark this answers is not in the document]
rwrzesien: Right, but I thought that is how we want it to work. So lets wait for Darek's response to use only strings which I think is best solution in that case, otherwise we would have to store extra info on the original type or use separate field for ints and strings. 
Btw. I don't see any casting of task_id in golem_messages, is there any option, apart from asking Darek, which you already did, to check how it works on Golem side ?

rwrzesien:
Ok, I have added validation which assumes only strings, however I will wait with adding tests until Darek will respond.

concent_api/core/tests/test_core_views.py
cameel: So to Golem there are no invalid IDs... that's bad because I think that using stuff like `[]`, `{}` `None`, `b'abc'`, `(1, 2, 3)`, `ForceReportComputedTask()` as IDs should not be allowed. It can easily lead to unexpected behavior if you don't carefully think about every possible corner case.

I think that you should at least add a test that checks these values. And you should test both `send/` and `receive/`.

rwrzesien: I will add them as soon as we clarify with Golem team to use only strings.

jiivan:
@rwrzesien @cameel I think the discussion got a little bit blurred. I've answered in https://github.com/golemfactory/concent/issues/55#issuecomment-359509983
It's a string and will be validated against 128char length in https://github.com/golemfactory/golem-messages/issues/99

rwrzesien:
@cameel I have added test for incorrect values, correct values are pretty much tested in other tests, which I have also needed to update. We don't currently do any validation in receive because we don't store any new message there.

concent_api/core/migrations/0005_auto_20180122_0044.py
```diff
@@ -0,0 +1,20 @@
...
+
+    dependencies = [
+        ('core', '0004_auto_20180119_1055'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='message',
+            name='task_id',
+            field=models.CharField(max_length=255, null=True),
```
cameel: So this should now be 128. Please regenerate the migration from scratch.

cameel: Also, string fields in Django should be non-nullable.

concent_api/concent_api/settings/base.py
```diff
@@ -209,3 +209,6 @@
 # - "Gatekeeper":  An internal helper that validates file transfer tokens.
 # - "admin-panel": Django admin panel that provides access to database content and service statistics.
 # CONCENT_FEATURES = []
+
+# Defines max length of task_id passed in Golem Messages.
+CONCENT_MAX_TASK_ID_LENGTH = 128
```
cameel: Oh, one more thing. This should not be a setting. We don't really support user changing this arbitrarily - we need it to match golem-messages. It should be a constant. Please put it in models.

concent_api/core/models.py
```diff
@@ -11,7 +12,7 @@ class Message(Model):
     type        = PositiveSmallIntegerField()
     timestamp   = DateTimeField()
     data        = BinaryField()
-    task_id     = IntegerField(null = True)
+    task_id     = CharField(max_length = settings.CONCENT_MAX_TASK_ID_LENGTH, null = True)
```
cameel: String fields should not be non-nullable.

rwrzesien:
@cameel Updated and rebased on master.

concent_api/core/models.py
```diff
@@ -1,17 +1,19 @@
...
 
+from .constants       import MESSAGE_TASK_ID_MAX_LENGTH
+
 
 class Message(Model):
     type        = PositiveSmallIntegerField()
     timestamp   = DateTimeField()
     data        = BinaryField()
-    task_id     = IntegerField(null = True)
+    task_id     = CharField(max_length = MESSAGE_TASK_ID_MAX_LENGTH, blank = True, default = '')
```
cameel: Please remove the default. The field should be assigned a value explicitly.

Also, now I see that this field should really be `blank = False` and `unique` too.

rwrzesien:
@cameel Updated and rebased on master.

concent_api/core/migrations/0005_auto_20180129_1639.py
```diff
@@ -0,0 +1,21 @@
...
+
+    dependencies = [
+        ('core', '0004_auto_20180119_1055'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='message',
+            name='task_id',
+            field=models.CharField(default='1517243982.234565', max_length=128),
```
cameel: WTF?

Issue #54, opened 2018-01-18
TaskToCompute.compute_task_def['deadline'] is float


jiivan:
In current golem core `ComputeTaskDef['deadline']` is `float`, but concent fails with:
```
Concent request failed with status 400 and response: '{"error": "Wrong type of deadline field."}'
```

Ideally validator should try to cast `deadline` to int and only raise exception if it fails.

cameel:
@jiivan OK. In general I prefer to avoid implicit conversions and be strict about types. Sometimes blindly converting stuff leads to a case where an obviously wrong value that should just make the code fail is converted and goes through silently still yielding wrong results. But in this particular case I don't see anything that could cause problems. `int()` supports only numbers and strings so we're covered.

One downside is that we'll have to remember to call `int()` every time we access a `timestamp` field which is error-prone. Maybe it would be a good idea to add a property on `Message` that performs the conversion internally?

jiivan:
@cameel _Be conservative in what you send, be liberal in what you accept_ :) Correct me if I'm wrong, but only use for this field is to compare it to current timestamp. It will work with any permutation of ints, floats and decimals. It will even work with `decimal.Decimal('infinity') :)

cameel:
But I did agree with you in my post above :) I meant that I don't see any problems with changing our code to do what you propose and we can modify it to add `int()` conversions. I just started by explaining that it's not my default approach because I prefer to have as many checks as possible to detect potential issues early. But conversions in this particular case seem harmless and even useful.

So what do you think about replacing `Message.timestamp` with a property and adding the conversion there? I agree that this coversion is a good idea but I think that requiring the user of the library to do this explicitly will lead to people forgetting about it.

jiivan:
`Message.timestamp` is int since https://github.com/golemfactory/golem-messages/commit/572b11ab87fc249901012ad4443251764dc80d82 but I believe it's outside of scope of this issue.

`ComputeTaskDef['deadline']` is float though. What I meant is that you don't even have to convert it to anything. You can just verify that it's some kind of a number. And I agree that verification should be a part of golem-messages.

cameel:
This is now fixed. It did not make it into v0.2 but will be included in the next (minor or major) release.

PR #60, opened 2018-01-22
Added positive int validator and tests

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/54
https://www.pivotaltracker.com/story/show/154471001

### DIFF
--- a/concent_api/core/tests/test_core_views.py
+++ b/concent_api/core/tests/test_core_views.py
@@ -327,6 +327,93 @@ class CoreViewSendTest(TestCase):
         self.assertEqual(response.status_code, 400)
         self.assertIn('error', response.json().keys())
 
+    @freeze_time("2017-11-17 10:00:00")
+    def test_send_should_return_http_400_if_task_to_compute_deadline_is_not_an_integer(self):
+        compute_task_def = message.ComputeTaskDef()
+        compute_task_def['task_id'] = '8'
+
+        invalid_values = [
+            -11,
+            'a11b',
+            {},
+            [],
+            (1, 2, 3),
+            None,
+        ]
+
+        for deadline in invalid_values:
+            Message.objects.all().delete()
+            compute_task_def['deadline'] = deadline
+            task_to_compute = message.TaskToCompute(
+                timestamp        = int(datetime.datetime.now().timestamp()),
+                compute_task_def = compute_task_def,
+            )
+
+            serialized_task_to_compute   = dump(task_to_compute,             REQUESTOR_PRIVATE_KEY,   PROVIDER_PUBLIC_KEY)
+            deserialized_task_to_compute = load(serialized_task_to_compute,  PROVIDER_PRIVATE_KEY,  REQUESTOR_PUBLIC_KEY, check_time = False)
+
+            with freeze_time("2017-11-17 10:00:00"):
+                force_report_computed_task = message.ForceReportComputedTask(
+                    timestamp = int(datetime.datetime.now().timestamp()),
+                )
+                force_report_computed_task.task_to_compute = deserialized_task_to_compute
+
+                response_400 = self.client.post(
+                    reverse('core:send'),
+                    data = dump(
+                        force_report_computed_task,
+                        PROVIDER_PRIVATE_KEY,
+                        CONCENT_PUBLIC_KEY
+                    ),
+                    content_type                   = 'application/octet-stream',
+                    HTTP_CONCENT_CLIENT_PUBLIC_KEY = b64encode(PROVIDER_PUBLIC_KEY).decode('ascii'),
+                )
+
+            self.assertEqual(response_400.status_code, 400)
+            self.assertIn('error', response_400.json().keys())
+
+    def test_send_should_return_http_202_if_task_to_compute_deadline_is_correct(self):
+        compute_task_def = message.ComputeTaskDef()
+        compute_task_def['task_id'] = '8'
+
+        invalid_values = [
+            11,
+            '1112',
+            True,
+            0,
+            False,
+        ]
+
+        for deadline in invalid_values:
+            Message.objects.all().delete()
+            compute_task_def['deadline'] = deadline
+            task_to_compute = message.TaskToCompute(
+                timestamp        = int(datetime.datetime.now().timestamp()),
+                compute_task_def = compute_task_def,
+            )
+
+            serialized_task_to_compute   = dump(task_to_compute,             REQUESTOR_PRIVATE_KEY,   PROVIDER_PUBLIC_KEY)
+            deserialized_task_to_compute = load(serialized_task_to_compute,  PROVIDER_PRIVATE_KEY,  REQUESTOR_PUBLIC_KEY, check_time = False)
+
+            with freeze_time("2017-11-17 10:00:00"):
+                force_report_computed_task = message.ForceReportComputedTask(
+                    timestamp = int(datetime.datetime.now().timestamp()),
+                )
+                force_report_computed_task.task_to_compute = deserialized_task_to_compute
+
+                response_202 = self.client.post(
+                    reverse('core:send'),
+                    data = dump(
+                        force_report_computed_task,
+                        PROVIDER_PRIVATE_KEY,
+                        CONCENT_PUBLIC_KEY
+                    ),
+                    content_type                   = 'application/octet-stream',
+                    HTTP_CONCENT_CLIENT_PUBLIC_KEY = b64encode(PROVIDER_PUBLIC_KEY).decode('ascii'),
+                )
+
+            self.assertIn(response_202.status_code, [200, 202])
+
 
 @override_settings(
     CONCENT_PRIVATE_KEY    = CONCENT_PRIVATE_KEY,
--- a/concent_api/core/views.py
+++ b/concent_api/core/views.py
@@ -455,8 +455,7 @@ def validate_golem_message_task_to_compute(data):
     if data.compute_task_def['task_id'] == '':
         raise Http400("task_id cannot be blank.")
 
-    if not isinstance(data.compute_task_def['deadline'], int):
-        raise Http400("Wrong type of deadline field.")
+    data.compute_task_def['deadline'] = validate_int_value(data.compute_task_def['deadline'])
 
 
 def validate_golem_message_reject(data):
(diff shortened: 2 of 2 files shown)


concent_api/core/tests/test_core_views.py
```diff
@@ -286,6 +286,70 @@ def test_send_reject_message_save_as_receive_out_of_band_status(self):
...
+        )
+
+        self.assertEqual(response_400.status_code, 400)
+        self.assertIn('error', response_400.json().keys())
+
+    @freeze_time("2017-11-17 10:00:00")
+    def test_send_should_return_http_400_if_task_to_compute_deadline_cant_be_casted_to_int(self):
+        compute_task_def = message.ComputeTaskDef()
+        compute_task_def['task_id'] = 8
+        compute_task_def['deadline'] = 'a11b'
```
cameel: You're missing a test with the timestamp in the future.

Actually, these tests are almost identical. Maybe replace them with one that has lists of valid and invalid values and checks them in a loop. We'll be able to test more values with less code this way.

concent_api/core/tests/test_core_views.py
```diff
@@ -286,6 +286,70 @@ def test_send_reject_message_save_as_receive_out_of_band_status(self):
         self.assertEqual(reject_response.status_code, 202)
         self.assertEqual(Message.objects.last().type, message.RejectReportComputedTask.TYPE)
 
+    @freeze_time("2017-11-17 10:00:00")
```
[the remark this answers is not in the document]

concent_api/core/tests/test_core_views.py
```diff
@@ -286,6 +286,70 @@ def test_send_reject_message_save_as_receive_out_of_band_status(self):
...
+    def test_send_should_return_http_400_if_task_to_compute_deadline_is_negative_integer(self):
+        compute_task_def = message.ComputeTaskDef()
+        compute_task_def['task_id'] = 8
```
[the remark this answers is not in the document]

concent_api/core/tests/test_core_views.py
```diff
@@ -286,6 +286,70 @@ def test_send_reject_message_save_as_receive_out_of_band_status(self):
...
+        task_to_compute = message.TaskToCompute(
+            timestamp = self.message_timestamp,
+            compute_task_def = self.compute_task_def,
+        )
+
+        serialized_task_to_compute      = dump(task_to_compute,             REQUESTOR_PRIVATE_KEY,   PROVIDER_PUBLIC_KEY)
+        deserialized_task_to_compute    = load(serialized_task_to_compute,  PROVIDER_PRIVATE_KEY,  REQUESTOR_PUBLIC_KEY, check_time = False)
+
+        ack_report_computed_task = message.AckReportComputedTask(
+            timestamp = self.message_timestamp,
```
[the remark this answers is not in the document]
rwrzesien: :+1: But personally I think it doesn't make any sense.

concent_api/core/tests/test_core_views.py
```diff
@@ -286,6 +286,70 @@ def test_send_reject_message_save_as_receive_out_of_band_status(self):
...
+            ),
+            content_type                    = 'application/octet-stream',
+            HTTP_CONCENT_CLIENT_PUBLIC_KEY  = b64encode(PROVIDER_PUBLIC_KEY).decode('ascii'),
+        )
+
+        self.assertEqual(response_400.status_code, 400)
+        self.assertIn('error', response_400.json().keys())
+
+    @freeze_time("2017-11-17 10:00:00")
+    def test_send_should_return_http_400_if_task_to_compute_deadline_cant_be_casted_to_int(self):
```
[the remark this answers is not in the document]

concent_api/core/views.py
```diff
@@ -370,8 +393,23 @@ def validate_golem_message_reject(data):
...
             raise Http400("task_id cannot be blank.")
 
-        if not isinstance(data.compute_task_def['deadline'], int):
+        data.compute_task_def['deadline'] = validate_int_value(data.compute_task_def['deadline'])
+
+
+def validate_int_value(value):
+    """
+    Checks if value is integer, if not it tries to cast it as integer.
+    Then it checks if value is positive integer.
```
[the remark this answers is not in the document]

concent_api/core/views.py
```diff
@@ -370,8 +393,23 @@ def validate_golem_message_reject(data):
...
+
+    """
+    if not isinstance(value, int):
+        try:
+            value = int(value)
+        except (ValueError, TypeError):
             raise Http400("Wrong type of deadline field.")
```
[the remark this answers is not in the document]

concent_api/core/tests/test_core_views.py

rwrzesien:
@cameel Updated.

concent_api/core/tests/test_core_views.py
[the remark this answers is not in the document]
cameel: Why? I'm not strongly opposed to doing it like this. And if there's more code to move to `setUp()` I'm all for it. But in this particular case moving just the timestamp to `setUp()` strikes me as something that only makes it hard to see what the timestamp is.

It would be fine if you moved the whole `AckReportComputed` task to `setUp()`. But why just the timestamp? It's not some generic constant that we can use for all messages. Whether the timestamp is correct or not depends on the other timestaps in the test and they're defined here and not in `setUp()`.

concent_api/core/tests/test_core_views.py
cameel: Looks like I got this one confused with #61. What I meant was to add more invalid values. Like dicts, objects, `None`, etc. I remember that I listed them in a comment in this pull request but I can't find that comment anywhere.

concent_api/core/views.py
```diff
@@ -370,8 +393,23 @@ def validate_golem_message_reject(data):
...
+    """
+    Checks if value is integer, if not it tries to cast it as integer.
+    Then it checks if value is non-negative integer.
+
+    """
+    if not isinstance(value, int):
+        try:
+            value = int(value)
+        except (ValueError, TypeError):
+            raise Http400("Wrong type of deadline field, expected integer or value that can be casted to integer.")
```
cameel: The field is not necessarily called `deadline`. This might be another integer field

concent_api/core/tests/test_core_views.py
```diff
@@ -286,6 +286,40 @@ def test_send_reject_message_save_as_receive_out_of_band_status(self):
         self.assertEqual(reject_response.status_code, 202)
         self.assertEqual(Message.objects.last().type, message.RejectReportComputedTask.TYPE)
 
+    def test_send_should_return_http_400_if_task_to_compute_deadline_is_wrong(self):
+        compute_task_def = message.ComputeTaskDef()
+        compute_task_def['task_id'] = '8'
+
+        for deadline in [-11, 'a11b', int(dateutil.parser.parse("2017-11-17 11:00:00").timestamp())]:
```
cameel: This test is broken. `int(dateutil.parser.parse("2017-11-17 11:00:00").timestamp())` is a valid deadline but it does not fail.

You should test both valid and invalid values.

concent_api/core/tests/test_core_views.py
```diff
@@ -286,6 +286,40 @@ def test_send_reject_message_save_as_receive_out_of_band_status(self):
...
+            task_to_compute = message.TaskToCompute(
+                timestamp = int(datetime.datetime.now().timestamp()),
+                compute_task_def = self.compute_task_def,
+            )
+
+            serialized_task_to_compute      = dump(task_to_compute,             REQUESTOR_PRIVATE_KEY,   PROVIDER_PUBLIC_KEY)
+            deserialized_task_to_compute    = load(serialized_task_to_compute,  PROVIDER_PRIVATE_KEY,  REQUESTOR_PUBLIC_KEY, check_time = False)
+
+            ack_report_computed_task = message.AckReportComputedTask(
+                timestamp = int(datetime.datetime.now().timestamp()),
```
cameel: Don't use `datetime.now()` outside of `freeze_time()`. If you need it here, make the `with freeze_time()` block a little bigger. It should cover everything that depends on current time.

rwrzesien:
@cameel Those two last deadlines are ok, but we are sending AckReportComputedTask without creating ForceReportComputedTask for it, so it returns 400 from:

```raise Http400("'ForceReportComputedTask' for this task has not been initiated yet. Can't accept your 'AckReportComputedTask'.")```

Test for correct deadlines is added. Btw. something must have changed in golem messages, because now giving negative timestamp changes it into now() during serialization.

File: concent_api/core/tests/test_core_views.py [hunk 11]
```diff
@@ -286,6 +286,93 @@ def test_send_reject_message_save_as_receive_out_of_band_status(self):
...
+            Message.objects.all().delete()
+            task_to_compute = message.TaskToCompute(
+                timestamp = int(datetime.datetime.now().timestamp()),
+                compute_task_def = self.compute_task_def,
+            )
+
+            serialized_task_to_compute      = dump(task_to_compute,             REQUESTOR_PRIVATE_KEY,   PROVIDER_PUBLIC_KEY)
+            deserialized_task_to_compute    = load(serialized_task_to_compute,  PROVIDER_PRIVATE_KEY,  REQUESTOR_PUBLIC_KEY, check_time = False)
+
+            deserialized_task_to_compute.compute_task_def['deadline'] = deadline
```
cameel: Why change this after serialization/deserialization and not in the original `task_to_compute`? You now have `deserialized_task_to_compute` that's different than the original `task_to_compute`. Fortunately you're not using it for anything in this test but if you were modify the test later, this can only lead to problems.

concent_api/core/tests/test_core_views.py
```diff
@@ -286,6 +286,93 @@ def test_send_reject_message_save_as_receive_out_of_band_status(self):
...
+            True,
+            0,
+        ]
+
+        for deadline in invalid_values:
+            Message.objects.all().delete()
+            compute_task_def['deadline'] = deadline
+            task_to_compute = message.TaskToCompute(
+                timestamp = int(datetime.datetime.now().timestamp()),
+                compute_task_def = self.compute_task_def,
```
cameel: You're putting the deadline in `compute_task_def` but using `self.compute_task_def`. I think this is why the test is passing with every value.

cameel:
> @cameel Those two last deadlines are ok, but we are sending AckReportComputedTask without creating ForceReportComputedTask for it, so it returns 400 from:

I think there's a problem in the test too. See my comments above. You're setting the deadline in one task def but using another to create the message.

> Btw. something must have changed in golem messages, because now giving negative timestamp changes it into now() during serialization.

If that's the case, please report an error in golem-messages. We really don't want it to do implicit conversions like this. If the timestamp is wrong, it's wrong. It should result in an error and the class should not second guess you and correct it silently because it can lead to bugs remaining undetected.

concent_api/core/tests/test_core_views.py
rwrzesien: The issue for that is what I have described
> Btw. something must have changed in golem messages, because now giving negative timestamp changes it into now() during serialization.

So when I set it before serializing, it gets changed and that would not really test the case. So I can either move it up front and assume the correctness because of serialization, or keep it that way until golem team will resolve bug.

cameel: Oh, OK. But add at least a FIXME and a comment explaining this.

rwrzesien: Ok, there is no bug, it was just my mistake to use self.compute_task_def.

rwrzesien:
@cameel There is no bug, it was my mistake from using self.compute_task_def, I have updated the tests and rebased on latest master.

Issue #45, opened 2018-01-17
Message timestamp checks in send/


cameel:
In #31 all timestamp checks in our endpoints have been disabled. This is not completely correct. We do need to disable timestamp checks when deserializing messages that have been archived in the database or embedded in other messages (e.g. for `TaskToCompute` nested in `ForceReportComputedTask`) but not for messages which are sent directly to our `send/` endpoint. These messages should be dropped if they're too old or have a timestamp too far into the future. Use the same rules and constants as golem-messages. Preferably just drop the `check_time=False` parameter where needed.

Add tests that check both cases - that old and future messages are rejected and current messages are processed.

rwrzesien:
@cameel I think we can do even better - there is a function `verify_time` in golem_messages which does exactly that. It raises golem_messages custom exceptions which I can catch and handle on our side.

rwrzesien:
@cameel While working on this I have discovered that in few of our tests we reach send() with client_message as None. Is that correct ? If not, I think we should create separate issue.

cameel:
You mean that `load()` returns `None`? I don't think that's correct.

I think we should do what I suggested to Karol long time ago but he did not do it: after every `load()` call where you expect that the result is not `None`, you should do `assert message is not None` to document that fact and let Python notify you about it by raising an error if it's not the case.

PR #61, opened 2018-01-23
Ignoring messages with timestamps in the past or in the future

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/45
https://www.pivotaltracker.com/story/show/154403513

### DIFF
--- a/concent_api/core/tests/test_core_views.py
+++ b/concent_api/core/tests/test_core_views.py
@@ -10,6 +10,7 @@ from django.http                    import HttpResponse
 from django.http                    import JsonResponse
 from django.utils                   import timezone
 
+from golem_messages                 import settings
 from golem_messages.shortcuts       import dump
 from golem_messages.shortcuts       import load
 from golem_messages                 import message
@@ -285,6 +286,46 @@ class CoreViewSendTest(TestCase):
 
         self.assertEqual(reject_response.status_code, 202)
         self.assertEqual(Message.objects.last().type, message.RejectReportComputedTask.TYPE)
+
+    def test_send_should_reject_message_when_timestamp_too_old(self):
+        with freeze_time("2017-11-17 10:00:00"):
+            timestamp = dateutil.parser.parse("2017-11-17 09:40:00")
+            assert datetime.datetime.now() - timestamp > settings.MSG_TTL
+            response = self.client.post(
+                reverse('core:send'),
+                data = dump(
+                    message.Ping(
+                        timestamp=int(timestamp.timestamp())
+                    ),
+                    PROVIDER_PRIVATE_KEY,
+                    CONCENT_PUBLIC_KEY
+                ),
+                content_type                   = 'application/octet-stream',
+                HTTP_CONCENT_CLIENT_PUBLIC_KEY = b64encode(PROVIDER_PUBLIC_KEY).decode('ascii')
+            )
+
+        self.assertEqual(response.status_code, 400)
+        self.assertIn('error', response.json().keys())
+
+    def test_send_should_reject_message_when_timestamp_too_far_in_future(self):
+        with freeze_time("2017-11-17 10:00:00"):
+            timestamp = dateutil.parser.parse("2017-11-17 10:10:00")
+            assert timestamp - datetime.datetime.now() > settings.FUTURE_TIME_TOLERANCE
+            response = self.client.post(
+                reverse('core:send'),
+                data = dump(
+                    message.Ping(
+                        timestamp=int(dateutil.parser.parse("2017-11-17 10:10:00").timestamp())
+                    ),
+                    PROVIDER_PRIVATE_KEY,
+                    CONCENT_PUBLIC_KEY
+                ),
+                content_type                   = 'application/octet-stream',
+                HTTP_CONCENT_CLIENT_PUBLIC_KEY = b64encode(PROVIDER_PUBLIC_KEY).decode('ascii')
+            )
+
+        self.assertEqual(response.status_code, 400)
+        self.assertIn('error', response.json().keys())
 
 
 @override_settings(
--- a/concent_api/core/views.py
+++ b/concent_api/core/views.py
@@ -6,6 +6,10 @@ from django.utils                   import timezone
 from django.conf                    import settings
 
 from golem_messages                 import message
+from golem_messages.message.base    import verify_time
+from golem_messages.exceptions      import MessageFromFutureError
+from golem_messages.exceptions      import MessageTooOldError
+from golem_messages.exceptions      import TimestampError
 
 from utils.api_view                 import api_view
 from utils.api_view                 import Http400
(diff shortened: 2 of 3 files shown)


concent_api/core/views.py
```diff
@@ -17,6 +21,8 @@
 @api_view
 @require_POST
 def send(_request, client_message):
+    if client_message:
```
cameel: Client message is not a boolean value.

concent_api/core/tests/test_core_views.py
```diff
@@ -286,6 +286,40 @@ def test_send_reject_message_save_as_receive_out_of_band_status(self):
         self.assertEqual(reject_response.status_code, 202)
         self.assertEqual(Message.objects.last().type, message.RejectReportComputedTask.TYPE)
 
+    @freeze_time("2017-11-17 10:00:00")
+    def test_send_reject_message_when_timestamp_too_old(self):
```
cameel: You're missing a `should` here.

concent_api/core/tests/test_core_views.py
```diff
@@ -286,6 +286,40 @@ def test_send_reject_message_save_as_receive_out_of_band_status(self):
...
+        response = self.client.post(
+            reverse('core:send'),
+            data = dump(
+                message.Ping(
+                    timestamp=int(dateutil.parser.parse("2017-11-17 09:40:00").timestamp())
+                ),
+                PROVIDER_PRIVATE_KEY,
+                CONCENT_PUBLIC_KEY),
```
cameel: Please fix your style here.

concent_api/core/tests/test_core_views.py
```diff
@@ -286,6 +286,40 @@ def test_send_reject_message_save_as_receive_out_of_band_status(self):
         self.assertEqual(reject_response.status_code, 202)
         self.assertEqual(Message.objects.last().type, message.RejectReportComputedTask.TYPE)
 
+    @freeze_time("2017-11-17 10:00:00")
```
cameel: Please use `with freeze_time()` around the tested call instead of wrapping the whole function.

concent_api/core/tests/test_core_views.py
```diff
@@ -286,6 +286,40 @@ def test_send_reject_message_save_as_receive_out_of_band_status(self):
...
+    def test_send_reject_message_when_timestamp_too_old(self):
+        response = self.client.post(
+            reverse('core:send'),
+            data = dump(
+                message.Ping(
+                    timestamp=int(dateutil.parser.parse("2017-11-17 09:40:00").timestamp())
```
cameel: You're missing an `assert` here. You should assert that the difference between the current time and your timestamp is greater than `MSG_TTL` defined in golem-messages. Currently it is but if this ever gets changed our tests will silently just stop working.

concent_api/core/tests/test_core_views.py
```diff
@@ -286,6 +286,40 @@ def test_send_reject_message_save_as_receive_out_of_band_status(self):
...
+        self.assertEqual(response.status_code, 400)
+        self.assertIn('error', response.json().keys())
+
+    @freeze_time("2017-11-17 10:00:00")
+    def test_send_reject_message_when_timestamp_too_far_in_future(self):
+        response = self.client.post(
+            reverse('core:send'),
+            data = dump(
+                message.Ping(
+                    timestamp=int(dateutil.parser.parse("2017-11-17 10:10:00").timestamp())
```
cameel: You're missing an `assert` here. You should assert that the difference between the current time and your timestamp is greater than `FUTURE_TIME_TOLERANCE` defined in golem-messages. Currently it is but if this ever gets changed our tests will silently just stop working.

concent_api/core/tests/test_core_views.py
```diff
@@ -286,6 +286,40 @@ def test_send_reject_message_save_as_receive_out_of_band_status(self):
...
+                CONCENT_PUBLIC_KEY),
+            content_type                   = 'application/octet-stream',
+            HTTP_CONCENT_CLIENT_PUBLIC_KEY = b64encode(PROVIDER_PUBLIC_KEY).decode('ascii')
+        )
+
+        self.assertEqual(response.status_code, 400)
+        self.assertIn('error', response.json().keys())
+
+    @freeze_time("2017-11-17 10:00:00")
+    def test_send_reject_message_when_timestamp_too_far_in_future(self):
```
cameel: See my remarks for the other test. They mostly apply to this one too.

concent_api/core/views.py
```diff
@@ -374,6 +403,13 @@ def validate_golem_message_reject(data):
             raise Http400("Wrong type of deadline field.")
 
 
+def validate_golem_message_timestamp(timestamp):
+    try:
+        verify_time(timestamp)
+    except (MessageFromFutureError, MessageTooOldError, TimestampError):
+        raise Http400('Message timestamp is wrong.')
```
cameel: Please return the original exception message instead of replacing it with a generic one that does not even say what's wrong with the timestamp.

concent_api/core/views.py
cameel: Also, `client_message` should never be `None` here (please add an `assert` that checks that). If `api_view` gets `None` after decoding a message, it should return an error. But this is not related to this branch. Please report and fix it separately.

concent_api/core/tests/test_core_views.py

concent_api/core/tests/test_core_views.py

concent_api/core/views.py
rwrzesien: Yes, lets make a separate issue out of it. Generally I think it should never hit the view and fail directly in api_views.py. I will work on that later.

concent_api/core/views.py

concent_api/core/tests/test_core_views.py

concent_api/core/tests/test_core_views.py

concent_api/core/tests/test_core_views.py

concent_api/core/tests/test_core_views.py

rwrzesien:
@cameel Updated.

concent_api/core/views.py
```diff
@@ -374,6 +403,17 @@ def validate_golem_message_reject(data):
...
 
+def validate_golem_message_timestamp(timestamp):
+    try:
+        verify_time(timestamp)
+    except MessageFromFutureError:
+        raise Http400('Message timestamp too much in the future.')
+    except MessageTooOldError:
+        raise Http400('Message timestamp too old.')
+    except TimestampError as e:
+        raise Http400(e)
```
cameel: Please don't use single-letter names for your variables.

concent_api/core/views.py
```diff
@@ -82,19 +88,24 @@ def send(_request, client_message):
         force_report_computed_task = message.Message.deserialize(
             force_report_computed_task_from_database.last().data.tobytes(),
             None,
-            check_time = False
```
cameel: Actually, now that I think about it this line is wrong.

You're removing a line that ignores the timestamp on a message deserialized from the database. We **do** want to ignore the timestamp here.

What you should be doing instead is to remove `check_time` in `api_view` decorator which decoded the outer message for the view.

concent_api/core/views.py

rwrzesien:
@cameel Yes, it works, and leads to message being None, for which we have separate issue. Tests I have written for this issue covers this case.

PR #65, opened 2018-01-24
Feature force get task results

rwrzesien:


concent_api/utils/api_view.py
```diff
@@ -38,7 +38,7 @@ def wrapper(request, *args, **kwargs):
         if request.content_type not in ['application/octet-stream', '', 'application/json']:
             return JsonResponse({'error': 'Concent supports only application/octet-stream and application/json.'}, status = 415)
 
-        if len(request.body) == 0:
+        if len(request.body) == 0 or request.body == b'{}':
```
cameel: Why is this change back? I thought we resolved that earlier. `{}` is not an empty message.

requirements-development.txt
```diff
@@ -9,6 +9,7 @@ pylint-django
 requests
 mypy==0.560
 coverage
+mock==2.0.0
```
cameel: Versions should not be hard-coded in the `.txt` file unless you know with certainty that no other version works.

By the way, please also fix the `mypy==0.560` line above. I thought we fixed it in #11?

concent_api/requirements.lock
```diff
@@ -18,6 +18,7 @@ python-dateutil==2.6.1
 pytz==2017.3
 PyYAML==3.12
 repoze.lru==0.7
+requests==2.18.4
```
cameel: Don't edit `requirements.lock` manually. Recreate it with `pip freeze` in a fresh virtualenv. You're missing the dependencies of `requests` here (`certifi`, `urllib3`, etc.)

concent_api/core/views.py
```diff
@@ -25,25 +29,34 @@ def send(_request, client_message):
...
         last_delivered_message_status = ReceiveStatus.objects.all().order_by('timestamp').last()
         if last_delivered_message_status is None:
             return None
-
         if last_delivered_message_status.message.type == message.ForceReportComputedTask.TYPE:
             return handle_receive_delivered_force_report_computed_task(last_delivered_message_status)
+        if last_delivered_message_status.message.type == message.concents.ForceGetTaskResultUpload.TYPE:
+            decoded_message_data = deserialize_message(last_delivered_message_status.message.data.tobytes())
+            file_status          = check_file_status(decoded_message_data.file_transfer_token)
+            if file_status is True:
```
cameel: Don't compare explicitly with `True`. It should be `if file_status` because `file_status` is a boolean.

By the way, the name is bad because it does not sound like it's a boolean value. It should be something like `all_files_uploaded`.

concent_api/core/views.py
```diff
@@ -346,21 +466,30 @@ def store_message_and_message_status(golem_message_type, task_id, raw_golem_mess
...
+            'Concent-Client-Public-Key':    b64encode(settings.CONCENT_PUBLIC_KEY).decode(),
+    }
+    request_http_address = settings.STORAGE_CLUSTER_ADDRESS + 'download/' + file_transfer_token.files[0]['path']
+
+    cluster_storage_response = requests.head(
+        request_http_address,
+        headers = headers
+    )
+
+    return bool(cluster_storage_response.status_code == 200)
```
cameel: This conversion is not necessary.

concent_api/core/views.py
```diff
@@ -346,21 +466,30 @@ def store_message_and_message_status(golem_message_type, task_id, raw_golem_mess
...
+
+
+def check_file_status(file_transfer_token):
+    file_transfer_token.operation = 'download'
+    dumped_file_transfer_token = shortcuts.dump(file_transfer_token, settings.CONCENT_PRIVATE_KEY, settings.CONCENT_PUBLIC_KEY)
+    headers = {
+            'Authorization': 'Golem ' + b64encode(dumped_file_transfer_token).decode(),
+            'Concent-Client-Public-Key':    b64encode(settings.CONCENT_PUBLIC_KEY).decode(),
+    }
+    request_http_address = settings.STORAGE_CLUSTER_ADDRESS + 'download/' + file_transfer_token.files[0]['path']
```
cameel: How do you know there's only one file in `files`?

Either fix it or add `assert len(file_transfer_token.files) == 1`. Otherwise the other files are silently ignored.

concent_api/core/views.py
```diff
@@ -346,21 +466,30 @@ def store_message_and_message_status(golem_message_type, task_id, raw_golem_mess
...
-            delivered   = True
-        )
-        receive_out_of_band_status.full_clean()
-        receive_out_of_band_status.save()
-    else:
-        return None
+    return None
+
+
+def check_file_status(file_transfer_token):
```
cameel: Please add a type annotation here that `file_transfer_token` is of type `FileTransferToken`.

cameel:
Please add type annotations (parameter types and return type) for all functions in `views`. We should start using them and this would be a good, easy first step.

requirements-development.txt
```diff
@@ -9,6 +9,7 @@ pylint-django
 requests
 mypy==0.560
 coverage
+mock
```
cameel: This should go to `requirements.txt`. You should be able to run unit tests even without stuff from `requirements-development.txt`. That's why for example `freezegun` is in `requirements.txt` and not here.

concent_api/requirements.lock
```diff
@@ -1,12 +1,17 @@
...
 cbor2==3.0.4
-cffi==1.11.2
+certifi==2018.1.18
+cffi==1.11.4
+chardet==3.0.4
 coincurve==6.0.0
 Django==1.11.9
 ethereum==1.6.1
+factory-boy==2.9.2
+Faker==0.8.10
```
cameel: I don't think that `requests` depends on `factory-boy` or `Faker`. This should not be here. Please regenerate this file in a clean virtuelenv.

concent_api/core/views.py
```diff
@@ -237,6 +289,83 @@ def handle_receive_ack_from_force_report_computed_task(decoded_message):
...
+        timestamp                       = current_time,
+        token_expiration_deadline       = current_time + settings.TOKEN_EXPIRATION_DEADLINE,
+        storage_cluster_address         = settings.STORAGE_CLUSTER_ADDRESS,
+        authorized_client_public_key    = request.META['HTTP_CONCENT_CLIENT_PUBLIC_KEY'],
+        operation                       = 'upload',
+    )
+
+    task_id     = decoded_message.report_computed_task.task_to_compute.compute_task_def['task_id']
+    part_id     = 0
+    file_path   = '{}/{}/result'.format(str(task_id), str(part_id))
```
cameel: These `str()` conversions should not be needed here.

concent_api/core/tests/utils.py
```diff
@@ -0,0 +1,65 @@
...
+from django.test            import TestCase
+
+from golem_messages         import dump
+from golem_messages         import message
+import dateutil.parser
+
+from utils.testing_helpers  import generate_ecc_key_pair
+
+
+class ConcentIntegrationTestCase(TestCase):
```
cameel: This should be in a separate commit. Currently it's hidden in a commit that adds a bunch of tests.

concent_api/core/tests/test_integration_get_task_result.py
```diff
@@ -0,0 +1,601 @@
...
+                HTTP_CONCENT_CLIENT_PUBLIC_KEY = self._get_encoded_requestor_public_key()
+            )
+
+        self.assertEqual(response.status_code,  200)
+
+        message_from_concent = load(response.content, self.REQUESTOR_PRIVATE_KEY, CONCENT_PUBLIC_KEY)
+
+        self.assertIsInstance(message_from_concent, message.ForceGetTaskResultRejected)
+        self.assertEqual(message_from_concent.timestamp, self._parse_iso_date_to_timestamp("2017-12-01 11:00:11"))
+        self.assertEqual(message_from_concent.reason, "ACCEPTANCE_TIME_LIMIT_EXCEEDED")
```
cameel: Don't hard-code strings here. Use values from enumeration.

concent_api/core/tests/test_integration_get_task_result.py
```diff
@@ -0,0 +1,601 @@
...
+
+    def test_requestor_forces_get_task_result_and_concent_immediately_sends_rejection_due_to_already_sent_message(self):
+        """
+        Tests if on requestor ForceGetTaskResult message Concent will return ForceGetTaskResultRejected
+        if ForceGetTaskResult was already sent by this requestor.
+
+        Expected message exchange:
+        Requestor -> Concent:    ForceGetTaskResult
+        Requestor -> Concent:    ForceGetTaskResult
```
cameel: This is done ony once, not twice.

concent_api/core/tests/test_integration_get_task_result.py
```diff
@@ -0,0 +1,601 @@
...
+        Concent   -> Requestor:  ForceGetTaskResultRejected
+        """
+
+        # STEP 1: Requestor forces get task result via Concent.
+        # Response is not important here TODO: Is it ?
```
cameel: Response indicates whether the submission was successful, so it is important. You should not get HTTP 4xx for example. You should test this response too.

concent_api/core/tests/test_integration_get_task_result.py
cameel: Sorry, my mistake. It's OK. That's the point of the test.

concent_api/core/tests/test_integration_get_task_result.py
cameel: OK. So now I see more clearly what this test is about.

This is a perfect situation that shows you why we need `assert` in tests and why it's different from `self.assertEqual()` and other tests assertions.

You should check the result but only with `assert`. This first call is not part of the test it's part of your setup. You do the first call and then you assume that it succeeded. If it did not, there's no point in continuting the test. If it returns an error it's not the code that's broken. It's your test because you should have constructed it in such a way that this first call succeeds.

So, in short, you need to `assert` that the first submission is successful.

concent_api/core/tests/test_integration_get_task_result.py
```diff
@@ -0,0 +1,601 @@
...
+                HTTP_CONCENT_CLIENT_PUBLIC_KEY = self._get_encoded_requestor_public_key()
+            )
+
+        self.assertEqual(response.status_code,  200)
+
+        message_from_concent = load(response.content, self.REQUESTOR_PRIVATE_KEY, CONCENT_PUBLIC_KEY)
+
+        self.assertIsInstance(message_from_concent, message.ForceGetTaskResultRejected)
+        self.assertEqual(message_from_concent.timestamp, self._parse_iso_date_to_timestamp("2017-12-01 11:00:09"))
+        self.assertEqual(message_from_concent.reason, "OPERATION_ALREADY_INITIATED")
```
cameel: Use enums, not strings.

concent_api/core/tests/test_integration_get_task_result.py
```diff
@@ -0,0 +1,601 @@
...
+            message_force_get_task_result = message_1_from_concent
+            message_file_transfer_token = message_2_from_concent
+        else:
+            message_force_get_task_result = message_2_from_concent
+            message_file_transfer_token = message_1_from_concent
+
+        # Test ForceGetTaskResult message
+        self.assertIsInstance(message_force_get_task_result, message.ForceGetTaskResult)
+        self.assertEqual(message_force_get_task_result.timestamp, self._parse_iso_date_to_timestamp("2017-12-01 11:00:02"))
+        self.assertTrue(message_force_get_task_result.task_to_compute)
```
cameel: `task_to_compute` is not a boolean. You should use `assertEqual()` and compare it with a specific value.

concent_api/core/tests/test_integration_get_task_result.py
```diff
@@ -0,0 +1,601 @@
...
+        self.assertTrue(message_force_get_task_result.task_to_compute.compute_task_def)
```
cameel: `compute_task_def` is not a boolean. Use `assertEqual()`.

concent_api/core/tests/test_integration_get_task_result.py
```diff
@@ -0,0 +1,601 @@
...
+        Concent   -> Provider:    FileTransferToken
+        Provider  -> Concent:     no response
+        Concent   -> Requestor:   ForceGetTaskResultFailed
+        """
+
+        # STEP 1: Requestor forces get task result via Concent.
+        # Concent accepts request if all conditions were met.
+        serialized_task_to_compute = self._get_serialized_task_to_compute(
+            timestamp   = "2017-12-01 10:00:00",
+            task_id     = 99,
```
cameel: Task IDs are now required to be strings.

concent_api/core/tests/test_integration_get_task_result.py
```diff
@@ -0,0 +1,601 @@
...
+            )
+
+        self.assertEqual(response_4.status_code,  200)
+
+        message_from_concent = load(response_4.content, self.REQUESTOR_PRIVATE_KEY, CONCENT_PUBLIC_KEY)
+
+        self.assertIsInstance(message_from_concent, message.ForceGetTaskResultFailed)
+        self.assertEqual(message_from_concent.timestamp, self._parse_iso_date_to_timestamp("2017-12-01 11:00:21"))
+        self.assertTrue(message_from_concent.task_to_compute)
+        self.assertTrue(message_from_concent.task_to_compute.compute_task_def)
```
cameel: These are not booleans. Use `assertEqual()`.

concent_api/core/tests/test_integration_get_task_result.py
```diff
@@ -0,0 +1,601 @@
...
+        Concent   -> Provider:    FileTransferToken
+        Provider  -> Concent:     TODO: Start Upload
+        Concent   -> Requestor:   ForceGetTaskResultFailed
+        """
+
+        # STEP 1: Requestor forces get task result via Concent.
+        # Concent accepts request if all conditions were met.
+        serialized_task_to_compute = self._get_serialized_task_to_compute(
+            timestamp   = "2017-12-01 10:00:00",
+            task_id     = 99,
```
cameel: task ID must be a string.

concent_api/core/tests/test_integration_get_task_result.py
```diff
@@ -0,0 +1,601 @@
...
+        """
+        Tests if on requestor ForceGetTaskResult message Concent will return ForceGetTaskResultFailed
+        if provider doesn't finish upload result.
+
+        Expected message exchange:
+        Requestor -> Concent:     ForceGetTaskResult
+        Concent   -> Requestor:   ForceGetTaskResultAck
+        Concent   -> Provider:    ForceGetTaskResult
+        Concent   -> Provider:    FileTransferToken
+        Provider  -> Concent:     TODO: Start Upload
```
cameel: What about this TODO?

concent_api/core/tests/test_integration_get_task_result.py
```diff
@@ -0,0 +1,601 @@
...
+            message_file_transfer_token = message_2_from_concent
+        else:
+            message_force_get_task_result = message_2_from_concent
+            message_file_transfer_token = message_1_from_concent
+
+        # Test ForceGetTaskResult message
+        self.assertIsInstance(message_force_get_task_result, message.ForceGetTaskResult)
+        self.assertEqual(message_force_get_task_result.timestamp, self._parse_iso_date_to_timestamp("2017-12-01 11:00:02"))
+        self.assertTrue(message_force_get_task_result.task_to_compute)
+        self.assertTrue(message_force_get_task_result.task_to_compute.compute_task_def)
```
cameel: These are not boolean values. Use `assertEqual()`.

concent_api/core/tests/test_integration_get_task_result.py
```diff
@@ -0,0 +1,601 @@
...
+        """
+        Tests if on requestor ForceGetTaskResult message Concent will return TaskResultHash
+        if provider uploads correct result.
+
+        Expected message exchange:
+        Requestor -> Concent:     ForceGetTaskResult
+        Concent   -> Requestor:   ForceGetTaskResultAck
+        Concent   -> Provider:    ForceGetTaskResult
+        Concent   -> Provider:    FileTransferToken
+        Provider  -> Concent:     TODO: Upload good files
```
cameel: What about this TODO?

concent_api/core/tests/utils.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def setUp(self):
+        super().setUp()
+        self.PROVIDER_PRIVATE_KEY, self.PROVIDER_PUBLIC_KEY = generate_ecc_key_pair()
+        self.REQUESTOR_PRIVATE_KEY, self.REQUESTOR_PUBLIC_KEY = generate_ecc_key_pair()
```
cameel: If these are now class variables there should not be named like constants. Use lower case.

concent_api/core/tests/utils.py
```diff
@@ -0,0 +1,59 @@
...
+            message_task_to_compute = task_to_compute,
+        )
+        return dump(
+            force_report_computed_task,
+            requestor_private_key or self.REQUESTOR_PRIVATE_KEY,
+            settings.CONCENT_PUBLIC_KEY
+        )
+
+    def _get_serialized_task_to_compute(self, timestamp, deadline, task_id=1, provider_private_key=None,
+                                        requestor_public_key=None):
```
cameel: Please fix your style here.

concent_api/api-integration-force-get-task-result-test.py
```diff
@@ -0,0 +1,152 @@
...
+            current_time,
+            size    = 1024,
+            checksum = '098f6bcd4621d373cade4e832627b4f6'
+        ),
+        headers = {
+            'Content-Type': 'application/octet-stream',
+            'concent-client-public-key': b64encode(REQUESTOR_PUBLIC_KEY).decode('ascii')
+        }
+    )
+    time.sleep(10)
```
cameel: These timings should be based on settings.

concent_api/api-integration-force-get-task-result-test.py
```diff
@@ -0,0 +1,152 @@
...
+import sys
+import datetime
+import time
+from base64                 import b64encode
+
+from golem_messages         import message
+
+from utils.testing_helpers  import generate_ecc_key_pair
+
+from api_testing_helpers            import api_request
```
cameel: Fix your style here.

concent_api/api-integration-force-get-task-result-test.py
```diff
@@ -0,0 +1,152 @@
...
+        CONCENT_PUBLIC_KEY,
+        get_force_get_task_result(
+            1,
+            current_time,
+            size    = 1024,
+            checksum = '098f6bcd4621d373cade4e832627b4f6'
+        ),
+        headers = {
+            'Content-Type': 'application/octet-stream',
+            'concent-client-public-key': b64encode(REQUESTOR_PUBLIC_KEY).decode('ascii')
```
cameel: Fix your style here.

concent_api/core/views.py
```diff
@@ -346,21 +481,50 @@ def store_message_and_message_status(golem_message_type, task_id, raw_golem_mess
...
-            timestamp   = message_timestamp,
-            delivered   = True
-        )
-        receive_out_of_band_status.full_clean()
-        receive_out_of_band_status.save()
+
+def get_file_status(file_transfer_token_from_database: message.concents.FileTransferToken) -> bool:
+    slash = '/'
+    assert file_transfer_token_from_database.files[0]['path'][0]    is not slash
+    assert settings.STORAGE_CLUSTER_ADDRESS[-1]                     is  slash
```
cameel: Use `str.endswith()` and `str.startswith()`.

concent_api/core/views.py
```diff
@@ -346,21 +481,50 @@ def store_message_and_message_status(golem_message_type, task_id, raw_golem_mess
...
+    assert len(file_transfer_token_from_database.files)             == 1
```
cameel: This assert should be at the top. If there are zero files the code will not even get here.

concent_api/core/tests/test_integration_get_task_result.py
rwrzesien: :+1:

concent_api/core/tests/test_integration_get_task_result.py
rwrzesien: It is mocked now.

concent_api/core/tests/test_integration_get_task_result.py
rwrzesien: It is mocked now.

PR #76, opened 2018-01-30
api_view: Add test for message with signature not matching the key sent in the headers

Jakub89:


concent_api/utils/tests/test_integration_api_view.py
```diff
@@ -106,3 +107,19 @@ def test_any_message_to_concent_report_empty_content_type_returns_400_error(self
 
         self.assertEqual(response.status_code,  400)
         self.assertIn('error', response.json().keys())
+
+    def test_any_message_to_concent_report_wrong_signature_returns_400_error(self):
+        """
+        Tests if a golem message to Concent signed with a wrong key returns HTTP 400 error.
+        """
+
+        assert type(load(self.serialized_dummy_message_to_concent, CONCENT_PRIVATE_KEY, PROVIDER_PUBLIC_KEY)) == message.Ping
```
cameel: This assert would do but it's not obvious why it's here. And it's bad because asserts are here partly as documentation - to tell someone reading the code what are your assumptions. It's here to ensure that the message is signed with provier's key but there isn't even a comment explaining it.

Calling `ecdsa_verify()` would be much clearer because this function exists exactly for this purpose:

``` python
from golem_messages.cryptography import ecdsa_verify

assert ecdsa_verify(PROVIDER_PUBLIC_KEY, self.serialized_dummy_message_to_concent.sig, self.serialized_dummy_message_to_concent.get_short_hash())
```

Jakub89: Done. `dummy_message_to_concent.sig`, not `serialized_dummy_message_to_concent.sig` ;)

cameel: Well, this kinda defeats the point. Now you're not checking the same data you're using in the test.

But OK. Let's stop at this point. It's good enough for now. We'll get back to this later in other tests.

Issue #63, opened 2018-01-23
View is called with client_message as None.


rwrzesien:
In few of our tests there is a case when view is called with client_message being None, which means it failed to be loaded in api_views.py, but we don't handle this type of situation. When working on different issue I have found out that 3 of our current tests cause this situation. 

We should add check to api_views.py after load() is performed to check if returned message is None, and if yes return Http400 error with proper message. Also an additional test which checks this specific case is needed.

cameel:
Looks like golem-messages has been updated to never return `None`: https://github.com/golemfactory/golem-messages/pull/98. This should be released soon. So we won't be getting `None` any more. 

We'll have to fix these tests after update though.

Just to make sure we really don't get `None`: please find all calls to `load()` or `deserialize()` in our code and add `assert decoded_message is not None` there as a sanity check.

PR #89, opened 2018-02-01
Fix issues with golem messages version switch from 1.6.0 to 1.7.0.

rwrzesien:
Resolves issues with golem_messages version switch from 1.6.0 to 1.6.2.
Resolves https://github.com/golemfactory/concent/issues/63.

### DIFF
--- a/concent_api/api-integration-force-report-computed-task-test.py
+++ b/concent_api/api-integration-force-report-computed-task-test.py
@@ -43,15 +43,12 @@ def force_report_computed_task(task_id, provider_private_key, provider_public_ke
     compute_task_def['task_id']     = task_id
     compute_task_def['deadline']    = current_time + 60
     task_to_compute = TaskToCompute(
-        timestamp = current_time - 3540,
         compute_task_def = compute_task_def)
 
     serialized_task_to_compute      = dump(task_to_compute,             requestor_private_key,  provider_public_key)
     deserialized_task_to_compute    = load(serialized_task_to_compute,  provider_private_key,   requestor_public_key, check_time = False)
 
-    force_report_computed_task = ForceReportComputedTask(
-        timestamp = current_time - 3540
-    )
+    force_report_computed_task = ForceReportComputedTask()
     force_report_computed_task.task_to_compute = deserialized_task_to_compute
 
     return force_report_computed_task
@@ -59,9 +56,7 @@ def force_report_computed_task(task_id, provider_private_key, provider_public_ke
 
 def ack_report_computed_task(task_id, provider_private_key, provider_public_key, requestor_private_key, requestor_public_key, current_time):
 
-    task_to_compute = TaskToCompute(
-        timestamp   = current_time - 3540
-    )
+    task_to_compute = TaskToCompute()
     task_to_compute.compute_task_def                = ComputeTaskDef()
     task_to_compute.compute_task_def['task_id']     = task_id
     task_to_compute.compute_task_def['deadline']    = current_time + 60
@@ -69,9 +64,7 @@ def ack_report_computed_task(task_id, provider_private_key, provider_public_key,
     serialized_task_to_compute      = dump(task_to_compute,             requestor_private_key,  provider_public_key)
     deserialized_task_to_compute    = load(serialized_task_to_compute,  provider_private_key,   requestor_public_key, check_time = False)
 
-    ack_report_computed_task = AckReportComputedTask(
-        timestamp  = current_time + 65,
-    )
+    ack_report_computed_task = AckReportComputedTask()
     ack_report_computed_task.task_to_compute = deserialized_task_to_compute
     return ack_report_computed_task
 
--- a/concent_api/core/tests/test_core_views.py
+++ b/concent_api/core/tests/test_core_views.py
@@ -41,13 +41,10 @@ class CoreViewSendTest(TestCase):
         self.compute_task_def['task_id'] = '8'
         self.compute_task_def['deadline'] = self.message_timestamp
         self.task_to_compute = message.TaskToCompute(
-            timestamp = self.message_timestamp,
             compute_task_def = self.compute_task_def,
         )
 
-        self.correct_golem_data = message.ForceReportComputedTask(
-            timestamp = self.message_timestamp,
-        )
+        self.correct_golem_data = message.ForceReportComputedTask()
         self.correct_golem_data.task_to_compute = self.task_to_compute
         self.want_to_compute = message.WantToComputeTask(
             node_name           = 1,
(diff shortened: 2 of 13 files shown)


concent_api/core/views.py
```diff
@@ -532,7 +536,6 @@ def get_file_status(file_transfer_token_from_database: message.concents.FileTran
 
     current_time = int(datetime.datetime.now().timestamp())
     file_transfer_token = message.concents.FileTransferToken(
-        timestamp                       = current_time,
         token_expiration_deadline       = current_time + settings.TOKEN_EXPIRATION_DEADLINE,
```
cameel: Please add an `assert` to check that timestamp <= deadline.

concent_api/core/views.py
```diff
@@ -358,20 +365,28 @@ def handle_receive_force_get_task_result_upload_for_requestor(decoded_message: m
     assert decoded_message.TYPE in message.registered_message_types
 
     current_time = int(datetime.datetime.now().timestamp())
-    decoded_message.timestamp                                      = current_time
-    decoded_message.file_transfer_token.timestamp                  = current_time
-    decoded_message.file_transfer_token.token_expiration_deadline  = current_time + settings.TOKEN_EXPIRATION_DEADLINE
-    decoded_message.file_transfer_token.operation                  = 'download'
+    file_transfer_token = message.concents.FileTransferToken(
+        token_expiration_deadline    = current_time + settings.TOKEN_EXPIRATION_DEADLINE,
```
cameel: Please add an `assert` to check that timestamp <= deadline.

concent_api/core/views.py
```diff
@@ -358,20 +365,28 @@ def handle_receive_force_get_task_result_upload_for_requestor(decoded_message: m
...
+        storage_cluster_address      = decoded_message.file_transfer_token.storage_cluster_address,
+        authorized_client_public_key = decoded_message.file_transfer_token.authorized_client_public_key,
+        operation                    = 'download',
+    )
+    file_transfer_token.files = decoded_message.file_transfer_token.files
+
+    force_get_task_result_upload = message.concents.ForceGetTaskResultUpload(
+        file_transfer_token   = file_transfer_token,
+    )
+    force_get_task_result_upload.force_get_task_result = decoded_message.force_get_task_result
```
cameel: Wouldn't it be possible to add this to the constructor call too?

concent_api/api-integration-force-report-computed-task-test.py
```diff
@@ -18,6 +18,7 @@
 
 from api_testing_helpers            import api_request
 
+from freezegun              import freeze_time
```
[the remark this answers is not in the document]
cameel: And to clarify these scripts are going to be executed against a real cluster.

concent_api/core/views.py
```diff
@@ -302,7 +314,6 @@ def handle_receive_force_get_task_result_upload_for_provider(request, decoded_me
 
     current_time            = int(datetime.datetime.now().timestamp())
     file_transfer_token     = message.concents.FileTransferToken(
-        timestamp                       = current_time,
         token_expiration_deadline       = current_time + settings.TOKEN_EXPIRATION_DEADLINE,
```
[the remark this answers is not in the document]
jiivan: I would refrain from using assert outside of tests. Mainly because it's disabled in optimized mode (`python -O`) https://docs.python.org/3/library/constants.html#__debug__

concent_api/api-integration-force-report-computed-task-test.py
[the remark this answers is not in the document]
rwrzesien: Right, I will remove it for now as we are going to refactor those anyway.

concent_api/core/views.py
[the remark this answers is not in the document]

concent_api/core/views.py

concent_api/core/views.py

concent_api/core/views.py

concent_api/core/views.py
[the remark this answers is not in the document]
cameel: @jiivan But that's how `assert` is supposed to be used. It's for development-time sanity checks. For explicitly stating your assumptions in a way that the interpreter can verify and call you out on it if you're wrong. The condition is never supposed to fail and the `-O` flag disables them so that they don't slow down your code in production. They're useless there because the code with broken assumptions should not even pass the testing stage without being repaired.

It's a development tool, not a replacement for the `if ...: raise ...` construct. Not for validating user input or handling errors. It's often used to check function parameters but only as a way to state _"this function does not support parameter values that do not satisfy these conditions and you should write your code in such a way that it never gets called with them"_. Which is of course not appropriate for a public function in an API that's supposed to work with all possible values but is fine in your internal code. If you call the function with bad values anyway, in optimized mode it will just continue and most likely give you wrong results, just like a function without the `assert` would. The difference is that with `assert` if you run into this case during development the interpreter will detect the problem early and signal it by raising `AssertionError` and interrupting the program.

In this case we're checking that our deadline is consistent with the timestamp created by golem-messages. It's always supposed to be and we don't want to add an `if` and check it at runtime. That would be silly. But it's always a good idea to document your assumptions. If we can also get the intepreter to verify them for us for free, that's even better.

Here's a good explanation of the concept: [Assert Statements in Python](https://dbader.org/blog/python-assert-tutorial). This is how it was originally used in the C language. It was a macro that you could disable in optimized code using a compiler flag. In Python the purpose is similar, though I find that many people are not familiar with the intended usage and later the testing frameworks appropriated `AssertionError` for their own purposes. Probably because the name fits really well. I recommend using it a lot because it's an invaluable tool when you're debugging. If you assert your assumptions a lot, your code tends to fail loudly rather than silently do the wrong thing.

rwrzesien:
@cameel updated.

concent_api/core/views.py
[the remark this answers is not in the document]
jiivan: Still, this means mixing tests with code, right? :)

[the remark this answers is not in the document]
cameel: I've never heard anyone refer to checking invariants or preconditions as testing. I think this association comes from the testing frameworks that use assertions too. The only thing in common is that both are comparing stuff and failing if it's not what's expected. But there's no contrived setup, no fixtures, no mocks and we're not running any additional code under test. It's not like, for example, doctests where you run short, one-line test cases and check the output. We're only checking the output of what the program normally does.

To me it's more similar to going through the code in a debugger and checking important values every few steps. Except that it happens automatically because you say what can go wrong in advance. It helps document the code because you're forced to express your implicit assumptions. Often this way you can replace verbose comments that state your assumptions with concise, executable checks at zero performance penalty.

concent_api/core/views.py
cameel: What about this one? Did you manage to do it?

concent_api/core/tests/test_core_views.py
```diff
@@ -408,47 +389,6 @@ def test_send_should_return_http_400_if_task_to_compute_deadline_is_not_an_integ
             self.assertEqual(response_400.status_code, 400)
             self.assertIn('error', response_400.json().keys())
 
-    def test_receive_should_return_http_400_if_task_id_is_not_valid_string(self):
```
cameel: Why is this test removed? Don't we still return HTTP 400 for invalid task IDs?

concent_api/core/views.py
rwrzesien: No, it raises `E:389, 8: Instance of 'ForceGetTaskResultUpload' has no 'force_get_task_result' member (no-member)`. I don't remember why I have marked it as ok.

concent_api/core/tests/test_core_views.py
rwrzesien: On golem_messages v1.7.0. I am not able to set wrong task_id. By set I mean doing this
```message.task_id = 11```
It fails instantly with validation exception.

File: concent_api/gatekeeper/views.py [hunk 7]
```diff
@@ -84,9 +85,7 @@ def parse_headers(request, path_to_file):
 
     try:
         loaded_golem_message = load(decoded_auth_header_content, settings.CONCENT_PRIVATE_KEY, settings.CONCENT_PUBLIC_KEY, check_time = False)
-    # FIXME: We want to catch only exceptions caused by malformed messages but golem-messages does not have specialized
-    # exception classes for that. It simply raises AttributeError.
-    except AttributeError:
+    except MessageError:
```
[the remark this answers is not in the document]
rwrzesien: Yes, it is:
```
class CryptoError(MessageError):
    pass

class InvalidSignature(CryptoError):
    pass
```

concent_api/core/tests/test_core_views.py
cameel: Well, OK. We still need this test but we'll do it in a separate issue. Even if the message class won't let you set `task_id` to an invalid value via the property it's definitely still possible to manually craft such a message. We'll need to ensure that such messages are correctly detected as invalid and result in HTTP 400 instead of going through and making the code behave in an unexpected way. I see that this is a bigger problem though and will need to be addressed separately.

PR #91, opened 2018-02-01
Add feature to gatekeeper which return headers in response

### LINKED ISSUE #77: Returning file checksum and size from gatekeeper on upload request
When gatekeeper gets an upload request it should, after validating everything, read the checksum and size of the requested file from the token and return it in HTTP headers of the response. The headers should be called `Concent-File-Checksum` and `Concent-File-Size`.

These headers should not be added to the response in case of a download request.

### Example
Client makes the following request to `nginx-storage`:
``` bash
curl -X POST http://storage.concent.golem.network/upload/123/456/result \
    --header "Content-Type: application/octet-stream" \
    --header "Authorization: Golem dG9rZW4K" \
    --header "Concent-Client-Public-Key: MTIzNDU2Cg==" \
    --data-binary @result.jpg
```

`nginx-storage` makes the following request to `gatekeeper`:

``` bash
curl -X POST http://gatekeeper/gatekeeper/upload/123/456/result \
    --header "Content-Type: application/octet-stream" \
    --header "Authorization: Golem dG9rZW4K" \
    --header "Concent-Client-Public-Key: MTIzNDU2Cg=="
```

The token contains the following values:

``` python
{
    'timestamp':                    1510410600, # 2017-11-11 15:30
    'token_expiration_deadline':    1510412400, # 2017-11-11 16:00
    'cluster_address':              "http://storage.concent.golem.network/",
    'authorized_client_public_key': b'123456',
    'operation':                    'upload',
    'files':                        [
        {
            'path':     '123/456/settings',
            'checksum': 'sha1:1741af78aee2cd01b40f1a33d8d551aeb0340eb2',
            'size':     154,  # bytes
        },
        {
            'path':     '123/456/result',
            'checksum': 'sha1:2bebc22296f03225617704d29d277c9e96fafcc2',
            'size':     1048576,  # bytes
        },
    ],
}
```

The HTTP response that `nginx-storage` receives back from `gatekeeper` should contain the following headers:

| header                  | value                     |
|-------------------------|---------------------------|
| `Concent-File-Checksum` | `sha1:2bebc22296f03225617704d29d277c9e96fafcc2`
| `Concent-File-Size`     | `1048576`

bartoszbetka:
Resolves #77

### DIFF
--- a/concent_api/gatekeeper/tests/test_gatekeeper_views.py
+++ b/concent_api/gatekeeper/tests/test_gatekeeper_views.py
@@ -90,6 +90,40 @@ class GatekeeperViewUploadTest(TestCase):
         self.assertEqual(response.status_code, 401)
         self.assertIn('message', response.json().keys())
 
+    @freeze_time("2018-12-30 11:00:00")
+    def test_upload_should_return_401_if_file_paths_are_not_unique(self):
+        file1 = FileTransferToken.FileInfo(
+            path     = 'blender/benchmark/test_task/scene-Helicopter-27-cycles.blend',
+            checksum = 'sha1:098f6bcd4621d373cade4e832627b4f6',
+            size     = 1024,
+        )
+
+        file2 = FileTransferToken.FileInfo(
+            path     = 'blender/benchmark/test_task/scene-Helicopter-27-cycles.blend',
+            checksum = 'sha1:098f6bcd4621d373cade4e832627b4f6',
+            size     = 1024,
+        )
+
+        self.upload_token.files = [file1, file2]
+        assert file1 == file2
+
+        golem_upload_token = dump(self.upload_token, settings.CONCENT_PRIVATE_KEY, settings.CONCENT_PUBLIC_KEY)
+        encoded_token = b64encode(golem_upload_token).decode()
+        response = self.client.post(
+            '{}{}'.format(
+                reverse('gatekeeper:upload'),
+                'blender/benchmark/test_task/scene-Helicopter-27-cycles.blend'
+            ),
+            HTTP_AUTHORIZATION             = 'Golem ' + encoded_token,
+            content_type                   = 'application/x-www-form-urlencoded',
+            HTTP_CONCENT_CLIENT_PUBLIC_KEY = self.public_key
+        )
+
+        self.assertIsInstance(response, JsonResponse)
+        self.assertEqual(response.status_code, 401)
+        self.assertIn('message', response.json().keys())
+        self.assertEqual("application/json", response["Content-Type"])
+
 
 @override_settings(
     CONCENT_PRIVATE_KEY = b'l\xcdh\x19\xeb$>\xbcG\xa1\xc7v\xe8\xd7o\x0c\xbf\x0e\x0fM\x89lw\x1e\xd7K\xd6Hnv$\xa2',
@@ -174,3 +208,36 @@ class GatekeeperViewDownloadTest(TestCase):
         self.assertIsInstance(response, JsonResponse)
         self.assertEqual(response.status_code, 401)
         self.assertIn('message', response.json().keys())
+
+    @freeze_time("2018-12-30 11:00:00")
+    def test_download_should_return_401_if_file_paths_are_not_unique(self):
+        file1 = FileTransferToken.FileInfo(
+            path     = 'blender/benchmark/test_task/scene-Helicopter-27-cycles.blend',
+            checksum = 'sha1:098f6bcd4621d373cade4e832627b4f6',
+            size     = 1024,
+        )
+
+        file2 = FileTransferToken.FileInfo(
+            path     = 'blender/benchmark/test_task/scene-Helicopter-27-cycles.blend',
+            checksum = 'sha1:098f6bcd4621d373cade4e832627b4f6',
+            size     = 1024,
+        )
+
+        self.upload_token.files = [file1, file2]
+        assert file1 == file2
+
+        golem_upload_token = dump(self.upload_token, settings.CONCENT_PRIVATE_KEY, settings.CONCENT_PUBLIC_KEY)
+        encoded_token = b64encode(golem_upload_token).decode()
+        response = self.client.get(
+            '{}{}'.format(
+                reverse('gatekeeper:download'),
+                'blender/benchmark/test_task/scene-Helicopter-27-cycles.blend'
+            ),
+            HTTP_AUTHORIZATION             = 'Golem ' + encoded_token,
+            HTTP_CONCENT_CLIENT_PUBLIC_KEY = self.public_key
+        )
+
+        self.assertIsInstance(response, JsonResponse)
+        self.assertEqual(response.status_code, 401)
+        self.assertIn('message', response.json().keys())
+        self.assertEqual("application/json", response["Content-Type"])
--- a/concent_api/gatekeeper/tests/test_gatekeeper_views.py
+++ b/concent_api/gatekeeper/tests/test_gatekeeper_views.py
@@ -30,7 +30,7 @@ class GatekeeperViewUploadTest(TestCase):
 
         self.upload_token.files                 = [FileTransferToken.FileInfo()]
         self.upload_token.files[0]['path']      = 'blender/benchmark/test_task/scene-Helicopter-27-cycles.blend'
-        self.upload_token.files[0]['checksum']  = '098f6bcd4621d373cade4e832627b4f6'
+        self.upload_token.files[0]['checksum']  = 'sha1:098f6bcd4621d373cade4e832627b4f6'
         self.upload_token.files[0]['size']      = 1024
         self.upload_token.operation             = 'upload'
 
@@ -50,6 +50,9 @@ class GatekeeperViewUploadTest(TestCase):
 
         self.assertIsInstance(response, JsonResponse)
         self.assertEqual(response.status_code, 200)
+        self.assertTrue(response.has_header("Concent-File-Size"))
+        self.assertTrue(response.has_header("Concent-File-Checksum"))
+        self.assertEqual("application/json", response["Content-Type"])
 
     @freeze_time("2018-12-30 11:00:00")
     def test_upload_should_return_401_if_wrong_request_content_type(self):
@@ -125,6 +128,9 @@ class GatekeeperViewDownloadTest(TestCase):
 
         self.assertIsInstance(response, JsonResponse)
         self.assertEqual(response.status_code, 200)
+        self.assertFalse(response.has_header("Concent-File-Size"))
+        self.assertFalse(response.has_header("Concent-File-Checksum"))
+        self.assertEqual("application/json", response["Content-Type"])
 
     @freeze_time("2018-12-30 11:00:00")
     def test_download_should_return_401_if_wrong_authorization_header(self):
--- a/concent_api/gatekeeper/views.py
+++ b/concent_api/gatekeeper/views.py
@@ -4,9 +4,11 @@ from base64                         import b64decode
 from base64                         import b64encode
 import json
 import logging
+from typing                         import Union
 
 from django.conf                    import settings
 from django.http                    import JsonResponse
+from django.core.handlers.wsgi      import WSGIRequest
 from django.core.validators         import URLValidator
 from django.core.exceptions         import ValidationError
 from django.urls                    import reverse
@@ -15,6 +17,7 @@ from django.views.decorators.http   import require_POST
 from django.views.decorators.http   import require_safe
 
 from golem_messages.exceptions      import MessageError
+from golem_messages.message         import FileTransferToken
 from golem_messages.message         import Message
 from golem_messages.shortcuts       import load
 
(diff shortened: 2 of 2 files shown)


concent_api/gatekeeper/tests/test_gatekeeper_views.py
```diff
@@ -88,6 +91,30 @@ def test_upload_should_return_401_if_message_token_deadline_pass(self):
...
+            ),
+            HTTP_AUTHORIZATION = 'Golem ' + encrypted_token,
+            content_type = '',
+            HTTP_CONCENT_CLIENT_PUBLIC_KEY = self.public_key
+        )
+
+        self.assertIsInstance(response, JsonResponse)
+        self.assertEqual(response.status_code, 401)
+        self.assertIn('message', response.json().keys())
+        self.assertIn("application/json", response["Content-Type"])
```
rwrzesien: Shouldn't it be equal ?

concent_api/gatekeeper/tests/test_gatekeeper_views.py
```diff
@@ -125,6 +152,9 @@ def test_download_should_accept_valid_message(self):
 
         self.assertIsInstance(response, JsonResponse)
         self.assertEqual(response.status_code, 200)
+        self.assertFalse(response.has_header("Concent-File-Size"))
+        self.assertFalse(response.has_header("Concent-File-Checksum"))
+        self.assertIn("application/json", response["Content-Type"])
```
rwrzesien: Shouldn't it be equal ?

concent_api/gatekeeper/tests/test_gatekeeper_views.py
```diff
@@ -168,3 +198,26 @@ def test_download_should_return_401_if_wrong_token_cluster_address(self):
...
+                'blender/benchmark/test_task/scene-Helicopter-27-cycles.blend'
+            ),
+            HTTP_AUTHORIZATION = 'Golem ' + encrypted_token,
+            HTTP_CONCENT_CLIENT_PUBLIC_KEY = self.public_key
+        )
+
+        self.assertIsInstance(response, JsonResponse)
+        self.assertEqual(response.status_code, 401)
+        self.assertIn('message', response.json().keys())
+        self.assertIn("application/json", response["Content-Type"])
```
rwrzesien: Shouldn't it be equal ?

concent_api/gatekeeper/views.py
```diff
@@ -143,10 +146,12 @@ def parse_headers(request, path_to_file):
     # -FILES
     if not all(isinstance(file, dict) for file in loaded_golem_message.files):
         return gatekeeper_access_denied_response('Wrong type of files variable.', path_to_file, loaded_golem_message.subtask_id, client_public_key)
-    transfer_token_paths_to_files = []
+    transfer_token_paths_to_files = [file["path"] for file in loaded_golem_message.files]
+    transfer_token_paths_to_files_without_repetitions = set(transfer_token_paths_to_files)
+    if (len(transfer_token_paths_to_files) != len(transfer_token_paths_to_files_without_repetitions)):
```
rwrzesien: There is no need of extra parenthesis.

concent_api/gatekeeper/tests/test_gatekeeper_views.py
bartoszbetka: Well, I not sure in this case.

rwrzesien: I have checked and this header is string looking exactly "application/json", so I don't see any reason to not compare it using assertEqual. If there is a reason for that, please let me know, if not, please change (in other places too).

bartoszbetka: Done.

concent_api/gatekeeper/views.py
```diff
@@ -31,14 +32,16 @@ def upload(request):
         return gatekeeper_access_denied_response('Unsupported content type.')
 
     path_to_file = request.get_full_path().partition(reverse('gatekeeper:upload'))[2]
-    response = parse_headers(request, path_to_file)
-    if response is not None:
-        logger.info(json.loads(response.content.decode())['message'])
-        return response
+    response_body = parse_headers(request, path_to_file)
```
cameel: This name is confusing because the function does not necessarily return a response. If it does, it's a full response not just body. If it does not, it's a `FileInfo` object.

1) Call it `response_or_file_info` for clarity.
2) Add type annotation on `parse_headers()` that indicates what it returns (Please ask @rwrzesien for help if you're not familiar with the syntax).

concent_api/gatekeeper/views.py
```diff
@@ -31,14 +32,16 @@ def upload(request):
...
+    if not isinstance(response_body, FileTransferToken.FileInfo):
+        logger.info(json.loads(response_body.content.decode())['message'])
+        return response_body
```
cameel: In this `if` you should `assert` that `response_or_file_info` is indeed a response object - because if it is not (e.g. because someone updated`parse_headers()` to return something else in some case and forgot to update the view), we have a bug and we won't easily detect it.

concent_api/gatekeeper/views.py
```diff
@@ -52,10 +55,10 @@ def download(request):
...
 
     path_to_file = request.get_full_path().partition(reverse('gatekeeper:download'))[2]
-    response = parse_headers(request, path_to_file)
-    if response is not None:
-        logger.info(json.loads(response.content.decode())['message'])
-        return response
+    response_body = parse_headers(request, path_to_file)
+    if not isinstance(response_body, FileTransferToken.FileInfo):
+        logger.info(json.loads(response_body.content.decode())['message'])
+        return response_body
```
cameel: See my remarks about `upload()`. Same applies here.

concent_api/gatekeeper/views.py
```diff
@@ -143,10 +146,12 @@ def parse_headers(request, path_to_file):
     # -FILES
     if not all(isinstance(file, dict) for file in loaded_golem_message.files):
         return gatekeeper_access_denied_response('Wrong type of files variable.', path_to_file, loaded_golem_message.subtask_id, client_public_key)
-    transfer_token_paths_to_files = []
+    transfer_token_paths_to_files = [file["path"] for file in loaded_golem_message.files]
+    transfer_token_paths_to_files_without_repetitions = set(transfer_token_paths_to_files)
```
cameel: This variable is superfluous. Just use `set()` in the `if`.

concent_api/gatekeeper/tests/test_gatekeeper_views.py
```diff
@@ -168,3 +198,26 @@ def test_download_should_return_401_if_wrong_token_cluster_address(self):
         self.assertIsInstance(response, JsonResponse)
         self.assertEqual(response.status_code, 401)
         self.assertIn('message', response.json().keys())
+
+    @freeze_time("2018-12-30 11:00:00")
+    def test_download_should_return_401_if_file_paths_are_not_unique(self):
+        self.upload_token.files.append(FileTransferToken.FileInfo())
```
cameel: You should assert that `self.upload_token.files` has exactly one element. Otherwise, when someone changes that in `setUp()` your test will just overwrite an existing `FileInfo`.

Also, I think it would be clearer to first create `FileInfo` and only then use `append()`. You would not have to repeatedly refer to its position on the list.

concent_api/gatekeeper/tests/test_gatekeeper_views.py
```diff
@@ -168,3 +198,26 @@ def test_download_should_return_401_if_wrong_token_cluster_address(self):
...
+        self.upload_token.files[1]['path']      = 'blender/benchmark/test_task/scene-Helicopter-27-cycles.blend'
+        self.upload_token.files[1]['checksum']  = 'sha1:098f6bcd4621d373cade4e832627b4f6'
+        self.upload_token.files[1]['size']      = 1024
+
+        golem_upload_token = dump(self.upload_token, settings.CONCENT_PRIVATE_KEY, settings.CONCENT_PUBLIC_KEY)
+        encrypted_token = b64encode(golem_upload_token).decode()
```
cameel: base64 is not encryption. The name is completely misleading.

concent_api/gatekeeper/tests/test_gatekeeper_views.py
```diff
@@ -168,3 +198,26 @@ def test_download_should_return_401_if_wrong_token_cluster_address(self):
         self.assertIsInstance(response, JsonResponse)
         self.assertEqual(response.status_code, 401)
         self.assertIn('message', response.json().keys())
+
+    @freeze_time("2018-12-30 11:00:00")
+    def test_download_should_return_401_if_file_paths_are_not_unique(self):
+        self.upload_token.files.append(FileTransferToken.FileInfo())
+        self.upload_token.files[1]['path']      = 'blender/benchmark/test_task/scene-Helicopter-27-cycles.blend'
```
cameel: You should `assert` that the file name in the existing `FileInfo` is the same as in your `FileInfo`. Otherwise the test is broken.

concent_api/gatekeeper/tests/test_gatekeeper_views.py
```diff
@@ -87,6 +90,30 @@ def test_upload_should_return_401_if_message_token_deadline_pass(self):
         self.assertEqual(response.status_code, 401)
         self.assertIn('message', response.json().keys())
 
+    @freeze_time("2018-12-30 11:00:00")
+    def test_upload_should_return_401_if_file_paths_are_not_unique(self):
```
cameel: See my remarks about the other test. Same applies here.

concent_api/gatekeeper/tests/test_gatekeeper_views.py
```diff
@@ -87,6 +90,30 @@ def test_upload_should_return_401_if_message_token_deadline_pass(self):
...
+        golem_upload_token = dump(self.upload_token, settings.CONCENT_PRIVATE_KEY, settings.CONCENT_PUBLIC_KEY)
+        encrypted_token = b64encode(golem_upload_token).decode()
+        response = self.client.post(
+            '{}{}'.format(
+                reverse('gatekeeper:upload'),
+                'blender/benchmark/test_task/scene-Helicopter-27-cycles.blend'
+            ),
+            HTTP_AUTHORIZATION = 'Golem ' + encrypted_token,
+            content_type = 'application/x-www-form-urlencoded',
+            HTTP_CONCENT_CLIENT_PUBLIC_KEY = self.public_key
```
cameel: Please fix your style here.

bartoszbetka: That style is in all previous method, so i used it, to keep convention. I should fix it ?

cameel: Fix it. I let some of this stuff slide in previous reviews because we needed to merge it and this was a small issue but you should not base your new code on that :)

concent_api/gatekeeper/tests/test_gatekeeper_views.py
bartoszbetka: I don't understand second part, you want create new `FileInfo` object in method and then create list with two filepaths? It rewrite `FileInfo` object in `setUp()` function.

bartoszbetka: Done.

cameel: I mean this:

``` python
file1 = FileTransferToken.FileInfo(
    path     = 'blender/benchmark/test_task/scene-Helicopter-27-cycles.blend',
    checksum = 'sha1:098f6bcd4621d373cade4e832627b4f6',
    size     = 1024,
)

file2 = FileTransferToken.FileInfo(
    path     = 'blender/benchmark/test_task/scene-Helicopter-27-cycles.blend',
    checksum = 'sha1:098f6bcd4621d373cade4e832627b4f6',
    size     = 1024,
)

self.upload_token.files = [file1, file2]
assert file1['name'] == file2['name']
```

concent_api/gatekeeper/tests/test_gatekeeper_views.py
cameel: Call it `encoded_token`.

concent_api/gatekeeper/views.py
cameel: Still no annotation for the return type. You only added annotations for the parameters.

concent_api/gatekeeper/tests/test_gatekeeper_views.py
cameel: Still no `assert` that the names are the same.

bartoszbetka: I create new `FileInfo` object, so it doesn't matter.

cameel: It does `assert` is also useful to indicate your assumptions. Adding it here shows clearly that you intended the names to be the same.

cameel:
There are warnings:

```
W:159, 4: Else clause on loop without a break statement (useless-else-on-loop)
```

concent_api/gatekeeper/views.py
```diff
@@ -142,10 +150,13 @@ def parse_headers(request, path_to_file):
...
         return gatekeeper_access_denied_response('Wrong type of files variable.', path_to_file, loaded_golem_message.subtask_id, client_public_key)
-    transfer_token_paths_to_files = []
-    for file in loaded_golem_message.files:
-        transfer_token_paths_to_files.append(file['path'])
-    if path_to_file not in transfer_token_paths_to_files:
+    transfer_token_paths_to_files = [file["path"] for file in loaded_golem_message.files]
+    if len(transfer_token_paths_to_files) != len(set(transfer_token_paths_to_files)):
+        return gatekeeper_access_denied_response('File paths in the token must be unique', path_to_file, loaded_golem_message.subtask_id, client_public_key)
+
+    get_file_from_token = [file for file in loaded_golem_message.files if path_to_file == file['path']]
```
cameel: This sounds like a function name. Please don't call your variables like that.

Issue #71, opened 2018-01-29
Concent API without gatekeeper feature enabled fails when trying to check if file has been uploaded


cameel:
`get_file_status()` function uses `reverse('gatekeeper:download')` to get the path of the `/download/` endpoint. This obviously won't work when gatekeeper URLs are disabled - and this is always the case in the `concent-api` container.

Fix this by extracting the `/download/` part from urlconfig into a constant that can be used even when URLs are disabled.

PR #96, opened 2018-02-05
Extract the gatekeeper 'download' part from urlconfig into a constant

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/71


cameel:
What about `urls.py`? This does not solve the problem yet, because someone updating the address in `urls.py` might not be even aware that this constant exists.

rwrzesien:
@cameel Updated.

Issue #93, opened 2018-02-05
Use non-naive datetimes.


rwrzesien:
Example from @jiivan:
``` python
>>> import time, calendar, datetime
>>> int(
    time.time()),
    calendar.timegm(time.gmtime()),
    int(datetime.datetime.utcnow().timestamp()),
    int(datetime.datetime.now().timestamp(),
)
(1518000142, 1518000142, 1517996542, 1518000142)
```

As you can see timestamps generated using `utcnow()` are not in UTC. But the problem is not in `utcnow()` but in the `timestamp()` method. [Docs for `datetime.timestamp()`](https://docs.python.org/3.6/library/datetime.html#datetime.datetime.timestamp) say:

> Naive datetime instances are assumed to represent local time and this method relies on the platform C mktime() function to perform the conversion.

This means that every call to `timestamp()` on a naive `datetime` object gives a result in the wrong timezone. Our timestamps must be in UTC. Otherwise timestamps created by clients in different timezones won't match.

To fix this:
1) Create a helper function for getting a timestamp that represents the current time (an int value). 
2) Find all places where we get current timestamp in our code and replace them with a call to the helper.
3) Create a helper function for converting a datetime object (should support both naive and timezone-aware) to a timestamp (an int value).
4) Find all calls to `timestamp()` in our code and replace them with a call to the helper.

rwrzesien:
Updated description.

rwrzesien:
@cameel What do you think about this one ? I see two options here:
1. Create timezone aware datetimes and store them in database with timezone info.
2. Create naive datetimes but in utc time, store them without timezone info, disable timezone support in django.

Do we even want/need to support timezones in the project ?

cameel:
Nomally we don't support timezones in projects where all timestamps are generated by ourselves. We store all timestamps in UTC and convert them to user's local timezone in the UI depending on his preferences.

But it's different when we have to support timestamps submitted by users (which is always problematic if you need to reply on those timestamps because of all the typical problems with time synchronization. That's the case here. Golem messages have timestamps created by clients. So it really depends on Golem - if they allow timestamps with timezones, we must allow them too.

But I think it would be better not to support timezones in messages. This would simplify things. Please talk with @jiivan about this.

Other than messages, we don't have to support timezones and we coud disable tiem.

cameel:
I've updated the description with the conclusions from our discussion with @jiivan and @shadeofblue on Rocket:
[...]

PR #102, opened 2018-02-08
Use non-naive datetimes

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/93

### DIFF
--- a/concent_api/api-integration-force-report-computed-task-test.py
+++ b/concent_api/api-integration-force-report-computed-task-test.py
@@ -2,7 +2,6 @@
 
 import os
 import sys
-import datetime
 import random
 import time
 from base64                 import b64encode
@@ -14,6 +13,7 @@ from golem_messages.message import ComputeTaskDef
 from golem_messages.message import ForceReportComputedTask
 from golem_messages.message import TaskToCompute
 
+from utils.helpers          import get_current_utc_timestamp
 from utils.testing_helpers  import generate_ecc_key_pair
 
 from api_testing_helpers            import api_request
@@ -72,7 +72,7 @@ def ack_report_computed_task(task_id, provider_private_key, provider_public_key,
 def main():
     cluster_url     = parse_command_line(sys.argv)
     task_id         = str(random.randrange(1, 100000))
-    current_time    = int(datetime.datetime.now().timestamp())
+    current_time    = get_current_utc_timestamp()
 
     api_request(
         cluster_url,
--- a/concent_api/utils/helpers.py
+++ b/concent_api/utils/helpers.py
@@ -1,5 +1,10 @@
+import calendar
 import base64
 import binascii
+import datetime
+import time
+
+from django.utils import timezone
 
 
 def is_base64(data: str) -> bool:
@@ -17,3 +22,27 @@ def is_base64(data: str) -> bool:
 
 def decode_key(key: str) -> bytes:
     return base64.b64decode(key.encode('ascii'), validate = True)
+
+
+def get_current_utc_timestamp() -> int:
+    """
+    Returns current timestamp as int.
+    Returned timestamp is guaranteed to be in UTC on systems where epoch is January 1, 1970, 00:00:00 (UTC).
+
+    """
+    current_timestamp = int(timezone.now().timestamp())
+    assert calendar.timegm(time.gmtime()) - current_timestamp <= 1
+    return current_timestamp
+
+
+def parse_datetime_to_timestamp(date_time: datetime.datetime) -> int:
+    """
+    Returns timestamp in UTC as int from given datetime.
+    Returned timestamp is relative to UNIX epoch and should works the same on all operating systems.
+    It works for both naive and timezone-aware datetime object. Naive datetime is interpreted as if it was in UTC.
+
+    """
+    if date_time.tzinfo is None:
+        return int(date_time.replace(tzinfo = timezone.utc).timestamp())
+    else:
+        return int(date_time.astimezone(timezone.utc).timestamp())
--- a/concent_api/utils/tests/test_helpers.py
+++ b/concent_api/utils/tests/test_helpers.py
@@ -1,0 +1,40 @@
+import datetime
+
+from django.test            import TestCase
+from django.utils           import timezone
+
+from utils.helpers          import parse_datetime_to_timestamp
+
+
+class HelpersTestCase(TestCase):
+
+    def test_parse_datetime_to_timestamp_should_return_correct_utc_timestamp(self):
+        """ Tests if parse_datetime_to_timestamp function works as expected. """
+
+        timestamp = 946684800  # 2000-01-01 00:00
+        assert datetime.datetime.fromtimestamp(timestamp) == datetime.datetime(2000, 1, 1, 0, 0)
+
+        for date_time in [
+            datetime.datetime(2000, 1, 1, 0, 0),
+            datetime.datetime(2000, 1, 1, 0, 0,    tzinfo = timezone.pytz.timezone('UTC')),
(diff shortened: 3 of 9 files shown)


concent_api/utils/helpers.py
```diff
@@ -0,0 +1,18 @@
+import datetime
+import time
+
+import calendar
+
+from django.utils import timezone
+
+
+def get_current_timestamp() -> int:
+    """ Returns current timestamp as int. """
```
cameel: Please add more detail here:
- The timestamp is guaranteed to be in UTC.
    - Maybe it would even be a good idea to add `utc` in the name? I'm not sure but that would make it very clear. What do you think?
- The timestamp is not OS-dependent (uses the UNIX epoch).
    - I'm not 100% sure that it's true but this is what is supposed to be returned so if it's not true it's a bug.

rwrzesien:
@cameel @kbeker @Jakub89 Rebased on latest master and ready for review.

concent_api/utils/helpers.py
```diff
@@ -0,0 +1,18 @@
...
+    current_timestamp = int(timezone.now().timestamp())
+    assert calendar.timegm(time.gmtime()) - current_timestamp <= 1
+    return current_timestamp
+
+
+def parse_datetime_to_timestamp(date_time: datetime.datetime) -> int:
+    """ Returns timestamp as int from given datetime. """
+    return int(date_time.replace(tzinfo=timezone.utc).timestamp())
```
cameel: I don't think this is correct for timezone-aware datetimes. It removes the existing `tzinfo` and puts UTC in its place without adjusting the values.

If you pass it a `datetime` with `tzinfo` indicating a timezone other than UTC it will simply ignore the timezone and convert to int as if it were in UTC.

This shows that we need tests for these helpers even though they may seem simple. Please add a test that passes various naive and tz-aware values (especially aware ones in different time zones) and compares resulting timestamps with our expectations. This will also come in handy if we decide to change the implementation later. We'll be sure that our assumptions are still satisfied.

concent_api/utils/helpers.py
```diff
@@ -0,0 +1,18 @@
...
+
+def get_current_timestamp() -> int:
+    """ Returns current timestamp as int. """
+    current_timestamp = int(timezone.now().timestamp())
+    assert calendar.timegm(time.gmtime()) - current_timestamp <= 1
+    return current_timestamp
+
+
+def parse_datetime_to_timestamp(date_time: datetime.datetime) -> int:
+    """ Returns timestamp as int from given datetime. """
```
cameel: Please expand this too:
- Say that we support both naive and timezone-aware `datetime`.
- Naive `datetime` is intepreted as it it were in UTC.
- The resulting timestamp is relative to UNIX epoch and it should work the same on all operating systems.
- The resulting timestamp is also in UTC (and not local time).

concent_api/utils/helpers.py
rwrzesien: Right, I have tested it before and thought it is ok, but I must have messed up something with testing. I will change it and add tests for it.

concent_api/utils/helpers.py
rwrzesien: :+1:

concent_api/utils/helpers.py
rwrzesien: :+1: 
```
The timestamp is not OS-dependent (uses the UNIX epoch).
I'm not 100% sure that it's true but this is what is supposed to be returned so if it's not true it's a bug.
```
It uses time.time under the hood, so it will work correctly on OSes that handle epoch correctly.

rwrzesien:
@cameel Updated, I have squashed fixups because almost every line changed.

concent_api/api-integration-force-report-computed-task-test.py
```diff
@@ -72,7 +72,7 @@ def ack_report_computed_task(task_id, provider_private_key, provider_public_key,
 def main():
     cluster_url     = parse_command_line(sys.argv)
     task_id         = random.randrange(1, 100000)
-    current_time    = int(datetime.datetime.now().timestamp())
+    current_time    = calendar.timegm(time.gmtime())
```
cameel: Please use the helpers in these scripts too.

concent_api/utils/helpers.py
```diff
@@ -17,3 +22,27 @@ def is_base64(data: str) -> bool:
 
 def decode_key(key: str) -> bytes:
     return base64.b64decode(key.encode('ascii'), validate = True)
+
+
+def get_current_utc_timestamp() -> int:
+    """
+    Returns current timestamp as int.
+    Returned timestamp is guaranteed to be in UTC as Django server is set to be in UTC.
```
cameel: 1) You mean that they will be in UTC only if `USE_TZ` is set to `True`?

    This function should return the time in UTC no matter what `USE_TZ` is set to. Or, if that's hard to do, we should simply say that we don't support `USE_TZ == False` by adding `assert settings.USE_TZ` as a precondition.

2) You did not mention in the docstring whether it's OS-dependent or not.

concent_api/utils/helpers.py
```diff
@@ -17,3 +22,27 @@ def is_base64(data: str) -> bool:
...
+    current_timestamp = int(timezone.now().timestamp())
+    assert calendar.timegm(time.gmtime()) - current_timestamp <= 1
+    return current_timestamp
+
+
+def parse_datetime_to_timestamp(date_time: datetime.datetime) -> int:
+    """
+    Returns timestamp in UTC as int from given datetime.
+    Returned timestamp is relative to UNIX epoch and should works the same on all operating systems.
+    It works for both naive and timezone-aware datetime object. Naive datetime is interpreted as it it was in UTC.
```
cameel: Typo: `it` -> `if`.

concent_api/utils/tests/test_helpers.py
```diff
@@ -0,0 +1,38 @@
...
+    def test_parse_datetime_to_timestamp_should_return_correct_utc_timestamp(self):
+        """ Tests if parse_datetime_to_timestamp function works as expected. """
+
+        timestamp = 946684800  # 2000-01-01 00:00
+        assert datetime.datetime.fromtimestamp(timestamp) == datetime.datetime(2000, 1, 1, 0, 0)
+
+        for date_time in [
+            datetime.datetime(2000, 1, 1, 0, 0),
+            datetime.datetime(2000, 1, 1, 4, 37,   tzinfo = timezone.pytz.timezone('Asia/Kabul')),
+            datetime.datetime(1999, 12, 31, 19, 4, tzinfo = timezone.pytz.timezone('US/Eastern')),
```
cameel: Please test with a `datetime` that has `tzinfo = timezone.utc` too. That's the most common case for us after all :)

concent_api/api-integration-force-report-computed-task-test.py
rwrzesien: :+1: I have checked, in current implementation cluster tests run from pure python doesn't parse settings but usage of timezone.now() without settings return time in UTC.

concent_api/utils/helpers.py
rwrzesien: :+1:

concent_api/utils/tests/test_helpers.py
rwrzesien: :+1:

concent_api/utils/helpers.py
rwrzesien: Ad 1) I have checked and without this setting it still returns utc, so I will update comment.
Ad 2) Added.

rwrzesien:
@cameel Updated.

PR #104, opened 2018-02-08
Validate checksum and size of fields in gatekeeper

### LINKED ISSUE #78: Validating file checksum and size fields in gatekeeper
We need to defend against invalid data in `checksum` and `size` values in `FileTransferToken`.

Add the following checks:
- `checksum`
    - is a string.
    - is not blank or `None`.
    - contains no whitespace.
    - contains exactly one colon (`:`).
    - The part before the colon must be one of the supported values (create an enum for them). Currently the only supported value is `sha1`.
    - The part after the colon contains exactly 40 characters from the range `a..f`, `A..F`, `0..9`.
- `size`
    - is not blank or `None`.
    - is a value that can be converted to an integer using `int()`.
    - is non-negative (zero is allowed).

Validate these fields for all files, not only the one being uploaded/downloaded. Run then both for upload and download.

bartoszbetka:
Resolves #78

### DIFF
--- a/concent_api/gatekeeper/enums.py
+++ b/concent_api/gatekeeper/enums.py
@@ -1,0 +1,7 @@
+from enum import Enum
+from enum import unique
+
+
+@unique
+class HashingAlgorithm(Enum):
+    SHA1 = 'sha1'
--- a/concent_api/gatekeeper/tests/test_gatekeeper_views.py
+++ b/concent_api/gatekeeper/tests/test_gatekeeper_views.py
@@ -125,235 +125,88 @@ class GatekeeperViewUploadTest(TestCase):
         self.assertEqual("application/json", response["Content-Type"])
 
     @freeze_time("2018-12-30 11:00:00")
-    def test_upload_should_return_401_if_checksum_is_not_string(self):
-        file1 = FileTransferToken.FileInfo(
-            path     = 'blender/benchmark/test_task/scene-Helicopter-27-cycles.blend',
-            checksum = 'sha1:95a0f391c7ad86686ab1366bcd519ba5ab3cce89',
-            size     = 1024,
-        )
+    def test_upload_should_return_401_if_checksum_is_wrong(self):
+        invalid_values = [
+            b'sha1:95a0f391c7ad86686ab1366bcd519ba5ab3cce89',
+            '',
+            '95a0f391c7ad86686ab1366bcd519ba5ab3cce89',
+            ':95a0f391c7ad86686ab1366bcd519ba5ab3cce89',
+            'sha1:95a0f391c7ad86686ab1366bcd519ba5amONj',
+            'sha1:',
+        ]
 
-        file2 = FileTransferToken.FileInfo(
-            path     = 'blenderrr/benchmark/test_task/scene-Helicopter-27-cycles.blend',
-            checksum = b'sha1:95a0f391c7ad86686ab1366bcd519ba5ab3cce89',
-            size     = 1024,
-        )
+        for invalid_value in invalid_values:
+            file1 = FileTransferToken.FileInfo(
+                path     = 'blender/benchmark/test_task/scene-Helicopter-27-cycles.blend',
+                checksum = 'sha1:95a0f391c7ad86686ab1366bcd519ba5ab3cce89',
+                size     = 1024,
+            )
 
-        self.upload_token.files = [file1, file2]
+            file2 = FileTransferToken.FileInfo(
+                path     = 'blenderrr/benchmark/test_task/scene-Helicopter-27-cycles.blend',
+                checksum = invalid_value,
+                size     = 1024,
+            )
 
-        golem_upload_token = dump(self.upload_token, settings.CONCENT_PRIVATE_KEY, settings.CONCENT_PUBLIC_KEY)
-        encrypted_token = b64encode(golem_upload_token).decode()
-        response = self.client.post(
-            '{}{}'.format(
-                reverse('gatekeeper:upload'),
-                'blender/benchmark/test_task/scene-Helicopter-27-cycles.blend'
-            ),
-            HTTP_AUTHORIZATION             = 'Golem ' + encrypted_token,
-            content_type                   = 'application/x-www-form-urlencoded',
-            HTTP_CONCENT_CLIENT_PUBLIC_KEY = self.public_key
-        )
+            self.upload_token.files = [file1, file2]
 
-        self.assertIsInstance(response, JsonResponse)
-        self.assertEqual(response.status_code, 401)
-        self.assertIn('message', response.json().keys())
-        self.assertEqual("application/json", response["Content-Type"])
+            golem_upload_token = dump(self.upload_token, settings.CONCENT_PRIVATE_KEY, settings.CONCENT_PUBLIC_KEY)
+            encrypted_token = b64encode(golem_upload_token).decode()
+            response = self.client.post(
+                '{}{}'.format(
+                    reverse('gatekeeper:upload'),
+                    'blender/benchmark/test_task/scene-Helicopter-27-cycles.blend'
+                ),
+                HTTP_AUTHORIZATION             = 'Golem ' + encrypted_token,
+                content_type                   = 'application/x-www-form-urlencoded',
+                HTTP_CONCENT_CLIENT_PUBLIC_KEY = self.public_key
+            )
 
-    @freeze_time("2018-12-30 11:00:00")
-    def test_upload_should_return_401_if_checksum_is_blank_or_none(self):
-        file1 = FileTransferToken.FileInfo(
-            path     = 'blender/benchmark/test_task/scene-Helicopter-27-cycles.blend',
-            checksum = 'sha1:95a0f391c7ad86686ab1366bcd519ba5ab3cce89',
-            size     = 1024,
-        )
-
-        file2 = FileTransferToken.FileInfo(
-            path     = 'blenderrr/benchmark/test_task/scene-Helicopter-27-cycles.blend',
-            checksum = '',
-            size     = 1024,
-        )
-
-        self.upload_token.files = [file1, file2]
-
-        golem_upload_token = dump(self.upload_token, settings.CONCENT_PRIVATE_KEY, settings.CONCENT_PUBLIC_KEY)
-        encrypted_token = b64encode(golem_upload_token).decode()
-        response = self.client.post(
-            '{}{}'.format(
-                reverse('gatekeeper:upload'),
-                'blender/benchmark/test_task/scene-Helicopter-27-cycles.blend'
-            ),
-            HTTP_AUTHORIZATION             = 'Golem ' + encrypted_token,
-            content_type                   = 'application/x-www-form-urlencoded',
-            HTTP_CONCENT_CLIENT_PUBLIC_KEY = self.public_key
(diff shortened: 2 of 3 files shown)


concent_api/gatekeeper/enum.py
```diff
@@ -0,0 +1,6 @@
+import enum
+
+
+@enum.unique
+class HashingAlgorytm(enum.Enum):
```
cameel: `HashingAlgorithm`

cameel: Also, please rename `enum.py` to `enums.py`. That's the usual convention in Django.

concent_api/gatekeeper/views.py
```diff
@@ -154,6 +158,30 @@ def parse_headers(request: WSGIRequest, path_to_file: str) -> Union[FileTransfer
...
+    if any((file["size"] is None) for file in loaded_golem_message.files):
+        return gatekeeper_access_denied_response('Wrong type of size variable. The variable is empty', path_to_file, loaded_golem_message.subtask_id, client_public_key)
+    file_sizes = [file["size"] for file in loaded_golem_message.files]
+    for file_size in file_sizes:
+        try:
+            int(file_size)
+        except ValueError:
+            return gatekeeper_access_denied_response('Wrong type of size variable. The variable must be digit', path_to_file, loaded_golem_message.subtask_id, client_public_key)
+        except TypeError:
+            return gatekeeper_access_denied_response('Wrong type of size variable. The variable must be integer', path_to_file, loaded_golem_message.subtask_id, client_public_key)
```
cameel: Both of these can have the same error message.

You can catch exceptions like this:

``` python
except (ValueError, TypeError):
```

Maybe even without the parenthesis.

concent_api/gatekeeper/views.py
```diff
@@ -154,6 +158,30 @@ def parse_headers(request: WSGIRequest, path_to_file: str) -> Union[FileTransfer
...
+
+    if any(int(file["size"]) < 0  for file in loaded_golem_message.files):
+        return gatekeeper_access_denied_response('Wrong type of size variable. The variable must contain negative digit', path_to_file, loaded_golem_message.subtask_id, client_public_key)
```
cameel: The message is wrong. The type is OK. It's just that the size cannot be negative.

concent_api/gatekeeper/views.py
```diff
@@ -21,11 +22,14 @@
 from golem_messages.message         import Message
 from golem_messages.shortcuts       import load
 
+from gatekeeper.enum                import HashingAlgorytm
 from gatekeeper.utils               import gatekeeper_access_denied_response
 
 
 logger = logging.getLogger(__name__)
 
+VALID_SHA1_HASH = re.compile(r"^[a-fA-F\d]{40}$")
```
cameel: Call it `VALID_SHA1_HASH_REGEX`.

concent_api/gatekeeper/views.py
```diff
@@ -5,6 +5,7 @@
 import json
 import logging
 from typing                         import Union
+import re
```
cameel: Keep imports sorted alphabetically.

concent_api/gatekeeper/views.py
```diff
@@ -154,6 +158,30 @@ def parse_headers(request: WSGIRequest, path_to_file: str) -> Union[FileTransfer
     if len(transfer_token_paths_to_files) != len(set(transfer_token_paths_to_files)):
         return gatekeeper_access_denied_response('File paths in the token must be unique', path_to_file, loaded_golem_message.subtask_id, client_public_key)
 
+    if any((file["checksum"] is None or len(file["checksum"]) == 0 or file["checksum"].isspace()) for file in loaded_golem_message.files):
+        return gatekeeper_access_denied_response('Wrong type of checksum variable. The variable is empty', path_to_file, loaded_golem_message.subtask_id, client_public_key)
```
cameel: These messages need to be improved. Not every error is a type error :)

concent_api/gatekeeper/views.py
```diff
@@ -154,6 +158,30 @@ def parse_headers(request: WSGIRequest, path_to_file: str) -> Union[FileTransfer
...
+    if any(not isinstance(file["checksum"], str) for file in loaded_golem_message.files):
+        return gatekeeper_access_denied_response('Wrong type of checksum variable. The variable must be string', path_to_file, loaded_golem_message.subtask_id, client_public_key)
```
cameel: This check should be above. Otherwise trying to call `isspace()` may result in an exception.

And if you check if it's `str` you don't have to specifically check for `None`.

concent_api/gatekeeper/views.py
```diff
@@ -154,6 +158,30 @@ def parse_headers(request: WSGIRequest, path_to_file: str) -> Union[FileTransfer
...
+    file_checksums = [tuple(file["checksum"].split(":")) for file in loaded_golem_message.files]
+    if any(file_checksum[0] != hashing_algorytm.value for file_checksum in file_checksums for hashing_algorytm in HashingAlgorytm):
```
cameel: `algorytm` -> `algorithm`

concent_api/gatekeeper/views.py
```diff
@@ -154,6 +158,30 @@ def parse_headers(request: WSGIRequest, path_to_file: str) -> Union[FileTransfer
...
+        return gatekeeper_access_denied_response('Wrong type of checksum variable. The hashing algorytm is not supported', path_to_file, loaded_golem_message.subtask_id, client_public_key)
```
cameel: This could be simpler:

``` python
if any(checksum_type not in HashingAlgorithm.__members__ for checksum_type, checksum in checksums):
```

or

``` python
if not set(checksum_type for checksum_type, checksum in checksums).issubset(set(HashingAlgorithm.__members__)):
```

concent_api/gatekeeper/views.py
```diff
@@ -154,6 +158,30 @@ def parse_headers(request: WSGIRequest, path_to_file: str) -> Union[FileTransfer
     if len(transfer_token_paths_to_files) != len(set(transfer_token_paths_to_files)):
         return gatekeeper_access_denied_response('File paths in the token must be unique', path_to_file, loaded_golem_message.subtask_id, client_public_key)
 
+    if any((file["checksum"] is None or len(file["checksum"]) == 0 or file["checksum"].isspace()) for file in loaded_golem_message.files):
+        return gatekeeper_access_denied_response('Wrong type of checksum variable. The variable is empty', path_to_file, loaded_golem_message.subtask_id, client_public_key)
+    if any(not isinstance(file["checksum"], str) for file in loaded_golem_message.files):
+        return gatekeeper_access_denied_response('Wrong type of checksum variable. The variable must be string', path_to_file, loaded_golem_message.subtask_id, client_public_key)
+    file_checksums = [tuple(file["checksum"].split(":")) for file in loaded_golem_message.files]
```
cameel: You're not checking if the checksum contains a colon before trying to split it.

concent_api/gatekeeper/enum.py
```diff
@@ -0,0 +1,6 @@
+import enum
```
cameel: The preferred style is to import `Enum` and `unique` explicitly from this module.

concent_api/gatekeeper/views.py
bartoszbetka: When I check `str(None)` , it return `'None'`, so it's work properly ?

concent_api/gatekeeper/views.py
bartoszbetka: Object `HashingAlgorithm.__members__` return order dictionary `OrderedDict([('SHA1', <HashingAlgorithm.SHA1: 'sha1'>)])` and not work with yours example I would stay with an earlier solution. What you think ?

concent_api/gatekeeper/enums.py
```diff
@@ -0,0 +1,6 @@
+from enum import Enum, unique
```
cameel: One import per line.

concent_api/gatekeeper/tests/test_gatekeeper_views.py
```diff
@@ -124,6 +124,237 @@ def test_upload_should_return_401_if_file_paths_are_not_unique(self):
         self.assertIn('message', response.json().keys())
         self.assertEqual("application/json", response["Content-Type"])
 
+    @freeze_time("2018-12-30 11:00:00")
```
cameel: All these tests are very repetitive. Could you replace them with a two tests (one for checksums, one for file sizes) where each has a list of valid and invalid values and goes over them in a loop?

concent_api/gatekeeper/views.py
```diff
@@ -154,6 +158,30 @@ def parse_headers(request: WSGIRequest, path_to_file: str) -> Union[FileTransfer
     if len(transfer_token_paths_to_files) != len(set(transfer_token_paths_to_files)):
         return gatekeeper_access_denied_response('File paths in the token must be unique', path_to_file, loaded_golem_message.subtask_id, client_public_key)
 
+    if any(not isinstance(file["checksum"], str) for file in loaded_golem_message.files):
```
cameel: This would be cleared if you did `not all(` instead of `any(not `

concent_api/gatekeeper/views.py
```diff
@@ -154,6 +158,30 @@ def parse_headers(request: WSGIRequest, path_to_file: str) -> Union[FileTransfer
...
+        return gatekeeper_access_denied_response('Wrong type of checksum variable. The variable must be string', path_to_file, loaded_golem_message.subtask_id, client_public_key)
+    if any((len(file["checksum"]) == 0 or file["checksum"].isspace()) for file in loaded_golem_message.files):
+        return gatekeeper_access_denied_response('Wrong value of checksum variable. The variable is empty or contain only space character', path_to_file, loaded_golem_message.subtask_id, client_public_key)
+    if any(":" not in file["checksum"] for file in loaded_golem_message.files):
```
cameel: This would be clearer if you did `not all(` instead of `any(not `

concent_api/gatekeeper/views.py
```diff
@@ -154,6 +158,30 @@ def parse_headers(request: WSGIRequest, path_to_file: str) -> Union[FileTransfer
...
+    file_checksums = [tuple(file["checksum"].split(":")) for file in loaded_golem_message.files]
+    if not set(checksum_type for checksum_type, checksum in file_checksums).issubset(set(HashingAlgorithm._value2member_map_)):
+        return gatekeeper_access_denied_response('Wrong type of checksum variable. The hashing algorithm is not supported', path_to_file, loaded_golem_message.subtask_id, client_public_key)
+    if any(VALID_SHA1_HASH_REGEX.match(file_checksum[1]) is None for file_checksum in file_checksums):
+        return gatekeeper_access_denied_response('Wrong type of checksum variable. The hash is incorrect', path_to_file, loaded_golem_message.subtask_id, client_public_key)
+
+    if any((file["size"] is None) for file in loaded_golem_message.files):
+        return gatekeeper_access_denied_response('Wrong type of size variable. The variable is empty', path_to_file, loaded_golem_message.subtask_id, client_public_key)
+    file_sizes = [file["size"] for file in loaded_golem_message.files]
```
cameel: You should use `file_sizes` in the `if` above.

Issue #110, opened 2018-02-09
[force subtask results] Don't allow messages other than accept and reject inside ForceSubtaskResultsResponse


rwrzesien:
In send() function there is a piece of code

```
elif isinstance(client_message, message.concents.ForceSubtaskResultsResponse):
    return handle_send_force_subtask_results_results_response(request, client_message)
```

and later inside that function

```
    if isinstance(client_message.subtask_results_accepted, message.tasks.SubtaskResultsAccepted):
        client_message_task_id = client_message.subtask_results_accepted.subtask_id
    else:
        client_message_task_id = client_message.subtask_results_rejected.report_computed_task.subtask_id
```

It will fail if message doesn't contain either subtask_results_accepted or subtask_results_rejected.

Please add check if one of those is present in client_message inside existing if in send() function.

PR #112, opened 2018-02-12
Fix dont allow messages other than accept and reject inside force subtask results response

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/110.

### DIFF
--- a/concent_api/core/tests/test_integration_accept_or_reject_results.py
+++ b/concent_api/core/tests/test_integration_accept_or_reject_results.py
@@ -1227,3 +1227,72 @@ class GetTaskResultIntegrationTest(ConcentIntegrationTestCase):
                 )
 
         self._test_400_response(response_2)
+
+    def test_requestor_send_subtask_results_without_accepted_or_rejected_should_return_http_400(self):
+        """
+        Test if Requestor wants to send ForceSubtaskResultsResponse without SubtaskResultsAccepted
+        or SubtaskResultsRejected then Concent should return HTTP 400 as not supported.
+
+        Expected message exchange:
+        Requestor   -> Concent: ForceSubtaskResultsResponse (empty)
+        Concent     -> Requestor: HTTP 400
+        """
+        deserialized_force_subtask_results = self._get_deserialized_force_subtask_results(
+            timestamp                = "2018-02-05 10:00:15",
+            ack_report_computed_task = self._get_deserialized_ack_report_computed_task(
+                subtask_id = "xxyyzz",
+                task_to_compute = self._get_deserialized_task_to_compute(
+                    timestamp = "2018-02-05 10:00:00",
+                    deadline  = "2018-02-05 10:00:10",
+                    task_id   = '2',
+                )
+            )
+        )
+
+        self._store_golem_messages_in_database(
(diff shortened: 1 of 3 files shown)


concent_api/core/tests/test_integration_accept_or_reject_results.py
```diff
@@ -1117,7 +1117,7 @@ def test_requstor_send_again_subtask_results_accepted_or_rejected_when_message_a
 
         self._test_400_response(response_2)
 
-    def test_requstor_send_again_subtask_results_accepted_or_rejected_when_message_already_rejected_concent_should_return_http_400(self):
+    def test_requestor_send_again_subtask_results_accepted_or_rejected_when_message_already_rejected_concent_should_return_http_400(self):
```
Jakub89: Maybe typo fixes should be separated commit? I think @cameel would point that.

rwrzesien: :+1:

concent_api/core/views.py
```diff
@@ -43,7 +43,10 @@ def send(request, client_message):
     elif isinstance(client_message, message.concents.ForceSubtaskResults) and client_message.ack_report_computed_task is not None:
         return handle_send_force_subtask_results(request, client_message)
 
-    elif isinstance(client_message, message.concents.ForceSubtaskResultsResponse):
+    elif (
+        isinstance(client_message, message.concents.ForceSubtaskResultsResponse) and
+        (client_message.subtask_results_accepted is not None or client_message.subtask_results_rejected is not None)
```
cameel: Shouldn't you also use `isinstance()` on `subtask_results_accepted` and `subtask_results_rejected`? What happens if they're different types of messages? Or not messages at all?

rwrzesien: @cameel As I thought, it will then fail during load(). Here is the output from console:

```
TypeError: Invalid nested message type <class 'int'> should be <class 'golem_messages.message.tasks.SubtaskResultsRejected'>
FieldError:  [subtask_results_rejected:1]
TypeError: Invalid nested message type <class 'golem_messages.message.concents.ServiceRefused'> should be <class 'golem_messages.message.tasks.SubtaskResultsRejected'>
FieldError:  [subtask_results_rejected:ServiceRefused(header=MessageHeader(type_=4000, timestamp=1518447887, encrypted=False), sig=b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00', slots=[['subtask_id', None], ['task_to_compute', None], ['reason', None]])]

```

cameel: OK. So we need to detect it before it happens.

As a general rule, the client should not be able to cause HTTP 500 by simply sending us invalid data.

rwrzesien: We are catching this in `api_views.py`.

Issue #106, opened 2018-02-09
The implementation of 'force subtask results' use case does not leave the requestor enough time for verification


cameel:
The documentation states that the provider can send `ForceSubtaskResults` message only after the time window in which requestor is supposed to accept or reject subtask results elapses. The correct time window starts at `AckReportComputedTask.timestamp` + `SUBTASK_VERIFICATION_TIME` and ends at `AckReportComputedTask.timestamp` + `SUBTASK_VERIFICATION_TIME` + `FORCE_ACCEPTANCE_TIME`.

The requestor should have a time window equal to `SUBTASK_VERIFICATION_TIME` before the provider can act but the current implementation only gives him only `CONCENT_MESSAGING_TIME` which is always much shorter and does not leave enough time for verification.

See this comment and the responses: https://github.com/golemfactory/concent/issues/41#issuecomment-364406657

PR #113, opened 2018-02-12
Fix for giving requestor enough time for verification and added SUBTA…

rwrzesien:
…SK_VERIFICATION_TIME to settings

Resolves https://github.com/golemfactory/concent/issues/106/

### DIFF
--- a/concent_api/core/views.py
+++ b/concent_api/core/views.py
@@ -421,8 +421,8 @@ def handle_send_force_subtask_results(request, client_message: message.concents.
                 reason      = message.concents.ServiceRefused.REASON.TooSmallProviderDeposit,
             )
 
-    client_message_send_to_soon = client_message.ack_report_computed_task.timestamp + settings.CONCENT_MESSAGING_TIME
-    client_message_send_to_late = client_message.ack_report_computed_task.timestamp + settings.CONCENT_MESSAGING_TIME + settings.FORCE_ACCEPTANCE_TIME
+    client_message_send_to_soon = client_message.ack_report_computed_task.timestamp + settings.SUBTASK_VERIFICATION_TIME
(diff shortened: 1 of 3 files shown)


concent_api/core/views.py
```diff
@@ -421,8 +421,8 @@ def handle_send_force_subtask_results(request, client_message: message.concents.
                 reason      = message.concents.ServiceRefused.REASON.TooSmallProviderDeposit,
             )
 
-    client_message_send_to_soon = client_message.ack_report_computed_task.timestamp + settings.CONCENT_MESSAGING_TIME
-    client_message_send_to_late = client_message.ack_report_computed_task.timestamp + settings.CONCENT_MESSAGING_TIME + settings.FORCE_ACCEPTANCE_TIME
+    client_message_send_to_soon = client_message.ack_report_computed_task.timestamp + settings.SUBTASK_VERIFICATION_TIME
+    client_message_send_to_late = client_message.ack_report_computed_task.timestamp + settings.SUBTASK_VERIFICATION_TIME + settings.FORCE_ACCEPTANCE_TIME
```
cameel: I have not noticed this earlier but you have an error in your names. It should be `..._too_soon` and `..._too_late`.

cameel:
Is that it? Are you sure we don't need to change the times in any other places?

concent_api/core/views.py
Jakub89: ups, my mistake.

rwrzesien: :+1:  I have fixed it.

rwrzesien:
> Is that it? Are you sure we don't need to change the times in any other places?

For this one, yes, thanks to good implementation by @Jakub89 we only check it in single place. The second issue for time windows require some more work.

PR #117, opened 2018-02-13
Fix api tests for cluster

### LINKED ISSUE #100: `api-integration-force-get-task-result-test.py` doesn't upload new file.
Right now script check file that's already uploaded. It should upload new file to cluster first, and then check if it was able to do that.
Curl that uploads file looks like:
```
curl -v -H "Authorization: Golem D6UAAAAAWmsOpAEoIMBNoHjDKvhObrK/bstagWtl5tiwu+yM8sR/mBRiY3f9+e9gHP0rBRoQjh8uHwklnBLciAc1frr9nNTKD8ufAQRlpFa6bJV3e5JiHSBEu7uu5RqrfZNUUV6tYAUtX2TgXIvdpmi4URBGgDOTAcCyB1qkNNaYuEOZw2bX2KDBanUEdvGhy+9EgcHNnoWo9GS51sm1W9pwAQpHLwHfvFqeC2VtbftYewSjoVm4GHMS50/YC2GE9PyEwWhtILEvSX2O/fJy09LPpLqRyftAIutEkWohwL/Z7FM5c5TQDlhKY91YcM3lp5jPVDn2QNk3NeBNo2UnuLXuc3zOWixzjUKJW5W/mT5ftwvCQh8sLjzgY/u2uY8tZ7ttvWANso5tAknqwRTACe+EmhdnwuSjWDRKS2u+6rZHp9FbIDFLIMxrMme9yOBYOqBPX1Uteg68jJAYMixa8q/sWxwYZXXDABpo56LXhTKYzvEChd5wfTDeqs1U4srObVkaj9h3TUCwXc9fikPjyhurOcmvA9CV86lRdAuuXkz6suY38S7u6+EAKXXv4PrRROyCV3vNlAVwxAmZjNjtHVKEWT6b81u0hZ6N10D7H+t/w4hG4/A47uj45s05YyWTXoTGjfvihlBG" 
-H "Concent-Client-Public-Key: 85cZzVjahnRpUBwm0zlNnqTdYom1LF1P1WNShLg17cmhN2UssnPrCjHKTi5susO3wrr/q07eswumbL82b4HgOw==" 
-X POST --data-binary "@result" http://devel.concent.golem.network/upload/ 
-H "Concent-upload-path: 3/0/result"
```
To scipt should be added:
- create locally new file with name: `result`
- check file checksum (if it starts with `sha1:`)
- check file size
- upload file with path consisted of `task_id/part_id/result` (example above in "Concent-upload-path")
- check if file is correctly uploaded
- delete local file

kbeker:
Resolves #100

### DIFF
--- a/concent_api/api-integration-force-accept-or-reject-results-test.py
+++ b/concent_api/api-integration-force-accept-or-reject-results-test.py
@@ -5,6 +5,7 @@ import sys
 import random
 import time
 from base64                 import b64encode
+from freezegun              import freeze_time
 
 from golem_messages         import message
 
@@ -12,6 +13,7 @@ from utils.helpers          import get_current_utc_timestamp
 from utils.testing_helpers  import generate_ecc_key_pair
 
 from api_testing_helpers    import api_request
+from api_testing_helpers    import timestamp_to_isoformat
 
 import requests
 
@@ -32,23 +34,26 @@ def parse_command_line(command_line):
     return cluster_url
 
 
-def force_subtask_results(ack_report_computed_task = None):
-    return message.concents.ForceSubtaskResults(
-        ack_report_computed_task = ack_report_computed_task,
-    )
+def force_subtask_results(timestamp = None, ack_report_computed_task = None):
+    with freeze_time(timestamp):
+        return message.concents.ForceSubtaskResults(
+            ack_report_computed_task = ack_report_computed_task,
+        )
 
 
-def ack_report_computed_task(subtask_id = None, task_to_compute = None):
-    return message.concents.AckReportComputedTask(
-        task_to_compute = task_to_compute,
-        subtask_id      = subtask_id,
-    )
+def ack_report_computed_task(timestamp = None, subtask_id = None, task_to_compute = None):
+    with freeze_time(timestamp):
+        return message.concents.AckReportComputedTask(
+            task_to_compute = task_to_compute,
+            subtask_id      = subtask_id,
+        )
 
 
-def task_to_compute(compute_task_def = None):
-    return message.tasks.TaskToCompute(
-        compute_task_def = compute_task_def,
-    )
+def task_to_compute(timestamp = None, compute_task_def = None):
+    with freeze_time(timestamp):
+        return message.tasks.TaskToCompute(
+            compute_task_def = compute_task_def,
+        )
 
 
 def compute_task_def(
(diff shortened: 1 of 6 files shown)


concent_api/api-integration-force-accept-or-reject-results-test.py
```diff
@@ -64,82 +69,90 @@ def compute_task_def(
...
+        force_subtask_results(
+            timestamp = get_strftime_from_timestamp(current_time),  # current_time
+            ack_report_computed_task = ack_report_computed_task(
+                timestamp = get_strftime_from_timestamp(current_time - 3601),  # current_time - 1:00:01
+                subtask_id = str(task_id) + '2A',
+                task_to_compute = task_to_compute(
+                    timestamp = get_strftime_from_timestamp(current_time - 3605),  # current_time - 1:00:05
+                    compute_task_def = compute_task_def(
+                        task_id     = str(task_id) + '2A',
+                        subtask_id  = str(task_id) + '2A',
```
rwrzesien: Please do that line 106 and reuse already prepared variable in all `task_id` and `subtask_id`.

concent_api/api-integration-force-get-task-result-test.py
```diff
@@ -125,8 +175,9 @@ def main():
         PROVIDER_PRIVATE_KEY,
         CONCENT_PUBLIC_KEY,
         headers = {
-            'Content-Type': 'application/octet-stream',
-            'concent-client-public-key': b64encode(PROVIDER_PUBLIC_KEY).decode('ascii')
+            'Content-Type':                     'application/octet-stream',
+            'concent-client-public-key':        b64encode(PROVIDER_PUBLIC_KEY).decode('ascii'),
+            'concent-other-party-public-key':   b64encode(REQUESTOR_PUBLIC_KEY).decode('ascii')
```
rwrzesien: You don't have to pass 'concent-other-party-public-key' to `receive`, only to `send`, and only when communication starts (first message in each use case).

concent_api/api-integration-force-get-task-result-test.py
```diff
@@ -32,6 +33,44 @@ def parse_command_line(command_line):
     return cluster_url
 
 
+def upload_new_file_on_cluster(task_id = '0', part_id = '0', current_time = 0):
+
+    os.system('touch result')
+    os.system('echo "{}" > result'.format(task_id))
+    file_size       = int(os.popen("wc -c < result").readlines()[0])
```
rwrzesien:
```
import os
os.path.getsize(file_name)
```

would be more pythonic.

concent_api/api-integration-force-get-task-result-test.py
```diff
@@ -32,6 +33,44 @@ def parse_command_line(command_line):
...
+    file_check_sum  = 'sha1:' + os.popen("sha1sum result | awk '{print $1}'").readlines()[0][:-1]
```
rwrzesien:
```
import hashlib
with open(file_name, 'rb') as f:
    file_check_sum = hashlib.sha1(f.read()).hexdigest()
```

concent_api/api-integration-force-get-task-result-test.py
```diff
@@ -32,6 +33,44 @@ def parse_command_line(command_line):
...
+    file_transfer_token.files[0]['path']        = file_path
+    file_transfer_token.files[0]['checksum']    = file_check_sum
+    file_transfer_token.files[0]['size']        = file_size
+
+    upload_token    = shortcuts.dump(file_transfer_token, CONCENT_PRIVATE_KEY, CONCENT_PUBLIC_KEY)
+    encrypted_token = b64encode(upload_token).decode()
+
+    authorized_golem_transfer_token = 'Golem ' + encrypted_token
+
+    file = open('result', 'rb').read()
```
rwrzesien: You can attach file to request without physically creating it, please check 3rd example from
http://docs.python-requests.org/en/latest/user/quickstart/#post-a-multipart-encoded-file

concent_api/core/views.py
```diff
@@ -854,7 +854,7 @@ def get_file_status(file_transfer_token_from_database: message.concents.FileTran
     )
     if cluster_storage_response.status_code == 200:
         return True
-    elif cluster_storage_response.status_code == 404:
+    elif cluster_storage_response.status_code == 404 or 401:
```
rwrzesien: It won't work this way. This will always return True. It will first check bool(cluster_storage_response.status_code == 404), and then do bool(401) which is True. Please do:
`elif cluster_storage_response.status_code in [401, 404]`

concent_api/api-integration-force-get-task-result-test.py
cameel: Yeah. Hundred times this. Please no `popen()`, `system()` or any other calls that spawn a separate process unless your fully aware of all the consequences of this. We should probably add this to coding style as a red flag that something is fishy.

As a general rule, if you're embedding bash commands in your python code, 99% of the time you're doing something wrong.

concent_api/api_testing_helpers.py
```diff
@@ -80,5 +81,9 @@ def api_request(host, endpoint, private_key, public_key, data = None, headers =
     print()
 
 
+def get_strftime_from_timestamp(timestamp):
+    return datetime.datetime.fromtimestamp(timestamp).strftime("%Y-%m-%d %H:%M:%S")
```
cameel: The name is misleading. Your function does not work like `strftime()` (which accepts an arbitrary format string, hence the `strf` in the name). It works more like [`isoformat()`](https://docs.python.org/3.6/library/datetime.html#datetime.datetime.isoformat), which uses a very specific format.

Please rename it to `timestamp_to_isoformat()`.

Actually, why not use `isoformat()` here?

kbeker: I'm using here strftime becuase I need time converted in `"%Y-%m-%d %H:%M:%S"` to freeze time to create message in specific time

cameel:
How is this pull request related to #100?

concent_api/api_testing_helpers.py
cameel: `strftime` does not freeze time. It just formats it.

And your format is exactly like the one `isoformat()` supports.

kbeker:
@cameel 
> How is this pull request related to #100?

This pull request adds function to upload new file. Also in this branch I repaird tests after Concent update v0.3

concent_api/api-integration-force-get-task-result-test.py
```diff
@@ -56,24 +94,36 @@ def get_force_get_task_result(task_id, current_time, size, checksum):
 
 
 def main():
-    cluster_url  = parse_command_line(sys.argv)
-    current_time = get_current_utc_timestamp()
-
+    cluster_url     = parse_command_line(sys.argv)
+    current_time    = get_current_utc_timestamp()
+    random_id       = random.randrange(1, 100000)
+    task_id         = str(random_id)
```
rwrzesien: Please combine those two into single line.

kbeker:
Change with task_id commited

concent_api/api_testing_helpers.py
```diff
@@ -80,5 +81,9 @@ def api_request(host, endpoint, private_key, public_key, data = None, headers =
     print()
 
 
+def timestamp_to_isoformat(timestamp):
+    return datetime.datetime.fromtimestamp(timestamp).isoformat()
```
cameel: You forgot about the argument. By default it inserts a `T` between the date and time.

`isoformat(' ')`.

Please make sure the output of this function is tested in some way.

kbeker: I tested it before and it was working this way, but now I changed it to: 
`return datetime.datetime.fromtimestamp(timestamp).isoformat(' ')`

Issue #111, opened 2018-02-09
[force subtask results] Add case 4A from blueprint with mocked payment API


rwrzesien:
Currently case 4A from https://github.com/golemfactory/concent/issues/41 is not implemented. In current implementation, when requestor is trying to receive `ForceSubtaskResultsResponse` after time passed, he will receive original `ForceSubtaskResults`. This is not ok and it should be treated as he doesn't provide response in during correct time window. So it would be best to add whole 4A with mocked payment API calls.

Please add:
* implementation 
* integration tests (please consider different responses from payments API)
* auth tests
* cluster tests

There is also test that checks current implementation that has to be updated:
```
test_requestor_should_not_receive_correct_subtask_results_from_concent_if_asked_concent_after_deadline
```

We should always return new message with new timestamp. Please update implementation and tests by removing checking time condition.

rwrzesien:
Updated description on how we handle receive ForceSubtaskResults outside time window.

Issue #75, opened 2018-01-29
Logic for requests for "Provider forces requestor to accept or reject results" use case in views


rwrzesien:
Add support for the message exchange defined in https://github.com/golemfactory/concent/issues/41.

**We are not doing 4A for now.**

General guidelines:
* Payment API should be mocked inside Concent code. Mock should allow to query for several users, at least as many to cover all cases.
* Follow existing strucute of `views.py`. 
* Do the basic logic first, then add identifying by key (it should be available at some point during implementation).
* Cases 1,2,3 are very similar to what was for ForceReportComputedTask case, most of the logic can be reused.
* Cases 4A, 4B, 5 are very similar to what was for ForceReportComputedTask which resulted in VerdictReportComputedTask.

Probably no changes in models will be needed for this issue. However, following objects should be created on each case:

* Step 1+ 2A: Concent rejects provider's request immediately if task was already submitted:
** No new objects created here.

* Step 1+ 2B: Concent rejects provider's request immediately if provider doesn't have enough funds on his account:
** No new objects created here.

* Step 1+ 2C: Concent rejects provider's request immediately if it was sent too soon or too late:
** No new objects created here.

* Step 1 + 2D: Concent accepts provider's request:
** Message object with ForceSubtaskResults data.
** MessageAuth objects with used keys for created Message.
** ReceiveStatus for receive for created Message.

* Step 3: Concent requests subtask results from requestor:
** Marks Messages from step 1 + 2D as delivered.

* Step 4A: Requestor doesn't provide response:
** Two message objects with SubtaskResultsSettled data.
** MessageAuth object with used keys for each created Message..
** ReceiveStatus for each created Message.

* Step 4B: Requestor responds to Concent with subtask results acceptance or rejection:
** Message objects with ForceSubtaskResultsResponse data.
** MessageAuth object with used keys for created Message..
** ReceiveStatus for created Message.

* Step 5: Concent notifies provider about subtask results acceptance or rejection:
** Marks Messages from step 4B as delivered.

Issue #41, opened 2018-01-15
[Blueprint] "Provider forces requestor to accept or reject results" use case in Concent API


rwrzesien:

rwrzesien:
@lukasz-glen I have several questions regarding this use case:
1. "Usługa Concent bez komunikacji z Requestorem sprawdza czy komunikat MessageTaskToCompute (zawarty w komunikacie MessageAckReportComputedTask) nie był już wcześniej procesowany." - what does it mean 'procesowany' in this context ?
2. What about scenario when provider sends ForceSubtaskResults for the same task again ? It is not documented, however for previous use cases when we did this scenario, it was also not  documented.
3. What will be in field 'payment_ts' in SubtaskResultsAccepted ?

cameel:
I think I can answer some of these:

> 1) "Usługa Concent bez komunikacji z Requestorem sprawdza czy komunikat MessageTaskToCompute (zawarty w komunikacie MessageAckReportComputedTask) nie był już wcześniej procesowany." - what does it mean 'procesowany' in this context ?

It means that for example if the client sends us `ForceSubtaskResults`, we only accept it if we have never accepted `ForceSubtaskResults` with the same `TaskToCompute` in `send()` before. We'll know that because we store information about all accepted tasks in the database.

> 2) What about scenario when provider sends ForceSubtaskResults for the same task again ? It is not documented, however for previous use cases when we did this scenario, it was also not documented.

The passage you mentioned in 1) answers that.

Jakub89:
According to new changes in documentation, we need to change step 2A. From `ForceSubtaskResultsRejected` to `ServiceRefused`.
```
Usługa Concent bez komunikacji z Requestorem sprawdza czy komunikat MessageTaskToCompute (zawarty w komunikacie MessageAckReportComputedTask) nie był już wcześniej procesowany. Jeśli tak to odpowiada Providerowi komunikatem MessageServiceRefused z podaniem przyczyny i Provider powinien zakończyć komunikację.
```

rwrzesien:
@Jakub89 I have updated blueprints and other related docs.

@lukasz-glen Which `reason` should we use in case 2A ?
Currently there are following values for `reason` field available in ServiceRefused (golem_messages version 1.7.0)
```
    class REASON(enum.Enum):
        TooSmallCommunicationPayment = 'TOO_SMALL_COMMUNICATION_PAYMENT'
[...]

PR #119, opened 2018-02-13
Add ForceSubtaskResults case 4A

Jakub89:
Resolves: #111 
Resolves: #75

### DIFF
--- a/concent_api/core/tests/test_integration_accept_or_reject_results.py
+++ b/concent_api/core/tests/test_integration_accept_or_reject_results.py
@@ -22,6 +22,10 @@ def _get_provider_account_status_false_mock(_):
     return False
 
 
+def _get_requestor_account_status(_provider, _requestor):
+    pass
+
+
 @override_settings(
     CONCENT_PRIVATE_KEY       = CONCENT_PRIVATE_KEY,
     CONCENT_PUBLIC_KEY        = CONCENT_PUBLIC_KEY,
@@ -1079,13 +1083,14 @@ class GetTaskResultIntegrationTest(ConcentIntegrationTestCase):
             }
         )
 
-        with freeze_time("2018-02-05 10:00:51"):
-            response_3 = self.client.post(
-                reverse('core:receive'),
-                data                            = '',
-                content_type                    = 'application/octet-stream',
-                HTTP_CONCENT_CLIENT_PUBLIC_KEY  = self._get_encoded_provider_public_key(),
-            )
+        with mock.patch('core.views.make_forced_payment', _get_requestor_account_status):
+            with freeze_time("2018-02-05 10:00:51"):
+                response_3 = self.client.post(
+                    reverse('core:receive'),
+                    data                            = '',
+                    content_type                    = 'application/octet-stream',
+                    HTTP_CONCENT_CLIENT_PUBLIC_KEY  = self._get_encoded_provider_public_key(),
+                )
         self._test_response(
             response_3,
             status       = 200,
--- a/concent_api/core/tests/test_integration_accept_or_reject_results.py
+++ b/concent_api/core/tests/test_integration_accept_or_reject_results.py
@@ -1025,7 +1025,9 @@ class GetTaskResultIntegrationTest(ConcentIntegrationTestCase):
         Concent     -> Requestor:   ForceSubtaskResults
         Requestor   -> Concent:     no response
         Concent     -> Provider:    SubtaskResultsSettled
-        Concent     -> Requestor    SubtaskResultsSettled
+        Concent     -> Provider:    HTTP 204
+        Concent     -> Requestor:   SubtaskResultsSettled
+        Concent     -> Requestor:   HTTP 204
         """
 
         serialized_force_subtask_results = self._get_serialized_force_subtask_results(
@@ -1085,14 +1087,14 @@ class GetTaskResultIntegrationTest(ConcentIntegrationTestCase):
 
         with mock.patch('core.views.make_forced_payment', _get_requestor_account_status):
             with freeze_time("2018-02-05 10:00:51"):
-                response_3 = self.client.post(
+                response_3a = self.client.post(
                     reverse('core:receive'),
                     data                            = '',
                     content_type                    = 'application/octet-stream',
                     HTTP_CONCENT_CLIENT_PUBLIC_KEY  = self._get_encoded_provider_public_key(),
                 )
         self._test_response(
-            response_3,
+            response_3a,
             status       = 200,
             key          = self.PROVIDER_PRIVATE_KEY,
             message_type = message.concents.SubtaskResultsSettled,
@@ -1102,15 +1104,25 @@ class GetTaskResultIntegrationTest(ConcentIntegrationTestCase):
             }
         )
 
+        with mock.patch('core.views.make_forced_payment', _get_requestor_account_status):
+            with freeze_time("2018-02-05 10:00:52"):
+                response_3b = self.client.post(
+                    reverse('core:receive'),
+                    data                            = '',
+                    content_type                    = 'application/octet-stream',
+                    HTTP_CONCENT_CLIENT_PUBLIC_KEY  = self._get_encoded_provider_public_key(),
+                )
+        self._test_204_response(response_3b)
+
         with freeze_time("2018-02-05 10:00:51"):
-            response_4 = self.client.post(
+            response_4a = self.client.post(
                 reverse('core:receive_out_of_band'),
                 data                            = '',
                 content_type                    = 'application/octet-stream',
                 HTTP_CONCENT_CLIENT_PUBLIC_KEY  = self._get_encoded_requestor_public_key(),
             )
         self._test_response(
-            response_4,
+            response_4a,
             status       = 200,
             key          = self.REQUESTOR_PRIVATE_KEY,
             message_type = message.concents.SubtaskResultsSettled,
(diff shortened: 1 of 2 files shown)


concent_api/core/tests/test_integration_accept_or_reject_results.py
```diff
@@ -1013,6 +1013,108 @@ def test_full_requestor_and_provider_communication_concent_should_accept_message
...
+        )
+        self._test_response(
+            response_2,
+            status       = 200,
+            key          = self.REQUESTOR_PRIVATE_KEY,
+            message_type = message.concents.ForceSubtaskResults,
+            fields       = {
+                'timestamp':                                            self._parse_iso_date_to_timestamp("2018-02-05 10:00:20"),
+                'ack_report_computed_task.subtask_id':                  'xxyyzz',
+                'ack_report_computed_task.task_to_compute.compute_task_def': deserialized_compute_task_def,
```
kbeker: Data on right are not aligned

Jakub89: :+1:

concent_api/core/tests/test_integration_accept_or_reject_results.py
```diff
@@ -1013,6 +1013,108 @@ def test_full_requestor_and_provider_communication_concent_should_accept_message
...
+        Concent     -> Requestor:   ForceSubtaskResults
+        Requestor   -> Concent:     no response
+        Concent     -> Provider:    SubtaskResultsSettled
+        Concent     -> Requestor    SubtaskResultsSettled
+        """
+
+        serialized_force_subtask_results = self._get_serialized_force_subtask_results(
+            timestamp                   = "2018-02-05 10:00:20",
+            ack_report_computed_task    = self._get_deserialized_ack_report_computed_task(
+                timestamp                   = "2018-02-05 10:00:20",
```
rwrzesien: Please align.

concent_api/core/views.py
```diff
@@ -69,6 +69,7 @@ def receive(request, _message):
         if last_delivered_message_status is None:
             return None
 
+        decoded_message_data = deserialize_message(last_delivered_message_status.message.data.tobytes())
```
rwrzesien: Please move to line 78 - it is not used before that line.

concent_api/core/views.py
```diff
@@ -716,6 +757,33 @@ def handle_receive_out_of_band_reject_report_computed_task(request, undelivered_
     return message_verdict
 
 
+def handle_receive_out_of_band_force_subtask_results(
```
rwrzesien: This function and `handle_receive_delivered_force_subtask_results` only differs by status passed to `store_message_and_message_status`. Please combine them into single function with status as parameter.

concent_api/core/tests/test_integration_accept_or_reject_results.py
Jakub89: :+1:

concent_api/core/views.py
Jakub89: :+1:

concent_api/core/views.py
Jakub89: :+1:

Jakub89:
The blueprint said " If requestor doesn't have enough funds on his deposit, only the available amount is payed." I didn't mock it anywhere, should I?

Jakub89:
I've rebase this branch on Karols " fix-api-tests-for-cluster", and I'm gonna add cluster test to this pull request.

concent_api/core/views.py
```diff
@@ -505,6 +521,40 @@ def handle_receive_delivered_force_report_computed_task(request, delivered_messa
...
 
 
+def handle_receive_force_subtask_results(
+    request,
+    decoded_message:                message.concents.ForceSubtaskResults,
+    previous_message_from_database: Union[ReceiveStatus, Message],
+    message_model
+) -> message.concents.ForceSubtaskResults:
+
+    assert decoded_message.TYPE in message.registered_message_types
```
cameel: You can be more specific here:

``` python
assert decoded_message.TYPE == message.concents.ForceSubtaskResults.TYPE
```

Your could also use `isinstance()` instead.

concent_api/core/views.py
```diff
@@ -505,6 +521,40 @@ def handle_receive_delivered_force_report_computed_task(request, delivered_messa
     return ack_report_computed_task
 
 
+def handle_receive_force_subtask_results(
+    request,
+    decoded_message:                message.concents.ForceSubtaskResults,
+    previous_message_from_database: Union[ReceiveStatus, Message],
```
cameel: Using the same parameter for both a message and a status makes things very confusing and error prone. The name says it's a message but it might be something entirely different. Please replace this with two separate arguments with distinct types. If one is specified, the other should be `None`.

Or, instead of passing two different objects just pass the key because that's the only piece of data you use and it's the same in both.

concent_api/core/views.py
```diff
@@ -93,6 +94,14 @@ def receive(request, _message):
                     decoded_message_data,
                     last_delivered_message_status,
                 )
+        subtask_results_settled_timing = settings.SUBTASK_VERIFICATION_TIME + settings.FORCE_ACCEPTANCE_TIME + settings.CONCENT_MESSAGING_TIME
```
cameel: Call it `acceptance_deadline`.

concent_api/core/views.py
```diff
@@ -93,6 +94,14 @@ def receive(request, _message):
...
+        if (
+            last_delivered_message_status.message.type                           == message.concents.ForceSubtaskResults.TYPE and
+            last_delivered_message_status.message.auth.provider_public_key_bytes == client_public_key and
+            current_time > decoded_message_data.ack_report_computed_task.timestamp + subtask_results_settled_timing
```
cameel: I think I have written somewhere before that we should look at the ack timestamp but after going through the docs recently I see that this should be the timestamp from `ReportComputedTask`. If there are more places to change than just this one, you can submit a separate ticket and leave this here as is.

concent_api/core/views.py
```diff
@@ -884,3 +934,7 @@ def tmp_is_provider_account_status_positive(request):
         return bool(request.META['HTTP_TEMPORARY_ACCOUNT_FUNDS'])
     except KeyError:
```
cameel: Don't catch `KeyError`. use `in` operator instead.

concent_api/core/views.py
Jakub89: I've change it to `isinstance()`.

concent_api/core/views.py
Jakub89: Right, I've pick second solution. Passing key to `handle_receive_force_subtask_results()`.

concent_api/core/views.py
Jakub89: :+1:

concent_api/core/views.py
Jakub89: This whole function was not added by me, but I made the fix.

concent_api/core/views.py
Jakub89: Yep, it's used in couple places, just as #41 suggested, so I'm not changing it right now.

concent_api/core/views.py
```diff
@@ -93,6 +94,14 @@ def receive(request, _message):
...
                     last_delivered_message_status,
                 )
+        acceptance_deadline = settings.SUBTASK_VERIFICATION_TIME + settings.FORCE_ACCEPTANCE_TIME + settings.CONCENT_MESSAGING_TIME
+        if (
+            last_delivered_message_status.message.type                           == message.concents.ForceSubtaskResults.TYPE and
+            last_delivered_message_status.message.auth.provider_public_key_bytes == client_public_key and
+            current_time > decoded_message_data.ack_report_computed_task.timestamp + acceptance_deadline
+        ):
+            make_forced_payment('provider', 'requestor')
+            return handle_receive_force_subtask_results(request, decoded_message_data, last_delivered_message_status.message.auth.provider_public_key_bytes, ReceiveStatus)
```
rwrzesien: In this usage the view is called by provider, but you are storing client_public_key as requestor key in 543.

cameel:
There's one very important thing you need to understand here: the `receive/` and `receive-out-of-band/` endpoints are not part of the state machine. Conceptually they do not and cannot have any effect on the processing. They're just means for the client to receive messages that have already been added to receive queue. Whether the client really gets a message is irrelevant to us. In a P2P network the message could just get lost in transit and we can't do anything about it.

There are rare cases where hitting receive triggers something that can result in a message being added to the receive queue and returned but it's just a shortcut to avoid having an active component that checks the state all the time. For example, instead of doing something when a timeout is exceeded we postpone doing it until the client calls `receive/` because we know that the result will be the same anyway.

In short: it only matters what `send/` gets from the clients and what we decide to send back and add to the receive queue. Not when of if the client comes to us to get it and if he actually gets it.

> In current implementation if requestor hits `\receive` after time window described in `Step 3: Concent requests subtask results from requestor.`

It does not matter when he hits `receive/`. If there's a message in the queue, he should get it (with a current timestamp) but it won't affect the deadline or task state in any way.

> in https://github.com/golemfactory/concent/issues/41 he receives original `ForceSubtaskResults` posted by provider (with old/original timestamp).

This has changed in the spec some time ago. We must never resend the original force message. We should create one from scratch.

> I think in this case it should behave like moving to step 4A, which means forcing payment, adding SubtaskResultsSettled for both requestor and provider and returning them.

The only thing that matters is whether the requestor submitted accept or reject in time. It does not matter at all whether he called `receive/` in the meantime or not.

concent_api/core/views.py
Jakub89: :+1:

rwrzesien:
@Jakub89 Looks better but we still need tests which performs step 4A in different order, when provider calls first, then requestor, and opposite.

Jakub89:
@rwrzesien Okey, branch: "chore-add-missing-tests-for-force-subtask-results-use-case"

Jakub89:
@rwrzesien Okey, I've added more tests.

PR #123, opened 2018-02-14
Chore 'force report computed task' use case to avoid resending messages

### LINKED ISSUE #87: Update 'force report computed task' use case to avoid resending messages
`ForceReportComputedTask` should always be created by Concent rather than taken from client's request and resent. Also, the content of this message has changed in golem messages (https://github.com/golemfactory/golem-messages/issues/116). It's now only a wrapper for `ReportComputedTask`. Update our `send` and `receive` views accordingly.

`AckReportComputedTask` and `RejectReportComputedTask` should never be returned from Concent directly. It must be wrapped in `ForceReportComputedTaskResponse`.

First update integration tests and make sure they fail. Then update the code.

### LINKED ISSUE #88: Update 'force get task result' use case to avoid resending messages
`ForceGetTaskResult` should always be created by Concent rather than taken from client's request and resent. Also, the content of this message has changed in golem messages (https://github.com/golemfactory/golem-messages/issues/116). It's now only a wrapper for `GetTaskResult`. Update our `send` and `receive` views accordingly.

First update integration tests and make sure they fail. Then update the code.

kbeker:
Resolves #87
Resolves #88


kbeker:
Opened new with different branch name

PR #128, opened 2018-02-16
Chore update first and second use cases to avoid resending messages

### LINKED ISSUE #87: Update 'force report computed task' use case to avoid resending messages

### LINKED ISSUE #88: Update 'force get task result' use case to avoid resending messages

kbeker:
Resolves #87
Resolves #88


concent_api/core/views.py
```diff
@@ -579,6 +655,33 @@ def handle_receive_force_get_task_result_failed(
     return force_get_task_result_failed
 
 
+def handle_receive_force_subtask_results_response(
+    request,
+    decoded_message,
+    message_auth,
+):
+    client_public_key = decode_client_public_key(request)
+    if hasattr(decoded_message.subtask_results_rejected, 'TYPE'):
```
rwrzesien: Why not checking this with `isinstance` like we do in other places ?

concent_api/core/views.py
```diff
@@ -99,13 +99,16 @@ def receive(request, _message):
...
         if client_public_key != last_undelivered_message_status.message.auth.requestor_public_key_bytes:
             return None
-        if decoded_message_data.task_to_compute.compute_task_def['deadline'] + settings.CONCENT_MESSAGING_TIME < current_time:
+        if decoded_message_data.report_computed_task.task_to_compute.compute_task_def['deadline'] + settings.CONCENT_MESSAGING_TIME < current_time:
             set_message_as_delivered(last_undelivered_message_status)
-            return handle_receive_ack_from_force_report_computed_task(decoded_message_data)
+            return handle_receive_ack_from_force_report_computed_task(
+                request,
+                decoded_message_data,
+                last_undelivered_message_status.message.auth
```
rwrzesien: I think that we should keep consistency here. With this changes we pass `.auth` in some functions and `last_undelivered_message_status` in others. I think we should, if its possible, try to keep it the same for all functions on the same level. @cameel What do you think ?

concent_api/core/views.py
```diff
@@ -198,14 +212,16 @@ def receive_out_of_band(request, _message):
...
-            return handle_receive_out_of_band_ack_report_computed_task(request, last_undelivered_receive_status)
-
         if last_undelivered_receive_status.type == message.ForceReportComputedTask.TYPE:
             return handle_receive_out_of_band_force_report_computed_task(request, last_undelivered_receive_status)
 
-        if last_undelivered_receive_status.type == message.RejectReportComputedTask.TYPE:
-            return handle_receive_out_of_band_reject_report_computed_task(request, last_undelivered_receive_status)
+        if last_undelivered_receive_status.type == message.concents.ForceReportComputedTaskResponse.TYPE:
+            decoded_message = deserialize_message(last_undelivered_receive_status.data.tobytes())
+            if hasattr(decoded_message.ack_report_computed_task, 'TYPE'):
```
rwrzesien: Why not checking it with `isinstance` ?

concent_api/requirements.lock
```diff
@@ -4,31 +4,29 @@ cbor2==3.0.4
 certifi==2018.1.18
 cffi==1.11.4
 chardet==3.0.4
-coincurve==6.0.0
-Django==1.11.9
+coincurve==7.1.0
+Django==1.11.10
```
cameel: Not a review, just a quick remark:

Please only update the stuff that comes from newer golem-messages. `Django` (and probably other packages that changed here) are not dependencies of golem-messages and updating them is not a part of this task. If you really want to update them, it should be a separate commit (so that we can easily see it and revert it if a newer version is broken without reverting other changes).

To get new dependencies of golem-messages do this:
- Create a clean virtualenv
- Install stuff from `requirements.lock`
- Install the new `golem-messages` with the `--upgrade` option
- Replace `requirements.lock` with the output of `pip freeze`

At this point in the diff you can see the packages that were added or updated because they are dependencies of `golem-messages`. To see what was removed you need additional steps:

- Run `pipdeptree` and see which top-level packages (packages listed without any indentation) are not in `requirements.txt`.
- Remove those packages from `requirements.lock`.


BTW review would be easier if updating golem-messages and fixing stuff broken after the update was a separate pull request.

kbeker: Updated. Only stuff that comes from golem-messages are updated.

cameel: OK. Please don't merge this until @bartoszbetka resolves the problem with pyellpitic update. Otherwise we won't be able to deploy code master to the cluster.

Issue #118, opened 2018-02-13
Gatekeeper sometimes tries to create headers containing newlines and crashes


cameel:
We received a crash report from `gatekeeper` that shows that it tried to insert a checksum ending with a newline character into a response HTTP header which resulted in Django raising `BadHeaderError` (see stack trace below).

1) Fix gatekeeper and make sure we never create such headers in our code.
2) Make sure Concent API will never create a token containing strings with newlines.
3) Add extra `FileTransferToken` validations that reject any string values containing newlines unless we want to explicitly allow them.

### Traceback
``` python
File “/srv/http/virtualenv/lib/python3.5/site-packages/django/core/handlers/exception.py” in inner
41.             response = get_response(request)
File “/srv/http/virtualenv/lib/python3.5/site-packages/django/core/handlers/base.py” in _get_response
187.                 response = self.process_exception_by_middleware(e, request)
File “/srv/http/virtualenv/lib/python3.5/site-packages/django/core/handlers/base.py” in _get_response
185.                 response = wrapped_callback(request, *callback_args, **callback_kwargs)
File “/srv/http/virtualenv/lib/python3.5/site-packages/django/views/decorators/csrf.py” in wrapped_view
58.         return view_func(*args, **kwargs)
File “/srv/http/virtualenv/lib/python3.5/site-packages/django/views/decorators/http.py” in inner
40.             return func(request, *args, **kwargs)
File “/srv/http/concent_api/gatekeeper/views.py” in upload
50.     response["Concent-File-Checksum"]  = response_or_file_info["checksum"]
File “/srv/http/virtualenv/lib/python3.5/site-packages/django/http/response.py” in _setitem_
152.         value = self._convert_to_charset(value, 'latin-1', mime_encode=True)
File “/srv/http/virtualenv/lib/python3.5/site-packages/django/http/response.py” in _convert_to_charset
125.             raise BadHeaderError("Header values can't contain newlines (got %r)" % value)

Exception Type: BadHeaderError at /gatekeeper/upload-auth/1/2/result Exception Value: Header values can't contain newlines (got ‘sha1:e5fa44f2b31c1fb553b6021e7360d07d5d91ff5e\n’)
```

rwrzesien:
I have added fix which covers 1 and 3 cases. About 2, the only option I see is to run explicitly .replace('\n', '') on each header in `get_file_status` function, but is this really the solution we want ? Those values are passed from client, so with this fix gatekeeper will respond with 401.

PR #132, opened 2018-02-16
Fix for gatekeeper sometimes tries to create headers containing newli…

rwrzesien:
…nes and crashes

Resolves https://github.com/golemfactory/concent/issues/118

Please see my note in issue description.

### DIFF
--- a/concent_api/gatekeeper/tests/test_gatekeeper_views.py
+++ b/concent_api/gatekeeper/tests/test_gatekeeper_views.py
@@ -209,6 +209,159 @@ class GatekeeperViewUploadTest(TestCase):
             self.assertIn('message', response.json().keys())
             self.assertEqual("application/json", response["Content-Type"])
 
+    @freeze_time("2018-12-30 11:00:00")
+    def test_upload_should_return_401_if_size_of_file_is_wrong(self):
+        invalid_values = [
+            None,
+            'number',
+            '-2',
+            '1.0',
+        ]
+
+        for invalid_value in invalid_values:
+            file1 = FileTransferToken.FileInfo(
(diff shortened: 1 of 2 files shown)


concent_api/gatekeeper/tests/test_gatekeeper_views.py
```diff
@@ -209,6 +209,159 @@ def test_upload_should_return_401_if_size_of_file_is_wrong(self):
...
+                HTTP_CONCENT_CLIENT_PUBLIC_KEY = self.public_key
+            )
+
+            self.assertIsInstance(response, JsonResponse)
+            self.assertEqual(response.status_code, 401)
+            self.assertIn('message', response.json().keys())
+            self.assertEqual("application/json", response["Content-Type"])
+
+    @freeze_time("2018-02-12 16:19:00")
+    def test_upload_should_return_401_if_specific_authorization_header(self):
```
cameel: What do you mean by 'specific'?

rwrzesien: It fails on exactly this header. Also if you manipulate a little bit with data which is used to create it, it also fails. But if you create a header from completely different data, it is ok. For example, in test `test_upload_should_return_401_if_newlines_in_message_checksum` it doesn't raised 500 but was handled correctly by gatekeeper validation.

cameel: Oh, OK. But I think it would have been better to try to describe what's so specific about it in the name of the test. I know it can be hard sometimes but names that are generic like this are often confusing. I'd have to read the whole test carefully to figure out what's going on here.

rwrzesien: :+1:

Issue #131, opened 2018-02-16
Machine-readable codes in JSON error messages


cameel:
Currently our JSON error messages only contain human-readable strings describing the issue. The only thing the client can do with such a string is to show it to the user. The messages are not stable enough to rely on them. They can change between versions without notice. Even worse if we ever enable translations. Messages can also contain embedded values from the request which would require the client to parse them according to some pattern making identifying specific errors even more complex. We have this issue even in our tests where we can't reliably test why the request fails - only that it does fail.

Update all JSON error responses to contain the following fields:
- `error_message`
- `error_code`

Example
``` json
{
    "error_message": "-123 is not a valid timestamp. Timestamp cannot be negative",
    "error_code":    "negative-timestamp"
}
```

Error codes can be arbitrary, unique strings. They should be granular enough for the client to be able to discern every important case but not so granular that every instance of an error has a different code.

We can come up with them ad-hoc but if you can suggest a sensible hierarchy that's even better. Maybe it would make sense to use lists of strings or dotted names instead of a single string. E.g. `["gatekeeper", "validation.checksum", "not-a-sha1-hash"]` or `"gatekeeper.validation.checksum.not-a-sha1-hash"`.

rwrzesien:
I suggest adding those codes as constants, as they will be repeatable along the code.

PR #133, opened 2018-02-19
Add Machine-readable codes in JSON error messages

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/131.


concent_api/gatekeeper/views.py
```diff
@@ -121,73 +157,193 @@ def parse_headers(request: WSGIRequest, path_to_file: str) -> Union[FileTransfer
...
+            loaded_golem_message.subtask_id,
+            client_public_key,
+        )
     current_time = get_current_utc_timestamp()
 
     if current_time > loaded_golem_message.token_expiration_deadline:
-        return gatekeeper_access_denied_response('token_expiration_deadline has passed.', path_to_file, loaded_golem_message.subtask_id, client_public_key)
+        return gatekeeper_access_denied_response(
+            'token_expiration_deadline has passed.',
+            constants.ERROR_CODE_GATEKEEPER_INVALID_REQUEST_MESSAGE,
```
cameel: Situations like deadlines and token expiration should have separate codes because it's likely that the client will have to react to that specific situation in a different way that for other errors. For example if you have a wrong type somewhere, your token is completely broken. But if the time runs out, there's sometimes a mechanism to renew the token.

concent_api/gatekeeper/views.py
```diff
@@ -121,73 +157,193 @@ def parse_headers(request: WSGIRequest, path_to_file: str) -> Union[FileTransfer
...
+            path_to_file,
+            loaded_golem_message.subtask_id,
+            client_public_key,
+        )
     if loaded_golem_message.storage_cluster_address != settings.STORAGE_CLUSTER_ADDRESS:
-        return gatekeeper_access_denied_response('Given storage_cluster_address is not defined in settings.', path_to_file, loaded_golem_message.subtask_id, client_public_key)
+        return gatekeeper_access_denied_response(
+            'Given storage_cluster_address is not defined in settings.',
+            constants.ERROR_CODE_GATEKEEPER_INVALID_REQUEST_MESSAGE,
```
cameel: This should have a separate code. There's a possibility that the cluster address was valid before but has been removed. We want to be able to distinguish this from obvious errors like a broken token.

In general codes for errors caused by invalid data should be distinguishable from codes for data that is syntactically correct but is rejected for some semantic reason.

concent_api/gatekeeper/views.py
```diff
@@ -121,73 +157,193 @@ def parse_headers(request: WSGIRequest, path_to_file: str) -> Union[FileTransfer
...
+            path_to_file,
+            loaded_golem_message.subtask_id,
+            client_public_key,
+        )
     client_public_key_base64 = (b64encode(loaded_golem_message.authorized_client_public_key)).decode('ascii')
     if client_public_key != client_public_key_base64:
-        return gatekeeper_access_denied_response('authorized_client_public_key is different then Concent-Client-Public-Key header.', path_to_file, loaded_golem_message.subtask_id, client_public_key)
+        return gatekeeper_access_denied_response(
+            'authorized_client_public_key is different then Concent-Client-Public-Key header.',
+            constants.ERROR_CODE_GATEKEEPER_INVALID_REQUEST_MESSAGE,
```
cameel: This should have a separate code.

concent_api/gatekeeper/views.py
```diff
@@ -121,73 +157,193 @@ def parse_headers(request: WSGIRequest, path_to_file: str) -> Union[FileTransfer
...
+            loaded_golem_message.subtask_id,
+            client_public_key,
+        )
 
     # -OPERATION
     if request.method == 'POST' and loaded_golem_message.operation != 'upload':
-        return gatekeeper_access_denied_response('Wrong operation variable for this request method.', path_to_file, loaded_golem_message.subtask_id, client_public_key)
+        return gatekeeper_access_denied_response(
+            'Wrong operation variable for this request method.',
+            constants.ERROR_CODE_GATEKEEPER_INVALID_REQUEST_MESSAGE,
```
cameel: We need to change this message. There are no 'variables' here.

concent_api/concent_api/constants.py
```diff
@@ -34,3 +34,23 @@
...
+ERROR_CODE_CORE_INVALID_REQUEST_MESSAGE                         = 'core.invalid_request.message'
+ERROR_CODE_CORE_INVALID_REQUEST_MESSAGE_TASK_ID                 = 'core.invalid_request.message.task_id'
+ERROR_CODE_CORE_INVALID_REQUEST_MESSAGE_VALUE_NOT_INTEGER       = 'core.invalid_request.message.value_not_integer'
+ERROR_CODE_CORE_QUEUE_MESSAGE_ALREADY_PROCESSED                 = 'core.queue.message_already_processed'
+ERROR_CODE_CORE_QUEUE_COMMUNICATION_NOT_STARTED                 = 'core.queue.communication_not_started'
+ERROR_CODE_CORE_QUEUE_RESPONSE_ALREADY_SUBMITTED                = 'core.queue.response_already_submitted'
+ERROR_CODE_CORE_QUEUE_TIMEOUT                                   = 'core.queue.timeout'
+ERROR_CODE_CORE_MESSAGE_UNKNOWN                                 = 'core.message.unknown'
+ERROR_CODE_CORE_MESSAGE_UNEXPECTED                              = 'core.message.unexpected'
+ERROR_CODE_CORE_MESSAGE_UNABLE_TO_DESERIALIZE                   = 'core.message.unable_to_deserialize'
```
cameel: I'm not sure if using app name as a prefix is a good idea. It's an implementation detail that the client should not care about. And this will be a problem when we start sharing more code between apps. We'll have to know which app is using the code which breaks encapsulation.

rwrzesien:
@cameel I agree with all things you have posted in your review, and actually most things were my doubts too. Just FYI about the code granularity I have followed the issue description which says there should not be a separate error code for everything.

concent_api/concent_api/constants.py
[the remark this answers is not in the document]
rwrzesien: :+1:

concent_api/gatekeeper/views.py
rwrzesien: :+1:

concent_api/concent_api/constants.py
rwrzesien: :+1:

rwrzesien:
@cameel I have updated the code, you can review and add comments, in meantime I will start adding checking of those to tests

rwrzesien:
Updated code and tests. In tests there are a lot of places where expected code might be different, but I left it for you to decide. If you want I can fix those in this ticket or we can create separate issue for each case.

cameel:
Since this is blocked waiting for my feedback and the pull request is big, maybe let's try to work with it on a smaller scale. Not in the whole project but in a single app or module.

Please extract changes related just to a single module (I suggest gatekeeper since it's small but does have a lot of error responses) into a separate pull request and then we'll discuss it on a smaller scale. Maybe it will be enough to agree on the granularity. If not, we'll discuss it further in subsequent pull requests.

rwrzesien:
Closing as this is outdated and will be done in smaller pull requests.

Issue #145, opened 2018-02-23
Don't clear message signatures when storing them in the database


cameel:
The current code ofen does something like this before storing a message in the database:
``` python
message.sig = None
```

This means that we're removing the original signature from the message. This should not be happening for messages submitted by clients. We won't be able to verify their authenticity later. And for messages created by Concent it should not be necessary at all.

Investigate all such cases and remove them from the code if possible. Each one that remains must have a very good reason to remain.

cameel:
@rwrzesien Let's keep discussions on what the best solution is out of pull requests. Reviews cause so much noise that it's hard to discuss anything that's not related to a particular line of code there.

> This is just a fix proposal. The original issue is caused when `.serialize()` method is called on message, and no `sign_func` parameter is provided, method is calling `_fake_sign` to set the `.sig` to:
>
> ```
> def _fake_sign(_):
>     return b'\0' * Message.SIG_LEN
> ```
>
> Which cause issues on loading that message. 
>
> In my solution we don't need to worry and remember about clearing this overwritten signature, but we still, on few occasions, store messages with this overwritten signature in database (when there is no original signature, so when the message was created by concent). However, there is no harm because we never send them back to client.
>
> The other solutions might be:
> 1. Report this as bug to golem-messages.
> 2. Keep the current solution.
> 3. Use different serialization method for storing messages in database (ie. pickle).
> 4. Use the same `sign_func` as `dump` function and sign message with concent keys
`functools.partial(cryptography.ecdsa_sign, privkey)`
>
> Please let me know what do you guys think about this.

So you're basically saying that `serialize()` without a parameter discards the original signature and replaces it with zeros? This is definitely unexpected behavior to me, maybe even a bug. It's both unexpected and inconsistent with what happens with the message if you let `golem-messages` serialize it. Please report it.

I see that your solution is simply letting golem-messages serialize the nested messages. That's good, less code for us. If we don't need to serialize explicitly, let's not.

For messages created by ourselves letting `golem-messages` replace the null sig with a string of zeros is fine as long as we never send such messages to the client. We can treat it as a special value, just like `None` or empty string.

We should add some assertions that make sure that a message returned from a view never has a signature that's a string of zeros. This applies to nested messages too (would be nice to have a function that checks this recursively).

rwrzesien:
> So you're basically saying that serialize() without a parameter discards the original signature and replaces it with zeros?

[...]

PR #150, opened 2018-02-26
Fix for message signature clearing

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/145.

This is just a fix proposal. The original issue is caused when `.serialize()` method is called on message, and no `sign_func` parameter is provided, method is calling `_fake_sign` to set the `.sig` to:

```
def _fake_sign(_):
    return b'\0' * Message.SIG_LEN
```

Which cause issues on loading that message. 

In my solution we don't need to worry and remember about clearing this overwritten signature, but we still, on few occasions, store messages with this overwritten signature in database (when there is no original signature, so when the message was created by concent). However, there is no harm because we never send them back to client.

The other solutions might be:
1. Report this as bug to golem-messages.
2. Keep the current solution.
3. Use different serialization method for storing messages in database (ie. pickle).
4. Use the same `sign_func` as `dump` function and sign message with concent keys
`functools.partial(cryptography.ecdsa_sign, privkey)`

Please let me know what do you guys think about this.

### DIFF
--- a/concent_api/utils/api_view.py
+++ b/concent_api/utils/api_view.py
@@ -81,6 +81,7 @@ def api_view(view):
             )
             return JsonResponse({'error': str(exception)}, status = 400)
         if isinstance(response_from_view, Message):
+            assert response_from_view.sig is None
             logging.log_message_returned(
                 response_from_view,
                 request.META['HTTP_CONCENT_CLIENT_PUBLIC_KEY'],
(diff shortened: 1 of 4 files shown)


rwrzesien:
@cameel 
Add 1) Added, there is only one place when should check it.
Add 2) Created, https://github.com/golemfactory/golem-messages/issues/162.
Add 3) I have found also few in old tests and removed them, but let rest of the team confirm.

concent_api/utils/api_view.py
```diff
@@ -81,6 +82,7 @@ def wrapper(request, *args, **kwargs):
             )
             return JsonResponse({'error': str(exception)}, status = 400)
         if isinstance(response_from_view, Message):
+            assert response_from_view.sig != _fake_sign(None)
```
cameel: This should be changed to `assert response_from_view.sig == None`. If we do need to return a signed message we should clear its signature explicitly.

Issue #126, opened 2018-02-15
Header with Concent version on all HTTP responses


cameel:
Add a header containing Concent version to all HTTP responses from Concent API. Call it `Concent-Version`.

Version should come from a git tag but not directly. Create a script that will store this version in a text file. The server should not use git but read the file instead. Running the script and creating the file will be part of our deployment. Version file should be listed in `.gitignore` to avoid committing it to the repo by accident.

Make sure version comes from a tag on the repository the code is located in. I.e. `concent` and not `concent-deployment`.

jiivan:
Here's and example what we is used in golem core and golem-messages to get a version from git tag: https://gist.github.com/jiivan/cd7f12200f34cfe3fa5ec8c9da8d98c0

PR #151, opened 2018-02-26
Feature header with concent version on all http responses

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/126.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -102,3 +102,6 @@ ENV/
 
 # Vim's temporary files
 *.sw[ponm]
+
+# Version
+/RELEASE-VERSION
--- a/concent_api/concent_api/middleware.py
+++ b/concent_api/concent_api/middleware.py
@@ -1,3 +1,7 @@
+import os
+
+from django.conf    import settings
+
 from golem_messages import __version__
 
 
@@ -13,4 +17,25 @@ class GolemMessagesVersionMiddleware(object):
     def __call__(self, request):
         response = self.get_response(request)
         response['Concent-Golem-Messages-Version'] = __version__
+        return response
+
+
+class ConcentVersionMiddleware(object):
+    """
+    Used to attach version of the Concent currently used to HTTP response header.
+
+    """
+
+    _concent_version = None
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        if self._concent_version is None:
+            with open(os.path.join(settings.BASE_DIR, '..', 'RELEASE-VERSION')) as f:
+                self._concent_version = f.read()
+
+        response = self.get_response(request)
+        response['Concent-Version'] = self._concent_version
         return response
--- a/concent_api/concent_api/settings/base.py
+++ b/concent_api/concent_api/settings/base.py
@@ -28,6 +28,7 @@ MIDDLEWARE = [
     'django.contrib.messages.middleware.MessageMiddleware',
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
     'concent_api.middleware.GolemMessagesVersionMiddleware',
+    'concent_api.middleware.ConcentVersionMiddleware',
 ]
 
 ROOT_URLCONF = 'concent_api.urls'
--- a/concent_api/concent_api/tests/test_middleware.py
+++ b/concent_api/concent_api/tests/test_middleware.py
@@ -1,4 +1,5 @@
 from base64                 import b64encode
+import mock
 
 from django.test            import override_settings
 from django.test            import TestCase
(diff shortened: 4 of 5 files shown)


version.py
```diff
@@ -0,0 +1,81 @@
+# Authors: Douglas Creager <dcreager@dcreager.net>
```
cameel: Please don't commit copied code into the repository.

It's already available in `golem-messages` anyway.

.gitignore
```diff
@@ -102,3 +102,6 @@ ENV/
 
 # Vim's temporary files
 *.sw[ponm]
+
+# Version
+RELEASE-VERSION
```
cameel: This should be
```
/RELEASE-VERSION
```

Otherwise this will ignore all files with that name, even in nested directories.

concent_api/concent_api/tests/test_middleware.py
```diff
@@ -47,3 +47,39 @@ def test_golem_messages_version_middleware_should_attach_http_header_to_response
             response._headers['concent-golem-messages-version'][1],
             __version__
         )
+
+
+@override_settings(
+    CONCENT_PRIVATE_KEY    = CONCENT_PRIVATE_KEY,
+    CONCENT_PUBLIC_KEY     = CONCENT_PUBLIC_KEY,
+    CONCENT_VERSION        = 'v1.0'
```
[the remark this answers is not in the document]
rwrzesien: :+1:

.gitignore
rwrzesien: :+1:

version.py
rwrzesien: :+1: I will use the golem_message version then, but I have used it only because Darek's proposed it himself.

concent_api/concent_api/settings/base.py
```diff
@@ -191,6 +192,13 @@
 }
 
 
+try:
+    with open(os.path.join(BASE_DIR, '..', 'RELEASE-VERSION')) as f:
+        CONCENT_VERSION = f.read()
+except FileNotFoundError:
```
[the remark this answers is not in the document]
rwrzesien: :+1:

concent_api/concent_api/settings/base.py
```diff
@@ -191,6 +192,13 @@
...
+    CONCENT_VERSION = 'Unknown'
```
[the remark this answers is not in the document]
rwrzesien: :+1:

rwrzesien:
@cameel Updated apart from the question with from where we want to get `version` script.

version.py
cameel: It's just a script that prints the version so it does not need to be importable.

concent_api/concent_api/middleware.py
```diff
@@ -14,3 +18,26 @@ def __call__(self, request):
...
         return response
+
+
+class ConcentVersionMiddleware(object):
+    """
+    Used to attach version of the Concent currently used to HTTP response header.
+
+    """
+
+    CONCENT_VERSION = None
```
cameel: This should be `_concent_version` because
- It's not really a constant (it's value is `None` and can stay at it for a non-negligible amount of time - until someone calls `get_concent_version()`)
- From the code below I see that you want the user of the class to get it via `get_concent_version()` and not directly. It should be private.

version.py
rwrzesien: So what are you suggesting to do ?

concent_api/concent_api/middleware.py
```diff
@@ -14,3 +18,26 @@ def __call__(self, request):
...
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        response = self.get_response(request)
+        response['Concent-Version'] = self.get_concent_version()
+        return response
+
+    def get_concent_version(self):
```
cameel: This should be a `@classmethod`. Otherwise you won't be able to get the version without creating an instance.

version.py
cameel: Just don't include this script here. We'll call it directly in deployment.

For convenience you might just add a helper script that detects the location of golem-messages and calls the script from there.

concent_api/concent_api/middleware.py
rwrzesien: Not exactly, I have added it this way because it was only way I could mock it in tests. This is for middleware usage only.

concent_api/concent_api/middleware.py
rwrzesien: :+1: But about second point please check explanation below. Actually I don't want anyone to use it in other way than through middleware. If we want a reusable function it should be in utils.

version.py
rwrzesien: :+1: I will add script but please note this is returned with every response so everyone need to run it on their environments to perform a request without failing.

rwrzesien:
@cameel Updated.

Issue #146, opened 2018-02-23
[Requestor does not pay for computation use case] Add missing tests


rwrzesien:
https://github.com/golemfactory/concent/issues/120 is missing some tests:

- auth tests,
- cluster integration tests,
- negative test scenarios,
- there is no test backend, just mocked functions. It was optional, but then there should be mocked specific backend functions, not the higher level functions.

Please add them.

rwrzesien:
Closing after merge.

PR #153, opened 2018-02-26
Api test for force payment + bugfix

Jakub89:
Add cluster integration tests for issue #146 , rest will be added in next pull request.

### DIFF
--- a/concent_api/core/payments/mock.py
+++ b/concent_api/core/payments/mock.py
@@ -1,3 +1,4 @@
+from django.conf import settings
 
 
 def sum_payments(list_of_payments):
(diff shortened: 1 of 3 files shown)


concent_api/core/payments/mock.py
```diff
@@ -8,7 +9,7 @@ def get_client_eth_account():
     return True
 
 
-def get_number_of_eth_block(request = None):  # pylint: disable=inconsistent-return-statements
+def get_number_of_eth_block(request = None, _subtask_results_accepted_list = None, _list_of_transactions = None, _list_of_forced_payments = None):  # pylint: disable=inconsistent-return-statements
```
rwrzesien: Please also change in `payments\base.py`. I see there are now two function calling `get_number_of_eth_block`. There should be only one, and second one should call `payment_summary()`. Please also make sure that both functions signature is correct in base and mock backends, and their usage in other code is correct.

Jakub89: Okey, I see. Function `get_number_of_eth_block` is currently not used at all in our code, and `payment_summary` is used only once. Should I maybe remove `get_number_of_eth_block` completely for right now?

rwrzesien: About removal, please discuss it with @kbeker .

This it not fixed yet:
1. `payment_summary` in base.py should call backend.payment_summary
2. `get_number_of_eth_block` in base.py should have only `backend, request` parameters.
3. `get_number_of_eth_block` in mock.py should have only `request` parameter.

Jakub89: Done, and I'll talk to @kbeker when I'll got a chance :)

rwrzesien: Ok, thanks.

Issue #86, opened 2018-02-01
Rename TOKEN_EXPIRATION_DEADLINE setting to TOKEN_EXPIRATION_TIME


cameel:
The current name is misleading. A deadline is a specific point in time and that's how it's used in Golem messages. This setting represents the amount of time so it should be called something like `TOKEN_EXPIRATION_TIME`

Rename all instances in `concent-api` and in our deployment scripts.

Issue #85, opened 2018-02-01
Endpoint that returns protocol constants


cameel:
For golem protocol constants will be indeed constant but for development we need an easy way to tell how the server is configured. We need an endpoint that will be returning this information.

- Add an endpoint under `/api/v1/protocol-constants/`.
- The endpoint should always return a JSON message with the following fields:
    - `concent_messaging_time`
    - `force_acceptance_time`
    - `token_expiration_time` (do #86 first)
    - In the future, other timing-related fields added from the Concent-Golem integration spec.
- The values should come from settings with the corresponding names.

PR #154, opened 2018-02-26
Feature endpoint that returns protocol constants

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/85.
Resolves https://github.com/golemfactory/concent/issues/86.


concent_api/core/views.py
```diff
@@ -326,6 +328,31 @@ def receive_out_of_band(request, _message):
...
 
 
+@require_GET
+def protocol_constants(_request):
+    """ Endpoint which returns Concent time settings. """
+    return JsonResponse(
+        data = {
+            'concent_messaging_time': {
+                'name':  'CONCENT_MESSAGING_TIME',
+                'value': getattr(settings, 'CONCENT_MESSAGING_TIME', 'undefined')
```
cameel: Please don't add default values here. We want this code to fail as soon as possible and as loudly as possible if these settings aren't defined so that we can fix this. The only case where they're undefined is if someone has completely botched his settings.

Might be a good idea to add a Django check for these settings too. And set them all to zeros in `testing.py` settings so that all tests that rely on them have to explicitly override them.

concent_api/core/tests/test_endpoint_protocol_constants.py
```diff
@@ -0,0 +1,78 @@
...
+        response = self.client.get(
+            reverse('core:protocol_constants'),
+        )
+
+        for setting_name in protocol_constants_settings:
+            self.assertIn(setting_name.lower(),                              response.json())
+            self.assertIn('name',                                            response.json()[setting_name.lower()])
+            self.assertIn('value',                                           response.json()[setting_name.lower()])
+            self.assertEqual(response.json()[setting_name.lower()]['name'],  setting_name)
+            self.assertEqual(response.json()[setting_name.lower()]['value'], 'undefined')
```
cameel: The code should just fail if the critical settings like those timings are undefined. I don't think we should test that case because we don't support it.

rwrzesien: :+1:

concent_api/core/views.py
rwrzesien: :+1: But I don't think adding them to Django checks is necessary. When you look at `system_check.py` code you will see that a most of it looks redundant. If we want to do this lets create a separate issue and do it more generic way.

cameel: Well, I'm OK with putting it in a separate issue. Could you create it?

It isn't really all that redundant. We don't have these checks anywhere else. We just write code as if this situation could never happen. At most we might have assertions in cases where we know the code would not just crash on its own but these don't count as checks. They're for programmers and run only in debug mode.

Checks are for the admin who sets up the application (that's still us for now ;)). Letting it crash when the setting is needed is much better than ignoring the problem but it's even better to add those checks so that bad settings fail immediately at `manage.py check` (which we always run before deployment) rather than long after the server has been deployed and just happens to enter the function that uses the setting that's missing. We did have deployment problems in the past caused by missing or bad settings. We need more of those checks because they're useful.

rwrzesien: :+1: Yes, I will add it tomorrow because it will need some more thinking what else can we check there.

rwrzesien:
@cameel Updated.
@Jakub89 @kbeker How about your review ?

Jakub89:
@rwrzesien 8 tests are failing on checking `token_expiration_deadline`:
```
Traceback (most recent call last):
  File "/home/jakub/Documents/work/projects/concent/concent_api/core/tests/test_integration_get_task_result.py", line 582, in test_concent_requests_task_result_from_provider_and_requestor_receives_task_result
    self.assertEqual(message_file_transfer_token.token_expiration_deadline, self._parse_iso_date_to_timestamp("2017-12-01 11:30:12"))
AssertionError: 1512126012 != 1512127812
```

rwrzesien:
@Jakub89 Fixed, good catch :+1:

rwrzesien:
@Jakub89 Fixed.

rwrzesien:
@Jakub89 Fixed.

PR #164, opened 2018-03-06
Feature script for testing force get task result use case

dybi:


concent_api/api-integration-force-accept-or-reject-results-test.py
```diff
@@ -5,35 +5,28 @@
 import random
 import time
 from base64                 import b64encode
+from collections import namedtuple
```
rwrzesien: Please follow the convention, `import namedtuple` should be aligned to other imports.

concent_api/api-integration-force-accept-or-reject-results-test.py
```diff
@@ -5,35 +5,28 @@
...
 from freezegun              import freeze_time
 
 from golem_messages         import message
 
 from utils.helpers          import get_current_utc_timestamp
 from utils.testing_helpers  import generate_ecc_key_pair
 
-from api_testing_helpers    import api_request
-from api_testing_helpers    import timestamp_to_isoformat
+from api_testing_common import api_request, parse_command_line
```
rwrzesien: Please follow the convention, import should be aligned to other imports.

Also, single import in single line, alphabetically.

concent_api/api-integration-force-accept-or-reject-results-test.py
```diff
@@ -5,35 +5,28 @@
...
+from api_testing_common    import timestamp_to_isoformat
 
 import requests
 
+HOUR = 3600
```
rwrzesien: I don't think this is necessary. First, you are using it only once. Second, after all it is hour, its value shouldn't change (if it can, it should be renamed).
But if you want the value `3600` you can use `int(timedelta(hours=1).total_seconds())`.

concent_api/api-integration-force-accept-or-reject-results-test.py
```diff
@@ -77,10 +70,10 @@ def force_subtask_results_response(timestamp = None, subtask_results_accepted =
         )
 
 
-def subtask_results_accepted(timestamp = None, subtask_id = None, payment_ts = None):
+def subtask_results_accepted(timestamp = None, task_to_compute = None, payment_ts = None):
     with freeze_time(timestamp):
         return message.tasks.SubtaskResultsAccepted(
-            subtask_id = subtask_id,
+            task_to_compute = task_to_compute,
```
rwrzesien: So basically we do the styling like this:
```
return message.tasks.SubtaskResultsAccepted(
    task_to_compute = task_to_compute,
    payment_ts          = payment_ts,
```

concent_api/api-integration-force-accept-or-reject-results-test.py
```diff
@@ -101,40 +94,87 @@ def report_computed_task(timestamp = None, subtask_id = None, task_to_compute =
         )
 
 
+def create_task(task_to_compute_timestamp, task_id, compute_task_def_deadline):
+    return task_to_compute(
```
rwrzesien: Please fix styling.

concent_api/api-integration-force-accept-or-reject-results-test.py
```diff
@@ -101,40 +94,87 @@ def report_computed_task(timestamp = None, subtask_id = None, task_to_compute =
...
+        timestamp=task_to_compute_timestamp,
+        compute_task_def=compute_task_def(
+            task_id=task_id,
+            subtask_id=task_id,
```
rwrzesien: `task_id` and `subtask_id` are two different things, `subtask_id` is even more important, please add it as separate parameter.

concent_api/api-integration-force-accept-or-reject-results-test.py
```diff
@@ -101,40 +94,87 @@ def report_computed_task(timestamp = None, subtask_id = None, task_to_compute =
...
+            deadline=compute_task_def_deadline,
+        )
+    )
+
+
+ProtocolConstants = namedtuple("ProtocolConstants",
```
rwrzesien: It should look like:
```
ProtocolConstants = namedtuple(
    "ProtocolConstants",
    [
        "concent_messaging_time",
        "subtask_verification_time",
        "force_acceptance_time",
        "token_expiration_time",
    ]
)

File: concent_api/api-integration-force-accept-or-reject-results-test.py [hunk 8]
```diff
@@ -101,40 +94,87 @@ def report_computed_task(timestamp = None, subtask_id = None, task_to_compute =
...
+                               ["concent_messaging_time",
+                                "subtask_verification_time",
+                                "force_acceptance_time",
+                                "token_expiration_time"])
+
+
+def get_protocol_constants(cluster_url):
+    url = f"{cluster_url}/api/v1/protocol-constants/"
```
rwrzesien: As I generally like this new feature, I would suggest to hold with it, I know that on some our environments we still use Python3.5, and that would block using whole script on this version.

concent_api/api-integration-force-accept-or-reject-results-test.py
```diff
@@ -101,40 +94,87 @@ def report_computed_task(timestamp = None, subtask_id = None, task_to_compute =
...
+    resp = requests.get(url)
```
rwrzesien: Please try to align this way:
```
...
json                       = resp.json()
concent_messaging_time = json['concent_messaging_time']['value']
subtask_verification_time = json['subtask_verification_time']['value']
```

concent_api/api-integration-force-accept-or-reject-results-test.py
```diff
@@ -101,40 +94,87 @@ def report_computed_task(timestamp = None, subtask_id = None, task_to_compute =
...
+    json = resp.json()
+    concent_messaging_time = json['concent_messaging_time']['value']
+    subtask_verification_time = json['subtask_verification_time']['value']
+    force_acceptance_time = json['force_acceptance_time']['value']
+    token_expiration_time = json['token_expiration_time']['value']
+    constants = ProtocolConstants(concent_messaging_time, subtask_verification_time, force_acceptance_time,
```
rwrzesien:
```
constants = ProtocolConstants(
    concent_messaging_time,
    subtask_verification_time, 
    force_acceptance_time,
    token_expiration_time,
)
```

concent_api/api-integration-force-accept-or-reject-results-test.py
```diff
@@ -101,40 +94,87 @@ def report_computed_task(timestamp = None, subtask_id = None, task_to_compute =
...
+
 def main():
     cluster_url     = parse_command_line(sys.argv)
     task_id         = str(random.randrange(1, 100000))
     current_time    = get_current_utc_timestamp()
 
+    cluster_consts = get_protocol_constants(cluster_url)
+    print_protocol_constants(cluster_consts)
+
+    current_timestamp = timestamp_to_isoformat(current_time)
```
rwrzesien: Please fix styling according to rules mentioned above.

concent_api/api-integration-force-accept-or-reject-results-test.py
```diff
@@ -101,40 +94,87 @@ def report_computed_task(timestamp = None, subtask_id = None, task_to_compute =
...
+    ack_report_computed_task_timestamp = timestamp_to_isoformat(
```
rwrzesien: We don't use hard line length limit, but suggested is 120, so it is ok to put it in single line.

concent_api/api_testing_common.py
```diff
@@ -0,0 +1,132 @@
+import sys
+
+from golem_messages.exceptions      import MessageError
+from golem_messages.message import Message, ComputeTaskDef, TaskToCompute
+from golem_messages.shortcuts       import dump
+from golem_messages.shortcuts       import load
+
+import datetime
```
rwrzesien: Those imports should be before golem_messages imports.

concent_api/api_testing_common.py
```diff
@@ -0,0 +1,132 @@
...
+    return cluster_url
+
+
+def create_task_to_compute(current_time, task_id):
+    compute_task_def = ComputeTaskDef()
+    compute_task_def['task_id'] = task_id
+    compute_task_def['deadline'] = current_time + 60
+    task_to_compute = TaskToCompute(
+        compute_task_def=compute_task_def)
+    return task_to_compute
```
rwrzesien: No new line at the end of the file.

concent_api/api-integration-force-accept-or-reject-results-test.py
```diff
@@ -101,40 +94,87 @@ def report_computed_task(timestamp = None, subtask_id = None, task_to_compute =
...
+                                "subtask_verification_time",
+                                "force_acceptance_time",
+                                "token_expiration_time"])
+
+
+def get_protocol_constants(cluster_url):
+    url = f"{cluster_url}/api/v1/protocol-constants/"
+    resp = requests.get(url)
+    json = resp.json()
+    concent_messaging_time = json['concent_messaging_time']['value']
```
rwrzesien: It has changed, there will be no `name` and `value`, just the value you can get directly from
`json['concent_messaging_time']` etc.

concent_api/api-integration-force-accept-or-reject-results-test.py
rwrzesien: Actually it is not used anywhere else beyond `print` function. Why don't you want to use dict from response directly ?

concent_api/api-integration-force-accept-or-reject-results-test.py
dybi: according to information from @cameel, we will need to upgrade to python3.6 before the release, so it should not be the problem ;)

concent_api/api-integration-force-accept-or-reject-results-test.py
dybi: has this change been introduced to master already?

rwrzesien: Yes.

concent_api/api-integration-force-accept-or-reject-results-test.py
rwrzesien: Ok, great.

concent_api/api-integration-force-accept-or-reject-results-test.py
```diff
@@ -101,40 +94,87 @@ def report_computed_task(timestamp = None, subtask_id = None, task_to_compute =
...
     task_id         = str(random.randrange(1, 100000))
     current_time    = get_current_utc_timestamp()
 
+    cluster_consts = get_protocol_constants(cluster_url)
+    print_protocol_constants(cluster_consts)
+
+    current_timestamp = timestamp_to_isoformat(current_time)
+    ack_report_computed_task_timestamp = timestamp_to_isoformat(
+        current_time - cluster_consts.subtask_verification_time - TIME_GUARD)  # now - SVT - TIME_GUARD (~4:00:10)
+    report_computed_task_timestamp = ack_report_computed_task_timestamp   # now - SVT - TIME_GUARD (~4:00:10)
```
rwrzesien: Is there a reason to create this extra variable ?

concent_api/api-integration-force-accept-or-reject-results-test.py
```diff
@@ -5,35 +5,28 @@
...
 
-from api_testing_helpers    import api_request
-from api_testing_helpers    import timestamp_to_isoformat
+from api_testing_common import api_request, parse_command_line
+from api_testing_common    import timestamp_to_isoformat
 
 import requests
 
+HOUR = 3600
+TIME_GUARD = 10
```
cameel: A comment explaining this would be nice because the name on its own does not say much.

concent_api/api-integration-force-accept-or-reject-results-test.py
cameel: Python 3.6 is now on the cluster. You can safely assume that it's at least 3.6.

concent_api/api-integration-force-accept-or-reject-results-test.py
cameel: Please use full names. `resp` -> `response`.

concent_api/api-integration-force-accept-or-reject-results-test.py
cameel: Also, with lists like this, I'd use named parameters because the order is not really set in stone and if someone changes it, it may be hard to notice that the times are now wrong.

``` python
constants = ProtocolConstants(
    concent_messaging_time    = concent_messaging_time,
    subtask_verification_time = subtask_verification_time, 
    force_acceptance_time     = force_acceptance_time,
    token_expiration_time     = token_expiration_time,
)
```

Or, to make things shorter with the same effect:
``` python
constants = ProtocolConstants(**json)
```

concent_api/api-integration-force-accept-or-reject-results-test.py
```diff
@@ -101,40 +94,87 @@ def report_computed_task(timestamp = None, subtask_id = None, task_to_compute =
...
+                               ["concent_messaging_time",
+                                "subtask_verification_time",
+                                "force_acceptance_time",
+                                "token_expiration_time"])
+
+
+def get_protocol_constants(cluster_url):
+    url = f"{cluster_url}/api/v1/protocol-constants/"
+    resp = requests.get(url)
+    json = resp.json()
```
cameel: We should not try to parse response as JSON unless `Content-Type` says it's JSON. If it's e.g. HTML or text (because something went wrong) it would be better to print it.

Also, this is WIP. We will be using these constants all over the place soon.

concent_api/api-integration-force-accept-or-reject-results-test.py
cameel: It's nice to have a namedtuple if your dict has strictly defined fields. It's similar in purpose to an enum. You could use strings instead of enum values but it you mistype it or use one that does not exist, Python can notice and report an error.

concent_api/api-integration-force-accept-or-reject-results-test.py
```diff
@@ -101,40 +94,87 @@ def report_computed_task(timestamp = None, subtask_id = None, task_to_compute =
...
-                        task_id     = task_id + '2A',
-                        subtask_id  = task_id + '2A',
-                        deadline    = current_time + 3600,  # current_time + 01:00:00
-                    )
-                )
+            timestamp=current_timestamp,  # current_time
+            ack_report_computed_task=ack_report_computed_task(
+                timestamp=ack_report_computed_task_timestamp,  # current_time - 4:00:10
+                subtask_id=task_id + '2A',
+                task_to_compute=task_1
```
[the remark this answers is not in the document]
cameel: I see that the code was using the same value for `subtask_id` and `task_id` before so I understand why it's confusing. We have realized the problem only recently and have to fix it in other places in our code too.

concent_api/api-integration-force-accept-or-reject-results-test.py
```diff
@@ -171,39 +205,35 @@ def main():
...
         cluster_url,
         'send',
         PROVIDER_PRIVATE_KEY,
         CONCENT_PUBLIC_KEY,
         force_subtask_results(
-            timestamp = timestamp_to_isoformat(current_time),  # current_time
+            timestamp = current_timestamp,  # current_time
             ack_report_computed_task = ack_report_computed_task(
-                timestamp = timestamp_to_isoformat(current_time - 14401),  # current_time - 4:00:01
+                timestamp = ack_report_computed_task_timestamp,  # current_time - 4:00:10
```
cameel: I see you changed the time in the comment. Was the comment wrong or is it an accidental change?

concent_api/api_testing_common.py
```diff
@@ -0,0 +1,132 @@
...
+        print('RAW RESPONSE: Reponse content is None')
+    elif len(response.content) != 0:
+        _print_messge_from_response(private_key, public_key, response)
+    else:
+        print('STATUS: {} {}'.format(response.status_code, http.client.responses[response.status_code]))
+        if response.text not in ['', None]:
+            print('RAW RESPONSE: {}'.format(response.text))
+
+
+def _print_messge_from_response(private_key, public_key, response):
```
cameel: Typo: `messge` -> `message`

concent_api/api_testing_common.py
```diff
@@ -0,0 +1,132 @@
...
+    for field, value in zip(fields, values):
+        if isinstance(value, Message):
+            print_golem_message(value, private_key, public_key, indent = indent + 4)
+        else:
+            print('{}{:30} = {}'.format(' ' * indent, field, value))
+
+
+def validate_response(actual_status_code, expected_status):
+    if expected_status is not None:
+        assert expected_status == actual_status_code, f"Expected:HTTP{expected_status}, actual:HTTP{actual_status_code}"
```
cameel: Could you replace this with a normal `if`? I don't think `assert` is what you meant here. It's for checking invariants and assuptions that are supposed to be true as long as the code is bug free. Not for validating input from user or an external server. Assertions are a debug feature, normally disabled in production.

concent_api/api-integration-force-accept-or-reject-results-test.py
dybi: @cameel, `constants = ProtocolConstants(**json)` might be unsafe if json contains additional params

concent_api/api-integration-force-accept-or-reject-results-test.py
dybi: I have changed the time, because I had introduced TIME_GUARD variable - a safety offset

concent_api/api-integration-force-accept-or-reject-results-test.py
dybi: @cameel - it's a time offset added to calculated times, to ensure tests will not fail because of 1 sec time difference on host, etc.

concent_api/api_testing_common.py
dybi: here `assert` is used in the same context as typically self.assertEqual from unittest is used - to check that software under test behaves as it should, that the response we have received is as inteded. Otherwise a particular test fails.

concent_api/api-integration-force-accept-or-reject-results-test.py
dybi: I disagree. Well named constant is batter than some magic naumber, even if it is used only once. As to calculation, I am pretty sure I can handle it without timedelta ;)

rwrzesien: I just don't see a difference between this and, for example `ONE = 1`. Do you expect this value to ever change ? If yes, then it makes sense, but then it should be called differently, more meaningful in context of the test.

concent_api/api-integration-force-accept-or-reject-results-test.py
```diff
@@ -12,28 +13,18 @@
 from utils.helpers          import get_current_utc_timestamp
 from utils.testing_helpers  import generate_ecc_key_pair
 
-from api_testing_helpers    import api_request
-from api_testing_helpers    import timestamp_to_isoformat
+from api_testing_common     import api_request, parse_command_line, get_protocol_constants, print_protocol_constants
```
rwrzesien: Each import should be in separate line.

concent_api/api-integration-force-accept-or-reject-results-test.py
```diff
@@ -12,28 +13,18 @@
...
+from api_testing_common     import timestamp_to_isoformat
 
-import requests
+HOUR = 3600
```
rwrzesien: I still think there is no need to keep that.

concent_api/api-integration-force-accept-or-reject-results-test.py
```diff
@@ -172,38 +174,34 @@ def main():
...
         headers = {
-            'Content-Type':                     'application/octet-stream',
-            'concent-client-public-key':        b64encode(PROVIDER_PUBLIC_KEY).decode('ascii'),
-            'concent-other-party-public-key':   b64encode(REQUESTOR_PUBLIC_KEY).decode('ascii'),
-            'temporary-account-funds':          ''
-        }
+            'Content-Type'                  : 'application/octet-stream',
+            'concent-client-public-key'     : b64encode(PROVIDER_PUBLIC_KEY).decode('ascii'),
+            'concent-other-party-public-key': b64encode(REQUESTOR_PUBLIC_KEY).decode('ascii'),
+            'temporary-account-funds'       :          ''
```
rwrzesien: Please remove extra spaces after `:`

concent_api/api-integration-force-accept-or-reject-results-test.py
```diff
@@ -172,38 +174,34 @@ def main():
...
-                        task_id     = task_id,
-                        subtask_id  = task_id,
-                        deadline    = current_time + 3600,  # current_time + 01:00:00
-                    )
-                )
+            timestamp = current_timestamp,  # current_time
+            ack_report_computed_task    = ack_report_computed_task(
+                    timestamp           = ack_report_computed_task_timestamp,  # current_time - 4:00:10
+                    subtask_id          = task_id,
+                    task_to_compute     = create_task(task_to_compute_timestamp, task_id, compute_task_def_deadline)
```
rwrzesien: Each parameter should be in separate line.

concent_api/api-integration-force-accept-or-reject-results-test.py
```diff
@@ -213,26 +211,20 @@ def main():
...
-                        task_id     = task_id + '2C',
-                        subtask_id  = task_id + '2C',
-                        deadline    = current_time,
-                    )
-                )
+            timestamp                   = timestamp_to_isoformat(current_time),
+            ack_report_computed_task    = ack_report_computed_task(
+                timestamp               = timestamp_to_isoformat(current_time - 2 * cluster_consts.subtask_verification_time - TIME_GUARD),  # current_time - 08:00:10
+                subtask_id              = task_id + '2C',
+                task_to_compute         = create_task(timestamp_to_isoformat(current_time - 28802), task_id + '2C', current_time)
```
rwrzesien: Line is too long, each parameter should be in separate line.

concent_api/api-integration-force-accept-or-reject-results-test.py
```diff
@@ -213,26 +211,20 @@ def main():
...
-                    timestamp = timestamp_to_isoformat(current_time - 28802),  # current_time - 08:00:02
-                    compute_task_def = compute_task_def(
-                        task_id     = task_id + '2C',
-                        subtask_id  = task_id + '2C',
-                        deadline    = current_time,
-                    )
-                )
+            timestamp                   = timestamp_to_isoformat(current_time),
+            ack_report_computed_task    = ack_report_computed_task(
+                timestamp               = timestamp_to_isoformat(current_time - 2 * cluster_consts.subtask_verification_time - TIME_GUARD),  # current_time - 08:00:10
```
rwrzesien: Line is too long, each parameter should be in separate line.

concent_api/api-integration-force-accept-or-reject-results-test.py
```diff
@@ -262,38 +255,35 @@ def main():
...
-            'Content-Type': 'application/octet-stream',
+            'Content-Type'             : 'application/octet-stream',
             'concent-client-public-key': b64encode(PROVIDER_PUBLIC_KEY).decode('ascii'),
-        }
+        },
+        expected_status = 200
     )
 
     # Test CASE 2D + 3 + 4B + 5. Requestor sends ForceSubtaskResultsResponse with SubtaskResultsRejected
+    task_2 = create_task(task_to_compute_timestamp, task_id + '2D', compute_task_def_deadline)
```
rwrzesien: Each parameter should be in separate line.

concent_api/api-integration-force-get-task-result-test.py
```diff
@@ -1,132 +1,147 @@
...
+from golem_messages.message.concents import AckForceGetTaskResult, ForceGetTaskResultUpload, ForceGetTaskResultFailed
 
-from api_testing_helpers    import api_request
-from api_testing_helpers    import timestamp_to_isoformat
+from utils.helpers                   import get_current_utc_timestamp
+from utils.testing_helpers           import generate_ecc_key_pair
 
-from freezegun              import freeze_time
-import requests
+from api_testing_common              import api_request, parse_command_line, create_task_to_compute, get_protocol_constants, \
```
rwrzesien: Each import should be in separate line.

concent_api/api-integration-force-get-task-result-test.py
```diff
@@ -1,132 +1,147 @@
...
+
 (PROVIDER_PRIVATE_KEY,  PROVIDER_PUBLIC_KEY)  = generate_ecc_key_pair()
 (REQUESTOR_PRIVATE_KEY, REQUESTOR_PUBLIC_KEY) = generate_ecc_key_pair()
 
 
-def parse_command_line(command_line):
-    if len(command_line) <= 1:
-        sys.exit('Not enough arguments')
+def prepare_hash(file_check_sum, hash_offset):
+    return file_check_sum[:-1] + hex(int(file_check_sum[-1], 16) + hash_offset)[2:]
```
rwrzesien: Please add a comment what this function does, name is good, but if this is some common method used to prepare hash it is worth describing.

concent_api/api-integration-force-get-task-result-test.py
```diff
@@ -1,132 +1,147 @@
...
+    response = upload_file(file_content, headers)
+    return (response.status_code, file_size, file_check_sum)
 
-    file_content    = task_id
-    file_size       = len(file_content)
-    file_check_sum  = 'sha1:' + hashlib.sha1(file_content.encode()).hexdigest()
-    file_path       = '{}/{}/result'.format(task_id, part_id)
 
+def prepare_file_for_transfer(part_id, task_id, time_offset):
+    file_content = task_id
```
rwrzesien: ` = ` should be aligned.

concent_api/api-integration-force-get-task-result-test.py
```diff
@@ -1,132 +1,147 @@
...
-    file_transfer_token.authorized_client_public_key    = CONCENT_PUBLIC_KEY
-    file_transfer_token.operation                       = 'upload'
-
-    file_transfer_token.files                   = [message.FileTransferToken.FileInfo()]
-    file_transfer_token.files[0]['path']        = file_path
-    file_transfer_token.files[0]['checksum']    = file_check_sum
-    file_transfer_token.files[0]['size']        = file_size
-
-    upload_token    = shortcuts.dump(file_transfer_token, CONCENT_PRIVATE_KEY, CONCENT_PUBLIC_KEY)
+    file_transfer_token.token_expiration_deadline = int(datetime.datetime.now().timestamp()) + time_offset
```
rwrzesien: ` = ` should be aligned.

concent_api/api-integration-force-get-task-result-test.py
```diff
@@ -1,132 +1,147 @@
...
-    (response_status_code, file_size, file_check_sum) = upload_new_file_on_cluster(
-        task_id = task_id,
-        part_id = '0',
-        current_time = current_time,
-    )
-    if response_status_code == 200:
-        print('\nCreated file with task_id {}. Checksum of this file is {}, and size of this file is {}.\n'.format(task_id, file_check_sum, file_size))
-    else:
-        print('File has not been stored on cluster')
+class count_fails(object):
```
rwrzesien: I think class name should be `CountFails` in this case.

concent_api/api-integration-force-get-task-result-test.py
```diff
@@ -1,132 +1,147 @@
...
-            size        = file_size,
-            checksum    = file_check_sum
-        ),
+        get_force_get_task_result(task_id, current_time, size = file_size, checksum = file_check_sum),
         headers = {
-            'Content-Type':                     'application/octet-stream',
-            'concent-client-public-key':        b64encode(REQUESTOR_PUBLIC_KEY).decode('ascii'),
-            'concent-other-party-public-key':   b64encode(PROVIDER_PUBLIC_KEY).decode('ascii'),
-        }
+            'Content-Type': 'application/octet-stream',
```
rwrzesien: Please align.

concent_api/api-integration-force-get-task-result-test.py
```diff
@@ -135,65 +150,201 @@ def main():
...
-            size    = 1024,
-            checksum = '098f6bcd4621d373cade4e832627b4f6'
-        ),
+        get_force_get_task_result(task_id, current_time, size = file_size, checksum = file_check_sum),
         headers = {
-            'Content-Type':                     'application/octet-stream',
-            'concent-client-public-key':        b64encode(REQUESTOR_PUBLIC_KEY).decode('ascii'),
-            'concent-other-party-public-key':   b64encode(PROVIDER_PUBLIC_KEY).decode('ascii'),
-        }
+            'Content-Type': 'application/octet-stream',
```
rwrzesien: Please align.

concent_api/api-integration-force-get-task-result-test.py
```diff
@@ -135,65 +150,201 @@ def main():
...
     api_request(
         cluster_url,
         'receive',
         PROVIDER_PRIVATE_KEY,
         CONCENT_PUBLIC_KEY,
         headers = {
-            'Content-Type':                 'application/octet-stream',
-            'concent-client-public-key':    b64encode(PROVIDER_PUBLIC_KEY).decode('ascii')
-        }
+            'Content-Type': 'application/octet-stream',
```
rwrzesien: Please align. Actually there is a more of those below, please fix them.

concent_api/api-integration-force-get-task-result-test.py
```diff
@@ -135,65 +150,201 @@ def main():
...
+    tests = [objects[name] for name in tests_to_execute]
+    for test in tests:
+        task_id = kwargs['task_id'] + test.__name__
+        kw = {k: v for k, v in kwargs.items() if k != 'task_id'}
+        test(task_id=task_id, **kw)
+        print("-" * 80)
+
+
+def main():
+    global CONCENT_MESSAGING_TIME
```
rwrzesien: Isn't it enough to use constants defined at the beginning of the file, and just set their values here ?

concent_api/api-integration-force-report-computed-task-test.py
```diff
@@ -8,15 +8,13 @@
...
-from golem_messages.message         import ComputeTaskDef
 from golem_messages.message         import ForceReportComputedTask
-from golem_messages.message         import TaskToCompute
 from golem_messages.message.tasks   import ReportComputedTask
 
 from utils.helpers                  import get_current_utc_timestamp
 from utils.testing_helpers          import generate_ecc_key_pair
 
-from api_testing_helpers            import api_request
+from api_testing_common import api_request, parse_command_line, create_task_to_compute
```
rwrzesien: Please fix imports.

concent_api/api_testing_common.py
```diff
@@ -0,0 +1,176 @@
+import datetime
+import http.client
+import json
+import requests
+import sys
+
+from collections                    import namedtuple
+
+from golem_messages.exceptions      import MessageError
+from golem_messages.message         import Message, ComputeTaskDef, TaskToCompute
```
rwrzesien: Each import should be in separate line.

concent_api/api_testing_common.py
```diff
@@ -0,0 +1,176 @@
...
+from golem_messages.shortcuts       import dump
+from golem_messages.shortcuts       import load
+
+
+ProtocolConstants = namedtuple("ProtocolConstants",
```
rwrzesien: I see this is still not fixed.

concent_api/api_testing_common.py
```diff
@@ -0,0 +1,176 @@
...
+                               ["concent_messaging_time",
+                                "subtask_verification_time",
+                                "force_acceptance_time",
+                                "token_expiration_time"])
+
+
+def get_protocol_constants(cluster_url):
+    url = f"{cluster_url}/api/v1/protocol-constants/"
```
rwrzesien: There is a lot of alignment issues in this file.

concent_api/api_testing_common.py
```diff
@@ -0,0 +1,176 @@
...
+    print_golem_message(decoded_response)
+
+
+def try_to_decode_golem_message(private_key, public_key, content):
+    try:
+        decoded_response = load(
+            content,
+            private_key,
+            public_key,
+            check_time=False
```
rwrzesien: Please add spaces between ` = `.

concent_api/utils/api_view.py
```diff
@@ -81,6 +81,7 @@ def wrapper(request, *args, **kwargs):
             )
             return JsonResponse({'error': str(exception)}, status = 400)
         if isinstance(response_from_view, Message):
+            assert response_from_view.sig is None
```
rwrzesien: This is already in master, please remove it and rebase on master if you need it.

concent_api/core/views.py
```diff
@@ -334,22 +335,10 @@ def protocol_constants(_request):
...
-            },
-            'subtask_verification_time': {
-                'name': 'SUBTASK_VERIFICATION_TIME',
-                'value': settings.SUBTASK_VERIFICATION_TIME,
-            },
-            'token_expiration_time': {
-                'name': 'TOKEN_EXPIRATION_TIME',
-                'value': settings.TOKEN_EXPIRATION_TIME,
-            },
+            'concent_messaging_time':    settings.CONCENT_MESSAGING_TIME,
```
rwrzesien: This is already in master, please remove and rebase.

concent_api/api-integration-force-accept-or-reject-results-test.py
```diff
@@ -172,38 +174,34 @@ def main():
...
-                    compute_task_def = compute_task_def(
-                        task_id     = task_id,
-                        subtask_id  = task_id,
-                        deadline    = current_time + 3600,  # current_time + 01:00:00
-                    )
-                )
+            timestamp = current_timestamp,  # current_time
+            ack_report_computed_task    = ack_report_computed_task(
+                    timestamp           = ack_report_computed_task_timestamp,  # current_time - 4:00:10
+                    subtask_id          = task_id,
```
rwrzesien: You are adding + '2A' to previous calls, I think it is worth adding '2B' here to for consistency, and for easier examination in logs/database if we will have to debug.

concent_api/utils/api_view.py
rwrzesien: Please rebase on latest master, those changes which are already in master should't be visible here.

concent_api/api-integration-force-get-task-result-test.py
```diff
@@ -1,132 +1,147 @@
...
+        'Concent-upload-path'      : '{}/{}/result'.format(task_id, part_id),
+        'Content-Type'             : 'application/x-www-form-urlencoded'
     }
+    return file_check_sum, file_content, file_size, headers
 
-    response = requests.post("{}".format(STORAGE_CLUSTER_ADDRESS + 'upload/'), headers = headers, data = file_content)
-    return (response.status_code, file_size, file_check_sum)
 
+def upload_file(file_content, headers):
+    response = requests.post("{}".format(STORAGE_CLUSTER_ADDRESS + 'upload/'), headers = headers, data = file_content,
```
rwrzesien:
```
requests.post(
    "{}".format(
        STORAGE_CLUSTER_ADDRESS + 'upload/'
    ), 
    headers = headers, 
    data = file_content,
    verify = False,
)
```

concent_api/api-integration-force-get-task-result-test.py
```diff
@@ -1,132 +1,147 @@
...
+    instances = []
+
+    def __init__(self, fun):
+        self._fun     = fun
+        self.__name__ = fun.__name__
+        self.failed  = False
+        count_fails.instances.append(self)
+
+    def __call__(self, *args, **kwargs):
+                try:
```
rwrzesien: Please align.

concent_api/api_testing_common.py
```diff
@@ -0,0 +1,176 @@
...
+                                "subtask_verification_time",
+                                "force_acceptance_time",
+                                "token_expiration_time"])
+
+
+def get_protocol_constants(cluster_url):
+    url = f"{cluster_url}/api/v1/protocol-constants/"
+    response = requests.get(url)
+    json = response.json()
+    concent_messaging_time = json['concent_messaging_time']
```
rwrzesien: Please review styling in this file, I see all issues that were here on previous review.

concent_api/api-integration-force-get-task-result-test.py
```diff
@@ -1,132 +1,147 @@
...
-from golem_messages         import shortcuts
+import requests
+import time
 
-from utils.helpers          import get_current_utc_timestamp
-from utils.testing_helpers  import generate_ecc_key_pair
+from base64                          import b64encode
+from golem_messages                  import message
+from golem_messages                  import shortcuts
+from golem_messages.message.concents import AckForceGetTaskResult, ForceGetTaskResultUpload, ForceGetTaskResultFailed
```
rwrzesien: Each import should be in separate line.

concent_api/api-integration-force-get-task-result-test.py
```diff
@@ -135,65 +150,201 @@ def main():
...
+    print_protocol_constants(cluster_consts)
+    CONCENT_MESSAGING_TIME = cluster_consts.concent_messaging_time
+    FORCE_ACCEPTANCE_TIME  = cluster_consts.force_acceptance_time
+    TOKEN_EXPIRATION_TIME  = cluster_consts.token_expiration_time
+    WAIT_TIME_1            = DEADLINE_OFFSET + FORCE_ACCEPTANCE_TIME  # recalculated as it could change due to cluster_consts
+
+    execute_tests(tests_to_execute, cluster_url=cluster_url, current_time=current_time, task_id=task_id)
+
+    total_fails = count_fails.get_fails()
+    if total_fails > 0:
```
rwrzesien: How about moving this into the class, and just call `count_fails.print_fails()` ?

concent_api/api-integration-force-accept-or-reject-results-test.py
dybi: done

concent_api/api-integration-force-accept-or-reject-results-test.py
dybi: @rwrzesien I still prefer to have it here ;)

concent_api/api-integration-force-accept-or-reject-results-test.py
dybi: done

concent_api/api-integration-force-accept-or-reject-results-test.py
dybi: done

concent_api/api-integration-force-accept-or-reject-results-test.py
dybi: done;

concent_api/api-integration-force-accept-or-reject-results-test.py
dybi: done

concent_api/api-integration-force-accept-or-reject-results-test.py
dybi: @rwrzesien , there is no such recommendation in our coding style guide. Why would you liketo have it like that?

concent_api/api-integration-force-get-task-result-test.py
dybi: done

concent_api/api-integration-force-get-task-result-test.py
dybi: done

concent_api/api-integration-force-accept-or-reject-results-test.py
rwrzesien: @dybi Once upon a time we had a discussion about it and @cameel told me he will add it there :)

concent_api/api-integration-force-get-task-result-test.py
dybi: done

concent_api/api-integration-force-get-task-result-test.py
dybi: done

concent_api/api-integration-force-get-task-result-test.py
dybi: done

concent_api/api-integration-force-get-task-result-test.py
dybi: done

concent_api/api-integration-force-get-task-result-test.py
dybi: to set a value of a global variable in a local scope you need to declare it:
`global CONCENT_MESSAGING_TIME`

concent_api/api-integration-force-payment.py
```diff
@@ -12,8 +12,8 @@
 from utils.helpers          import get_current_utc_timestamp
 from utils.testing_helpers  import generate_ecc_key_pair
 
-from api_testing_helpers    import api_request
-from api_testing_helpers    import timestamp_to_isoformat
+from api_testing_common import api_request, parse_command_line
```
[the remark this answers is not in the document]
dybi: done

concent_api/api-integration-force-report-computed-task-test.py
dybi: done

concent_api/api_testing_common.py
dybi: done

concent_api/api_testing_common.py
dybi: done

concent_api/api_testing_common.py
dybi: done

concent_api/api-integration-force-get-task-result-test.py
dybi: done

concent_api/api-integration-force-get-task-result-test.py
dybi: done

concent_api/api-integration-force-payment.py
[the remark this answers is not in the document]
dybi: done

concent_api/api-integration-force-get-task-result-test.py
dybi: to use global for writing in a local scope, you have to declare it first; otherwise you will create local variable of the same name as global

concent_api/api-integration-force-report-computed-task-test.py
dybi: done

concent_api/api_testing_common.py
dybi: done

concent_api/api_testing_common.py
dybi: done

concent_api/api_testing_common.py
dybi: done

concent_api/api_testing_common.py
dybi: done

concent_api/api-integration-force-accept-or-reject-results-test.py
dybi: done

concent_api/api-integration-force-get-task-result-test.py
dybi: done

concent_api/api-integration-force-get-task-result-test.py
dybi: done

concent_api/api_testing_common.py
dybi: done

concent_api/api-integration-force-get-task-result-test.py
dybi: done

concent_api/api-integration-force-get-task-result-test.py
dybi: @rwrzesien  when the fails should be printed? How could one know when all tests have been executed?

rwrzesien: @dybi What I have meant here is to move 

```
if total_fails > 0:
    print(f'Total failed tests : {total_fails} out of {number_of_tests}')
..

into some method inside `count_fails` class, and instead of doing this check + print in test code, do it inside method of `count_fails` class, and in test code only call this method. In example (pseudo code):

```
class count_fails(object):

    def print_total_fails(self):
         if self.get_total_fails() > 0:
              print(f'Total failed tests : {total_fails} out of {number_of_tests}')

......

count_fails.print_total_fails()
```

File: concent_api/api-integration-force-accept-or-reject-results-test.py (hunk 17 above)
Comment (cameel) [reply concent_api/api-integration-force-accept-or-reject-results-test.py]: I know, you explained it to me earlier. My point was that you should put this explanation in a comment in the code.

File: concent_api/api_testing_common.py [hunk 53]
```diff
@@ -0,0 +1,187 @@
...
+    concent_messaging_time      = json['concent_messaging_time']
+    subtask_verification_time   = json['subtask_verification_time']
+    force_acceptance_time       = json['force_acceptance_time']
+    token_expiration_time       = json['token_expiration_time']
+    constants                   = ProtocolConstants(
+            concent_messaging_time      = concent_messaging_time,
+            subtask_verification_time   = subtask_verification_time,
+            force_acceptance_time       = force_acceptance_time,
+            token_expiration_time       = token_expiration_time
+    )
```
cameel: We talked about it earlier and there was something that you made me realize. If the server returns constants not listed in `ProtocolConstants`, this code won't notice. We want the code to crash in that case so that we can fix it. So please add a check that ensures that only the constants listed here are in the JSON response.

Alternatively you could use the form I showed in my previous comment - it's more concise and at the same time gives you that check:
``` python
constants = ProtocolConstants(**json)
```

By the way, we use 4 spaces per indent level. You have 6 here.

concent_api/api-integration-force-accept-or-reject-results-test.py
cameel: I also think it's not needed and I would not define it myself but it does not break anything so if Piotr prefers to keep it, I'm for keeping it. It looks like a matter of opinion to me.

concent_api/api-integration-force-accept-or-reject-results-test.py
cameel: I was objecting to something like this:
``` python
task_2 = create_task(
    task_to_compute_timestamp, task_id + '2D', compute_task_def_deadline
)
```
because it suggests that it's a single parameter. But this:

``` python
task_2 = create_task(task_to_compute_timestamp, task_id + '2D', compute_task_def_deadline)
```
is typical and completely fine. It's still hard to distinguish the parameters but it does not mislead you into thinking it's a single one when you're just skimming through the code.

About the style guide: I'm sorry, I have a list of minor stuff that came up in reviews to add there. I was a bit too overloaded recently to update it but I'm going to get to if after the talk about style we have planned.

concent_api/api-integration-force-get-task-result-test.py
cameel: But if we are changing its value, is it still a constant?

I agree with @rwrzesien here. There should be no reason to use `global` here. Either it's a constant and we don't need to change it or it's not a constant and we should be passing it everywhere as a parameter (possibly inside `ProtocolConstants`).

Third option, which I'm against because I don't see a good reason for using global variables (that are not constants) here, we rename it to `concent_messaging_time` so that it's obvious it's not a constant.

concent_api/api-integration-force-get-task-result-test.py
```diff
@@ -1,197 +1,362 @@
...
+        self._fun     = fun
+        self.__name__ = fun.__name__
+        self.failed  = False
+        count_fails.instances.append(self)
+
+    def __call__(self, *args, **kwargs):
+        try:
+            print("Running TC: " + self.__name__)
+            return self._fun(*args, **kwargs)
+        except AssertionError as e:
```
cameel: Please do not use `AssertionError` here. This conflicts with the typical use of `assert` - for checking assumptions. If an `assert` fails it means that we have a bug in the testing code itself, not that the server we're testing behaved in a wrong way. In that case we want to interrupt the script immediately and fix it. Not include it in the report.

Please define a custom exception class and catch that. And do not use `assert` for testing expectations.

concent_api/api-integration-force-get-task-result-test.py
```diff
@@ -1,197 +1,362 @@
...
-    else:
-        print('File has not been stored on cluster')
+class count_fails(object):
+    """
+    Decorator that wraps a test functions for intercepting assertions and counting them.
+    """
+    instances = []
+
+    def __init__(self, fun):
+        self._fun     = fun
```
cameel: Please use whole words for variable names. I suspect this was meant to be `function`?

concent_api/api-integration-force-get-task-result-test.py
cameel: Please use full words for variable names. `e` -> `exception`.

concent_api/api-integration-force-get-task-result-test.py
```diff
@@ -1,197 +1,362 @@
...
+            self.failed = True
+
+    @classmethod
+    def get_fails(cls):
+        return sum([instance.failed for instance in cls.instances])
+
+
+@count_fails
+def case_4d_concent_notifies_the_provider_that_task_results_are_ready(cluster_url, current_time, task_id):
+    file_check_sum, file_content, file_size, headers = prepare_file_for_transfer('0', task_id, TOKEN_EXPIRATION_TIME)
```
cameel: Please use parentheses around tuples:

``` python
(file_check_sum, file_content, file_size, headers) = prepare_file_for_transfer('0', task_id, TOKEN_EXPIRATION_TIME)
```
This makes it clear at a glance that you're assigning multiple values here.

concent_api/api-integration-force-get-task-result-test.py
```diff
@@ -1,197 +1,362 @@
...
+
+    response = upload_file(file_content, headers)
+
+    assert response.status_code == 200, 'File has not been stored on cluster'
```
cameel: Please do not use `assert` for checking expectations. We use `assert` for checking assumptions about the code. If we do `assert X` then `X` is expected to be universally true unless the code has a bug. This is not the case here. This may be false when server returns an unexpected response.

concent_api/api-integration-force-get-task-result-test.py
```diff
@@ -1,197 +1,362 @@
...
-
+DEADLINE_OFFSET        = 5
+CONCENT_MESSAGING_TIME = 30
+FORCE_ACCEPTANCE_TIME  = 30
+TOKEN_EXPIRATION_TIME  = 3600
 
-def parse_command_line(command_line):
-    if len(command_line) <= 1:
-        sys.exit('Not enough arguments')
+WAIT_TIME_1            = DEADLINE_OFFSET + FORCE_ACCEPTANCE_TIME
```
cameel: Please give it a more meaningful name. When I was reading the test code I had to go up here to check what this is. If it had a self-explanatory name I would not have to do so.

Actually, any good name I can come up with is the combination of those two so I think that we should remove this constant. It would be more self-explanatory to have just `DEADLINE_OFFSET + FORCE_ACCEPTANCE_TIME` in the code.

concent_api/api-integration-force-get-task-result-test.py
```diff
@@ -1,197 +1,362 @@
...
+from api_testing_common              import create_task_to_compute
+from api_testing_common              import get_protocol_constants
+from api_testing_common              import print_protocol_constants
 
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "concent_api.settings")
 
-(PROVIDER_PRIVATE_KEY,  PROVIDER_PUBLIC_KEY)  = generate_ecc_key_pair()
-(REQUESTOR_PRIVATE_KEY, REQUESTOR_PUBLIC_KEY) = generate_ecc_key_pair()
-
+DEADLINE_OFFSET        = 5
```
cameel: What is `DEADLINE_OFFSET`? This needs an explanatory comment because it's not obvious.

If I had to guess, I'd say it's the difference between `TaskToCompute.deadline` and `TaskToCompute.timestamp`. Is that right? If that's the case then I think it needs a better name. For example `DEFAULT_TASK_DURATION`, though maybe you can come up with a better one. Something that would suggest that (1) it's not really a protocol constant but just a default you're using here (tasks do not have constant durations; it completely depends on what requestor and provider agree) and (2) that it's the task duration, not some offset that you add on top of the task deadline.

concent_api/api-integration-force-accept-or-reject-results-test.py
rwrzesien: Off course, I was just wondering if this can get more meaningful name in context of the usage in test.

concent_api/api-integration-force-accept-or-reject-results-test.py
cameel: This is still not fixed.

If the server returns HTML (with `Content-Type` of `text/html`), this code will try to parse it as JSON.

concent_api/api_testing_common.py
cameel: But this is not `unittest`. Using `assert` here means that we're checking an invariant, precondition, etc. That we're stating something that universally holds unless the code has a bug.

In `unittest` we basically have no choice because of their unfortunate design choice to reuse `AssertionError` instead of creating a custom exception for reporting test failures. This conflates errors caused by failing assumptions (bugs in the testing code) with actual test failures (bugs in the code under test). I think that `unittest` was just written by someone who never seriously used `assert` for its intended purpose and did not care. We do care though.

We still keep that distinction in automated tests. You'll notice that we only ever use bare `assert` when we're making sure that all the data and environment satisfies our assumptions before we run the code under test.

concent_api/api_testing_common.py
```diff
@@ -0,0 +1,187 @@
...
+
+def validate_response_status(actual_status_code, expected_status):
+    if expected_status is not None:
+        assert expected_status == actual_status_code, f"Expected:HTTP{expected_status}, actual:HTTP{actual_status_code}"
+
+
+def validate_response_message(encoded_message, expected_message, private_key, public_key,):
+    if expected_message is not None:
+        decoded_message = try_to_decode_golem_message(private_key, public_key, encoded_message)
+        assert isinstance(decoded_message, expected_message), f"Expected:{expected_message}, actual:{decoded_message}"
```
cameel: This code should use `Content-Type` header when determining how to decode it. Please look at how this code worked before the rewrite.

concent_api/api_testing_common.py
```diff
@@ -0,0 +1,187 @@
...
+        sys.exit('Not enough arguments')
+
+    if len(command_line) >= 3:
+        sys.exit('Too many arguments')
+
+    cluster_url = command_line[1]
+    return cluster_url
+
+
+def create_task_to_compute(current_time, task_id, deadline_offset=60):
```
cameel: There is a global constant called `DEADLINE_OFFSET`. Why is it hard-coded to 60 here instead of coming from the constant?

concent_api/api_testing_common.py
```diff
@@ -0,0 +1,187 @@
...
+        assert expected_status == actual_status_code, f"Expected:HTTP{expected_status}, actual:HTTP{actual_status_code}"
+
+
+def validate_response_message(encoded_message, expected_message, private_key, public_key,):
+    if expected_message is not None:
+        decoded_message = try_to_decode_golem_message(private_key, public_key, encoded_message)
+        assert isinstance(decoded_message, expected_message), f"Expected:{expected_message}, actual:{decoded_message}"
+
+
+def api_request(host, endpoint, private_key, public_key, data=None, headers=None, expected_status=None, expected_message=None):
```
cameel: We need `expected_content_type` here. Or it might be assumed based on the data type. Right now it's not checked at all.

concent_api/api_testing_common.py
```diff
@@ -0,0 +1,187 @@
...
+            token_expiration_time       = token_expiration_time
+    )
+    return constants
+
+
+def print_protocol_constants(constants):
+    print("PROTOCOL_CONSTANTS: ")
+    print(f"concent_messaging_time = {constants.concent_messaging_time}")
+    print(f"subtask_verification_time = {constants.subtask_verification_time}")
+    print(f"force_acceptance_time = {constants.force_acceptance_time}")
```
cameel: Why not do it in a loop? We won't have to update it each time we add a new constant.

concent_api/api_testing_common.py
```diff
@@ -0,0 +1,187 @@
...
+
+from golem_messages.exceptions      import MessageError
+from golem_messages.message         import ComputeTaskDef
+from golem_messages.message         import Message
+from golem_messages.message         import TaskToCompute
+from golem_messages.shortcuts       import dump
+from golem_messages.shortcuts       import load
+
+
+ProtocolConstants = namedtuple(
```
cameel: This could be [typing.NamedTuple](https://docs.python.org/3/library/typing.html#typing.NamedTuple).

Sorry for not pointing it out before but only recetly realized that there are typed named tuples in Python.

concent_api/api_testing_common.py
dybi: done

concent_api/api-integration-force-get-task-result-test.py
dybi: done

concent_api/api-integration-force-get-task-result-test.py
dybi: done

concent_api/api-integration-force-get-task-result-test.py
dybi: ... done

concent_api/api-integration-force-get-task-result-test.py
dybi: done

concent_api/api-integration-force-get-task-result-test.py
dybi: done

concent_api/api-integration-force-get-task-result-test.py
dybi: a more meaningful name has been given

concent_api/api-integration-force-get-task-result-test.py
dybi: agreed; changed to DEFAULT_TASK_DURATION

concent_api/api_testing_common.py
dybi: changed to constant

concent_api/api_testing_common.py
dybi: done

concent_api/api_testing_common.py
dybi: done

concent_api/api_testing_common.py
dybi: done

concent_api/api_testing_common.py
dybi: done

concent_api/api-integration-force-accept-or-reject-results-test.py
```diff
@@ -262,38 +267,35 @@ def main():
...
-            )
         ),
         headers = {
-            'Content-Type':                     'application/octet-stream',
-            'concent-client-public-key':        b64encode(PROVIDER_PUBLIC_KEY).decode('ascii'),
-            'concent-other-party-public-key':   b64encode(REQUESTOR_PUBLIC_KEY).decode('ascii'),
-            'temporary-account-funds':          'True'
-        }
+            'Content-Type'                  : 'application/octet-stream',
+            'concent-client-public-key'     : b64encode(PROVIDER_PUBLIC_KEY).decode('ascii'),
```
cameel: Ekhm... our style is not that eccentric :) I don't know where it's coming from but it was fine before this change.

concent_api/api-integration-force-get-task-result-test.py
```diff
@@ -1,197 +1,341 @@
...
-
-    authorized_golem_transfer_token = 'Golem ' + encrypted_token
-
+    file_transfer_token.files                           = [message.FileTransferToken.FileInfo()]
+    file_transfer_token.files[0]['path']                = file_path
+    file_transfer_token.files[0]['checksum']            = file_check_sum  # prepare_hash(file_check_sum, wrong_hash)
+    file_transfer_token.files[0]['size']                = file_size
+    upload_token                                        = shortcuts.dump(file_transfer_token,
+                                                                         CONCENT_PRIVATE_KEY,
+                                                                         CONCENT_PUBLIC_KEY
```
cameel: This script should not assume that it has access to Concent's private key. It's meant to run against a live cluster where it acts as a normal client without any special knowledge about server's secrets.

I see that the key was being used here already before your changes so it's your mistake but please fix it as a part of this task. It's necessary if we want to be able to run these tests against staging.

concent_api/api-integration-force-get-task-result-test.py
dybi: done

cameel:
This pull request got big and has too many comments. It's Piotr's first pull request in our project and most remarks are related to hashing out our initial coding differences while there are still problems that need to be fixed. We must do this on something smaller.

I talked with Piotr today and we've decided that we'll scrap it and he'll resubmit it as a series of smaller pull requests that can be reviewed and merged quickly and efficiently.

I suggest splitting it the following way:
- Renaming `api_testing_helpers.py` to `api_testing_common.py` and extracting repeating code into it.
- Any updates needed for a newer golem-messages version.
- `ProtocolConstants` and any code related to using them and getting them from the server.
- `count_fails` and any code related to validating results returned from the sever and making the existing scripts behave like test cases.
- Testing file uploads.
- Any minor refactoring or style changes not related to the above.

This is just my suggestion. @dybi Please use your own judgement in splitting it but try to keep the changes small, complete and self-consistent. I think it would be beneficial to split it in even smaller parts if you can. Don't hesitate to do pull requests that are even just a few lines long if you think they're parts that are straightforward and can be merged without much discussion.

cameel:
Oh, and by the way, please use dashes (`-`), not underscores (`_`) when creating new branches as stated in our git style.

dybi:
Decalined, is being done in parts in separate PRs

PR #175, opened 2018-03-12
Integration sentry with django

### LINKED ISSUE #167: Allow using Sentry for error reports
Configure the application to be able to send error reports to https://sentry.io

Currently we're using mail reports and we want the same information to be sent to Sentry. It should be possible to use either one or even both. The default config should use none and we want to be able to enable it via `local_settings.py`.

See [Sentry docs > SDKs > Python > Integrations > Django](https://sentry.io)

- Add the plugin to requirements and `INSTALLED_APPS`.
- Use [Integration with `logging`](https://docs.sentry.io/clients/python/integrations/django/#integration-with-logging)
    - Add a `sentry` handler.
    - Add the handler to all loggers that use `mail_admins`.
- Enable `Sentry404CatchMiddleware` middleware too.
- I don't think we need WSGI Middleware

bartoszbetka:
Resolves #167

### DIFF
--- a/concent_api/concent_api/settings/base.py
+++ b/concent_api/concent_api/settings/base.py
@@ -6,12 +6,16 @@ BASE_DIR = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__fil
 # Application definition
 
 INSTALLED_APPS = [
+    # Built-in apps:
     'django.contrib.admin',
     'django.contrib.auth',
     'django.contrib.contenttypes',
     'django.contrib.sessions',
     'django.contrib.messages',
     'django.contrib.staticfiles',
+
+    # Third-party apps:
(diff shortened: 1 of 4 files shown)


concent_api/concent_api/settings/base.py
```diff
@@ -17,6 +17,9 @@
     'concent_api',
     'core',
     'gatekeeper',
+
+    # Add raven plugin to support sentry.io
+    'raven.contrib.django.raven_compat',
```
cameel: Put this above our apps. Let's keep this order here: django apps, external apps, our apps.

bartoszbetka: Done.

cameel:
Has it been tested in development?

Issue #180, opened 2018-03-16
TypeError when a client key in TaskToCompute is empty


Jakub89:
Error happens for example when we send AckReportComputedTask, but nested TaskToCompute have empty `provider_public_key` field. Then in `store_pending_message()` method we're trying to  create Client using empty key, which ends in crush. `store_pending_message()` takes `client_public_key` form a lot of different sources, so we're going to need some additional validation.

```python
13:05:40 ERROR    | Internal Server Error: /api/v1/send/
Traceback (most recent call last):
  File "/home/jakub/.virtualenvs/test/lib/python3.6/site-packages/django/core/handlers/exception.py", line 41, in inner
    response = get_response(request)
  File "/home/jakub/.virtualenvs/test/lib/python3.6/site-packages/django/core/handlers/base.py", line 187, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/home/jakub/.virtualenvs/test/lib/python3.6/site-packages/django/core/handlers/base.py", line 185, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/lib/python3.6/contextlib.py", line 52, in inner
    return func(*args, **kwds)
  File "/home/jakub/.virtualenvs/test/lib/python3.6/site-packages/django/views/decorators/csrf.py", line 58, in wrapped_view
    return view_func(*args, **kwargs)
  File "/home/jakub/Documents/work/projects/concent/concent_api/utils/api_view.py", line 75, in wrapper
    response_from_view = view(request, message, *args, **kwargs)
  File "/home/jakub/.virtualenvs/test/lib/python3.6/site-packages/django/views/decorators/http.py", line 40, in inner
    return func(request, *args, **kwargs)
  File "/home/jakub/Documents/work/projects/concent/concent_api/core/views.py", line 42, in send
    logging.log_message_received(
  File "/home/jakub/Documents/work/projects/concent/concent_api/core/views.py", line 217, in handle_send_ack_report_computed_task
    subtask             = subtask,
  File "/home/jakub/Documents/work/projects/concent/concent_api/core/views.py", line 858, in store_pending_message
    client          = Client.objects.get_or_create_full_clean(client_public_key)
  File "/home/jakub/Documents/work/projects/concent/concent_api/core/models.py", line 44, in get_or_create_full_clean
    instance = self.get(public_key = base64.b64encode(public_key))
[...]

PR #183, opened 2018-03-19
Fix for TypeError when a client key in TaskToCompute is empty

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/180.

### DIFF
--- a/concent_api/core/tests/test_core_views.py
+++ b/concent_api/core/tests/test_core_views.py
@@ -40,6 +40,7 @@ class CoreViewSendTest(ConcentIntegrationTestCase):
 
     @freeze_time("2017-11-17 10:00:00")
     def setUp(self):
+        super().setUp()
         self.stored_message_counter         = 0
         self.message_timestamp              = get_current_utc_timestamp()  # 1510912800
         self.compute_task_def               = message.ComputeTaskDef()
@@ -47,7 +48,9 @@ class CoreViewSendTest(ConcentIntegrationTestCase):
         self.compute_task_def['subtask_id'] = '8'
         self.compute_task_def['deadline']   = self.message_timestamp
         self.task_to_compute                = message.TaskToCompute(
-            compute_task_def = self.compute_task_def,
+            compute_task_def     = self.compute_task_def,
+            provider_public_key  = self.PROVIDER_PUBLIC_KEY,
+            requestor_public_key = self.REQUESTOR_PUBLIC_KEY,
         )
 
         self.correct_golem_data                                         = message.ForceReportComputedTask()
@@ -64,12 +67,14 @@ class CoreViewSendTest(ConcentIntegrationTestCase):
         )
 
         self.task_to_compute_for_cannot_compute_message = message.TaskToCompute(
-            provider_public_key = PROVIDER_PUBLIC_KEY,
+            provider_public_key  = self.PROVIDER_PUBLIC_KEY,
+            requestor_public_key = self.REQUESTOR_PUBLIC_KEY,
         )
 
         self.cannot_compute_task = message.CannotComputeTask()
         self.cannot_compute_task.task_to_compute = message.TaskToCompute(
-            provider_public_key = PROVIDER_PUBLIC_KEY,
+            provider_public_key  = self.PROVIDER_PUBLIC_KEY,
+            requestor_public_key = self.REQUESTOR_PUBLIC_KEY,
         )
 
         self.cannot_compute_task.task_to_compute.compute_task_def               = message.ComputeTaskDef()
@@ -150,7 +155,9 @@ class CoreViewSendTest(ConcentIntegrationTestCase):
         compute_task_def['subtask_id']  = 'ABC00XYZ'
         compute_task_def['deadline']    = self.message_timestamp
         task_to_compute                 = message.TaskToCompute(
-            compute_task_def = compute_task_def,
+            compute_task_def     = compute_task_def,
+            provider_public_key  = PROVIDER_PUBLIC_KEY,
+            requestor_public_key = REQUESTOR_PUBLIC_KEY,
         )
 
         report_computed_task = message.tasks.ReportComputedTask(
@@ -188,7 +195,9 @@ class CoreViewSendTest(ConcentIntegrationTestCase):
     def test_send_should_return_http_400_if_data_is_incorrect(self):
         compute_task_def    = message.ComputeTaskDef()
         task_to_compute     = message.TaskToCompute(
-            compute_task_def = compute_task_def
+            compute_task_def     = compute_task_def,
+            provider_public_key  = PROVIDER_PUBLIC_KEY,
+            requestor_public_key = REQUESTOR_PUBLIC_KEY,
         )
         report_computed_task = message.tasks.ReportComputedTask(
             task_to_compute = task_to_compute
@@ -293,7 +302,9 @@ class CoreViewSendTest(ConcentIntegrationTestCase):
 
         with freeze_time(datetime.datetime.fromtimestamp(self.message_timestamp - 10000)):
             task_to_compute = message.TaskToCompute(
-                compute_task_def = self.compute_task_def,
+                compute_task_def     = self.compute_task_def,
+                provider_public_key  = PROVIDER_PUBLIC_KEY,
+                requestor_public_key = REQUESTOR_PUBLIC_KEY,
             )
 
         serialized_task_to_compute      = dump(task_to_compute,             REQUESTOR_PRIVATE_KEY,   PROVIDER_PUBLIC_KEY)
@@ -424,7 +435,9 @@ class CoreViewSendTest(ConcentIntegrationTestCase):
             StoredMessage.objects.all().delete()
             compute_task_def['deadline'] = deadline
             task_to_compute = message.TaskToCompute(
-                compute_task_def = compute_task_def,
+                compute_task_def     = compute_task_def,
+                provider_public_key  = PROVIDER_PUBLIC_KEY,
+                requestor_public_key = REQUESTOR_PUBLIC_KEY,
             )
 
             serialized_task_to_compute   = dump(task_to_compute,             REQUESTOR_PRIVATE_KEY,   PROVIDER_PUBLIC_KEY)
@@ -468,7 +481,9 @@ class CoreViewSendTest(ConcentIntegrationTestCase):
             compute_task_def['subtask_id']  = str(i)
             compute_task_def['deadline']    = deadline
             task_to_compute                 = message.TaskToCompute(
-                compute_task_def = compute_task_def,
+                compute_task_def     = compute_task_def,
+                provider_public_key  = PROVIDER_PUBLIC_KEY,
+                requestor_public_key = REQUESTOR_PUBLIC_KEY,
             )
 
             serialized_task_to_compute   = dump(task_to_compute,             REQUESTOR_PRIVATE_KEY,   PROVIDER_PUBLIC_KEY)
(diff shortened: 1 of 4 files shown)


concent_api/core/views.py
```diff
@@ -619,6 +620,9 @@ def validate_golem_message_task_to_compute(data):
 
     validate_id_value(data.compute_task_def['task_id'], 'task_id')
 
+    validate_public_key(data.provider_public_key,  'provider_public_key')
+    validate_public_key(data.requestor_public_key, 'requestor_public_key')
```
cameel: @rwrzesien Unrelated to this pull request but this should be refactored later: I think `data` is a left-over from the early JSON implementation. We should rename it to `message` and add a type annotation.

concent_api/core/views.py
```diff
@@ -661,6 +665,16 @@ def validate_id_value(value, field_name):
...
 
 
+def validate_public_key(value, field_name):
+    assert isinstance(field_name, str)
+
+    if not isinstance(value, bytes):
+        raise Http400("{} must be string.".format(field_name))
+
+    if len(value) != GOLEM_PUBLIC_KEY_LENGTH:
+        raise Http400("{} length cannot be different than {} chars.".format(field_name, GOLEM_PUBLIC_KEY_LENGTH))
```
cameel: Maybe `"The length of {} must be exactly {} characters."`?

concent_api/core/views.py

concent_api/core/views.py

concent_api/core/tests/test_core_views.py
```diff
@@ -496,6 +510,98 @@ def test_send_should_return_http_202_if_task_to_compute_deadline_is_correct(self
 
             self.assertIn(response_202.status_code, [200, 202])
 
+    @freeze_time("2017-11-17 10:00:00")
+    def test_send_task_to_compute_without_public_key_should_return_http_400(self):
+        assert StoredMessage.objects.count() == 0
+
+        compute_task_def = message.ComputeTaskDef()
```
dybi: @rwrzesien, why wouldn't you use `self.task_to_compute`? This boilerplate code seems unnecessary ;)

concent_api/core/tests/test_core_views.py
```diff
@@ -496,6 +510,98 @@ def test_send_should_return_http_202_if_task_to_compute_deadline_is_correct(self
...
+            'provider_public_key',
+            'requestor_public_key',
+        ]:
+            setattr(task_to_compute, field_name, None)
+
+            report_computed_task = message.tasks.ReportComputedTask(
+                task_to_compute = task_to_compute
+            )
+
+            correct_golem_data = message.ForceReportComputedTask(
```
dybi: please change the name as it is misleading. Perhaps `message`? Or better - just inline the variable

concent_api/core/tests/test_core_views.py
```diff
@@ -496,6 +510,98 @@ def test_send_should_return_http_202_if_task_to_compute_deadline_is_correct(self
...
+        compute_task_def['subtask_id']  = 'ABC00XYZ'
+        compute_task_def['deadline']    = self.message_timestamp
+
+        task_to_compute                 = message.TaskToCompute(
+            compute_task_def     = compute_task_def,
+            provider_public_key  = PROVIDER_PUBLIC_KEY,
+            requestor_public_key = REQUESTOR_PUBLIC_KEY,
+        )
+
+        for field_name in [
```
dybi: I would prefer to use:
```
    @freeze_time("2017-11-17 10:00:00")
    @pytest.mark.parametrize('attribute', [
        'provider_public_key',
        'requestor_public_key',
    ])
    def test_send_task_to_compute_without_public_key_should_return_http_400(self, attribute):
        assert StoredMessage.objects.count() == 0
        setattr(self.task_to_compute, attribute, None)
        ...
```
instead of `for` loops in the tests ;)

concent_api/core/tests/test_core_views.py
```diff
@@ -496,6 +510,98 @@ def test_send_should_return_http_202_if_task_to_compute_deadline_is_correct(self
...
+                HTTP_CONCENT_CLIENT_PUBLIC_KEY = b64encode(PROVIDER_PUBLIC_KEY).decode('ascii'),
+                HTTP_CONCENT_OTHER_PARTY_PUBLIC_KEY=b64encode(REQUESTOR_PUBLIC_KEY).decode('ascii'),
+            )
+
+            self.assertEqual(response.status_code, 400)
+            self._assert_stored_message_counter_not_increased()
+            self._assert_client_count_is_equal(0)
+
+    @freeze_time("2017-11-17 10:00:00")
+    def test_send_task_to_compute_with_public_key_with_wrong_length_should_return_http_400(self):
```
dybi: The same as above:
1) please use `self.task_to_compute` to get rid of biolerplate
2) please remove `for` loop and use `pytest` instead
3) inline unnecesary variable

concent_api/core/tests/test_core_views.py
```diff
@@ -496,6 +510,98 @@ def test_send_should_return_http_202_if_task_to_compute_deadline_is_correct(self
...
+                data = dump(
+                    correct_golem_data,
+                    PROVIDER_PRIVATE_KEY,
+                    CONCENT_PUBLIC_KEY),
+                content_type = 'application/octet-stream',
+                HTTP_CONCENT_CLIENT_PUBLIC_KEY = b64encode(PROVIDER_PUBLIC_KEY).decode('ascii'),
+                HTTP_CONCENT_OTHER_PARTY_PUBLIC_KEY=b64encode(REQUESTOR_PUBLIC_KEY).decode('ascii'),
+            )
+
+            self.assertEqual(response.status_code, 400)
```
dybi: This code seems the same as in the test above - please extract it to a well name method.

concent_api/core/tests/test_core_views.py
```diff
@@ -496,6 +510,98 @@ def test_send_should_return_http_202_if_task_to_compute_deadline_is_correct(self
 
             self.assertIn(response_202.status_code, [200, 202])
 
+    @freeze_time("2017-11-17 10:00:00")
+    def test_send_task_to_compute_without_public_key_should_return_http_400(self):
```
dybi: There is some code duplication with the function below; with some small refactoring you can get sth like this:
```
@freeze_time("2017-11-17 10:00:00")
    @pytest.mark.parametrize('attribute', [
        'provider_public_key',
        'requestor_public_key',
    ])
    def test_send_task_to_compute_without_public_key_should_return_http_400(self, attribute):
        assert StoredMessage.objects.count() == 0
        setattr(self.task_to_compute, attribute, None)

        response = self.send_force_report_computed_task(self.task_to_compute)

        self.assert_bad_request(response)
```

concent_api/core/tests/test_core_views.py

concent_api/core/tests/test_core_views.py

concent_api/core/tests/test_core_views.py
rwrzesien: Right, but we don't use pytest for now, if it should be added lets create separate issue for adding it and changing all tests like this at once, because there are more of them.

concent_api/core/tests/test_core_views.py
rwrzesien: 1,3. done
2. answered above.

concent_api/core/tests/test_core_views.py
rwrzesien: Not exactly, first test is setting attribute to None, while second is cutting its length.

concent_api/core/tests/test_core_views.py
rwrzesien: I have changed 
`self.assertEqual(response.status_code, 400)`
into
`self._test_400_response(response)`
but this level of granulation and flexibility is currently what we want in tests.

concent_api/core/tests/test_core_views.py
dybi: second test could be like this:
```
@freeze_time("2017-11-17 10:00:00")
@pytest.mark.parametrize('attribute', [
        'provider_public_key',
        'requestor_public_key',
])
def test_send_task_to_compute_with_public_key_with_wrong_length_should_return_http_400(self, attribute):
        assert StoredMessage.objects.count() == 0
        setattr(
                self.task_to_compute,
                attribute,
                getattr(self.task_to_compute, attribute)[:-1]
            )
        response = self.send_force_report_computed_task(self.task_to_compute)

        self.assert_bad_request(response)
```

concent_api/core/tests/test_core_views.py
dybi: Why not changing them gradually. Right now, changing these two tests is like ~0 effort. Having a separate task for changing everything at once might:
1) be time consuming
2) has low priority
3) introduce some bugs (the smaller the change, the easier it is to control ;)

rwrzesien: This is also answer for https://github.com/golemfactory/concent/pull/183#discussion_r175580697. I am referring to this part of code:

```
response = self.send_force_report_computed_task(self.task_to_compute)
self.assert_bad_request(response)
```

> introduce some bugs (the smaller the change, the easier it is to control ;)

Doing it partially drives to a different issue - lack of consistency. Currently we agreed on some level of granularity in tests, and we try to keep it everywhere, so it is easier for everyone involved to quickly read a single test or modify a bunch of them, because they are built from similar steps. When we want to change something like this, we should change it globally. There is still some mess because not all test files were refactored to use our custom test methods - and `test_core_views.py` is an example of that. So currently we have already two levels of abstractions in tests - lets not add third to it. Besides, our current custom test methods have the right level of flexibility, and adding/removing them is quite easy. I don't think we want a huge test functions that have a lot of parameters and does 90% of test job, because that can lead to errors too, and in quickly changing environment/requirements it is easier to add those small functions instead of modifying big ones in a way they still cover and pass for all cases.

> be time consuming
I think it takes more time if you sum time spent on modifying tests one by one from time to time instead of changing them all at once.

> has low priority
That is actually our internal case, so I would not treat this as a problem.

One more thing about `test_core_views.py` itself. I think there were first ever tests in this project, and there were two tasks for them:
- Test if api and api views wrapper is working correctly.
- Test partially request and responses from first use case (report computed task).
After some time, I see them as a good candidate to get rid of. All cases from first use case are covered by first use case tests `test_integration_report_computed_task` and `test_auth_report_computed_task`. Some others test validation, some other other different things. I found them difficult to maintain and they need some general cleaning, which might be done as a separate issue.

About pytest, feel free for adding issue to add pytest and modify existing tests with it.

cameel: I talked about this with @dybi. For the record, here's my opinion:
- This improvement seems contentious so let's do it in a separate pull request like @rwrzesien suggests not to stall fixing of the original issue.
- This requires adding a new dependency (`pytest`) so in my opinion this does not count as a simple refactoring. But I am not against small refactoring in general as long as it is not mixed into the same commits with more relevant changes.
- I talked with @dybi and I think that his concern with putting off refactoring is valid but I also share @rwrzesien's concerns about this leading to inconsistent patchwork of various approaches in the code. We need a compromise. I think that our move towards smaller and quicker pull requests might be the solution here. Being able to quickly refactor something small in the whole module and have it merged in a reasonable time sounds like a good middle ground to me.
- @dybi seems averse to the overhead created by adding new issues but I'm with @rwrzesien on this one. Yeah, it does add overhead but this is the cost of being able to quickly drop or put off part of work if it's less important and takes much more time than expected. Not everything can be done perfectly in the time we have so sometimes some specific piece of refactoring suffers but it's fine as long as a reasonable amount of refactoring is getting done as we go along.
- We need more smaller, non-integration tests. @dybi calls them unit tests, I'd say functional tests because to me a pure unit test means something that's mocked-out to the point of being useless but anyway, we mean the same thing. Every public function should have a set of small tests that check just the code inside it and the private code it calls, in isolation from other, unrelated parts of the system. Integration tests should come after that and not necessarily deal with all the tiny corner cases, because as we all see they're pretty costly to maintain and add significant overhead. I think that the decision to focus on them was a good one given the circumstances but now we're quickly getting to the point were we add integration tests for things that could be more efficiently tested with unit/functional tests. `test_core_views.py` was meant to contain such tests for `core.views`, though, like other parts of the code, it went its own way. At this point it's probably better to throw this file away and start from scratch but we do need this type of tests in general.

Issue #176, opened 2018-03-15
Missing rollbacks on HTTP 4xx and 5xx errors


cameel:
We're writing code with the assumption that all our views are wrapped in database transactions and that in case of an exception the transaction is rolled back. `ATOMIC_REQUESTS` in database settings ensures that the transactions are there but I have noticed that rollbacks are currently broken. That's because we catch exceptions ourselves in `api_view` instead of letting Django handle them.

Ensure that a HTTP 4xx or HTTP 5xx error returned to the client always involves a rollback of all the (incomplete) database changes. If necessary, roll back the transaction manually.

Review also views that do not use `api_view` such as the ones in gatekeeper.

Add a simple test that defines a view based on `api_view`, raises an exception in it and checks that changes have been rolled back. Note that this test case must be based on [TransactionTestCase](https://docs.djangoproject.com/en/2.0/topics/testing/tools/#django.test.TransactionTestCase) because the ordinary `TestCase` uses transactions as a way to quickly clear the database between test runs.

rwrzesien:
This is how it works currently and after my changes:
- For gatekeeper, there is no concern because it is not using database at the moment. If it start, we will have to approach to it individually depending on how it will be used.
- For 5xx errors there is no problem in rollbacks as long as we have `ATOMIC_REQUESTS` set to True.
- For 4xx errors there is no problem when we don't handle errors in our own way.
- For 4xx errors which we handle (like in `@api_view`.) there was an issue, which is currently fixed by using transaction savepoint before starting a view and rolling back to it when handing 4xx error.

I have added tests if changes are rolled back for each case:
- 400 in view with `@api_view`
- 500 in view with `@api_view`
- 400 in view without `@api_view`
- 500 in view without `@api_view`

PR #187, opened 2018-03-20
Fix missing rollbacks on http 4xx and 5xx errors

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/176.

Needs https://github.com/golemfactory/concent/pull/186 merged to work correctly.
Please check my comment in issue.

### DIFF
--- a/concent_api/utils/tests/test_api_view.py
+++ b/concent_api/utils/tests/test_api_view.py
@@ -1,18 +1,28 @@
-import json
 from base64                         import b64encode
+import json
+import mock
 
 from django.conf                    import settings
-from django.test                    import TestCase, RequestFactory, override_settings
+from django.http.response           import HttpResponse
+from django.shortcuts               import reverse
+from django.test                    import override_settings
+from django.test                    import RequestFactory
+from django.test                    import TestCase
+from django.test                    import TransactionTestCase
 from django.views.decorators.http   import require_POST
-
-from golem_messages.message         import WantToComputeTask
+from golem_messages                 import message
 from golem_messages                 import dump, load
 
+from core.models                    import Client
 from utils.api_view                 import api_view
+from utils.api_view                 import Http400
+from utils.helpers                  import get_current_utc_timestamp
 from utils.testing_helpers          import generate_ecc_key_pair
 
 
-(CONCENT_PRIVATE_KEY, CONCENT_PUBLIC_KEY) = generate_ecc_key_pair()
+(CONCENT_PRIVATE_KEY,   CONCENT_PUBLIC_KEY)   = generate_ecc_key_pair()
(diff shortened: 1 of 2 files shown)


concent_api/utils/tests/test_api_view.py
```diff
@@ -115,6 +123,120 @@ def dummy_view(request, message):                                       # pylint
...
+                self.client.post(
+                    reverse('core:send'),
+                    data                                = '',
+                    content_type                        = 'application/octet-stream',
+                    HTTP_CONCENT_CLIENT_PUBLIC_KEY      = b64encode(settings.CONCENT_PUBLIC_KEY).decode('ascii'),
+                )
+            except TypeError:
+                pass
+
+        self.assertEqual(Client.objects.count(), 0)
```
cameel: I think we need tests for the HTTP 200 cases when there is no rollback too.

Using `patch()` is tricky because you have to specify the module where the function is being used, not where it's defined. It's easy to make a mistake and break it silently. If this happens, we won't notice because the code still won't be creating any objects and the tests will pass. We will detect this only if we also have a test for HTTP 200 that uses the same mock - this one will break if the object is not created when it should.

rwrzesien: :+1:

concent_api/utils/tests/test_api_view.py
```diff
@@ -1,17 +1,28 @@
...
+from utils.testing_helpers          import generate_ecc_key_pair
 
 
-(CONCENT_PRIVATE_KEY, CONCENT_PUBLIC_KEY) = generate_ecc_key_pair()
-
+(CONCENT_PRIVATE_KEY,   CONCENT_PUBLIC_KEY)   = generate_ecc_key_pair()
+(PROVIDER_PRIVATE_KEY,  PROVIDER_PUBLIC_KEY)  = generate_ecc_key_pair()
+(REQUESTOR_PRIVATE_KEY, REQUESTOR_PUBLIC_KEY) = generate_ecc_key_pair()
 
 @override_settings(
```
Jakub89: "./concent_api/utils/tests/test_api_view.py:27:1: E302 expected 2 blank lines, found 1"

concent_api/utils/tests/test_api_view.py
```diff
@@ -49,9 +60,9 @@ def test_api_view_should_return_golem_message_as_octet_stream(self):
         decoded_message = None
 
         @api_view
-        def dummy_view(request, message):                                       # pylint: disable=unused-argument
+        def dummy_view(request, _message):                                       # pylint: disable=unused-argument
```
dybi: why `_` added?

cameel: This is pylint's convention for ignoring warnings about unused parameters: https://pylint.readthedocs.io/en/latest/faq.html#i-have-a-callback-function-where-i-have-no-control-over-received-arguments-how-do-i-avoid-getting-unused-argument-warnings

PR #188, opened 2018-03-21
Delete empty RAVEN setting from development settings because it causes an error

bartoszbetka:


concent_api/concent_api/settings/development.py
```diff
@@ -17,8 +17,4 @@
     "admin-panel",
 ]
 
-RAVEN_CONFIG = {
-    'dsn': 'https://',
-}
-
```
cameel: OK, but what about the warnings?

bartoszbetka: There is no warning, the `sentry` only adds information that the `raven` is not configured.

cameel: OK. I thought it was a warning. We really want to avoid having warnings when there is no issue because it teaches people to ignore them.

PR #190, opened 2018-03-21
Remove hardcoded values from payments

### LINKED ISSUE #159: Remove hardcoded values from ForcePaymentCommitted
Currently in `handle_send_force_payment` when `ForcePaymentCommitted` is created, we use hardcoded values:
```
amount_paid             = 10.99,
amount_pending          = 0.01,
```

`amount_paid` should be value returned from call to `base.make_payment_to_provider`. 
`amount_pending` should be (value returned from call to `base.make_payment_to_provider` - sum_of_payments)

Hardcoded values can be returned from mock backend `make_payment_to_provider` function instead of `return True`.

In tests, please:
- mock `make_payment_to_provider` to return different value then hardcoded to check if it is passed correctly
- create a test where sum_of_payments is different than value returned from `make_payment_to_provider` to check if `amount_pending` is calculated correctly.

kbeker:
Resolves #159

### DIFF
--- a/concent_api/core/tests/test_integration_force_payment.py
+++ b/concent_api/core/tests/test_integration_force_payment.py
@@ -1,6 +1,7 @@
 from unittest               import skip
-import mock
+from decimal                import Decimal
 
+import mock
 from django.test            import override_settings
 from django.urls            import reverse
 from freezegun              import freeze_time
@@ -29,8 +30,16 @@ def _get_payment_summary_negative(request, subtask_results_accepted_list, list_o
     return -1
 
 
-def _get_payment_summary_positive(request, subtask_results_accepted_list, list_of_transactions, list_of_forced_payments):  # pylint: disable=unused-argument
-    return 1
+def _get_payment_summary_value_100(request, subtask_results_accepted_list, list_of_transactions, list_of_forced_payments):  # pylint: disable=unused-argument
+    return Decimal('100')
+
+
+def _get_payment_summary_value_90(request, subtask_results_accepted_list, list_of_transactions, list_of_forced_payments):  # pylint: disable=unused-argument
+    return Decimal('90')
+
+
+def _make_payment_to_provider(_sum_of_payments, _payment_ts, _requestor_ethereum_public_key, _client_public_key):
+    return Decimal('90')
 
 
 @override_settings(
@@ -359,14 +368,15 @@ class ForcePaymentIntegrationTest(ConcentIntegrationTestCase):
 
         with freeze_time("2018-02-05 12:00:20"):
             with mock.patch('core.views.base.get_list_of_transactions', _get_requestor_valid_list_of_transactions):
-                with mock.patch('core.views.base.payment_summary', _get_payment_summary_positive):
-                    response_1 = self.client.post(
-                        reverse('core:send'),
-                        data                                = serialized_force_payment,
-                        content_type                        = 'application/octet-stream',
-                        HTTP_CONCENT_CLIENT_PUBLIC_KEY      = self._get_encoded_provider_public_key(),
-                        HTTP_CONCENT_OTHER_PARTY_PUBLIC_KEY = self._get_encoded_requestor_public_key(),
-                    )
+                with mock.patch('core.views.base.payment_summary', _get_payment_summary_value_100):
+                    with mock.patch('core.views.base.make_payment_to_provider', _make_payment_to_provider):
+                        response_1 = self.client.post(
+                            reverse('core:send'),
+                            data                                = serialized_force_payment,
+                            content_type                        = 'application/octet-stream',
+                            HTTP_CONCENT_CLIENT_PUBLIC_KEY      = self._get_encoded_provider_public_key(),
+                            HTTP_CONCENT_OTHER_PARTY_PUBLIC_KEY = self._get_encoded_requestor_public_key(),
+                        )
         self._test_response(
             response_1,
             status       = 200,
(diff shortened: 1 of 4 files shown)


concent_api/core/tests/test_integration_force_payment.py
```diff
@@ -24,14 +25,18 @@ def _get_requestor_invalid_list_of_transactions(current_time, request):  # pylin
     return [{'timestamp': current_time - 22}, {'timestamp': current_time - 23}, {'timestamp': current_time - 24}]
 
 
-def _get_payment_summary_negative(request, subtask_results_accepted_list, list_of_transactions, list_of_forced_payments):  # pylint: disable=unused-argument
+def _get_payment_summary_negative(request, subtask_results_accepted_list, list_of_transactions, list_of_forced_payments):   # pylint: disable=unused-argument
```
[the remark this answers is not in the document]
cameel: Also, you can just start parameter name with `_` instead of disabling the warning.

concent_api/core/views.py
```diff
@@ -562,22 +562,22 @@ def handle_send_force_payment(request, client_message: message.concents.ForcePay
             reason = message.concents.ForcePaymentRejected.REASON.NoUnsettledTasksFound
         )
     elif sum_of_payments > 0:
-        base.make_payment_to_provider(sum_of_payments, payment_ts, requestor_ethereum_public_key, client_public_key)
+        (amount_paid, amount_pending)   = base.make_payment_to_provider(sum_of_payments, payment_ts, requestor_ethereum_public_key, client_public_key)
```
rwrzesien: I know it was like this before but could you please split arguments into separate lines ? This is getting a little bit too long.

rwrzesien: I think that `make_payment_to_provider` should return only `amount_paid`, and `amount_pending` should be calculated here.

concent_api/core/tests/test_integration_force_payment.py
```diff
@@ -584,3 +591,86 @@ def test_provider_send_force_payment_with_same_subtasks_id_concent_should_refuse
             }
         )
         self._assert_stored_message_counter_not_increased()
+
+    def test_sum_of_payments_return_diffrent_value_then_return_from_mocked_backend(self):
```
rwrzesien: We need also test when `amount_pending` is 0.

concent_api/core/tests/test_integration_force_payment.py
```diff
@@ -584,3 +591,86 @@ def test_provider_send_force_payment_with_same_subtasks_id_concent_should_refuse
...
+                        )
+        self._test_response(
+            response_1,
+            status       = 200,
+            key          = self.PROVIDER_PRIVATE_KEY,
+            message_type = message.concents.ForcePaymentCommitted,
+            fields       = {
+                'recipient_type': message.concents.ForcePaymentCommitted.Actor.Provider,
+                'timestamp':      self._parse_iso_date_to_timestamp("2018-02-05 12:00:20"),
+                'amount_pending': Decimal('10.01'),
```
dybi: you could make variables out of those two values -> `amount_pending`, `amount_paid` and reuse them ;)

concent_api/core/tests/test_integration_force_payment.py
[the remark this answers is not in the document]
kbeker: @rwrzesien Actually there is only 2 spaces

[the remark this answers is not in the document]
kbeker: `_` added to these variables

concent_api/core/views.py
kbeker: Done

concent_api/core/tests/test_integration_force_payment.py
kbeker: Done

concent_api/core/tests/test_integration_force_payment.py
kbeker: Done

concent_api/core/tests/test_integration_force_payment.py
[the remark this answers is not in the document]
rwrzesien: :+1: Ok, I don't know why but the diff looks like there were two and one extra was added, nevermind.

concent_api/core/tests/test_integration_force_payment.py
```diff
@@ -608,6 +616,96 @@ def test_sum_of_payments_return_diffrent_value_then_return_from_mocked_backend(s
...
+            fields       = {
+                'recipient_type': message.concents.ForcePaymentCommitted.Actor.Requestor,
+                'timestamp':      self._parse_iso_date_to_timestamp("2018-02-05 12:00:21"),
+                'amount_pending': amount_pending,
+                'amount_paid':    amount_paid,
+            }
+        )
+        self._assert_stored_message_counter_not_increased()
+
+    def test_sum_of_payments_when_amount_pending_and_amount_paid_is_0(self):
```
rwrzesien: `amount_pending`.

concent_api/core/views.py
rwrzesien: @kbeker What I have meant is that `amount_pending` should be calculated `here`, in view. Sorry, maybe I should write this more precisely. I guess there will be no API method for getting something like `amount_pending`. In this view you have `sum_of_payments` and `amount_paid` so you can calculate the value. This is actually quite important to be done this way so please fix. In tests you can manipulate `base.payment_summary` and `base.make_payment_to_provider` to test different scenarios.

concent_api/core/tests/test_integration_force_payment.py
kbeker: Done

Issue #152, opened 2018-02-26
Soft shutdown for Concent service


rwrzesien:
@cameel Adding this to discuss implementation details.

I would see it this way:
1. Somehow shutdown is triggered (how ? this is the part I would like to ask you).
2. We set the flag on Concent that is it in shutdown mode (django-live-settings might be good for this).
3. Flag is checked in `send`, no new request is accepted, error message is returned (what code could it be ?)
4. In `receive` and `receive-out-of-band`, according to https://github.com/golemfactory/concent/issues/139 we will be checking all pending messages for timeout for current client. In that place, if the flag is raised, we can also check if there are any messages left globally, for all clients. If no, then perform a shutdown (does it mean stop django server ? If yes, how we want to do it ? By stopping it via supervisorctl ?). Also, should we treat timeouted messages, which can be still received by clients, as those which hold server from shutting down ?

The problem with this solution I see is that server instance might be never stopped, if there will be unreceived messages by client, or clients will stop sending requests to server when it is in shutdown mode. We might also introduce some graceful shutdown time as setting, and check it in point 4.

cameel:
I'm quoting your whole description because we'll need to replace it with the actual task description and my answer may not make sense without context:

> @cameel Adding this to discuss implementation details.
>
> I would see it this way:
> 1. Somehow shutdown is triggered (how ? this is the part I would like to ask you).
> 2. We set the flag on Concent that is it in shutdown mode (django-live-settings might be good for this).
> 3. Flag is checked in `send`, no new request is accepted, error message is returned (what code could it be ?)
> 4. In `receive` and `receive-out-of-band`, according to https://github.com/golemfactory/concent/issues/139 we will be checking all pending messages for timeout for current client. In that place, if the flag is raised, we can also check if there are any messages left globally, for all clients. If no, then perform a shutdown (does it mean stop django server ? If yes, how we want to do it ? By stopping it via supervisorctl ?). Also, should we treat timeouted messages, which can be still received by clients, as those which hold server from shutting down ?
>
> The problem with this solution I see is that server instance might be never stopped, if there will be unreceived messages by client, or clients will stop sending requests to server when it is in shutdown mode. We might also introduce some graceful shutdown time as setting, and check it in point 4.

For now I see the whole feature as a very passive thing. Flip a setting and add something that lets the admin see what's left in the system. It's the admin that flips it, monitors remaining tasks and makes a decision whether to pull the plug.

#### The switch
[...]

PR #191, opened 2018-03-21
Feature soft shutdown for concent service

rwrzesien:
Proposed solution for https://github.com/golemfactory/concent/issues/152.

### DIFF
--- a/concent_api/concent_api/settings/base.py
+++ b/concent_api/concent_api/settings/base.py
@@ -16,6 +16,8 @@ INSTALLED_APPS = [
 
     # Third-party apps:
     'raven.contrib.django.raven_compat',
+    'constance',
+    'constance.backends.database',
 
     # Our apps:
     'concent_api',
@@ -200,6 +202,16 @@ LOGGING = {
     },
 }
 
+# Django constance config
+CONSTANCE_BACKEND = 'constance.backends.database.DatabaseBackend'
+CONSTANCE_CONFIG = {
+    # Defines if Concent is in shutdown mode.
+    'SHUTDOWN_MODE': (False, 'Shutdown mode', bool),
+}
+
+CONSTANCE_CONFIG_FIELDSETS = {
+    'Concent Options': ('SHUTDOWN_MODE',),
+}
 
 # Private and public keys to be used by Concent to sign and encrypt its own messages.
 # Stored in a 'bytes' array, e.g. b'\xf3\x97\x19\xcdX\xda...'
--- a/concent_api/concent_api/tests/test_shutdown_mode.py
+++ b/concent_api/concent_api/tests/test_shutdown_mode.py
@@ -1,0 +1,329 @@
+from base64                     import b64encode
+
+from django.conf                import settings
+from django.core                import mail
+from django.http.response       import JsonResponse
+from django.test                import override_settings
+from django.urls                import reverse
+
+from constance                  import config
+from constance.test             import override_config
+from freezegun                  import freeze_time
+from golem_messages             import message
+from golem_messages.shortcuts   import dump
+
+from core.message_handlers      import store_subtask
+from core.models                import PendingResponse
+from core.models                import Subtask
+from core.tests.utils           import ConcentIntegrationTestCase
+from utils.helpers              import get_current_utc_timestamp
+from utils.testing_helpers      import generate_ecc_key_pair
+
+
+(CONCENT_PRIVATE_KEY, CONCENT_PUBLIC_KEY) = generate_ecc_key_pair()
+
+
+@override_config(
+    SHUTDOWN_MODE = True
+)
+@override_settings(
+    CONCENT_PRIVATE_KEY         = CONCENT_PRIVATE_KEY,
+    CONCENT_PUBLIC_KEY          = CONCENT_PUBLIC_KEY,
+    PAYMENT_DUE_TIME            = 10,  # seconds
+    PAYMENT_GRACE_PERIOD        = 10,  # seconds
+    CONCENT_MESSAGING_TIME      = 10,  # seconds
+    FORCE_ACCEPTANCE_TIME       = 10,  # seconds
+    SUBTASK_VERIFICATION_TIME   = 10,  # seconds
+    ADMINS                      = ['admin@localhost']
+)
+class ShutdownModeTest(ConcentIntegrationTestCase):
+
+    def test_in_shutdown_mode_concent_should_not_accept_messages_that_would_cause_transition_to_active_state(self):
+        """
+        Tests if in shutdown mode
+        Concent will not accept any new messages which create or update subtasks in active state.
+        """
+
+        compute_task_def = self._get_deserialized_compute_task_def(
+            task_id     = '1',
+            subtask_id  = '8',
+            deadline    = "2017-12-01 11:00:00"
+        )
+
+        task_to_compute = self._get_deserialized_task_to_compute(
+            timestamp           = "2017-12-01 10:00:00",
+            compute_task_def    = compute_task_def,
+        )
+
+        report_computed_task = self._get_deserialized_report_computed_task(
+            timestamp = "2017-12-01 10:59:00",
+            task_to_compute = task_to_compute,
+        )
+
+        serialized_force_report_computed_task = self._get_serialized_force_report_computed_task(
+            timestamp = "2017-12-01 10:59:00",
+            force_report_computed_task = self._get_deserialized_force_report_computed_task(
+                timestamp               = "2017-12-01 10:59:00",
+                report_computed_task    = report_computed_task
+            ),
+            provider_private_key = self.PROVIDER_PRIVATE_KEY
+        )
+
+        with freeze_time("2017-12-01 10:59:00"):
+            response = self.client.post(
+                reverse('core:send'),
+                data                                = serialized_force_report_computed_task,
+                content_type                        = 'application/octet-stream',
+                HTTP_CONCENT_CLIENT_PUBLIC_KEY      = self._get_encoded_provider_public_key(),
+                HTTP_CONCENT_OTHER_PARTY_PUBLIC_KEY = self._get_encoded_requestor_public_key(),
+            )
+
+        self.assertIsInstance(response, JsonResponse)
+        self.assertEqual(response.status_code, 503)
+        self._assert_stored_message_counter_not_increased()
+
+    def test_in_shutdown_mode_concent_should_accept_messages_that_would_cause_transition_to_passive_state(self):
+        """
+        Tests if in shutdown mode
+        Concent will accept new messages which update subtasks in passive state.
+        It also checks if email to admins is sent when all subtasks are turned into passive states.
+        """
+
+        compute_task_def = self._get_deserialized_compute_task_def(
+            task_id     = '1',
+            subtask_id  = '8',
+            deadline    = "2017-12-01 11:00:00"
+        )
+        task_to_compute = self._get_deserialized_task_to_compute(
+            timestamp           = "2017-12-01 10:00:00",
+            compute_task_def    = compute_task_def,
+        )
+        report_computed_task = self._get_deserialized_report_computed_task(
+            timestamp       = "2017-12-01 10:59:00",
+            task_to_compute = task_to_compute,
+        )
+
+        with freeze_time("2017-12-01 11:00:00"):
+            config.SHUTDOWN_MODE = False
+            store_subtask(
+                task_id                 = '1',
+                subtask_id              = '8',
+                provider_public_key     = self.PROVIDER_PUBLIC_KEY,
+                requestor_public_key    = self.REQUESTOR_PUBLIC_KEY,
+                state                   = Subtask.SubtaskState.FORCING_REPORT,
+                next_deadline           = get_current_utc_timestamp() + settings.CONCENT_MESSAGING_TIME,
+                task_to_compute         = task_to_compute,
+                report_computed_task    = report_computed_task,
+            )
+            config.SHUTDOWN_MODE = True
+
+        self.stored_message_counter = 2
+
+        serialized_ack_report_computed_task = self._get_serialized_ack_report_computed_task(
+            timestamp                = "2017-12-01 11:00:05",
+            ack_report_computed_task = self._get_deserialized_ack_report_computed_task(
+                timestamp       = "2017-12-01 11:00:05",
+                subtask_id      = '8',
+                task_to_compute = task_to_compute
+            ),
+            requestor_private_key    = self.REQUESTOR_PRIVATE_KEY
+        )
+
+        with freeze_time("2017-12-01 11:00:05"):
+            response = self.client.post(
+                reverse('core:send'),
+                data                           = serialized_ack_report_computed_task,
+                content_type                   = 'application/octet-stream',
+                HTTP_CONCENT_CLIENT_PUBLIC_KEY = self._get_encoded_requestor_public_key(),
+            )
+
+        self.assertEqual(response.status_code,  202)
+        self.assertEqual(len(response.content), 0)
+        self._assert_stored_message_counter_increased(increased_by = 1)
+        self._test_subtask_state(
+            task_id                  = '1',
+            subtask_id               = '8',
+            subtask_state            = Subtask.SubtaskState.REPORTED,
+            provider_key             = self._get_encoded_provider_public_key(),
+            requestor_key            = self._get_encoded_requestor_public_key(),
+            expected_nested_messages = {'task_to_compute', 'report_computed_task', 'ack_report_computed_task'},
+        )
+        self._test_last_stored_messages(
+            expected_messages= [
+                message.concents.AckReportComputedTask,
+            ],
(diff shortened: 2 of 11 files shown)


concent_api/gatekeeper/views.py
```diff
@@ -191,3 +201,15 @@ def parse_headers(request: WSGIRequest, path_to_file: str) -> Union[FileTransfer
     else:
         assert len(matching_files) == 0
         return gatekeeper_access_denied_response('Path to specified file is not listed in files variable.', path_to_file, loaded_golem_message.subtask_id, client_public_key)
+
+
+def is_request_from_concent_api(request) -> bool:
+    """
+    Checks if given request is sent from Concent API by checking HTTP_CONCENT_CLIENT_PUBLIC_KEY header.
+    """
+    if (
```
dybi: it could be simplified to
```
return 'HTTP_CONCENT_CLIENT_PUBLIC_KEY' in request.META and
    decode_key(request.META['HTTP_CONCENT_CLIENT_PUBLIC_KEY']) == settings.CONCENT_PUBLIC_KEY
```

concent_api/gatekeeper/views.py
```diff
@@ -35,6 +37,10 @@
 @require_POST
 def upload(request):
     logger.debug("Upload request received.")
+
+    if config.SHUTDOWN_MODE is True and not is_request_from_concent_api(request):
+        return JsonResponse({'error': 'Concent is in shutdown mode.'}, status=503)
```
dybi: Should be extracted to constant, i. e.:
`SHUTDOWN_MODE_ERROR_RESPONSE` or something similar

concent_api/concent_api/tests/test_shutdown_mode.py
```diff
@@ -0,0 +1,329 @@
...
+            timestamp       = "2017-12-01 11:00:05"
+        )
+        self._test_undelivered_pending_responses(
+            subtask_id                         = '8',
+            client_public_key                  = self._get_encoded_provider_public_key(),
+            expected_pending_responses_receive = [
+                PendingResponse.ResponseType.ForceReportComputedTaskResponse,
+            ]
+        )
+        self.assertEqual(len(mail.outbox), len(settings.ADMINS))
```
dybi: Sorry, I am not sure what is being compared here...

concent_api/concent_api/tests/test_shutdown_mode.py
```diff
@@ -0,0 +1,329 @@
...
+
+        self.assertIsInstance(response, JsonResponse)
+        self.assertEqual(response.status_code, 503)
+
+    def test_in_shutdown_mode_gatekeeper_should_not_accept_download_tokens_from_clients(self):
+        """
+        Tests if in shutdown mode Gatekeeper will not accept new download tokens from clients.
+        """
+
+        upload_token                               = message.FileTransferToken()
```
dybi: Did you mean `download_token`? ;)

concent_api/concent_api/tests/test_shutdown_mode.py
```diff
@@ -0,0 +1,329 @@
...
+        upload_token.storage_cluster_address       = 'http://devel.concent.golem.network/'
+        upload_token.authorized_client_public_key  = self.PROVIDER_PUBLIC_KEY
+
+        upload_token.files                 = [message.FileTransferToken.FileInfo()]
+        upload_token.files[0]['path']      = 'blender/benchmark/test_task/scene-Helicopter-27-cycles.blend'
+        upload_token.files[0]['checksum']  = 'sha1:95a0f391c7ad86686ab1366bcd519ba5ab3cce89'
+        upload_token.files[0]['size']      = 1024
+        upload_token.operation             = 'download'
+
+        golem_upload_token  = dump(upload_token, settings.CONCENT_PRIVATE_KEY, settings.CONCENT_PUBLIC_KEY)
```
dybi: `golem_download_token`

concent_api/concent_api/tests/test_shutdown_mode.py
```diff
@@ -0,0 +1,329 @@
...
+
+        self.assertIsInstance(response, JsonResponse)
+        self.assertEqual(response.status_code, 503)
+
+    def test_in_shutdown_mode_gatekeeper_should_accept_download_tokens_from_concent(self):
+        """
+        Tests if in shutdown mode Gatekeeper will accept new download tokens from Concent.
+        """
+
+        upload_token                               = message.FileTransferToken()
```
dybi: `download_token`

concent_api/concent_api/tests/test_shutdown_mode.py
```diff
@@ -0,0 +1,329 @@
...
+        upload_token.storage_cluster_address       = 'http://devel.concent.golem.network/'
+        upload_token.authorized_client_public_key  = self.PROVIDER_PUBLIC_KEY
+
+        upload_token.files                 = [message.FileTransferToken.FileInfo()]
+        upload_token.files[0]['path']      = 'blender/benchmark/test_task/scene-Helicopter-27-cycles.blend'
+        upload_token.files[0]['checksum']  = 'sha1:95a0f391c7ad86686ab1366bcd519ba5ab3cce89'
+        upload_token.files[0]['size']      = 1024
+        upload_token.operation             = 'download'
+
+        golem_upload_token  = dump(upload_token, settings.CONCENT_PRIVATE_KEY, settings.CONCENT_PUBLIC_KEY)
```
dybi: `golem_download_token`

concent_api/concent_api/tests/test_shutdown_mode.py
```diff
@@ -0,0 +1,329 @@
+from base64                 import b64encode
+
+from django.conf                import settings
```
Jakub89: One...more...tab...

rwrzesien: done :+1:

concent_api/concent_api/tests/test_shutdown_mode.py
rwrzesien: If each admin get an email. In issue there is

> An e-mail notification when the last active subtask switches to a passive state would also be useful. But just once so that the admin is not flooded with e-mails when the system stops.

concent_api/concent_api/tests/test_shutdown_mode.py
rwrzesien: Yes, I was trying to do this too quickly. Will be fixed :+1:

concent_api/concent_api/tests/test_shutdown_mode.py
rwrzesien: done :+1:

concent_api/concent_api/tests/test_shutdown_mode.py
rwrzesien: done :+1:

concent_api/concent_api/tests/test_shutdown_mode.py
rwrzesien: done :+1:

concent_api/gatekeeper/views.py
rwrzesien: Off course, but in our coding style we have:

`Eksplicite zwracamy wartość z funkcji za pomocą return, chyba że funkcja z założenia nie ma zwracać wartości (tzn. wszystkie ścieżki kończą się zwróceniem None).`

But I am not 100% sure if this is the case. @cameel can you remind me how this should be ?

concent_api/gatekeeper/views.py
rwrzesien: It could and there is a issue (I think) for doing that globally.

concent_api/gatekeeper/views.py
```diff
@@ -56,6 +62,10 @@ def upload(request):
 @require_safe
 def download(request):
     logger.debug("Download request received.")
+
+    if config.SHUTDOWN_MODE is True and not is_request_from_concent_api(request):
+        return JsonResponse({'error': 'Concent is in shutdown mode.'}, status=503)
```
cameel: Have you checked what nginx-storage will return if you return HTTP 503 here? Keep in mind that this JSON message does not go directly to the client. It goes to nginx and [ngx_http_auth_request](https://nginx.org/en/docs/http/ngx_http_auth_request_module.html) module decides what to return. The docs say that it expects either HTTP 200 or HTTP 401. We need to check what happens if you return HTTP 503. Will it return it or will it replace it with a HTTP 500 saying that it received something unexpected? Ask @bartoszbetka for help in checking that on the cluster.

concent_api/gatekeeper/views.py
rwrzesien: :+1:

concent_api/gatekeeper/views.py
```diff
@@ -56,6 +62,10 @@ def upload(request):
 @require_safe
 def download(request):
     logger.debug("Download request received.")
+
+    if config.SHUTDOWN_MODE is True and not is_request_from_concent_api(request):
```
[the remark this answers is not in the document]
rwrzesien: I think that adding this extra check won't hurt.

concent_api/core/models.py
```diff
@@ -275,6 +277,10 @@ def from_db(cls, db, field_names, values):
     def clean(self):
         super().clean()
 
+        # Concent should not accept anything that cause a transition to an active state in shutdown mode.
+        if config.SHUTDOWN_MODE is True and self.state_enum in self.ACTIVE_STATES:
+            raise ConcentInShutdownMode
```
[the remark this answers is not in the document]
rwrzesien: Yes, I agree this is a good example.

concent_api/core/exceptions.py
```diff
@@ -1,3 +1,7 @@
 
 class UnexpectedResponse(Exception):
     pass
+
+
+class ConcentInShutdownMode(Exception):
+    pass
```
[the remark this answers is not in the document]
rwrzesien: Why ? Currently we handle each of them separately. 

What would be nice, is if we add custom error classes for everything, and if they include error code inside them. This way we will have exceptions and error codes in single place.

concent_api/core/admin.py
```diff
@@ -1,6 +1,119 @@
...
+    get_requestor_public_key.short_description = 'Requestor public key'  # type: ignore
+
+
+class PendingResponseAdmin(ReadOnlyMixin, admin.ModelAdmin):
+
+    list_display = [
+        'response_type',
+        'queue',
+        'get_subtask_subtask_id',
+        'get_client_public_key',
```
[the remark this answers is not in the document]
rwrzesien: It is not working like this. It will not create link automatically. When you define list items like this, django admin evaluates those models `__str__` method, which represents the model. Currently we don't have those, and even if we want to have I don't think `Subtask` model should be represented only by `subtask_id`. I have used those because I wanted to show explicitly those values there.

concent_api/core/admin.py
```diff
@@ -1,6 +1,119 @@
...
+
+
+class SubtaskAdmin(ReadOnlyMixin, admin.ModelAdmin):
+
+    list_display = [
+        'subtask_id',
+        'task_id',
+        'state',
+        'get_provider_public_key',
+        'get_requestor_public_key',
```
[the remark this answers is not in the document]
rwrzesien: Please check my answer above.

concent_api/core/admin.py
```diff
@@ -1,6 +1,119 @@
 from django.contrib import admin
 
+from .models import PendingResponse
 from .models import StoredMessage
+from .models import Subtask
 
 
+class ReadOnlyMixin(object):
+    """ Disables all editing capabilities. """
```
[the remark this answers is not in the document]
rwrzesien: :+1: Actually it is only for admin, but I agree, if our admin will grow we should split it a little bit.

concent_api/concent_api/settings/base.py
```diff
@@ -200,6 +202,16 @@
     },
 }
 
+# Django constance config
+CONSTANCE_BACKEND = 'constance.backends.database.DatabaseBackend'
+CONSTANCE_CONFIG = {
+    # Defines if Concent is in shutdown mode.
+    'SHUTDOWN_MODE': (False, 'Shutdown mode', bool),
+}
```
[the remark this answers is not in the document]
rwrzesien: It only does initial migration, there are no changes when you change this. I think it holds this data inside database fields.

concent_api/concent_api/tests/test_shutdown_mode.py
```diff
@@ -0,0 +1,329 @@
...
+            ),
+            HTTP_AUTHORIZATION              = 'Golem ' + encrypted_token,
+            content_type                    = 'application/x-www-form-urlencoded',
+            HTTP_CONCENT_CLIENT_PUBLIC_KEY  = self._get_encoded_key(settings.CONCENT_PUBLIC_KEY)
+        )
+
+        self.assertIsInstance(response, JsonResponse)
+        self.assertNotEqual(response.status_code, 503)
+
+    def test_in_shutdown_mode_gatekeeper_should_not_accept_upload_tokens_from_concet(self):
```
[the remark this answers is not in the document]
rwrzesien: :+1:

concent_api/concent_api/tests/test_shutdown_mode.py
```diff
@@ -0,0 +1,329 @@
...
+        upload_token.authorized_client_public_key  = self.PROVIDER_PUBLIC_KEY
+
+        upload_token.files                 = [message.FileTransferToken.FileInfo()]
+        upload_token.files[0]['path']      = 'blender/benchmark/test_task/scene-Helicopter-27-cycles.blend'
+        upload_token.files[0]['checksum']  = 'sha1:95a0f391c7ad86686ab1366bcd519ba5ab3cce89'
+        upload_token.files[0]['size']      = 1024
+        upload_token.operation             = 'upload'
+
+        golem_upload_token  = dump(upload_token, settings.CONCENT_PRIVATE_KEY, settings.CONCENT_PUBLIC_KEY)
+        encrypted_token     = b64encode(golem_upload_token).decode()
```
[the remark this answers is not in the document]
rwrzesien: :+1: Yes, exactly, I have taken it from different test. I have added an issue for this https://github.com/golemfactory/concent/issues/196.

concent_api/gatekeeper/views.py
rwrzesien: No, I didn't, but it would be good to check.

@bartoszbetka can you help with that ?

concent_api/concent_api/settings/base.py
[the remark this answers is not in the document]
cameel: But is the data preserved after a complex migration? It's not necessarily a problem with this single setting but it we add this plugin we're going to want to use for more and more stuff. I'd like to be sure that it can migrate settings. If not, I'd prefer to use a different one or just create the tables manually.

concent_api/core/admin.py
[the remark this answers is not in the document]
cameel: We definitely should have `__str__()` methods and they should print more than just `subtask_id`. It worked quite nicely in previous projects that way in the admin. But OK, we probably don't want it to include the key. Let's leave it as is in that case.

concent_api/core/exceptions.py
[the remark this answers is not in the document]
cameel: Yeah, combining error codes and exceptions is a good idea.

As to the base class - I was just asking. I'm not saying we have to have one. But I think that it might be useful to be able to selectively catch (or not catch) all the non-built-in stuff. If you think we don't need one, let's leave it as is.

cameel:
Two issues remain:
- Checking how nginx-storage will react to HTTP 503
- Checking how the settings plugin handles migrations

concent_api/core/exceptions.py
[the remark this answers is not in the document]
rwrzesien: We might need it, but not yet. Currently our exception handling is doing ok, but in the future I see possibility that we will need it.

concent_api/concent_api/settings/base.py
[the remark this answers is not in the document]
rwrzesien: The migration is inside the `django-constance` app. Concent will not create any additional migration unless you edit the original code of extra app. Which will not happen, because it is installed from requirements. I don't really see what could cause any problems here.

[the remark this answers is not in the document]
cameel: We talked about it today. @rwrzesien's opinion is that the live settings are not meant to be preserved long-term so resetting them to defaults on migration is not a problem.

I'm not sure. I have no opinion one way or another yet. We'll leave it as is and start worrying about it when we have settings that we need to migrate.

concent_api/gatekeeper/views.py
cameel: @bartoszbetka says that nginx returns HTTP 500. So this needs to be changed.

cameel: This can't easily be done in nginx config. We'd have to stop using the module and rewrite this logic in lua which is not what we really want to do.

Return HTTP 401.

rwrzesien: done :+1:

Issue #179, opened 2018-03-16
Use correct file names in 'force get task result' use case


cameel:
Currently `FileTransferToken` generated by Concent always contains a single path and the name is hard-coded as `{task_id}/0/result`.

This needs to be changed:
- The file name used in the token should be `{task_id}.{subtask_id}.zip`.
- Result files should all be placed in `blender/result/{task_id}/` subdirectories.
    - Note that `{task_id}` is both in the directory name and in the file name. That's to prevent having all files in a single directory which often degrades filesystem performance. This should be enough in the short term and we'll address this in a more general way in a separate issue for the long term.

You can keep assuming that there's always a single result file because the actual result is inside the zip file. There can be more results (if multiple frames are rendered) but they're all supposed to be inside (frame names are defined by [get_frame_name()](https://github.com/golemfactory/golem/blob/develop/apps/rendering/task/framerenderingtask.py#L427-L435)).

### Token example
``` json
{
    "subtask_id":                   "zzz523",
    "token_expiration_deadline":    1512122400,  # "2017-12-01 11:00"
    "cluster_address":              "https://storage.concent.golem.network/",
    "authorized_client_public_key": 123456,
    "operation":                    "upload",
    "files":                        [
        {
            "path":     "blender/result/ef0dc1/ef0dc1.zzz523.zip",
            "checksum": "sha1:95a0f391c7ad86686ab1366bcd519ba5ab3cce89",
            "size":     1024
        }
    ]
}
```

The path where the client should upload the file after receiving such a token would be:
```
https://storage.concent.golem.network/upload/blender/result/ef0dc1/ef0dc1.zzz523.zip
```

PR #192, opened 2018-03-22
Use correct file names in 'force get task result' use case

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/179.

### DIFF
--- a/concent_api/api-integration-force-get-task-result-test.py
+++ b/concent_api/api-integration-force-get-task-result-test.py
@@ -37,12 +37,12 @@ def parse_command_line(command_line):
     return cluster_url
 
 
-def upload_new_file_on_cluster(task_id = '0', part_id = '0', current_time = 0):
+def upload_new_file_on_cluster(task_id, subtask_id):
 
     file_content    = task_id
     file_size       = len(file_content)
     file_check_sum  = 'sha1:' + hashlib.sha1(file_content.encode()).hexdigest()
-    file_path       = '{}/{}/result'.format(task_id, part_id)
+    file_path       = 'blender/result/{}/{}.{}.zip'.format(task_id, task_id, subtask_id)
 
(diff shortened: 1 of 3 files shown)


concent_api/api-integration-force-get-task-result-test.py
```diff
@@ -42,7 +42,7 @@ def upload_new_file_on_cluster(task_id = '0', part_id = '0', current_time = 0):
     file_content    = task_id
     file_size       = len(file_content)
     file_check_sum  = 'sha1:' + hashlib.sha1(file_content.encode()).hexdigest()
-    file_path       = '{}/{}/result'.format(task_id, part_id)
+    file_path       = 'blender/result/{}/{}.{}.zip'.format(task_id, task_id, part_id)
```
cameel: Please rename `part_id` to `subtask_id` because it's supposed to be the subtask ID.

If there's any other place in code that calls it `path_id`, please find it and rename it too.

concent_api/api-integration-force-get-task-result-test.py
```diff
@@ -42,7 +42,7 @@ def upload_new_file_on_cluster(task_id = '0', part_id = '0', current_time = 0):
     file_content    = task_id
```
cameel: I think it would be better to remove defaults `task_id` and `part_id` here. I think that in most cases we'll need to pass a value matching the task anyway. And having a default makes it very easy to forget that you have to set this.

concent_api/api-integration-force-get-task-result-test.py
rwrzesien: done :+1:

concent_api/api-integration-force-get-task-result-test.py
rwrzesien: done :+1:

Issue #165, opened 2018-03-06
[Blueprint] Concent timing


cameel:
### Settings/constants
- CMT: `CONCENT_MESSAGING_TIME`
- MDT: `MAXIMUM_DOWNLOAD_TIME`
- SVT: `SUBTASK_VERIFICATION_TIME`
- FAT: `FORCE_ACCEPTANCE_TIME`
- AVCT: `ADDITIONAL_VERIFICATION_CALL_TIME`
- PDT: `PAYMENT_DUE_TIME`

These values are meant to be adjustable in Concent settings.

#### Relations between time window lengths

- `SVT > 4 CMT + 3 MDT`

#### Relations between message timestamps

Any message that does not satisfy any of these conditions must be rejected as invalid:
- `TaskToCompute.timestamp` <= `ReportComputedTask.timestamp` <= `TaskToCompute.deadline`
- `AckReportComputedTask.timestamp`    <= `TaskToCompute.deadline` + CMT
- `RejectReportComputedTask.timestamp` <= `TaskToCompute.deadline` + CMT

### Communication windows

![concent-timing](https://user-images.githubusercontent.com/137030/37044570-1cf77746-2164-11e8-8cbc-5c8edb98d87e.png)

1. REPORT
    - Submission of `ForceReportComputedTask` by the provider:
        - start: `ReportComputedTask.timestamp`
        - end:   `TaskToCompute.deadline`
    - Submission of `AckReportComputedTask` or `RejectReportComputedTask` by the requestor:
        - start: `ForceReportComputedTask.timestamp`
        - end:   `TaskToCompute.deadline` + CMT

2. RESULT TRANSFER
    - Submission of `ForceGetTaskResult` by the requestor:
        - start: `ReportComputedTask.timestamp`
        - end:   `TaskToCompute.deadline`
    - Upload of task results to the storage cluster by the provider:
        - start: `ForceGetTaskResult.timestamp`
        - end:   `TaskToCompute.deadline` + 3 CMT + 2 MDT
            - (this should be the value used in `FileTransferToken.deadline` returned in `ForceGetTaskResultUpload`)
    - Download of task results from the storage cluster by the requestor:
        - start: `FileTransferToken.timestamp`
        - end:   `TaskToCompute.deadline` + SVT
            - (this should be the value used in `FileTransferToken.deadline` returned in `ForceGetTaskResultDownload`)

3. ACCEPTANCE
    - Submission of `ForceSubtaskResults` by the provider:
[...]

PR #198, opened 2018-03-23
Update time windows force report computed task 

Jakub89:
Part 1/5 of #165

### DIFF
--- a/concent_api/core/tests/test_core_views.py
+++ b/concent_api/core/tests/test_core_views.py
@@ -464,6 +464,7 @@ class CoreViewSendTest(ConcentIntegrationTestCase):
             self.assertEqual(response_400.status_code, 400)
             self.assertIn('error', response_400.json().keys())
 
+    @freeze_time("2017-11-17 10:00:00")
     def test_send_should_return_http_202_if_task_to_compute_deadline_is_correct(self):
         compute_task_def = message.ComputeTaskDef()
 
@@ -480,34 +481,31 @@ class CoreViewSendTest(ConcentIntegrationTestCase):
             compute_task_def['task_id']     = str(i)
             compute_task_def['subtask_id']  = str(i)
             compute_task_def['deadline']    = deadline
-            task_to_compute                 = message.TaskToCompute(
+
+            deserialized_task_to_compute = self._get_deserialized_task_to_compute(
                 compute_task_def     = compute_task_def,
-                provider_public_key  = PROVIDER_PUBLIC_KEY,
-                requestor_public_key = REQUESTOR_PUBLIC_KEY,
+                provider_public_key  = self.PROVIDER_PUBLIC_KEY,
+                requestor_public_key = self.REQUESTOR_PUBLIC_KEY,
             )
 
-            serialized_task_to_compute   = dump(task_to_compute,             REQUESTOR_PRIVATE_KEY,   PROVIDER_PUBLIC_KEY)
-            deserialized_task_to_compute = load(serialized_task_to_compute,  PROVIDER_PRIVATE_KEY,  REQUESTOR_PUBLIC_KEY, check_time = False)
-
-            with freeze_time("2017-11-17 10:00:00"):
-                report_computed_task = message.tasks.ReportComputedTask(
+            force_report_computed_task = self._get_deserialized_force_report_computed_task(
+                report_computed_task = self._get_deserialized_report_computed_task(
+                    subtask_id      = str(i),
                     task_to_compute = deserialized_task_to_compute
                 )
-                force_report_computed_task = message.ForceReportComputedTask(
-                    report_computed_task = report_computed_task
-                )
+            )
 
-                response_202 = self.client.post(
-                    reverse('core:send'),
-                    data = dump(
-                        force_report_computed_task,
-                        PROVIDER_PRIVATE_KEY,
-                        CONCENT_PUBLIC_KEY
-                    ),
-                    content_type                        = 'application/octet-stream',
-                    HTTP_CONCENT_CLIENT_PUBLIC_KEY      = b64encode(PROVIDER_PUBLIC_KEY).decode('ascii'),
-                    HTTP_CONCENT_OTHER_PARTY_PUBLIC_KEY = b64encode(REQUESTOR_PUBLIC_KEY).decode('ascii'),
-                )
+            response_202 = self.client.post(
+                reverse('core:send'),
+                data = dump(
+                    force_report_computed_task,
+                    self.PROVIDER_PRIVATE_KEY,
+                    CONCENT_PUBLIC_KEY
+                ),
+                content_type                        = 'application/octet-stream',
+                HTTP_CONCENT_CLIENT_PUBLIC_KEY      = self._get_encoded_provider_public_key(),
+                HTTP_CONCENT_OTHER_PARTY_PUBLIC_KEY = self._get_encoded_requestor_public_key(),
+            )
 
             self.assertIn(response_202.status_code, [200, 202])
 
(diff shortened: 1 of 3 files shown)


concent_api/core/tests/test_core_views.py
```diff
@@ -464,6 +464,7 @@ def test_send_should_return_http_400_if_task_to_compute_deadline_is_not_an_integ
             self.assertEqual(response_400.status_code, 400)
             self.assertIn('error', response_400.json().keys())
 
+    @freeze_time("2017-11-17 10:00:00")
```
rwrzesien: Why this changed ?

concent_api/core/tests/test_core_views.py
```diff
@@ -558,6 +558,44 @@ def test_send_with_empty_data_should_return_http_400_error(self):
 
         self.assertEqual(response.status_code, 400)
 
+    def test_send_should_return_http_400_if_task_to_compute_younger_then_report_computed(self):
+        compute_task_def = message.ComputeTaskDef()
+
+        StoredMessage.objects.all().delete()
```
rwrzesien: This is not needed - database is rolled back after each test.

concent_api/core/tests/test_core_views.py
```diff
@@ -558,6 +558,44 @@ def test_send_with_empty_data_should_return_http_400_error(self):
...
+        compute_task_def['task_id']     = '1'
+        compute_task_def['subtask_id']  = '3'
+        compute_task_def['deadline']    = get_current_utc_timestamp() + (60 * 37)
+        task_to_compute                 = message.TaskToCompute(
```
rwrzesien: Please use predefined function for creating message, they are already there.

concent_api/core/tests/test_core_views.py
```diff
@@ -489,25 +490,24 @@ def test_send_should_return_http_202_if_task_to_compute_deadline_is_correct(self
...
-                )
+            response_202 = self.client.post(
+                reverse('core:send'),
+                data = dump(
+                    force_report_computed_task,
+                    PROVIDER_PRIVATE_KEY,
+                    CONCENT_PUBLIC_KEY
+                ),
+                content_type                        = 'application/octet-stream',
+                HTTP_CONCENT_CLIENT_PUBLIC_KEY      = b64encode(PROVIDER_PUBLIC_KEY).decode('ascii'),
```
rwrzesien: There are methods for encoding both of those keys.

concent_api/core/tests/test_core_views.py
```diff
@@ -558,6 +558,44 @@ def test_send_with_empty_data_should_return_http_400_error(self):
...
+
+            response = self.client.post(
+                reverse('core:send'),
+                data = dump(
+                    force_report_computed_task,
+                    PROVIDER_PRIVATE_KEY,
+                    CONCENT_PUBLIC_KEY
+                ),
+                content_type                        = 'application/octet-stream',
+                HTTP_CONCENT_CLIENT_PUBLIC_KEY      = b64encode(PROVIDER_PUBLIC_KEY).decode('ascii'),
```
rwrzesien: There are methods for encoding both of those keys.

concent_api/core/views.py
```diff
@@ -617,6 +618,16 @@ def deserialize_message(raw_message_data):
         raise Http400("Unable to deserialize Golem Message: {}.".format(exception))
 
 
+def validate_report_computed_task_time_window(report_computed_task):
+    if not isinstance(report_computed_task, message.ReportComputedTask):
```
rwrzesien: In `handle_send_force_report_computed_task` you are already after passing this condition in `send()`, so I think only `assert` should be here.

concent_api/core/views.py
```diff
@@ -617,6 +618,16 @@ def deserialize_message(raw_message_data):
...
+        raise Http400("Expected ReportComputedTask")
+
+    task_to_compute_younger_then_report_computed = report_computed_task.task_to_compute.timestamp > report_computed_task.timestamp
```
rwrzesien: I don't think creating extra variable is needed here, this calculation is not too complex to name it.

concent_api/core/tests/test_core_views.py
Jakub89: So that all messages are created with this `timestamp` instead of previous current time.

concent_api/core/views.py
```diff
@@ -623,6 +624,13 @@ def deserialize_message(raw_message_data):
         raise Http400("Unable to deserialize Golem Message: {}.".format(exception))
 
 
+def validate_report_computed_task_time_window(report_computed_task):
+    assert isinstance(report_computed_task, message.ReportComputedTask)
+
+    if report_computed_task.task_to_compute.timestamp > report_computed_task.timestamp:
+        raise Http400("ReportComputedTask timestamp is older then nested TaskToCompute.")
```
cameel: Please reverse this condition for readability:
``` python
report_computed_task.timestamp < report_computed_task.task_to_compute.timestamp
```

It's easier to parse mentally when it's the same direction in the condition and in the message. For a moment I even thought that there was an error here and wanted to report it.

concent_api/core/tests/test_core_views.py
```diff
@@ -558,6 +558,37 @@ def test_send_with_empty_data_should_return_http_400_error(self):
 
         self.assertEqual(response.status_code, 400)
 
+    def test_send_should_return_http_400_if_task_to_compute_younger_then_report_computed(self):
```
cameel: `then` -> `than`

Jakub89:
@cameel No removal needed. In this usecase we have:
```
if client_message.report_computed_task.task_to_compute.compute_task_def['deadline'] < current_time:
```
which is correct in my opinion and returns RejectReportComputedTask.

concent_api/core/tests/test_core_views.py
```diff
@@ -480,34 +481,31 @@ def test_send_should_return_http_202_if_task_to_compute_deadline_is_correct(self
...
-                report_computed_task = message.tasks.ReportComputedTask(
-                    task_to_compute = deserialized_task_to_compute
-                )
-                force_report_computed_task = message.ForceReportComputedTask(
-                    report_computed_task = report_computed_task
-                )
+            report_computed_task = message.tasks.ReportComputedTask(
+                task_to_compute = deserialized_task_to_compute
+            )
+            force_report_computed_task = message.ForceReportComputedTask(
```
dybi: you can extract it to a function / method and use throughout tests, to avoid duplication

concent_api/core/tests/test_core_views.py
```diff
@@ -558,6 +556,37 @@ def test_send_with_empty_data_should_return_http_400_error(self):
...
+                    force_report_computed_task,
+                    self.PROVIDER_PRIVATE_KEY,
+                    CONCENT_PUBLIC_KEY
+                ),
+                content_type                        = 'application/octet-stream',
+                HTTP_CONCENT_CLIENT_PUBLIC_KEY      = self._get_encoded_provider_public_key(),
+                HTTP_CONCENT_OTHER_PARTY_PUBLIC_KEY = self._get_encoded_requestor_public_key(),
+            )
+
+        self.assertEqual(response.status_code, 400)
```
dybi: I think I have already seen a method for this. Sth like `self._assert_400`.

concent_api/core/validation.py
```diff
@@ -69,3 +69,10 @@ def validate_golem_message_task_to_compute(golem_message: message.base.Message):
 
     validate_public_key(golem_message.provider_public_key, 'provider_public_key')
     validate_public_key(golem_message.requestor_public_key, 'requestor_public_key')
+
+
+def validate_report_computed_task_time_window(report_computed_task):
```
dybi: please make unit tests for it

concent_api/core/tests/test_core_views.py
Jakub89: Done

concent_api/core/tests/test_core_views.py
Jakub89: `self._test_400_response()` Done

concent_api/core/validation.py
Jakub89: `test_send_should_return_http_400_if_task_to_compute_younger_than_report_computed` in `test_core_views.py` is to check that method. It's not enough?

concent_api/core/tests/test_core_views.py
Jakub89: Done

cameel:
@Jakub89 
> No removal needed. In this usecase we have:
> (...)
> which is correct in my opinion and returns RejectReportComputedTask.

Oh. So you're saying that the end time did not change in this case. OK then.

concent_api/core/validation.py
dybi: Ok, we will talk about that on the next Tech Talk ;)

PR #200, opened 2018-03-23
Update time windows force get task result

Jakub89:
Part 2/5 of #165


concent_api/core/views.py
```diff
@@ -326,6 +327,7 @@ def handle_send_force_get_task_result(request, client_message: message.concents.
     validate_golem_message_task_to_compute(client_message.report_computed_task.task_to_compute)
     validate_id_value(client_message.report_computed_task.task_to_compute.compute_task_def['task_id'],      'task_id')
     validate_id_value(client_message.report_computed_task.task_to_compute.compute_task_def['subtask_id'],   'subtask_id')
+    force_get_task_result_acceptance_time = 2 * settings.CONCENT_MESSAGING_TIME + settings.MAXIMUM_DOWNLOAD_TIME
```
rwrzesien: Please move this after `if` below, just before this is used for the first time.

concent_api/core/views.py
```diff
@@ -1275,13 +1287,21 @@ def create_file_transfer_token(
...
-    current_time    = get_current_utc_timestamp()
     task_id         = report_computed_task.task_to_compute.compute_task_def['task_id']
     subtask_id      = report_computed_task.task_to_compute.compute_task_def['subtask_id']
     file_path       = '{}/{}/result'.format(subtask_id, task_id)
 
+    if operation == 'upload':
+        token_expiration_deadline = (
+            report_computed_task.task_to_compute.compute_task_def['deadline'] + 3 * settings.CONCENT_MESSAGING_TIME + 2 * settings.MAXIMUM_DOWNLOAD_TIME
+        )
+    else:  # It has to be 'download'
```
rwrzesien: Please add `if operation == 'download'` and assert before that if that `operation in ['upload', 'download']

concent_api/core/views.py
```diff
@@ -617,6 +619,16 @@ def deserialize_message(raw_message_data):
         raise Http400("Unable to deserialize Golem Message: {}.".format(exception))
 
 
+def validate_report_computed_task_time_window(report_computed_task):
+    if not isinstance(report_computed_task, message.ReportComputedTask):
+        raise Http400("Expected ReportComputedTask")
+
+    task_to_compute_younger_then_report_computed = report_computed_task.task_to_compute.timestamp > report_computed_task.timestamp
```
dybi: please either rename it to boolean format name (is*, has*,) -> `is_task_to_compute_younger_then_report_computed` or just inline variable

concent_api/core/views.py
dybi: I would do sth like:
```
if operation == 'upload':
    # some code
elif if operation == 'download':
    # code
else:
 raise SomeException()
```

BTW: Is there a unit test for it?

concent_api/core/views.py
```diff
@@ -617,6 +619,16 @@ def deserialize_message(raw_message_data):
         raise Http400("Unable to deserialize Golem Message: {}.".format(exception))
 
 
+def validate_report_computed_task_time_window(report_computed_task):
```
dybi: Is there a unittest for it?

concent_api/core/views.py
cameel: I agree with @rwrzesien. I think that using a [contract](https://en.wikipedia.org/wiki/Design_by_contract) to state which parameter values are supported is better than defensively checking for values that make no sense. By adding an `assert` at the beginning we state that any values other than `'upload'` or `'download'` are simply not supported and trying to call it with such a value is a bug in the calling code. The caller is responsible for ensuring that `operation` is one of these allowed values. It the caller gets it from the outside, he should validate it and report an error if the input is invalid. If he generates the values himself, then validation is not required. Only the caller can discern between these two cases.

concent_api/core/views.py
cameel: Yeah. Would be a good idea to start adding unit tests right now. This pull request is so it's a good place to start.

concent_api/core/views.py
```diff
@@ -325,6 +326,7 @@ def handle_send_force_get_task_result(request, client_message: message.concents.
     validate_golem_message_task_to_compute(client_message.report_computed_task.task_to_compute)
     validate_id_value(client_message.report_computed_task.task_to_compute.compute_task_def['task_id'],      'task_id')
     validate_id_value(client_message.report_computed_task.task_to_compute.compute_task_def['subtask_id'],   'subtask_id')
+    force_get_task_result_acceptance_time = 2 * settings.CONCENT_MESSAGING_TIME + settings.MAXIMUM_DOWNLOAD_TIME
```
cameel: Using `FORCE_ACCEPTANCE_TIME` in this use case was a mistake. Please don't name stuff after it.

I'd replace this with
``` python
force_get_task_result_deadline = (
    client_message.report_computed_task.task_to_compute.compute_task_def['deadline'] +
    2 * settings.CONCENT_MESSAGING_TIME +
    settings.MAXIMUM_DOWNLOAD_TIME
)
```

concent_api/core/views.py
```diff
@@ -1281,13 +1290,21 @@ def create_file_transfer_token(
...
     """
-    current_time    = get_current_utc_timestamp()
     task_id         = report_computed_task.task_to_compute.compute_task_def['task_id']
     subtask_id      = report_computed_task.task_to_compute.compute_task_def['subtask_id']
     file_path       = '{}/{}/result'.format(subtask_id, task_id)
 
+    if operation == 'upload':
+        token_expiration_deadline = (
+            report_computed_task.task_to_compute.compute_task_def['deadline'] + 3 * settings.CONCENT_MESSAGING_TIME + 2 * settings.MAXIMUM_DOWNLOAD_TIME
+        )
```
cameel: For readability, if you're splitting it into multiple lines anyway, I'd suggest this:

``` python
token_expiration_deadline = (
    report_computed_task.task_to_compute.compute_task_def['deadline'] +
    3 * settings.CONCENT_MESSAGING_TIME +
    2 * settings.MAXIMUM_DOWNLOAD_TIME
)
```

When the names are long it's easier to see what the expression is composed of if you put each component on a separate line.

concent_api/core/views.py
Jakub89: Yes there is. `test_send_should_return_http_400_if_task_to_compute_younger_than_report_computed` in `test_core_views.py`

concent_api/core/views.py
Jakub89: It's used in `elif` after that `if` so I can move it only if change that `elif` to `if`.

concent_api/core/transfer_operations.py
```diff
@@ -108,13 +107,25 @@ def create_file_transfer_token(
...
     file_path       = 'blender/result/{}/{}.{}.zip'.format(task_id, task_id, subtask_id)
 
+    assert operation in ['upload', 'download']
+    if operation == 'upload':
+        token_expiration_deadline = (
+            report_computed_task.task_to_compute.compute_task_def['deadline'] +
+            3 * settings.CONCENT_MESSAGING_TIME +
+            2 * settings.MAXIMUM_DOWNLOAD_TIME
+        )
+    if operation == 'download':
```
rwrzesien: Please do `elif` here. It won't change anything in logic but will look more readable, meaning that the code always goes either this way or another.

Jakub89: Done

concent_api/core/transfer_operations.py
```diff
@@ -119,11 +119,13 @@ def create_file_transfer_token(
             3 * settings.CONCENT_MESSAGING_TIME +
             2 * settings.MAXIMUM_DOWNLOAD_TIME
         )
-    if operation == 'download':
+    elif operation == 'download':
         token_expiration_deadline = (
             report_computed_task.task_to_compute.compute_task_def['deadline'] +
             settings.SUBTASK_VERIFICATION_TIME
         )
+    else:
```
rwrzesien: This code will never be reach (there is assert above), I will remove it.

Jakub89: Isn't it an error, at least in pylint. That we have `if` than `elif` without `else`?

rwrzesien: No, at least it doesn't show any warnings when I was running it.

Issue #203, opened 2018-03-23
[Authentication] Add utility functions for authentication


rwrzesien:
We need to add several utility functions which will help in implementing https://github.com/golemfactory/concent/issues/138. This functions might be also reused in different cases.

* Function for loading message without public key.
* Function for validating if passed list of TaskToCompute contains identical instances.
* Function for validating if passed golem message is itself signed with the keys it contains.
* Function for validating ClientAuthorization message.

Add unit tests for each of those.

PR #222, opened 2018-03-27
Authentication: Add utility functions for authentication

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/203.

This part can be reviewed and optionally merged, if it won't be before I finish next issue related with authentication I will use the same PR for it.

### DIFF
--- a/concent_api/core/tests/test_unit_authentication_utilities.py
+++ b/concent_api/core/tests/test_unit_authentication_utilities.py
@@ -1,0 +1,209 @@
+from django.test                    import override_settings
+from django.test                    import TestCase
+
+from golem_messages                 import message
+from golem_messages.shortcuts       import dump
+from golem_messages.shortcuts       import load
+
+from core.exceptions                import Http400
+from core.validation                import validate_golem_message_client_authorization
+from core.validation                import validate_golem_message_signed_with_key
+from core.validation                import validate_list_of_identical_task_to_compute
+from utils.shortcuts                import load_without_public_key
+from utils.testing_helpers          import generate_ecc_key_pair
+
+
+(CONCENT_PRIVATE_KEY, CONCENT_PUBLIC_KEY)       = generate_ecc_key_pair()
+(REQUESTOR_PRIVATE_KEY, REQUESTOR_PUBLIC_KEY)   = generate_ecc_key_pair()
+
+
+@override_settings(
+    CONCENT_PRIVATE_KEY       = CONCENT_PRIVATE_KEY,
+    CONCENT_PUBLIC_KEY        = CONCENT_PUBLIC_KEY,
+)
+class LoadWithoutPublicKeyUnitTest(TestCase):
+
+    def test_load_without_public_key_should_load_message(self):
+        """
+        Test that message loaded with load_without_public_key function will be the same as load
+        with golem_messages load function.
+        """
+
+        # Create and fill some data into ComputeTaskDef
+        compute_task_def = message.ComputeTaskDef()
+        compute_task_def['task_id']     = '1'
+        compute_task_def['subtask_id']  = '2'
+        compute_task_def['deadline']    = 1510912800
+
+        # Create TaskToCompute
+        task_to_compute = message.TaskToCompute(compute_task_def = compute_task_def)
+
+        # Dump TaskToCompute to make it signed
+        dumped_task_to_compute = dump(task_to_compute, REQUESTOR_PRIVATE_KEY, CONCENT_PUBLIC_KEY)
+
+        loaded_task_to_compute_with_utility_function    = load_without_public_key(
+            dumped_task_to_compute,
+        )
+
+        loaded_task_to_compute_with_golem_messages_load = load(
+            dumped_task_to_compute,
+            CONCENT_PRIVATE_KEY,
+            REQUESTOR_PUBLIC_KEY,
+        )
+
+        self.assertEqual(loaded_task_to_compute_with_utility_function, loaded_task_to_compute_with_golem_messages_load)
+
+
+class ValidateGolemMessageClientAuthorizationUnitTest(TestCase):
+
+    def test_validate_golem_message_client_authorization_should_not_raise_error_if_correct_message_is_used(self):
+        client_authorization = message.concents.ClientAuthorization(
+            client_public_key = CONCENT_PUBLIC_KEY
+        )
+
+        try:
+            validate_golem_message_client_authorization(client_authorization)
+        except Http400:
+            self.fail()
+
+    def test_validate_golem_message_client_authorization_should_raise_400_error_when_wrong_message_is_used(self):
+        ping = message.Ping()
+
+        with self.assertRaises(Http400):
+            validate_golem_message_client_authorization(ping)
+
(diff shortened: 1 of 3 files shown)


concent_api/core/tests/test_unit_authentication_utilities.py
```diff
@@ -0,0 +1,214 @@
...
+
+        loaded_task_to_compute_with_golem_messages_load = load(
+            dumped_task_to_compute,
+            CONCENT_PRIVATE_KEY,
+            self.REQUESTOR_PUBLIC_KEY,
+        )
+
+        self.assertEqual(loaded_task_to_compute_with_utility_function, loaded_task_to_compute_with_golem_messages_load)
+
+        for slot in message.TaskToCompute.__slots__:
```
dybi: is it necessary? is any of the slots differs, would the assertion above pass?

concent_api/core/tests/test_unit_authentication_utilities.py
```diff
@@ -0,0 +1,214 @@
...
+                requestor_id = 1
+            ),
+        ]
+
+        with self.assertRaises(Http400):
+            validate_list_of_identical_task_to_compute(
+                list_of_not_identical_task_to_compute
+            )
+
+    def test_validate_list_of_identical_task_to_compute_should_not_return_http400_when_all_task_to_compute_are_identical(self):
```
dybi: "should_not_raise_HTTP400_"

concent_api/core/tests/test_unit_authentication_utilities.py
```diff
@@ -0,0 +1,214 @@
...
+
+        self.assertTrue(
+            validate_list_of_identical_task_to_compute(
+                [
+                    message.TaskToCompute(),
+                ]
+            )
+        )
+
+    def test_validate_list_of_identical_task_to_compute_should_return_http400_when_not_all_task_to_compute_are_identical(self):
```
dybi: ""should_raise_HTTP400_"

concent_api/core/tests/test_unit_authentication_utilities.py
```diff
@@ -0,0 +1,214 @@
...
+from golem_messages.shortcuts       import load
+
+from core.exceptions                import Http400
+from core.validation                import validate_golem_message_client_authorization
+from core.validation                import validate_golem_message_signed_with_key
+from core.validation                import validate_list_of_identical_task_to_compute
+from core.tests.utils               import ConcentIntegrationTestCase
+from utils.shortcuts                import load_without_public_key
+from utils.testing_helpers          import generate_ecc_key_pair
+
```
dybi: Generally, all the tests above are just simple unit tests (which is great!), so it would be better to make them more lightweight by inheriting from `unittest.TestCase` and not `ConcentIntegrationTestCase`

concent_api/core/tests/test_unit_authentication_utilities.py
```diff
@@ -0,0 +1,214 @@
...
+        ping = load(dumped_ping, CONCENT_PRIVATE_KEY, CONCENT_PUBLIC_KEY)
+
+        assert ping.sig is not None
+
+        try:
+            validate_golem_message_signed_with_key(
+                ping,
+                CONCENT_PUBLIC_KEY,
+            )
+        except Exception:  # pylint: disable=broad-except
```
cameel: Plase don't catch `Exception` unless there's absolutely no other way. I agree with pylint here.

An exception will make the test fail anyway so it's fine not to catch it. If you really want it to be a test failure (rather than an error), catch specific exception types that you know can be raised.

Stuff like `SyntaxError`, `TypeError`, etc, should always be let through unless you're writing a top-level catch-all handler that can be made comprehensive enough to sensily handle different kinds of errors.

cameel:
> This part can be reviewed and optionally merged, if it won't be before I finish next issue related with authentication I will use the same PR for it.

I think you should make a separate PR. We want to review and merge every branch separately.

concent_api/core/tests/test_unit_authentication_utilities.py
rwrzesien: The test checks if the function doesn't raise any exception. If it does, the test fails. It is a different thing that when exception is raised by test - it results in error, not in fail, and gives a different output. 

In this case, I want to test if function was executed correctly. It is constructed this way, because function doesn't return any value which can be checked. If you know better construction for checking that case please let me know.

concent_api/core/tests/test_unit_authentication_utilities.py
rwrzesien: I think that extra checking of this won't hurt.

concent_api/core/tests/test_unit_authentication_utilities.py
rwrzesien: :+1:

concent_api/core/tests/test_unit_authentication_utilities.py
rwrzesien: :+1:

concent_api/core/tests/test_unit_authentication_utilities.py
rwrzesien: :+1:

rwrzesien:
@cameel 
> I think you should make a separate PR. We want to review and merge every branch separately.

Ok, then this one is ready to go.

concent_api/core/tests/test_unit_authentication_utilities.py
cameel: I understand that the output is different but I don't think it's different enough to be a problem.

Basically, I think that it's better to catch too little than too much and for me this applies both to tests and to normal code. If we catch too little, there's a very high chance that we find out in testing and update the handler to include that particular exception type. If we catch too much, we might never know about the error. I think it's better to avoid an overly general handler and miss some exceptions because catching stuff like `SyntaxError` is worse.

I think that there's rarely a real need to catch and not immediately re-raise `Exception` or `BaseException`, especially in tests, and any such case should should be reviewed very thoughly because misuse can cause a lot of trouble.

I know that this particular code will fail anyway and we won't really overlook the error but the cause of the problem will be misrepresented and it may still be a minor problem in specific circumstances. For example you might think that after a refactoring a part of code is failing because it's just not ready yet and skip the test while it's actually a simple problem in the test itself. A problem that would be evident if exceptions were not being caught indiscriminately.

There's definitely an exception type you expect this code to raise if there's an error. Add this exception (or exceptions) to the handler and let everything else escape and just be treated as an error. If you find out that there's more that can be raised, update the handler when you notice. That's better than catching everything.

concent_api/core/tests/test_unit_authentication_utilities.py
dybi: I disagree. It distracts you and gets into thinking that somehow these objects cannot be compared normally. It also adds more logic to the test's code which should be avoided (the simpler the tests, the better). If it is not needed, please remove it.

rwrzesien: done :+1: Ok I have checked how golem_messages are compared and there is a check like this there already.

PR #223, opened 2018-03-28
Fix operation type in request upload status

dybi:
Fix in function: operation type in token creation changed 'upload' -> 'download' for HEAD request

### DIFF
--- a/concent_api/core/transfer_operations.py
+++ b/concent_api/core/transfer_operations.py
@@ -28,8 +28,9 @@ def verify_file_status(
     Function to verify existence of a file on cluster storage
     """
 
+    encoded_client_public_key = b64encode(client_public_key)
     force_get_task_result_list = Subtask.objects.filter(
-        requestor__public_key  = b64encode(client_public_key),
+        requestor__public_key  =encoded_client_public_key,
         state                  = Subtask.SubtaskState.FORCING_RESULT_TRANSFER.name,  # pylint: disable=no-member
     )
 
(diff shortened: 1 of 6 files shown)


concent_api/core/transfer_operations.py
```diff
@@ -127,19 +128,16 @@ def create_file_transfer_token(
...
     assert settings.STORAGE_CLUSTER_ADDRESS.endswith(slash)
 
-    current_time = get_current_utc_timestamp()
-    file_transfer_token = message.concents.FileTransferToken(
-        token_expiration_deadline       = current_time + settings.TOKEN_EXPIRATION_TIME,
-        storage_cluster_address         = settings.STORAGE_CLUSTER_ADDRESS,
-        authorized_client_public_key    = settings.CONCENT_PUBLIC_KEY,
-        operation                       = 'upload',
-    )
+    file_transfer_token = create_file_transfer_token(report_computed_task, settings.CONCENT_PUBLIC_KEY, 'download')
```
rwrzesien: @kbeker I remember that we have some discussion about this part in pull request when you have used 'upload' here and I remember you have some explanation for it. Do you maybe remember what was it ?

kbeker: @rwrzesien @dybi Sorry, but I don't remember why I needed there excalty 'upload'. If @dybi tested it locally and on cluster and everything is ok then everything is good. 
@dybi have you tested checking file existance and uploading file to cluster storage?

dybi: Tested on storage and updated all our tests

Issue #141, opened 2018-02-23
Storing contents of Golem messages in dedicated tables


cameel:
Create the following tables for storing the contents of the corresponding Golem messages:
- `StoredTaskToCompute`
- `StoredReportComputedTask`
- `StoredAckReportComputedTask`
- `StoredRejectReportComputedTask`
- `StoredSubtaskResultsAccepted`
- `StoredSubtaskResultsRejected`

Update the subtask table (#140) to replace foreign keys to `StoredMessage` with foreign keys to specific tables.

Create a base model class to keep the common data like timestamp and signature. Decide if it's better to have an abstract or a concrete model as a base.

All data should be stored in a relational manner. Nested messages should be replaced with foreign keys to other tables (create more tables if needed). Nested lists and dictonaries should be separate tables.

Create helpers for converting Golem messages to model instances and vice versa.

kbeker:
I think we should add few more tables to store more messages. For this moment we store in database messages for example such as `ForceGetTaskResultUpload` which contains `FileTransferToken`. Also `RejectReportComputedTask` have 3 nested messages: `TaskToCompute` which you mentioned above, but also `CannotComputeTask` and `TaskFailure`. 

For this moment I found just these four more tables to add:
`StoredForceGetTaskResultUpload`,
`StoredFileTransferToken`,
`StoredCannotComputeTask`,
`StoredTaskFailure`

kbeker:
@cameel 
What about these tables which I mentioned above? I see that we don't have in Subtask table place for messages such as `TaskFailure` or `CannotComputeTask`. Do we want to store in separate table FileTransferToken message?

cameel:
The issue lists only the tables needed to store messages that the `Subtask` table links to. Because those are the only messages Concent is supposed to store.

For other messages, e.g. kept for audit purposes it's optional. I also have nothing against creating a table for every message but we don't have to create them if we're not storing them.

As for nested messages, like `TaskFailure`, of course these should have tables too. That's what I meant by "All data should be stored in a relational manner. Nested messages should be replaced with foreign keys to other tables (create more tables if needed). Nested lists and dictonaries should be separate tables".

As for `ComputeTaskDef`, I said in the description that you should have separate tables for dicts, but honestly it's just fine to flatten the structure. Generally, if you have something like this in a message:

``` python
{
    'a': {
        'c': {
            'f': 1,
        },
    },
    'b': {
        'd': 2,
        'e': 3,
    }
}
```
[...]

PR #233, opened 2018-03-29
Feature soft shutdown for concent service

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/152.

### DIFF
--- a/concent_api/concent_api/settings/base.py
+++ b/concent_api/concent_api/settings/base.py
@@ -18,6 +18,8 @@ INSTALLED_APPS = [
 
     # Third-party apps:
     'raven.contrib.django.raven_compat',
+    'constance',
+    'constance.backends.database',
 
     # Our apps:
     'concent_api',
@@ -89,6 +91,9 @@ DATABASES = {
 }
 
 DATABASE_ROUTERS = ['concent_api.database_router.DatabaseRouter']
+
+# Defines database used by Constance app.
+CONSTANCE_DBS = ['default']
 
 # Password validation
 # https://docs.djangoproject.com/en/1.11/ref/settings/#auth-password-validators
@@ -215,6 +220,16 @@ LOGGING = {
     },
 }
 
+# Django constance config
+CONSTANCE_BACKEND = 'constance.backends.database.DatabaseBackend'
+CONSTANCE_CONFIG = {
+    # Defines if Concent is in soft shutdown mode.
+    'SOFT_SHUTDOWN_MODE': (False, 'Soft shutdown mode', bool),
+}
+
+CONSTANCE_CONFIG_FIELDSETS = {
+    'Concent Options': ('SOFT_SHUTDOWN_MODE',),
+}
 
 # Private and public keys to be used by Concent to sign and encrypt its own messages.
 # Stored in a 'bytes' array, e.g. b'\xf3\x97\x19\xcdX\xda...'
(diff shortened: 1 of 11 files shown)


cameel:
Please mark it as `[WIP]` (unless you think you fixed the issue already). I think we should make it work with storage cluster too as a part of this pull request but I still don't have a clear idea which solution to choose (I could not find a moment to think about it in the last few days).

rwrzesien:
@cameel Right, let me write down possible quickly invented solutions:
* When `SHUTDOWN_MODE` is on, concent-api is making this information public by exposing it via endpoint (like protocol constants), gatekeeper is checking in on every download/upload request (lame, because of a need to perform an extra request on each request)
* When `SHUTDOWN_MODE` is turned on, concent-api sends request to gatekeeper endpoint, gatekeeper is storing this information in a file.

Every of those solutions is far from perfect and has obvious issues

rwrzesien:
@cameel Rebased on latest master and updated on the reverting gatekeeper changes. I will create separate issue for changes in admin panel which need to wait for https://github.com/golemfactory/concent/issues/141.

concent_api/concent_api/settings/base.py
```diff
@@ -90,6 +92,9 @@
 
 DATABASE_ROUTERS = ['concent_api.database_router.DatabaseRouter']
 
+# Defines database used by Constance app.
+CONSTANCE_DBS = ['default']
```
cameel: Please remember to update this if we merge #330 first.

concent_api/utils/admin.py
```diff
@@ -0,0 +1,28 @@
+
+
+class ReadOnlyMixin:
```
cameel: I think the name should be more specific. It can't be mixed into anything. Only into admin related stuff. It's in the `admin` module, but still.

concent_api/core/message_handlers.py
```diff
@@ -1001,6 +1010,13 @@ def update_subtask(
         next_deadline,
     )
 
+    # Concent should send e-mail notification when the last active subtask switches to a passive state.
+    if config.SHUTDOWN_MODE is True and not Subtask.objects.filter(state__in=Subtask.ACTIVE_STATES).exists():
+        mail_admins(
+            subject = 'Concent Shutdown Ready',
+            message = 'All subtasks in Concent instance being in shutdown mode turned into passive states',
```
cameel: Please reword:

``` python
subject = "Concent soft shutdown complete",
mesage = (
    "All subtasks tracked by this Concent instance are now in passive states.\n"
    "It's safe to turn off the control cluster.\n"
    "Note that there may still be downloads in progress on the storage cluster."
),
```

concent_api/utils/admin.py
rwrzesien: :+1: done

concent_api/core/message_handlers.py

rwrzesien:
@cameel 

> I'd also rename 'shutdown mode' to 'soft shutdown mode' but it's probably to late to change this now.

Not with Pycharm :) done.

> Before merging it, please test it on the dev cluster and make sure #236 is not happening again.

I have tested it locally with cluster tests, first with SOFT_SHUTDOWN mode turned off, then turned on, but if you think it is not enough I will have to wait until Monday.

cameel:
> Not with Pycharm :) done.

Great!

> I have tested it locally with cluster tests, first with SOFT_SHUTDOWN mode turned off, then turned on, but if you think it is not enough I will have to wait until Monday.

I'd rather have it tested more thoroughly than risk that we'll have to revert it again :) There's not going to be much development during the long weekend so I think waiting for Monday should not be a big problem. Your code won't get out of sync in the meantime :)

PR #238, opened 2018-04-03
Fix logging text with subtasks state

### LINKED ISSUE #182: "None of subtask changed state" in concent_api logs
We're getting lots of lines like this in logs on staging:

```
textPayload: "17:08:56 INFO     | None of subtask changed state -- CLIENT PUBLIC KEY: b'07byqHcsE3uMqRRw3q5GykDQyK0YNXjUO1N1JnZpHiPykajphbddrrTdSw8PJ79BO2masDxXm2SXgep/RovsYQ=='
```

I think that `None` was meant to be a number. Please find and fix the error that's causing this.

kbeker:
Resolves #182

### DIFF
--- a/concent_api/utils/logging.py
+++ b/concent_api/utils/logging.py
@@ -162,17 +162,11 @@ def log_stored_message_added_to_subtask(
     ))
 
 
-def log_no_changes_in_subtask_states(client_public_key: bytes):
-    logger.info('None of subtask changed state -- CLIENT PUBLIC KEY: {}'.format(
-        client_public_key,
-    ))
-
-
 def log_changes_in_subtask_states(client_public_key: bytes, count: int):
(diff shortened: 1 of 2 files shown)


concent_api/utils/logging.py
```diff
@@ -163,7 +163,7 @@ def log_stored_message_added_to_subtask(
 
 
 def log_no_changes_in_subtask_states(client_public_key: bytes):
-    logger.info('None of subtask changed state -- CLIENT PUBLIC KEY: {}'.format(
+    logger.info('0 subtasks changed state -- CLIENT PUBLIC KEY: {}'.format(
```
rwrzesien: I think you should merge it with `log_changes_in_subtask_states`.

PR #239, opened 2018-04-03
Update time window in force payment usecase

### LINKED ISSUE #166: Update time windows and settings to match the current blueprint
Blueprint: #165

- Review and update time windows in all use cases to match the ones listed in the blueprint. Things to pay particular attention to:
    - Forced result transfer now starts after task deadline (not after `ReportTaskToCompute.timestamp`)
    - Time windows in REPORT and RESULT TRANSFER use cases now overlap. They both start at `ReportTaskToCompute.timestamp`. 
    - Time windows in the RESULT TRANSFER use case have been greatly simplified. They definitely don't match the current implementation that was based on the wrong setting (FAT) that I specified in #12.
    - PGP is no longer necessary in the PAYMENT use case.
    - All time windows where Concent expects the other client to submit or upload something now start in the moment when the force message is submitted.
- Make sure that the deadline in the `FileTransferToken` matches the blueprint.
- Remove unnecessary settings. We need only the ones specified in the blueprint.
- Update the `protocol-constants/` endpoint
- Update values in existing blueprints: #12, #41, #116, and the one that's already in Sphinx docs.

For now you can assume that use cases never run simultaneously. If one is running, clients cannot start another one until Concent enters a passive state.

Jakub89:
Final part of #165 
Resolves #166

### DIFF
--- a/concent_api/api-integration-force-payment.py
+++ b/concent_api/api-integration-force-payment.py
@@ -86,7 +86,7 @@ def main():
         subtask_results_accepted_list = [
             subtask_results_accepted(
                 timestamp       = timestamp_to_isoformat(current_time),
-                payment_ts      = current_time - PAYMENT_DUE_TIME - PAYMENT_GRACE_PERIOD - 1,
+                payment_ts      = current_time - PAYMENT_DUE_TIME - 1,
                 task_to_compute = task_to_compute(
                     timestamp                       = timestamp_to_isoformat(current_time),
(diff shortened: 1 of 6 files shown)


Jakub89:
@rwrzesien I've only found more `PAYMENT_GRACE_PERIOD` in `api-integration-force-payment.py`. Did I miss something else?

rwrzesien:
@Jakub89 I see there are some usages in `payments/mock.py`. You can ask @kbeker for details.

PR #242, opened 2018-04-04
Feature add auth decorator

### LINKED ISSUE #204: [Authentication] Changes in core 'receive' function
With the implementation of https://github.com/golemfactory/concent/issues/138 there must be following changes in `receive` and `receive_out_of_band` views:

* Create a decorator which will load `ClientAuthorization` message, validate public key and pass key into the views. 
* Decorator should handle exceptions related with this action.

In tests:
* Add ClientAuthorization to every call to `receive` and `receive_out_of_band` in tests.
* In all auth tests, add ClientAuthorization with wrong keys to every call to `receive` and `receive_out_of_band` in tests.

kbeker:
Resolves #204

### DIFF
--- a/concent_api/core/tests/test_auth_accept_or_reject_results.py
+++ b/concent_api/core/tests/test_auth_accept_or_reject_results.py
@@ -240,7 +240,7 @@ class AuthAcceptOrRejectIntegrationTest(ConcentIntegrationTestCase):
         with freeze_time("2018-02-05 10:00:29"):
             response = self.client.post(
                 reverse('core:receive'),
-                data                            = '',
+                data                            = self._create_diff_requestor_auth_message(),
                 content_type                    = 'application/octet-stream',
                 HTTP_CONCENT_CLIENT_PUBLIC_KEY  = self._get_encoded_key(self.DIFFERENT_REQUESTOR_PUBLIC_KEY),
             )
@@ -251,7 +251,7 @@ class AuthAcceptOrRejectIntegrationTest(ConcentIntegrationTestCase):
         with freeze_time("2018-02-05 10:00:29"):
             response = self.client.post(
                 reverse('core:receive'),
-                data                            = '',
+                data                            = self._create_provider_auth_message(),
                 content_type                    = 'application/octet-stream',
                 HTTP_CONCENT_CLIENT_PUBLIC_KEY  = self._get_encoded_provider_public_key(),
             )
@@ -263,7 +263,7 @@ class AuthAcceptOrRejectIntegrationTest(ConcentIntegrationTestCase):
         with freeze_time("2018-02-05 10:00:29"):
             response = self.client.post(
                 reverse('core:receive'),
-                data                            = '',
+                data                            = self._create_requestor_auth_message(),
                 content_type                    = 'application/octet-stream',
                 HTTP_CONCENT_CLIENT_PUBLIC_KEY  = self._get_encoded_requestor_public_key(),
             )
@@ -365,7 +365,7 @@ class AuthAcceptOrRejectIntegrationTest(ConcentIntegrationTestCase):
         with freeze_time("2018-02-05 10:00:24"):
             response = self.client.post(
                 reverse('core:receive'),
-                data                            = '',
+                data                            = self._create_diff_requestor_auth_message(),
                 content_type                    = 'application/octet-stream',
                 HTTP_CONCENT_CLIENT_PUBLIC_KEY  = self._get_encoded_key(self.DIFFERENT_REQUESTOR_PUBLIC_KEY),
             )
@@ -376,7 +376,7 @@ class AuthAcceptOrRejectIntegrationTest(ConcentIntegrationTestCase):
         with freeze_time("2018-02-05 10:00:24"):
             response = self.client.post(
                 reverse('core:receive'),
-                data                            = '',
+                data                            = self._create_provider_auth_message(),
                 content_type                    = 'application/octet-stream',
                 HTTP_CONCENT_CLIENT_PUBLIC_KEY  = self._get_encoded_provider_public_key(),
             )
@@ -388,7 +388,7 @@ class AuthAcceptOrRejectIntegrationTest(ConcentIntegrationTestCase):
         with freeze_time("2018-02-05 10:00:24"):
             response = self.client.post(
                 reverse('core:receive'),
-                data                            = '',
+                data                            = self._create_requestor_auth_message(),
                 content_type                    = 'application/octet-stream',
                 HTTP_CONCENT_CLIENT_PUBLIC_KEY  = self._get_encoded_requestor_public_key(),
             )
@@ -517,7 +517,7 @@ class AuthAcceptOrRejectIntegrationTest(ConcentIntegrationTestCase):
         with freeze_time("2018-02-05 11:00:02"):
             response = self.client.post(
                 reverse('core:receive'),
-                data                            = '',
+                data                            = self._create_diff_requestor_auth_message(),
                 content_type                    = 'application/octet-stream',
                 HTTP_CONCENT_CLIENT_PUBLIC_KEY  = self._get_encoded_key(self.DIFFERENT_PROVIDER_PUBLIC_KEY),
             )
@@ -528,7 +528,7 @@ class AuthAcceptOrRejectIntegrationTest(ConcentIntegrationTestCase):
         with freeze_time("2018-02-05 11:00:02"):
             response = self.client.post(
                 reverse('core:receive'),
-                data                            = '',
+                data                            = self._create_requestor_auth_message(),
                 content_type                    = 'application/octet-stream',
                 HTTP_CONCENT_CLIENT_PUBLIC_KEY  = self._get_encoded_requestor_public_key(),
             )
@@ -540,7 +540,7 @@ class AuthAcceptOrRejectIntegrationTest(ConcentIntegrationTestCase):
         with freeze_time("2018-02-05 11:00:02"):
             response = self.client.post(
                 reverse('core:receive'),
-                data                            = '',
+                data                            = self._create_provider_auth_message(),
                 content_type                    = 'application/octet-stream',
                 HTTP_CONCENT_CLIENT_PUBLIC_KEY  = self._get_encoded_provider_public_key(),
             )
@@ -637,7 +637,7 @@ class AuthAcceptOrRejectIntegrationTest(ConcentIntegrationTestCase):
         with freeze_time("2018-02-05 10:00:24"):
             response = self.client.post(
                 reverse('core:receive'),
-                data='',
+                data = self._create_diff_requestor_auth_message(),
                 content_type='application/octet-stream',
                 HTTP_CONCENT_CLIENT_PUBLIC_KEY=self._get_encoded_key(self.DIFFERENT_REQUESTOR_PUBLIC_KEY),
             )
@@ -648,7 +648,7 @@ class AuthAcceptOrRejectIntegrationTest(ConcentIntegrationTestCase):
         with freeze_time("2018-02-05 10:00:24"):
             response = self.client.post(
                 reverse('core:receive'),
-                data                           = '',
+                data                           = self._create_provider_auth_message(),
                 content_type                   = 'application/octet-stream',
                 HTTP_CONCENT_CLIENT_PUBLIC_KEY = self._get_encoded_provider_public_key(),
             )
@@ -660,7 +660,7 @@ class AuthAcceptOrRejectIntegrationTest(ConcentIntegrationTestCase):
         with freeze_time("2018-02-05 10:00:24"):
             response = self.client.post(
                 reverse('core:receive'),
-                data                            = '',
+                data                            = self._create_requestor_auth_message(),
                 content_type                    = 'application/octet-stream',
                 HTTP_CONCENT_CLIENT_PUBLIC_KEY  = self._get_encoded_requestor_public_key(),
             )
@@ -802,7 +802,7 @@ class AuthAcceptOrRejectIntegrationTest(ConcentIntegrationTestCase):
         with freeze_time("2018-02-05 11:00:02"):
             response = self.client.post(
                 reverse('core:receive'),
-                data                            = '',
+                data                            = self._create_diff_requestor_auth_message(),
                 content_type                    = 'application/octet-stream',
                 HTTP_CONCENT_CLIENT_PUBLIC_KEY  = self._get_encoded_key(self.DIFFERENT_PROVIDER_PUBLIC_KEY),
             )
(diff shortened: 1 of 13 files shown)


concent_api/utils/message_handler.py
```diff
@@ -0,0 +1,72 @@
+from functools                      import wraps
```
rwrzesien: You can put those two decorators in a single file `utils\decorators.py`. It would be more in python/django convention and less repetitive on imports.

concent_api/utils/client_auth.py
```diff
@@ -0,0 +1,42 @@
...
+from golem_messages.exceptions      import MessageTooOldError
+from golem_messages.exceptions      import TimestampError
+
+from core.validation                import validate_golem_message_signed_with_key
+from core.exceptions                import Http400
+
+from utils.shortcuts                import load_without_public_key
+
+
+def client_auth(message_handler):
```
rwrzesien: This variable should be still called `view`, because it should not matter if we are wrapping just view or view wrapped in some other decorator. I think it should be good to write some tests that check this.

concent_api/utils/message_handler.py
```diff
@@ -0,0 +1,72 @@
...
+from golem_messages                 import dump
+from golem_messages.message         import Message
+
+from core.exceptions                import Http400
+
+from utils                          import logging
+
+
+def message_handler(view):
+    @wraps(view)
```
rwrzesien: Please add a comment what this decorator does, and as with previous one write some tests if possible (and reuse existing ones for `api_view`).

concent_api/utils/client_auth.py
rwrzesien: The name is little bit misleading - it actually does more with loading golem message and handling its exceptions. Please call this one `require_golem_message` or something more descriptive.

concent_api/utils/message_handler.py
```diff
@@ -0,0 +1,72 @@
...
+
+from golem_messages                 import dump
+from golem_messages.message         import Message
+
+from core.exceptions                import Http400
+
+from utils                          import logging
+
+
+def message_handler(view):
```
rwrzesien: Same here, we need more descriptive name, as it actually handles transactions, errors and responses. How about `handle_errors_and_responses` ? If you have better one feel free to replace it.

concent_api/utils/message_handler.py
```diff
@@ -0,0 +1,72 @@
...
+            logging.log_empty_queue(
+                view.__name__,
+                message.client_public_key,
+            )
+            return HttpResponse("", status = 204)
+        elif isinstance(response_from_view, bytes):
+            return HttpResponse(response_from_view)
+
+        assert False, "Invalid response type"
+        raise Exception("Invalid response type")
```
dybi: This code is unreachable. I would delete `assert` above ;)

cameel: It's been added here only because pylint complained about missing return.

If anything, we could delete the `raise` but I think it's better than disabling the warning.

cameel: Because this part is never meant to run in production. Code that would reach this `raise` is so broken that it should never make it to production. Doing anything here is simply not necessary unless we want to code defensively and add additional logic to handle "what if" situations resulting from bugs.

This `assert` is like a white line on the street. It informs us that this is a boundary that should not be crossed and makes it apparent that we stepped past it. But that's it. It's not meant to hold our backs if we actually cross it in a real situation.

To be honest this case is borderline because the "handling" is simply a single `raise` so it's not that different from `assert False`. As long as it's not just a silent `return`, doing it instead of `assert` is not harmful, just unnecessary in my opinion. It would be different if we needed more code to handle this. Anything more than an `assert` would be a waste of time.

For me `assert` is an important tool to avoid coding defensively - i.e. for every nonsensical case I can think of - while still being sure that if I'm wrong or if the code around changes this will not go unnoticed. I can just write an `assert` that makes it clear that I'm sure that a situation is impossible (or a result of a bug elsewhere) and move on to more important parts of the code. In other words, please don't think about it as assert vs no assert. More often it's assert vs lots of unnecessary code to handle something that cannot happen.

concent_api/core/tests/test_auth_accept_or_reject_results.py
```diff
@@ -240,7 +240,7 @@ def test_requestor_should_receive_subtask_results_from_concent_with_correct_keys
         with freeze_time("2018-02-05 10:00:29"):
             response = self.client.post(
                 reverse('core:receive'),
-                data                            = '',
+                data                            = self._create_client_auth_message(self.DIFFERENT_REQUESTOR_PRIVATE_KEY, self.DIFFERENT_REQUESTOR_PUBLIC_KEY),
```
rwrzesien: One more thing that about the tests. How about you write 4 wrapping functions for `_create_client_auth_message`, like this:
```
def _create_client_auth_message_provider():
    return self._create_client_auth_message(self.DIFFERENT_REQUESTOR_PRIVATE_KEY, self.DIFFERENT_REQUESTOR_PUBLIC_KEY)
```

etc. It shouldn't take much to write them, and they can be replaced in all filed with single grep.

concent_api/utils/client_auth.py
```diff
@@ -0,0 +1,42 @@
...
+    @wraps(message_handler)
+    def wrapper(request, *args, **kwargs):
+        if request.content_type == '':
+            return JsonResponse({'error': 'Content-Type is missing.'}, status = 400)
+        elif request.content_type == 'application/octet-stream':
+            try:
+                message = load_without_public_key(request.body, _check_time = False)
```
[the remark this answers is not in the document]
kbeker: Done

concent_api/utils/client_auth.py
```diff
@@ -0,0 +1,42 @@
...
+from golem_messages.exceptions      import TimestampError
+
+from core.validation                import validate_golem_message_signed_with_key
+from core.exceptions                import Http400
+
+from utils.shortcuts                import load_without_public_key
+
+
+def client_auth(message_handler):
+    @wraps(message_handler)
```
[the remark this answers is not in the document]
kbeker: Done

concent_api/utils/message_handler.py
kbeker: Done

concent_api/core/tests/test_auth_accept_or_reject_results.py
kbeker: Done

rwrzesien: Missing in this file.

concent_api/core/tests/test_auth_force_payment.py
```diff
@@ -95,9 +95,8 @@ def test_provider_send_force_payment_and_concent_should_return_it_to_requestor_w
         with freeze_time("2018-02-05 12:00:21"):
             response_2 = self.client.post(
                 reverse('core:receive_out_of_band'),
-                data                            = '',
+                data                            = self._create_client_auth_message(self.PROVIDER_PRIVATE_KEY, self.PROVIDER_PUBLIC_KEY),
```
rwrzesien: Missing in this file.

concent_api/utils/decorators.py
```diff
@@ -0,0 +1,115 @@
...
+from golem_messages.exceptions      import TimestampError
+
+from core.validation                import validate_golem_message_signed_with_key
+from core.exceptions                import Http400
+
+from utils.shortcuts                import load_without_public_key
+
+from utils                          import logging
+
+"""
```
rwrzesien: Comment should be below function name.

concent_api/utils/decorators.py
```diff
@@ -0,0 +1,115 @@
...
+                return JsonResponse({'error': "Error in Golem Message. {}".format(exception)}, status = 400)
+        else:
+            return JsonResponse({'error': "Concent supports only application/octet-stream."}, status = 400)
+
+        return view(request, auth_message, *args, *kwargs)
+    return wrapper
+
+
+def handle_errors_and_responses(view):
+    @wraps(view)
```
rwrzesien: Please add comment here too.

kbeker: Done

concent_api/utils/decorators.py
kbeker: Done

concent_api/core/tests/test_integration_report_computed_task.py
```diff
@@ -152,9 +152,8 @@ def test_provider_forces_computed_task_report_and_requestor_sends_acknowledgemen
         with freeze_time("2017-12-01 11:00:05"):
             response_2 = self.client.post(
                 reverse('core:receive'),
-                data         = '',
-                content_type = '',
-                HTTP_CONCENT_CLIENT_PUBLIC_KEY = self._get_encoded_requestor_public_key(),
+                data         = self._create_client_auth_message(self.REQUESTOR_PRIVATE_KEY, self.REQUESTOR_PUBLIC_KEY),
```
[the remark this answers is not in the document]
kbeker: Done

concent_api/core/tests/test_integration_accept_or_reject_results.py
```diff
@@ -327,9 +327,8 @@ def test_requestor_should_receive_subtask_results_from_concent(self):
         with freeze_time("2018-02-05 10:00:29"):
             response_2 = self.client.post(
                 reverse('core:receive'),
-                data                            = '',
+                data                            = self._create_client_auth_message(self.REQUESTOR_PRIVATE_KEY, self.REQUESTOR_PUBLIC_KEY),
```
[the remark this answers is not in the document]
kbeker: Done

concent_api/core/tests/test_core_views.py
```diff
@@ -671,8 +676,7 @@ def test_receive_should_accept_valid_message(self):
         response = self.client.post(
             reverse('core:receive'),
             content_type                   = 'application/octet-stream',
-            data                           = '',
-            HTTP_CONCENT_CLIENT_PUBLIC_KEY = b64encode(REQUESTOR_PUBLIC_KEY).decode('ascii')
+            data                           = self.create_client_auth_message(REQUESTOR_PRIVATE_KEY, REQUESTOR_PUBLIC_KEY),
```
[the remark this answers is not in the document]
kbeker: Done

concent_api/core/tests/test_auth_report_computed_task.py
```diff
@@ -116,9 +112,8 @@ def test_provider_forces_computed_task_report_and_requestor_sends_acknowledgemen
         with freeze_time("2017-12-01 11:00:05"):
             response = self.client.post(
                 reverse('core:receive'),
-                data                           = '',
-                content_type                   = '',
-                HTTP_CONCENT_CLIENT_PUBLIC_KEY = b64encode(DIFFERENT_REQUESTOR_PUBLIC_KEY).decode('ascii'),
+                data                           = self._create_client_auth_message(self.DIFFERENT_REQUESTOR_PRIVATE_KEY, self.DIFFERENT_REQUESTOR_PUBLIC_KEY),
```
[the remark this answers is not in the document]
kbeker: Done

concent_api/core/tests/test_auth_force_payment.py
kbeker: Done

concent_api/core/tests/test_auth_accept_or_reject_results.py
kbeker: Done

concent_api/utils/tests/test_decorators.py
```diff
@@ -0,0 +1,222 @@
...
+        def dummy_view_handle_http_response(_request, _message):
+            http_response = HttpResponse(status = 200)
+            return http_response
+
+        response = dummy_view_handle_http_response(request, self.client_auth)
+
+        self.assertIsInstance(response, HttpResponse)
+        self.assertEqual(response.status_code, 200)
+
+    def test_handle_errors_and_responses_should_return_http_response_if_view_passed_none(self):
```
dybi: please rename -> `...should_return_empty_http_response...`

concent_api/utils/tests/test_decorators.py
```diff
@@ -0,0 +1,222 @@
...
+        @handle_errors_and_responses
+        def dummy_view_handle_none_response(_request, _message):
+            return None
+
+        response = dummy_view_handle_none_response(request, self.client_auth)  # pylint: disable=assignment-from-none
+
+        self.assertEqual(response.status_code, 204)
+        self.assertEqual(len(response.content), 0)
+
+    def test_handle_errors_and_responses_should_return_json_response_if_view_passed_dict(self):
```
dybi: suggestion: you might want to use a bit more sophisticated dicts here - empty dict is a quite specific case;)

concent_api/utils/tests/test_decorators.py
```diff
@@ -0,0 +1,222 @@
...
+        def dummy_view_handle_dict(_request, _message):
+            return {}
+
+        response = dummy_view_handle_dict(request, self.client_auth)
+
+        self.assertEqual(response.status_code, 200)  # pylint: disable=no-member
+        self.assertIsInstance(response, JsonResponse)
+        self.assertEqual(json.loads(response.content), {})  # pylint: disable=no-member
+
+    def test_handle_errors_and_responses_should_return_http_response_if_view_passed_http_400_exception(self):
```
dybi: `passed` -> `raised`

concent_api/utils/tests/test_decorators.py
```diff
@@ -0,0 +1,222 @@
...
+        self.assertEqual(response.status_code, 400)  # pylint: disable=no-member
+        self.assertIn('error', json.loads(response.content))  # pylint: disable=no-member
+
+    def test_handle_errors_and_responses_should_return_http_response_not_allowed_if_view_passed_it(self):
+
+        request = self.request_factory.post("/dummy-url/", content_type = 'application/octet-stream')
+
+        @handle_errors_and_responses
+        def dummy_view_handle_http_response_not_allowed(_request, _message):
+            return HttpResponseNotAllowed({}, status = 405)
```
dybi: I do not mean to be picky, but we have agreed upon: `status=405` style formatting ;)

concent_api/core/tests/test_core_views.py
```diff
@@ -938,14 +939,18 @@ def setUp(self):
         new_message_inbox.full_clean()
         new_message_inbox.save()
 
+    def create_client_auth_message(self, client_priv_key, client_public_key):  # pylint: disable=no-self-use
```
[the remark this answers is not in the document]
kbeker: Done

concent_api/core/tests/utils.py
```diff
@@ -625,3 +625,24 @@ def _add_signature_to_message(self, golem_message, priv_key):  # pylint: disable
         golem_message = golem_message.serialize(sign_func = signature)
         golem_message = message.Message.deserialize(golem_message, decrypt_func = None, check_time = False)
         return golem_message.sig
+
+    def _create_client_auth_message(self, client_priv_key, client_public_key):  # pylint: disable=no-self-use
```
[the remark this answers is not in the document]
kbeker: Done

PR #246, opened 2018-04-05
Add sign to nested FileTransferToken message

### LINKED ISSUE #245: FileTransferToken is not signed separately from the message that contains it
`FileTransferToken` stored inside `ForceGetTaskResultDownload` and `ForceGetTaskResultUpload` must be signed separately from the message that contains it. That's because the client extracts it from the containing message and sends to Concent separately when downloading/uploading files. If it's not signed separately, its signature is all zeros and authentication fails.

https://github.com/golemfactory/concent/blob/ce1b6bfc77ae9206a29d9c65fe68bf23c7e4bcec/concent_api/core/message_handlers.py#L713-L720

The code creates a token, inserts it into the outer message and returns them. The outer message is later signed by our `api_view` decorator but the token is not (nested messages are not signed separately by default).

Also, make sure that gatekeeper logs something when it gets a token that's not signed properly (e.g. signature is all zeros). Currently it appears to detect this case but, unlike in case of other errors, it either does not log anything or it does it at the `DEBUG` level.

kbeker:
Resolves #245

### DIFF
--- a/concent_api/core/tests/test_integration_get_task_result.py
+++ b/concent_api/core/tests/test_integration_get_task_result.py
@@ -356,6 +356,7 @@ class GetTaskResultIntegrationTest(ConcentIntegrationTestCase):
 
         # Test FileTransferToken message
         self.assertIsInstance(message_file_transfer_token, message.FileTransferToken)
+        self.assertEqual(message_file_transfer_token.sig,  self._get_signature(message_file_transfer_token, CONCENT_PRIVATE_KEY, CONCENT_PUBLIC_KEY))
         self.assertEqual(message_file_transfer_token.timestamp, self._parse_iso_date_to_timestamp("2017-12-01 11:00:12"))
         self.assertEqual(message_file_transfer_token.token_expiration_deadline, self._parse_iso_date_to_timestamp("2017-12-01 11:00:50"))
         self.assertEqual(message_file_transfer_token.operation, 'upload')
@@ -613,6 +614,7 @@ class GetTaskResultIntegrationTest(ConcentIntegrationTestCase):
 
         # Test FileTransferToken message
         self.assertIsInstance(message_file_transfer_token,                      message.FileTransferToken)
+        self.assertEqual(message_file_transfer_token.sig,                       self._get_signature(message_file_transfer_token, CONCENT_PRIVATE_KEY, CONCENT_PUBLIC_KEY))
         self.assertEqual(message_file_transfer_token.timestamp,                 self._parse_iso_date_to_timestamp("2017-12-01 11:00:06"))
         self.assertEqual(message_file_transfer_token.token_expiration_deadline, self._parse_iso_date_to_timestamp("2017-12-01 11:00:50"))
         self.assertEqual(message_file_transfer_token.operation,                 'upload')
@@ -645,6 +647,7 @@ class GetTaskResultIntegrationTest(ConcentIntegrationTestCase):
         self.assertEqual(message_from_concent.timestamp,                                                    self._parse_iso_date_to_timestamp("2017-12-01 11:00:21"))
         self.assertEqual(message_from_concent.force_get_task_result.report_computed_task.task_to_compute,   deserialized_task_to_compute)
         self.assertIsInstance(message_from_concent.file_transfer_token,                                     message.FileTransferToken)
+        self.assertEqual(message_from_concent.file_transfer_token.sig,                                      self._get_signature(message_from_concent.file_transfer_token, CONCENT_PRIVATE_KEY, CONCENT_PUBLIC_KEY))
         self.assertEqual(message_from_concent.file_transfer_token.timestamp,                                self._parse_iso_date_to_timestamp("2017-12-01 11:00:21"))
         self.assertEqual(message_from_concent.file_transfer_token.token_expiration_deadline,                self._parse_iso_date_to_timestamp("2017-12-01 11:30:00"))
         self.assertEqual(message_from_concent.file_transfer_token.operation,                                'download')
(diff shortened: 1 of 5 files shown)


concent_api/core/transfer_operations.py
```diff
@@ -136,6 +138,10 @@ def create_file_transfer_token(
     file_transfer_token.files[0]['checksum']  = report_computed_task.package_hash
     file_transfer_token.files[0]['size']      = report_computed_task.size
 
+    sign = partial(cryptography.ecdsa_sign, settings.CONCENT_PRIVATE_KEY)
+    seriallized_ftt = file_transfer_token.serialize(sign_func=sign)
```
cameel: typo: `seriallized` -> `serialized`

concent_api/core/transfer_operations.py
```diff
@@ -136,6 +138,10 @@ def create_file_transfer_token(
     file_transfer_token.files[0]['checksum']  = report_computed_task.package_hash
     file_transfer_token.files[0]['size']      = report_computed_task.size
 
+    sign = partial(cryptography.ecdsa_sign, settings.CONCENT_PRIVATE_KEY)
```
cameel: Call it `ecdsa_sign_with_concent_key`

concent_api/core/tests/test_integration_get_task_result.py
```diff
@@ -356,6 +356,7 @@ def test_concent_requests_task_result_from_provider_and_requestor_receives_failu
 
         # Test FileTransferToken message
         self.assertIsInstance(message_file_transfer_token, message.FileTransferToken)
+        self.assertIsNot(message_file_transfer_token.sig,  None)
```
cameel: You should also check that `sig` is not equal to the output of `fake sign` (i.e. a byte string containing just zeros). Actually it would be best to verify the signature here. Otherwise you can't be sure it's correct.

concent_api/core/transfer_operations.py
```diff
@@ -186,3 +188,10 @@ def send_request_to_cluster_storage(headers, request_http_address):
             request_http_address,
             headers=headers
     )
+
+
+def add_message_signature(golem_message, priv_key, pub_key):
```
rwrzesien: Please rename to `sign_message`.

concent_api/core/tests/utils.py
```diff
@@ -616,3 +616,9 @@ def _assert_stored_message_counter_not_increased(self):
 
     def _assert_client_count_is_equal(self, count):
         self.assertEqual(Client.objects.count(), count)
+
+    def _get_signature(self, golem_message, priv_key, pub_key):  # pylint: disable=no-self-use
+        golem_message.sig = None
+        serialized_message = dump(golem_message, priv_key, pub_key)
```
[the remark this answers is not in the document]
cameel: You should post it in a comment in that other issue so that we can catch it during review if you miss it.

concent_api/core/tests/test_unit_transfer_operations.py
```diff
@@ -0,0 +1,44 @@
...
+        self.assertEqual(self.ping_message.sig, None)
+
+        ping_message = sign_message(self.ping_message, CONCENT_PRIVATE_KEY, CONCENT_PUBLIC_KEY)
+
+        self.assertIsNot(ping_message.sig, None)
+        self.assertIsInstance(ping_message.sig, bytes)
+
+    def test_empty_arguments_to_add_signature(self):
+
+        with self.assertRaises(AssertionError):
```
cameel: You're not supposed to test `assert`s. They're not part of functions behavior. It's like trying to get a value of a mathematical function outside of its domain. An assertion on function's input (i.e. a [precondition](https://en.wikipedia.org/wiki/Precondition)) means that the function does not implement any specific behavior the input that fails the assertion and you, as a caller, are responsible for ensuring that it never gets input that triggers it. If the function is meant to handle this case you should not have put that assertion inside it. You should have implemented some well defined behavior for that input instead.

concent_api/gatekeeper/views.py
```diff
@@ -119,6 +119,10 @@ def parse_headers(request: WSGIRequest, path_to_file: str) -> Union[FileTransfer
     )
 
     # Check ConcentFileTransferToken each field:
+    # -SIGNATURE
+    if not isinstance(loaded_golem_message.sig, bytes):
+        logger.info(f'FileTransferToken with subtask_id {loaded_golem_message.subtask_id}, has empty signature -- CLIENT PUBLIC KEY: {client_public_key}')
+        return gatekeeper_access_denied_response('Empty signature field.', path_to_file, loaded_golem_message.subtask_id, client_public_key)
```
cameel: Isn't `gatekeeper_access_denied_response()` supposed to log the error? I think that the problem here is that that function does not log properly and adding an extra `logger.info()` here is just a workaround that does not solve the cause of the problem.

concent_api/core/tests/utils.py
```diff
@@ -616,3 +616,10 @@ def _assert_stored_message_counter_not_increased(self):
 
     def _assert_client_count_is_equal(self, count):
         self.assertEqual(Client.objects.count(), count)
+
+    # TODO: Merge with '_sign_message' after authentication is merged
+    def _get_signature(self, golem_message, priv_key, pub_key):  # pylint: disable=no-self-use
```
cameel: Calling it `get_xxx` is misleading because the function modifies the message it gets as input.

concent_api/core/tests/utils.py
```diff
@@ -616,3 +616,10 @@ def _assert_stored_message_counter_not_increased(self):
...
+        golem_message.sig = None
+        serialized_message = dump(golem_message, priv_key, pub_key)
+        loaded_message = load(serialized_message, priv_key, pub_key, check_time = False)
+        return loaded_message.sig
```
cameel: You don't have to `load()` and `dump()`. That's an overkill. golem-messages has functions for computing the signature.

PR #250, opened 2018-04-05
Fix function add message sig

kbeker:


concent_api/core/tests/utils.py
```diff
@@ -618,8 +619,9 @@ def _assert_client_count_is_equal(self, count):
         self.assertEqual(Client.objects.count(), count)
 
     # TODO: Merge with '_sign_message' after authentication is merged
-    def _get_signature(self, golem_message, priv_key, pub_key):  # pylint: disable=no-self-use
+    def _add_signature_to_message(self, golem_message, priv_key):  # pylint: disable=no-self-use
         golem_message.sig = None
-        serialized_message = dump(golem_message, priv_key, pub_key)
-        loaded_message = load(serialized_message, priv_key, pub_key, check_time = False)
-        return loaded_message.sig
+        signature = functools.partial(cryptography.ecdsa_sign, priv_key)
```
rwrzesien: @cameel I saw you comment in outdated PR about not using load/dump as an overkill. Don't you think that the way which golem signs the message might change, but it will always be doable using those shortcuts, so using them protects as from this kind of change ?

concent_api/core/transfer_operations.py
```diff
@@ -191,12 +193,12 @@ def send_request_to_cluster_storage(headers, request_http_address):
     )
 
 
-def sign_message(golem_message, priv_key, pub_key):
+def sign_message(golem_message, priv_key):
```
dybi: this function deosn't belong here; please move it to `utils/helpers.py`

concent_api/core/tests/utils.py
```diff
@@ -13,6 +13,7 @@
 
 from golem_messages         import dump
 from golem_messages         import load
+from golem_messages         import cryptography
 from golem_messages         import message
```
Jakub89: We should remember about the alphabetical order.

concent_api/core/transfer_operations.py
kbeker: Done

concent_api/core/tests/utils.py
kbeker: Done

concent_api/core/tests/utils.py
cameel: Well, now I see why you wanted it that way but I still think that doing a full dump/load cycle just to get a signature should not be necessary. I would agree if we wanted to perform some obscure and very specialized operation but to me signing a message is something that should be a first class operation in the API of golem-messages.

If we're afraid that golem-messages treats this as an implementation detail (rather than a part of its public API), in my opinion, we should raise our concern on Rocket rather than try to work around it.

rwrzesien: Depending on response on rocket chat we will either leave it as it is or make a separate issue out of it.

Issue #201, opened 2018-03-23
Concent: Accepting additional verification requests from provider


cameel:
### Sequence of operation
Implement the following part of the #193 blueprint:
- Provider sends `SubtaskResultsVerify` in the right time window (see #165)
- Concent core responds synchronously if:
    - Additional verification was done for this subtask before (`ServiceRefused`)
    - The request is successfully recognized as an additional verification request but the `SubtaskResultsRejected` message inside it is invalid for any reason, e.g.:
        - The field is empty (`ServiceRefused`).
        - The field has a wrong type (`ServiceRefused`).
        - Data inside the message is contradictory or does not pass our validations (please look at the message structure and see if there are any sensible validations we could do) (`ServiceRefused`).
        - Errors during decoding, such as invalid signature or malformed message can be reported with HTTP 400 if they happen outside of the code that implements the use case (e.g. inside golem-messages or in `api_view`).
    - `SubtaskResultsRejected` was not issued by the requestor (`ServiceRefused`)
    - `SubtaskResultsRejected` was issued for a reason other than negative verification on requestor's part  (`ServiceRefused`)
    - Requestor does not have enough funds to pay for the task in case of positive verification (`ServiceRefused` with `REASON.TooSmallRequestorDeposit`)
- Otherwise Concent accepts the request
- Concent sends `blender_verification_request` message (see #193) to the work queue.

### Detecting duplicate requests
Check subtask state:
1) Subtask state is `VERIFICATION RESULT TRANSFER` or `ADDITIONAL VERIFICATION`: this is a duplicate request. Respond with `ServiceRefused`.
2) Subtask state is `ACCEPTED` or `FAILED`: it might or might not be a duplicate request. Does not matter because we don't allow additional verification in that state. Respond the same way as in any other state that does not allow it.

In case 3) you need more information than just the state. It should be possible to infer it based on what we have in the `Subtask` table but if there are problems with that we'll consider adding extra info to the table.

### Setup for local development
You can assume that Celery is already configured (#197).

For local development you'll need to run a RabbitMQ server. You can run it in a docker container with a single command. See https://github.com/golemfactory/concent/issues/197#issuecomment-375620295. It should work out of the box because `settings/development.py` is configured to connect to a RabbitMQ server on default port running on localhost.

You don't need a running worker for this part of the implementation.

dybi:
@cameel, as agreed, please elaborate a bit more on 

[...]

PR #213, opened 2018-03-25
Feature integrate with smart contract interface

kbeker:
Finished backend with real SCI. Still need to do:
1. Add unit tests to check if it's works properly.
2. Repair old tests for Payment Use Case.
3. Resolve problem with RPC synchronization (Every time SCI needs about 20 seconds to synchronize with geth).
4. Add cluster test.
5. Change variables `amount_paid` and `amount_pending` in `PaymentInfo` from `Decimal` to maybe `IntegerField`? As I investigate amount always will be `integer` value, I'm not sure for this moment so it needs to be cleared up.
6. Also we still waiting for https://github.com/golemfactory/golem-messages/issues/157 to be resolved. So this is our blocker. (I've already added this to Pivotal)

Please verify if this backend do what it should do with payments. 

@cameel Also please check whole Payment Use Case.

PR #251, opened 2018-04-06
Feature accepting additional verification request from provider

dybi:
Resolves https://github.com/golemfactory/concent/issues/201

### DIFF
--- a/concent_api/concent_api/settings/base.py
+++ b/concent_api/concent_api/settings/base.py
@@ -220,6 +220,9 @@ MAXIMUM_DOWNLOAD_TIME = 1800  # 30 minutes
 # A global constant defining the length of the time window within which a requestor can verify subtask.
 SUBTASK_VERIFICATION_TIME = 4 * 3600  # 4 hours
 
+# A global constant defining the length of the time window within which a provider can ask concent for verifying subtask results
+ADDITIONAL_VERIFICATION_CALL_TIME = 4 * 3600  # 4 hours
+
 # A global constant defining the lenght of the time window within which a requestor must pay
 PAYMENT_DUE_TIME = 12 * 3600  # 12 hours
 
--- a/concent_api/core/message_handlers.py
+++ b/concent_api/core/message_handlers.py
@@ -1050,13 +1050,17 @@ def handle_send_subtask_results_verify(
         compute_task_def['subtask_id'],
         'subtask_id'
     )
+    validate_golem_message_task_to_compute(report_computed_task.task_to_compute)
+
     client_public_key = decode_client_public_key(request)
     other_party_public_key = decode_other_party_public_key(request)
 
     if Subtask.objects.filter(
-        Q(subtask_id=compute_task_def['subtask_id']),
-        Q(state=Subtask.SubtaskState.VERIFICATION_FILE_TRANSFER.name) |  # pylint: disable=no-member
-        Q(state=Subtask.SubtaskState.ADDITIONAL_VERIFICATION.name),  # pylint: disable=no-member
+        subtask_id=compute_task_def['subtask_id'],
+        state__in=[
+            Subtask.SubtaskState.VERIFICATION_FILE_TRANSFER.name,
+            Subtask.SubtaskState.ADDITIONAL_VERIFICATION.name
+        ]
     ).exists():
         return message.concents.ServiceRefused(
             reason=message.concents.ServiceRefused.REASON.DuplicateRequest,
--- a/concent_api/core/message_handlers.py
+++ b/concent_api/core/message_handlers.py
@@ -1033,7 +1033,10 @@ def store_message(
     return stored_message
 
 
-def handle_send_subtask_results_verify(request, client_message):
+def handle_send_subtask_results_verify(
+    request,
+    client_message: message.concents.SubtaskResultsVerify
+):
     compute_task_def_subtask_id = client_message.subtask_results_rejected.report_computed_task.task_to_compute.compute_task_def['subtask_id']
     validate_id_value(
         compute_task_def_subtask_id,
@@ -1051,10 +1054,13 @@ def handle_send_subtask_results_verify(request, client_message):
         return message.concents.ServiceRefused(
             reason=message.concents.ServiceRefused.REASON.DuplicateRequest,
         )
-
     if not base.is_requestor_account_status_positive(request):  # pylint: disable=no-value-for-parameter
         return message.concents.ServiceRefused(
             reason=message.concents.ServiceRefused.REASON.TooSmallRequestorDeposit,
+        )
+    if client_message.subtask_results_rejected.reason != message.tasks.SubtaskResultsRejected.REASON.VerificationNegative:
+        return message.concents.ServiceRefused(
+            reason=message.concents.ServiceRefused.REASON.InvalidRequest,
         )
     return None
 
--- a/concent_api/core/message_handlers.py
+++ b/concent_api/core/message_handlers.py
@@ -5,13 +5,11 @@ from base64 import b64encode
 from typing import Optional
 
 from django.conf import settings
-from django.db.models import Q
 from django.http import HttpResponse
 from django.utils import timezone
 
 from golem_messages import message
 from golem_messages.datastructures import MessageHeader
-from golem_messages.message.tasks import SubtaskResultsRejected
 from golem_messages.message import FileTransferToken
 
 from core.exceptions import Http400
@@ -1044,13 +1042,15 @@ def handle_send_subtask_results_verify(
     request,
     subtask_results_verify: message.concents.SubtaskResultsVerify
 ):
-    report_computed_task = subtask_results_verify.subtask_results_rejected.report_computed_task
+    subtask_results_rejected = subtask_results_verify.subtask_results_rejected
+    report_computed_task = subtask_results_rejected.report_computed_task
     compute_task_def = report_computed_task.task_to_compute.compute_task_def
     validate_id_value(
         compute_task_def['subtask_id'],
         'subtask_id'
     )
-    validate_golem_message_task_to_compute(report_computed_task.task_to_compute)
+    validate_task_to_compute(report_computed_task.task_to_compute)
+    validate_golem_message_subtask_results_rejected(subtask_results_rejected)
 
     client_public_key = decode_client_public_key(request)
     other_party_public_key = decode_other_party_public_key(request)
@@ -1058,8 +1058,8 @@ def handle_send_subtask_results_verify(
     if Subtask.objects.filter(
         subtask_id=compute_task_def['subtask_id'],
         state__in=[
-            Subtask.SubtaskState.VERIFICATION_FILE_TRANSFER.name,
-            Subtask.SubtaskState.ADDITIONAL_VERIFICATION.name
+            Subtask.SubtaskState.VERIFICATION_FILE_TRANSFER.name,  # pylint: disable=no-member
+            Subtask.SubtaskState.ADDITIONAL_VERIFICATION.name,     # pylint: disable=no-member
         ]
     ).exists():
         return message.concents.ServiceRefused(
@@ -1069,7 +1069,7 @@ def handle_send_subtask_results_verify(
         return message.concents.ServiceRefused(
             reason=message.concents.ServiceRefused.REASON.TooSmallRequestorDeposit,
         )
-    if subtask_results_verify.subtask_results_rejected.reason != SubtaskResultsRejected.REASON.VerificationNegative:
+    if subtask_results_rejected.reason != SubtaskResultsRejected.REASON.VerificationNegative:
         return message.concents.ServiceRefused(
             reason=message.concents.ServiceRefused.REASON.InvalidRequest,
         )
@@ -1080,10 +1080,10 @@ def handle_send_subtask_results_verify(
         provider_public_key=other_party_public_key,
         requestor_public_key=client_public_key,
         state=Subtask.SubtaskState.VERIFICATION_FILE_TRANSFER,
-        next_deadline=subtask_results_verify.subtask_results_rejected.timestamp + settings.ADDITIONAL_VERIFICATION_CALL_TIME,
+        next_deadline=subtask_results_rejected.timestamp + settings.ADDITIONAL_VERIFICATION_CALL_TIME,
         set_next_deadline=True,
         task_to_compute=report_computed_task.task_to_compute,
-        subtask_results_rejected=subtask_results_verify.subtask_results_rejected
+        subtask_results_rejected=subtask_results_rejected
     )
 
     send_blender_verification_request()
--- a/concent_api/core/message_handlers.py
+++ b/concent_api/core/message_handlers.py
@@ -9,6 +9,7 @@ from django.http import HttpResponse
 from django.utils import timezone
 from golem_messages import message
 from golem_messages.datastructures import MessageHeader
+from golem_messages.message.tasks import SubtaskResultsRejected
 from golem_messages.message import FileTransferToken
 
 from core.exceptions import Http400
@@ -1029,6 +1030,35 @@ def store_message(
     return stored_message
 
 
+def handle_send_subtask_results_verify(request, client_message):
+    compute_task_def_subtask_id = client_message.subtask_results_rejected.report_computed_task.task_to_compute.compute_task_def['subtask_id']
+    validate_id_value(
+        compute_task_def_subtask_id,
+        'subtask_id'
+    )
+    # current_time = get_current_utc_timestamp()
+    # client_public_key = decode_client_public_key(request)
+    # other_party_public_key = decode_other_party_public_key(request)
+
+    if Subtask.objects.filter(
+        Q(subtask_id=compute_task_def_subtask_id),
+        Q(state=Subtask.SubtaskState.VERIFICATION_FILE_TRANSFER.name) |
+        Q(state=Subtask.SubtaskState.ADDITIONAL_VERIFICATION.name),  # pylint: disable=no-member)
+    ).exists():
+        return message.concents.ServiceRefused(
+            reason=message.concents.ServiceRefused.REASON.DuplicateRequest,
+        )
+
+    send_verification_request()
+
+    encoded_client_public_key = b64encode(decode_client_public_key(request))
+    ack_subtask_results_verify = message.concents.AckSubtaskResultsVerify(
+        subtask_results_verify=subtask_results_verify,
+        file_transfer_token=create_file_transfer_token(report_computed_task, encoded_client_public_key, "upload"),
+    )
+    return ack_subtask_results_verify
+
+
 def handle_message(client_message, request):
     if isinstance(client_message, message.ForceReportComputedTask):
         return handle_send_force_report_computed_task(client_message)
@@ -1059,6 +1089,7 @@ def handle_message(client_message, request):
 
     elif isinstance(client_message, message.concents.ForcePayment):
         return handle_send_force_payment(request, client_message)
-
+    elif isinstance(client_message, message.concents.SubtaskResultsVerify):
+        return handle_send_subtask_results_verify(request, client_message)
     else:
         return handle_unsupported_golem_messages_type(client_message)
--- a/concent_api/core/message_handlers.py
+++ b/concent_api/core/message_handlers.py
@@ -5,8 +5,10 @@ from base64 import b64encode
 from typing import Optional
 
 from django.conf import settings
+from django.db.models import Q
 from django.http import HttpResponse
 from django.utils import timezone
+
 from golem_messages import message
 from golem_messages.datastructures import MessageHeader
 from golem_messages.message.tasks import SubtaskResultsRejected
@@ -20,6 +22,7 @@ from core.models import StoredMessage
 from core.models import Subtask
 from core.payments import base
 from core.validation import validate_golem_message_reject
+from core.validation import validate_id_value
 from core.validation import validate_task_to_compute
 from core.validation import validate_report_computed_task_time_window
 from core.validation import validate_list_of_identical_task_to_compute
@@ -1042,21 +1045,18 @@ def handle_send_subtask_results_verify(request, client_message):
 
     if Subtask.objects.filter(
         Q(subtask_id=compute_task_def_subtask_id),
-        Q(state=Subtask.SubtaskState.VERIFICATION_FILE_TRANSFER.name) |
-        Q(state=Subtask.SubtaskState.ADDITIONAL_VERIFICATION.name),  # pylint: disable=no-member)
+        Q(state=Subtask.SubtaskState.VERIFICATION_FILE_TRANSFER.name) |  # pylint: disable=no-member
+        Q(state=Subtask.SubtaskState.ADDITIONAL_VERIFICATION.name),  # pylint: disable=no-member
     ).exists():
         return message.concents.ServiceRefused(
             reason=message.concents.ServiceRefused.REASON.DuplicateRequest,
         )
 
-    send_verification_request()
-
-    encoded_client_public_key = b64encode(decode_client_public_key(request))
-    ack_subtask_results_verify = message.concents.AckSubtaskResultsVerify(
-        subtask_results_verify=subtask_results_verify,
-        file_transfer_token=create_file_transfer_token(report_computed_task, encoded_client_public_key, "upload"),
-    )
-    return ack_subtask_results_verify
+    if not base.is_requestor_account_status_positive(request):  # pylint: disable=no-value-for-parameter
+        return message.concents.ServiceRefused(
+            reason=message.concents.ServiceRefused.REASON.TooSmallRequestorDeposit,
+        )
+    return None
 
 
 def handle_message(client_message, request):
--- a/concent_api/core/message_handlers.py
+++ b/concent_api/core/message_handlers.py
@@ -1066,6 +1066,14 @@ def handle_send_subtask_results_verify(
         return message.concents.ServiceRefused(
             reason=message.concents.ServiceRefused.REASON.DuplicateRequest,
         )
+    if is_message_recieved_in_wrong_state(
+        compute_task_def['subtask_id'],
+        [
+            Subtask.SubtaskState.ACCEPTED.name,  # pylint: disable=no-member
+            Subtask.SubtaskState.FAILED.name,  # pylint: disable=no-member
+        ]
+    ):
+        raise Http400("SubtaskResultsVerify is not allowed in current state")
     if not current_time <= subtask_results_rejected.timestamp + settings.ADDITIONAL_VERIFICATION_CALL_TIME:
         return message.concents.ServiceRefused(
             reason=message.concents.ServiceRefused.REASON.InvalidRequest,
@@ -1135,3 +1143,24 @@ def handle_message(client_message, request):
         return handle_send_subtask_results_verify(request, client_message)
     else:
         return handle_unsupported_golem_messages_type(client_message)
+
+
+def is_signed_by_right_party(
+    subtask_results_rejected: message.tasks.SubtaskResultsRejected,
+    other_party_public_key: bytes
+) -> bool:
+    try:
+        validate_golem_message_signed_with_key(
+            subtask_results_rejected,
+            other_party_public_key,
+        )
+        return True
+    except Http400:
+        return False
+
+
+def is_message_recieved_in_wrong_state(subtask_id, forbidden_states):
+    return Subtask.objects.filter(
+        subtask_id=subtask_id,
+        state__in=forbidden_states
+    ).exists()
--- a/concent_api/core/message_handlers.py
+++ b/concent_api/core/message_handlers.py
@@ -24,8 +24,10 @@ from core.queue_operations import send_blender_verification_request
 from core.subtask_helpers import verify_message_subtask_results_accepted
 from core.transfer_operations import create_file_transfer_token
 from core.transfer_operations import store_pending_message
-from core.validation import validate_report_computed_task_time_window, validate_golem_message_subtask_results_rejected
 from core.validation import validate_golem_message_reject
+from core.validation import validate_golem_message_subtask_results_rejected
+from core.validation import validate_golem_message_signed_with_key
+from core.validation import validate_golem_message_task_to_compute
 from core.validation import validate_id_value
 from golem_messages.message.tasks import SubtaskResultsRejected
 from core.validation import validate_task_to_compute
@@ -35,6 +37,7 @@ from core.validation import validate_golem_message_signed_with_key
 from core.subtask_helpers import verify_message_subtask_results_accepted
 from core.transfer_operations import store_pending_message
 from core.transfer_operations import create_file_transfer_token
+from core.validation import validate_report_computed_task_time_window
 from utils import logging
 from utils.helpers import deserialize_message
 from utils.helpers import get_current_utc_timestamp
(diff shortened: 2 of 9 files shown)


concent_api/concent_api/settings/base.py
```diff
@@ -220,6 +220,8 @@
 # A global constant defining the length of the time window within which a requestor can verify subtask.
 SUBTASK_VERIFICATION_TIME = 4 * 3600  # 4 hours
 
+ADDITIONAL_VERIFICATION_CALL_TIME = 4 * 3600  # 4 hours
```
rwrzesien: Please add comment above, as there is in every other Concet custom setting.

concent_api/core/message_handlers.py
```diff
@@ -1,33 +1,37 @@
...
-from core.subtask_helpers           import verify_message_subtask_results_accepted
-from core.transfer_operations       import store_pending_message
-from core.transfer_operations       import create_file_transfer_token
-from utils                          import logging
-from utils.helpers                  import decode_client_public_key
-from utils.helpers                  import decode_other_party_public_key
-from utils.helpers                  import deserialize_message
-from utils.helpers                  import get_current_utc_timestamp
-from utils.helpers                  import parse_timestamp_to_utc_datetime
+import datetime
```
rwrzesien: Didn't we agreed to keep this formatting for imports ?

concent_api/core/message_handlers.py
```diff
@@ -954,6 +958,58 @@ def store_message(
     return stored_message
 
 
+def handle_send_subtask_results_verify(
+    request,
+    subtask_results_verify: message.concents.SubtaskResultsVerify
+):
+    report_computed_task = subtask_results_verify.subtask_results_rejected.report_computed_task
+    compute_task_def = report_computed_task.task_to_compute.compute_task_def
+    validate_id_value(
```
rwrzesien: Please add validation for:
1. task_id
2. task_to_compute

concent_api/core/message_handlers.py
```diff
@@ -954,6 +958,58 @@ def store_message(
...
+        compute_task_def['subtask_id'],
+        'subtask_id'
+    )
+    client_public_key = decode_client_public_key(request)
+    other_party_public_key = decode_other_party_public_key(request)
+
+    if Subtask.objects.filter(
+        Q(subtask_id=compute_task_def['subtask_id']),
```
rwrzesien: I think `Q` is not needed in this line. Generally this can be rewritten to:
```
Subtask.objects.filter(
    subtask_id=compute_task_def['subtask_id'],
    Q(state__in=[
        Subtask.SubtaskState.VERIFICATION_FILE_TRANSFER.name,
        Subtask.SubtaskState.ADDITIONAL_VERIFICATION.name
    ]
).exists()
```

concent_api/core/message_handlers.py
```diff
@@ -954,6 +958,58 @@ def store_message(
...
+        Q(state=Subtask.SubtaskState.VERIFICATION_FILE_TRANSFER.name) |  # pylint: disable=no-member
+        Q(state=Subtask.SubtaskState.ADDITIONAL_VERIFICATION.name),  # pylint: disable=no-member
+    ).exists():
```
rwrzesien: I think this condition only checks if Subtask is currently in verification. In issue there is 

> Additional verification was done for this subtask before (ServiceRefused)

So I think there should be different check if verification was already done (please ask @cameel I don't really know how to check this). Or update issue description if it is wrong.

concent_api/concent_api/settings/base.py

concent_api/core/payments/base.py
```diff
@@ -64,3 +64,8 @@ def is_provider_account_status_positive(backend, request = None):
 @_add_backend
 def calculate_amount_pending(backend):
     return backend.calculate_amount_pending()
+
+
+@_add_backend
+def is_requestor_account_status_positive(backend, request):
```
rwrzesien: This function will be added in https://github.com/golemfactory/concent/pull/213 but called `is_provider_account_status_positive`. But actually it uses `requestor_eth_address` :) We should make it independent, so please just call it `is_account_status_positive`, I will add comment to the @kbeker PR.

concent_api/core/queue_operations.py
```diff
@@ -0,0 +1,2 @@
+def send_verification_request():  # TODO implement sending real verivication_request to Work Queue
```
rwrzesien: Is there going to be something more or this is the part when Conductor starts ?

concent_api/core/tests/test_integration_subtask_results_verify.py
```diff
@@ -0,0 +1,228 @@
...
+            timestamp=subtask_result_rejected_time_str,
+            report_computed_task=self.report_computed_task)
+        subtask_results_verify_time_str = self._add_time_offset_to_date(subtask_result_rejected_time_str,
+                                                                        settings.ADDITIONAL_VERIFICATION_CALL_TIME / 2)
+        subtask_results_verify = self._get_deserialized_subtask_results_verify(
+            timestamp=subtask_results_verify_time_str,
+            subtask_results_rejected=subtask_results_rejected)
+        return subtask_results_verify, subtask_results_verify_time_str
+
+    def _create_report_computed_task(self):
```
rwrzesien: I think there is a function in test utils which already does that.

concent_api/core/tests/test_integration_subtask_results_verify.py
```diff
@@ -0,0 +1,228 @@
...
+            subtask_id=self.subtask_id
+        )
+        report_computed_task = self._get_deserialized_report_computed_task(
+            timestamp="2018-04-01 10:01:00",
+            subtask_id=self.subtask_id,
+            task_to_compute=task_to_compute
+        )
+        return report_computed_task
+
+    def _add_time_offset_to_date(self, base_time, offset):
```
rwrzesien: This looks very reusable, please move it to `ConcentIntegrationTestCase`

concent_api/core/tests/test_integration_subtask_results_verify.py
```diff
@@ -0,0 +1,228 @@
...
+)
+class SubtaskResultsVerifyIntegrationTest(ConcentIntegrationTestCase):
+    def setUp(self):
+        super().setUp()
+        self.task_id = "task_1"
+        self.subtask_id = "subtask_1"
+        self.report_computed_task = self._create_report_computed_task()
+
+    def test_that_concent_responds_with_service_refused_when_verification_for_this_subtask_done_before(self):
+        # given
```
rwrzesien: We have a non-written convention to describe message exchange in test comment. In several cases it gives very quick view of what the test actually does, please add them, check in other test_integration_*.

concent_api/core/tests/test_integration_subtask_results_verify.py
```diff
@@ -0,0 +1,228 @@
...
+            response,
+            status=200,
+            key=self.PROVIDER_PRIVATE_KEY,
+            message_type=message.concents.ServiceRefused,
+            fields={
+                'reason': message.concents.ServiceRefused.REASON.InvalidRequest,
+            }
+        )
+        self._assert_stored_message_counter_not_increased()
+
```
rwrzesien: Looks good, just remember to add more tests when more checks are added in view.

concent_api/core/message_handlers.py
```diff
@@ -954,6 +958,58 @@ def store_message(
...
+        state=Subtask.SubtaskState.VERIFICATION_FILE_TRANSFER,
+        next_deadline=subtask_results_verify.subtask_results_rejected.timestamp + settings.ADDITIONAL_VERIFICATION_CALL_TIME,
+        set_next_deadline=True,
+        task_to_compute=report_computed_task.task_to_compute,
+        subtask_results_rejected=subtask_results_verify.subtask_results_rejected
+    )
+
+    send_verification_request()
+    
+    encoded_client_public_key = b64encode(decode_client_public_key(request))
```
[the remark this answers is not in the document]
dybi: it will `request['HTTP_CONCENT_CLIENT_KEY']` as `bytes`, ie:
`'ifLken5xe8tIgHerZqk0q8pzHYFsFRIwLaX+xSQeIeeYk5Bf9vOKBvJoNeeaiyTfaVso5ToBPPzzegFycrFY1A=='` -> 
`b'ifLken5xe8tIgHerZqk0q8pzHYFsFRIwLaX+xSQeIeeYk5Bf9vOKBvJoNeeaiyTfaVso5ToBPPzzegFycrFY1A=='`

concent_api/concent_api/settings/base.py
rwrzesien: I don't see it :) Probably you still have it locally. That's why I am doing two things:
1. :+1: - for marking it for myself as done, or 'accepted/agreed to do'. When you scroll through comments you can easily see visually which one you have already resolved.
2. Add 'done' when it is pushed to github.

concent_api/core/message_handlers.py
rwrzesien: Ok, I just thought we will keep them for imports.

concent_api/core/tests/test_integration_subtask_results_verify.py
dybi: :+1:

concent_api/core/tests/test_integration_subtask_results_verify.py
dybi: done

concent_api/core/tests/test_integration_subtask_results_verify.py
dybi: I am using it in my function - here `_create_report_computed_task` is used for easier setup

concent_api/core/queue_operations.py
dybi: this is the part where Conductor starts, so when your part is in `master`, I will create and send a real `verification_request` here ;)

concent_api/core/payments/base.py
dybi: ok, I have added `TODO` not to forget to change it once `concent_rpc` becomes available

concent_api/core/message_handlers.py
dybi: done

concent_api/core/message_handlers.py
dybi: done;
'task_id' is checked within `validate_golem_message_task_to_compute(r` -> I have added:
`validate_golem_message_task_to_compute(report_computed_task.task_to_compute)`

concent_api/core/tests/test_integration_subtask_results_verify.py
dybi: done

concent_api/core/message_handlers.py
```diff
@@ -954,6 +958,58 @@ def store_message(
...
+        )
+    if not base.is_requestor_account_status_positive(request):  # pylint: disable=no-value-for-parameter
+        return message.concents.ServiceRefused(
+            reason=message.concents.ServiceRefused.REASON.TooSmallRequestorDeposit,
+        )
+    if subtask_results_verify.subtask_results_rejected.reason != SubtaskResultsRejected.REASON.VerificationNegative:
+        return message.concents.ServiceRefused(
+            reason=message.concents.ServiceRefused.REASON.InvalidRequest,
+        )
+
```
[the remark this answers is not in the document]
dybi: done;
additional validation added

concent_api/core/validation.py
```diff
@@ -131,3 +135,9 @@ def validate_golem_message_signed_with_key(
                 public_key
             )
         )
+
+
+def validate_golem_message_subtask_results_rejected(subtask_results_rejected : message.tasks.SubtaskResultsRejected):
```
cameel: The type annotation is wrong. This function handles types other than `SubtaskResultsRejected` too (by raising `Http400`).

concent_api/core/tests/utils.py
```diff
@@ -569,6 +567,32 @@ def _get_serialized_force_payment(
...
+
+    def _get_serialized_subtask_results_verify(
+        self,
+        timestamp=None,
+        subtask_results_verify=None,
+        provider_private_key=None
+    ):
+        return dump(
+            subtask_results_verify or self._get_deserialized_subtask_results_verify(timestamp),
+            provider_private_key or self.PROVIDER_PRIVATE_KEY,
```
cameel: These are not boolean values. Please compare them explicitly.

concent_api/core/message_handlers.py
```diff
@@ -954,6 +957,69 @@ def store_message(
...
+        subtask_id=compute_task_def['subtask_id'],
+        state__in=[
+            Subtask.SubtaskState.VERIFICATION_FILE_TRANSFER.name,  # pylint: disable=no-member
+            Subtask.SubtaskState.ADDITIONAL_VERIFICATION.name,     # pylint: disable=no-member
+        ]
+    ).exists():
+        return message.concents.ServiceRefused(
+            reason=message.concents.ServiceRefused.REASON.DuplicateRequest,
+        )
+    if current_time > subtask_results_rejected.timestamp + settings.ADDITIONAL_VERIFICATION_CALL_TIME:
```
rwrzesien: Please reverse the condition.

concent_api/core/queue_operations.py
```diff
@@ -0,0 +1,2 @@
+def send_blender_verification_request():  # TODO implement sending real verivication_request to Work Queue
+    pass
```
cameel: Sending a verification request is also a part of this task. There's no issue for adding it so if you don't do it here, no one else will.

concent_api/core/payments/base.py
```diff
@@ -64,3 +64,8 @@ def is_provider_account_status_positive(backend, request = None):
 @_add_backend
 def calculate_amount_pending(backend):
     return backend.calculate_amount_pending()
+
+
+@_add_backend
+def is_requestor_account_status_positive(backend, request):  # TODO use real concent_rpc when available
+    return backend.is_requestor_account_status_positive(request)
```
cameel: We need to ensure that someone remembers to come back and add this. If you can't implement something because a part you depend on is missing I think you should add an issue to actually use that part when it's finally implemented. Or at least add it in the description of the issue that adds this part.

concent_api/core/message_handlers.py
cameel: I would also be for keeping it for imports but unfortutely I don't think PyCharm can handle that.

concent_api/core/message_handlers.py
```diff
@@ -954,6 +957,69 @@ def store_message(
...
+
+    client_public_key = decode_client_public_key(request)
+    other_party_public_key = decode_other_party_public_key(request)
+    current_time = get_current_utc_timestamp()
+
+    if Subtask.objects.filter(
+        subtask_id=compute_task_def['subtask_id'],
+        state__in=[
+            Subtask.SubtaskState.VERIFICATION_FILE_TRANSFER.name,  # pylint: disable=no-member
+            Subtask.SubtaskState.ADDITIONAL_VERIFICATION.name,     # pylint: disable=no-member
```
cameel: @rwrzesien I thought we solved the problem with `no-member` popping up everywhere where we use an enum? If not we'll need to add `Subtask.SubtaskState` to the list of classes for which pylint should not check members. Disabled warnings should be a rare exeption rather than a common occurrence.

concent_api/core/message_handlers.py
```diff
@@ -954,6 +957,69 @@ def store_message(
...
+        task_to_compute=report_computed_task.task_to_compute,
+        subtask_results_rejected=subtask_results_rejected
+    )
+
+    send_blender_verification_request()
+
+    encoded_client_public_key = b64encode(decode_client_public_key(request))
+    ack_subtask_results_verify = message.concents.AckSubtaskResultsVerify(
+        subtask_results_verify=subtask_results_verify,
+        file_transfer_token=create_file_transfer_token(report_computed_task, encoded_client_public_key, "upload"),
```
cameel: golem-messages has recently introduced an enum for `"download"` and `"upload"`. Might be a good idea to start using it.

https://github.com/golemfactory/golem-messages/blob/master/golem_messages/message/concents.py#L161-L164

concent_api/core/message_handlers.py
```diff
@@ -954,6 +957,69 @@ def store_message(
...
+        'subtask_id'
+    )
+    validate_golem_message_task_to_compute(report_computed_task.task_to_compute)
+    validate_golem_message_subtask_results_rejected(subtask_results_rejected)
+
+    client_public_key = decode_client_public_key(request)
+    other_party_public_key = decode_other_party_public_key(request)
+    current_time = get_current_utc_timestamp()
+
+    if Subtask.objects.filter(
```
cameel: State transitions are not being validated here. For example if a client sends this message when state is `FORCING REPORT` the handler will proceed and fail only on model validations. This will result in HTTP 500. There are validations in the model but they should be treated like a last line of defense or sanity check. Views need to check for this before we even get to these validations.

cameel:
I don't think this is implemented:

> - Concent core responds synchronously if:
>     (...)
>     - `SubtaskResultsRejected` was not issued by the requestor (`ServiceRefused`)

`SubtaskResultsRejected` must be signed separately from the message that contains it and be signed with a key that belongs to requestor (not the client, who is submitting the request).

concent_api/core/message_handlers.py
cameel: I have just added a section that describes how to deal with duplicate requests: https://github.com/golemfactory/concent/issues/201#issue-308021944

We talked about it today but I realized that there are more cases so I have decided to include it in the issue.

concent_api/core/message_handlers.py
```diff
@@ -954,6 +958,58 @@ def store_message(
...
+        Q(subtask_id=compute_task_def['subtask_id']),
+        Q(state=Subtask.SubtaskState.VERIFICATION_FILE_TRANSFER.name) |  # pylint: disable=no-member
+        Q(state=Subtask.SubtaskState.ADDITIONAL_VERIFICATION.name),  # pylint: disable=no-member
+    ).exists():
+        return message.concents.ServiceRefused(
+            reason=message.concents.ServiceRefused.REASON.DuplicateRequest,
+        )
+    if not base.is_requestor_account_status_positive(request):  # pylint: disable=no-value-for-parameter
+        return message.concents.ServiceRefused(
+            reason=message.concents.ServiceRefused.REASON.TooSmallRequestorDeposit,
```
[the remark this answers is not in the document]
dybi: done;
it should be `TooSmallRequestorDeposit`; changed the specs.

concent_api/core/message_handlers.py
dybi: do you mean:
`if not subtask_results_rejected.timestamp + settings.ADDITIONAL_VERIFICATION_CALL_TIME <= current_time` ?

concent_api/core/validation.py
dybi: it is not. I have created this function, so I know what argument it is supposed to take ;)
It is for checking `SubtaskResultsRejected`

concent_api/core/validation.py
```diff
@@ -66,6 +66,10 @@ def validate_golem_message_task_to_compute(golem_message: message.base.Message):
     if not isinstance(golem_message, message.TaskToCompute):
         raise Http400("Expected TaskToCompute.")
 
+    if not all(getattr(golem_message, attribute) for attribute in ['compute_task_def',
+                                                                   'provider_public_key',
+                                                                   'requestor_public_key']):
```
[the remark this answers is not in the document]
dybi: done

concent_api/core/tests/utils.py
dybi: this conventione is widely used throughout the rest of the code.

concent_api/core/payments/base.py
dybi: done;
issue: https://github.com/golemfactory/concent/issues/255

concent_api/core/message_handlers.py
dybi: Since it is used in many places, there should be a sparate issue for it.

concent_api/core/message_handlers.py
dybi: Again, as discussed yesterday, there are no transition checks at all in the `views` (handling functions). There should be a separate story/issue for enhancing message handlers with state transtion checks.

concent_api/core/validation.py
cameel: It's a function that validates data received from the client. You cannot guarantee that the client will send you only `SubtaskResultsRejected` and nothing else. So it must handle other types too and the annotation is wrong (it states that the function takes only this type).

If you could assume that `subtask_results_rejected` is always of this type then this `if` would have to be an `assert`:

``` python
if not isinstance(subtask_results_rejected,  message.tasks.SubtaskResultsRejected):
    raise Http400("subtask_results_rejected should be of type:  SubtaskResultsRejected")
```

concent_api/core/message_handlers.py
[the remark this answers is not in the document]
dybi: done

concent_api/core/queue_operations.py
dybi: separate issue: https://github.com/golemfactory/concent/issues/271

concent_api/core/message_handlers.py
```diff
@@ -1019,6 +1025,82 @@ def store_message(
...
 
 
+def handle_send_subtask_results_verify(
+    request,
+    subtask_results_verify: message.concents.SubtaskResultsVerify
+):
+    subtask_results_rejected = subtask_results_verify.subtask_results_rejected
+    report_computed_task = subtask_results_rejected.report_computed_task
+    compute_task_def = report_computed_task.task_to_compute.compute_task_def
+    validate_id_value(
```
rwrzesien: Wouldn't it be better to use validation for task_to_compute, which already has this check plus few more ?

concent_api/core/message_handlers.py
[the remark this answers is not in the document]
rwrzesien: Where is it ?

concent_api/core/message_handlers.py
```diff
@@ -1019,6 +1025,82 @@ def store_message(
...
+        compute_task_def['subtask_id'],
+        'subtask_id'
+    )
+    validate_golem_message_subtask_results_rejected(subtask_results_rejected)
+    validate_golem_message_signed_with_key(
+        report_computed_task.task_to_compute,
+        report_computed_task.task_to_compute.requestor_public_key,
+
```
rwrzesien: Please remove blank line.

concent_api/core/tests/utils.py
rwrzesien: Where ? :) We are actually trying to avoid this.

concent_api/core/message_handlers.py
```diff
@@ -1019,6 +1025,82 @@ def store_message(
...
+        return message.concents.ServiceRefused(
+            reason=message.concents.ServiceRefused.REASON.DuplicateRequest,
+        )
+    if is_subtask_results_verified_recieved_in_wrong_state(compute_task_def['subtask_id']):
+        raise Http400("SubtaskResultsVerify is not allowed in current state")
+    if not current_time <= subtask_results_rejected.timestamp + settings.ADDITIONAL_VERIFICATION_CALL_TIME:
+        return message.concents.ServiceRefused(
+            reason=message.concents.ServiceRefused.REASON.InvalidRequest,
+        )
+    if not is_signed_by_right_party(
```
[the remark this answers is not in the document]
dybi: @rwrzesien, I agree and I do not like this solution either. However it strictly implements specification:
>    SubtaskResultsRejected was not issued by the requestor (ServiceRefused)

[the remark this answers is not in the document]
rwrzesien: Then it is ok for me, but I don't think extra function for it is necessary.

concent_api/core/message_handlers.py
dybi: it was there, I must have lost it during rebase :/, will fix

dybi: EDIT: Actually I forgot I have refactoed it and now `validate_golem_message_subtask_results_rejected` does it inside

concent_api/core/message_handlers.py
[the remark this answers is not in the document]
dybi: I guess we will need to refactor it one day (and make it cohesive among all handlers/use cases) ;)

concent_api/core/message_handlers.py
```diff
@@ -1019,6 +1025,82 @@ def store_message(
...
+        provider_public_key,
+    ):
+        return message.concents.ServiceRefused(
+            reason=message.concents.ServiceRefused.REASON.InvalidRequest,
+        )
+    if not base.is_requestor_account_status_positive(request):  # pylint: disable=no-value-for-parameter
+        return message.concents.ServiceRefused(
+            reason=message.concents.ServiceRefused.REASON.TooSmallRequestorDeposit,
+        )
+    if subtask_results_rejected.reason != SubtaskResultsRejected.REASON.VerificationNegative:
```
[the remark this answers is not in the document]
dybi: good call! :) done

concent_api/core/message_handlers.py
```diff
@@ -1019,6 +1025,82 @@ def store_message(
     return stored_message
 
 
+def handle_send_subtask_results_verify(
+    request,
+    subtask_results_verify: message.concents.SubtaskResultsVerify
+):
+    subtask_results_rejected = subtask_results_verify.subtask_results_rejected
```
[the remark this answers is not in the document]
dybi: done

concent_api/core/message_handlers.py
```diff
@@ -1049,6 +1131,31 @@ def handle_message(client_message, request):
...
     elif isinstance(client_message, message.concents.ForcePayment):
         return handle_send_force_payment(request, client_message)
-
+    elif isinstance(client_message, message.concents.SubtaskResultsVerify):
+        return handle_send_subtask_results_verify(request, client_message)
     else:
         return handle_unsupported_golem_messages_type(client_message)
+
+
+def is_signed_by_right_party(
```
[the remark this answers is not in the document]
dybi: please see my naswer above

concent_api/core/message_handlers.py
```diff
@@ -1049,6 +1131,31 @@ def handle_message(client_message, request):
...
+        validate_golem_message_signed_with_key(
+            subtask_results_rejected,
+            other_party_public_key,
+        )
+        return True
+    except Http400:
+        return False
+
+
+def is_subtask_results_verified_recieved_in_wrong_state(subtask_id):
```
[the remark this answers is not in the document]
dybi: I extracted it for the sake of readability/clarity

concent_api/core/message_handlers.py
```diff
@@ -1030,53 +1028,49 @@ def handle_send_subtask_results_verify(
...
+    requestor_public_key = task_to_compute.requestor_public_key
+    provider_public_key = task_to_compute.provider_public_key
+    current_time = get_current_utc_timestamp()
+
     validate_golem_message_subtask_results_rejected(subtask_results_rejected)
     validate_golem_message_signed_with_key(
-        report_computed_task.task_to_compute,
-        report_computed_task.task_to_compute.requestor_public_key,
+        task_to_compute,
+        task_to_compute.requestor_public_key,
```
rwrzesien: Please use `requestor_public_key` directly.

concent_api/core/message_handlers.py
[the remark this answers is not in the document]
rwrzesien: Do you mean we should use `is_signed_by_right_party` in other places or remove it from here ?

concent_api/core/message_handlers.py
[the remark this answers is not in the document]
rwrzesien: It would make more sense if you can pass a list of states into that function. Then it could be reused in several places.

concent_api/core/tests/utils.py
rwrzesien: @dybi What about this one ?

concent_api/core/message_handlers.py
dybi: good point :); done

concent_api/core/tests/utils.py
rwrzesien: @dybi

concent_api/core/message_handlers.py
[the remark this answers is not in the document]
rwrzesien: @dybi

concent_api/core/message_handlers.py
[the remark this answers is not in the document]
dybi: or else ;) I think we should try to move this kind of verification to some generic decorator or sth like that

concent_api/core/message_handlers.py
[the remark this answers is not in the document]
dybi: `VERIFICATION_FILE_TRANSFER` and `ADDITIONAL_VERIFICATION` -> *duplicated*
`ACCEPTED` and `FAILED` -> *wrong state* (`Http400`)

concent_api/core/tests/utils.py
dybi: in `ConcentIntegrationTestCase` class. Actually, I see we use two patterns:
```python
    def _get_serialized_force_get_task_result(
        self,
        report_computed_task,
        timestamp,
        requestor_private_key = None
    ):
        """ Returns MessageForceGetTaskResult serialized. """
        with freeze_time(timestamp):
            force_get_task_result = message.concents.ForceGetTaskResult(
                report_computed_task    = report_computed_task,
            )
        return dump(
            force_get_task_result,
            requestor_private_key or self.REQUESTOR_PRIVATE_KEY,
            settings.CONCENT_PUBLIC_KEY
        )
```
or
```python
    def _sign_message(self, golem_message, client_private_key = None):
        return sign_message(
            golem_message,
            self.REQUESTOR_PRIVATE_KEY if client_private_key is None else client_private_key,
        )
```
I am fine with both and would use the one which is more readible in given place.
@rwrzesien, what's you opinion?

rwrzesien: We should compare explicitly if value is not bool, please fix the place you have found too.

concent_api/core/message_handlers.py
rwrzesien: We fixed it everywhere apart from queries.
@dybi can you add it as part of this pull request ?

concent_api/core/message_handlers.py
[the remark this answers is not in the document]
rwrzesien: I still think that you should put the code of this function directly here for better readability and less complexity but if you want to keep that way it is ok.

concent_api/core/message_handlers.py
```diff
@@ -1029,6 +1033,85 @@ def store_message(
...
+
+    requestor_public_key = task_to_compute.requestor_public_key
+    provider_public_key = task_to_compute.provider_public_key
+    current_time = get_current_utc_timestamp()
+
+    validate_golem_message_subtask_results_rejected(subtask_results_rejected)
+    validate_golem_message_signed_with_key(
+        task_to_compute,
+        requestor_public_key,
+
```
rwrzesien: Please remove blank line.

concent_api/core/message_handlers.py
```diff
@@ -1029,6 +1033,85 @@ def store_message(
...
+        task_to_compute=task_to_compute,
+        subtask_results_rejected=subtask_results_rejected
+    )
+
+    send_blender_verification_request()
+
+    encoded_client_public_key = b64encode(provider_public_key)
+    ack_subtask_results_verify = message.concents.AckSubtaskResultsVerify(
+        subtask_results_verify=subtask_results_verify,
+        file_transfer_token=create_file_transfer_token(subtask_results_rejected.report_computed_task,
```
rwrzesien: Please fix styling.

concent_api/core/message_handlers.py
Jakub89: Done.

concent_api/core/message_handlers.py
Jakub89: Done.

Issue #210, opened 2018-03-23
[Authentication] Changes in api_view


rwrzesien:
With the implementation of #138 there must be following changes in api_view tests:

* Remove loading of golem_message from api_view and handling of related exceptions. 
* Remove validation of `HTTP_CONCENT_CLIENT_PUBLIC_KEY` from `api_view`.

In tests:
* Remove usage of 'HTTP_CONCENT_CLIENT_PUBLIC_KEY' and 'HTTP_CONCENT_OTHER_PARTY_PUBLIC_KEY' from all core tests.
* Rewrite all tests which test usage of 'HTTP_CONCENT_CLIENT_PUBLIC_KEY', as this will now happen differently.
* Rewrite all tests which test handling exceptions in loading golem message in api_view, as this will now happen in differently in different views.

PR #254, opened 2018-04-09
[WIP] Feature full authentication changes in api view

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/210


concent_api/core/views.py
```diff
@@ -36,7 +36,7 @@ def send(request, client_message, client_public_key):
 @require_golem_auth_message
 @handle_errors_and_responses
 @require_POST
-def receive(_request, message):
+def receive(_request, message, _client_public_key):
```
dybi: I see that in `receive` and `receive_out_of_band` endpoints,`client_public_key` from message is used to update subtasks, whereas in `send`, the one that comes from wrapper. What is the reason for that?

Issue #256, opened 2018-04-10
`store_subtask` fails when receives `next_deadline=None`


dybi:
Function `store_subtask` should be capable of handling tasks in PASSIVE states (there is an assert:
`assert (state in Subtask.PASSIVE_STATES) == (next_deadline is None)`). However, when `next_deadline is None`, function failes at: `next_deadline   = parse_timestamp_to_utc_datetime(next_deadline),`

PR #257, opened 2018-04-10
Fix store_subtask to correctly handle None as next_deadline

dybi:
Fixes: https://github.com/golemfactory/concent/issues/256

### DIFF
--- a/concent_api/core/message_handlers.py
+++ b/concent_api/core/message_handlers.py
@@ -2,6 +2,8 @@ import datetime
 
 import copy
 from base64                         import b64encode
+from typing import Optional
+
 from django.conf                    import settings
 from django.http                    import HttpResponse
 from django.utils                   import timezone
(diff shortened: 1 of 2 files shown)


concent_api/core/message_handlers.py
```diff
@@ -566,7 +566,7 @@ def store_subtask(
         provider        = provider,
         requestor       = requestor,
         state           = state.name,
-        next_deadline   = parse_timestamp_to_utc_datetime(next_deadline),
+        next_deadline   = next_deadline and parse_timestamp_to_utc_datetime(next_deadline),
```
rwrzesien: In this case it will also fail (because `next_deadline` will be passed to `parse_timestamp_to_utc_datetime`) for ie. `True`, `[0]`etc. Please:
1. Add assert, that if next_deadline is not None, it is int.
2. Change function signature typing for `next_deadline` to be `Union[int, None]`.
3. Change the condition to check explicitly if `isinstance(next_deadline, int) parse_timestamp_to_utc_datetime(next_deadline) else None`

dybi: as discussed:
ad 1) assert will be changed (in `models` as well)
ad 2) I will use `Optional[int]`
ad 3) will do! ;)

dybi: done

PR #259, opened 2018-04-10
Fix bug authorized client public key in file transfer token

kbeker:


concent_api/gatekeeper/views.py
```diff
@@ -36,6 +36,7 @@
 def upload(request):
     logger.debug("Upload request received.")
     if request.content_type != 'application/x-www-form-urlencoded':
+        logger.info(f'Unsupported content type: {request.content_type}')
```
dybi: :+1: for using f-strings ;)

concent_api/core/transfer_operations.py
```diff
@@ -38,7 +38,7 @@ def verify_file_status(
         report_computed_task    = deserialize_message(get_task_result.report_computed_task.data.tobytes())
         file_transfer_token     = create_file_transfer_token(
             report_computed_task,
-            encoded_client_public_key,
+            client_public_key,
```
dybi: are you sure, that `client_public_key` should be here in not-encoded format?

kbeker: It should be here client_public_key giving in `bytes` because after it `FileTransferToken` must have field `authorized_client_public_key` with public key in bytes

PR #260, opened 2018-04-11
Change gatekeeper content_type to 'multipart/form-data'

kbeker:


concent_api/gatekeeper/tests/test_gatekeeper_views.py
```diff
@@ -45,7 +45,7 @@ def test_upload_should_accept_valid_message(self):
                 'blender/benchmark/test_task/scene-Helicopter-27-cycles.blend'
             ),
             HTTP_AUTHORIZATION = 'Golem ' + encrypted_token,
-            content_type = 'application/x-www-form-urlencoded',
+            content_type = 'multipart/form-data',
```
rwrzesien: Shouldn't this be added to other tests ? They may pass but we just check `400` response without checking error messages, so there is a chance all of them are returning `400` because of missing this, even if they should return `400` from different reason. Please check.

concent_api/gatekeeper/views.py
```diff
@@ -35,7 +35,7 @@
 @require_POST
 def upload(request):
     logger.debug("Upload request received.")
-    if request.content_type != 'application/x-www-form-urlencoded':
+    if request.content_type != 'multipart/form-data':
```
bartoszbetka: I think this should like this `if request.content_type == 'multipart/form-data'` or `if request.content_type != 'application/octet-stream'`

kbeker: Done

concent_api/gatekeeper/tests/test_gatekeeper_views.py
kbeker: Checked. Everything is ok :)

Issue #262, opened 2018-04-11
Change golem_messages version to 2.3.1


rwrzesien:
Change golem_messages version to 2.3.1 and make sure all tests pass.

cameel:
Also update code to handle this change in the spec: https://github.com/golemfactory/golem-messages/issues/172

PR #263, opened 2018-04-11
Chore change golem messages from v2.0.0 to v2.3.1

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/262.


kbeker:
I think we should refactor some utils helpers to enforce them making already full working messages and then change just variables in tests. In may be much helpful if in the future will be some new changes in protocol. But it should be separate issue.

rwrzesien:
@dybi Have you reinstalled requirements ?

PR #266, opened 2018-04-12
Add csrf_exempt to decorators

kbeker:


Jakub89:
Because api-integration-tests are responding with `Forbidden (CSRF cookie not set.): /api/v1/receive/`

cameel:
@rwrzesien See my response on Slack. Removing the middleware would disable CSRF in the admin panel.

rwrzesien:
Ok, makes sense, but those two decorators are meant to be used together, so with this solution we will have something like:
```
csrf_exempt
require_golem_auth_message
csrf_exempt
handle_errors_and_responses
```

and it makes this tightly coupled too. I would suggest using `csrf_exempt` directly on the views on which we want it to be disabled.

cameel:
@rwrzesien I'm fine with doing it either way as long as it's not disabled globally.

I agree that having a `csrf_sandwich` is superfluous. Since we don't have many views putting it on every one would not be a problem in my opinion.

Issue #207, opened 2018-03-23
Conductor app and database configuration on the storage cluster


cameel:
This task implements a fragment of the #193 blueprint

### Conductor app
- Create a new app within the `concent_api` project, called `conductor`.
- Add support for a new value in `settings.CONCENT_FEATURES`: `'conductor-urls'`. `conductor` app should include its URLs in the main urlconfig only if `'conductor-urls'` is present in that setting (see how it's currently done for `gatekeeper`).
- Create database models listed in #193.

### Database configuration
Conductor's models must be stored in a database separate from the one the rest of Concent uses. For now both databases will be on the same server but eventually each storage cluster will have its own.

Read [Multiple Databases](https://docs.djangoproject.com/en/2.0/topics/db/multi-db/) in Django docs. Then:
- Configure database routing so that every model is stored in the right database
- Add a database called `storage` to `settings.DATABASES`.
- Rember that each database needs to be migrated separately (separate call to `manage.py migrate` with the `--database` parameter).

cameel:
Update: models have changed in the blueprint.

cameel:
Update: `'conductor-api'` in `settings.CONCENT_FEATURES` renamed to `'conductor-urls'`.

PR #217, opened 2018-03-26
Feature verification conductor

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/207.
Resolves https://github.com/golemfactory/concent/issues/208.
Resolves https://github.com/golemfactory/concent/issues/224.

PR #268, opened 2018-04-12
[Verification 1] Feature verification conductor app and database configuration on storage cluster

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/207.

Was a part of https://github.com/golemfactory/concent/pull/217


concent_api/conductor/models.py
```diff
@@ -0,0 +1,77 @@
...
+                'result_package_path': 'result_package_path cannot be used as source_package_path in any other VerificationRequest.'
+            })
+
+
+class BlenderSubtaskDefinition(Model):
+    """
+    For each VerificationRequest there must be exactly one BlenderSubtaskDefinition in the database.
+    """
+
+    class OutputFormat(ChoiceEnum):
```
dybi: aren't there more formats allowed?[blender docs]( https://docs.blender.org/manual/en/dev/data_system/files/media/image_formats.html#image-formats)

dybi: Are we supposed to support just those two formats? In the [blender docs]( https://docs.blender.org/manual/en/dev/data_system/files/media/image_formats.html#image-formats) seems to be more of it ;)

rwrzesien: Off course we should, I was looking for this but couldn't find it, thanks for the link !

cameel: Let's drop EXR.

Also, please make sure that both uppercase and lowercase versions are supported.

rwrzesien: done :+1: 

What do you mean by `are supported` ?

concent_api/conductor/models.py
```diff
@@ -0,0 +1,90 @@
...
+    """
+
+    # Relative path of the file. Relative to the same directory that paths listed in FileTransferTokens are relative to.
+    path = CharField(max_length=MESSAGE_PATH_LENGTH)
+
+    # Foreign key to VerificationRequest. Can be NULL if there's no corresponding request.
+    verification_request = ForeignKey(VerificationRequest, related_name = 'upload_reports', blank=True, null=True)
+
+    # Indicates when conductor has been notified about the upload.
+    created_at      = DateTimeField(default = timezone.now)
```
Jakub89: There seams to be a lot of "space" issue i this whole PR. Which style we should use anyway? With or without spaces in function call?

rwrzesien: @Jakub89 All should be updated by now.

concent_api/conductor/models.py
cameel: I mean that `TaskToCompute` can specify the format as either `PNG` or `png` and both are accepted by Golem (see https://github.com/golemfactory/golem/issues/218). And maybe things like `PnG` are accepted too. We should be able to handle all values accepted by Golem.

Though not necessarily in this enum - we could assume that Concent e.g. converts extension to lower case when it passes the format in a Celery task.

rwrzesien: :+1: Off course, but that will be handled in the following pull requests, here is only definition.

PR #288, opened 2018-04-17
Chore refactor api cluster tests: force-accept-or-reject-results

dybi:
This is (almost) final version of `force-accept-or-reject-results` - once other tests are brought to similiar level, further refactoring is possible (extracing more common items, reusing helper method/functions form django integration tests)

### DIFF
--- a/concent_api/api-integration-force-accept-or-reject-results-test.py
+++ b/concent_api/api-integration-force-accept-or-reject-results-test.py
@@ -15,6 +15,7 @@ from utils.testing_helpers import generate_ecc_key_pair
 
 from api_testing_common import api_request
 from api_testing_common import create_client_auth_message
+from api_testing_common import parse_command_line
 from api_testing_common import timestamp_to_isoformat
 
 from protocol_constants import get_protocol_constants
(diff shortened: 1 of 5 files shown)


concent_api/api-integration-force-accept-or-reject-results-test.py
```diff
@@ -109,181 +103,147 @@ def report_computed_task(timestamp = None, task_to_compute = None):
...
-    #  Test CASE 2A + 2D + 3 - Send ForceSubtaskResults with same task_id as stored by Concent before
-    #  Step 1. Send ForceSubtaskResults first time
-    subtask_id = str(random.randrange(1, 100000))
-    task_id = subtask_id + '2A'
+    (cluster_url, patterns) = parse_arguments()
+    cluster_consts = get_protocol_constants(cluster_url)
+    print_protocol_constants(cluster_consts)
+
+    test_id = str(random.randrange(1, 100000))
+    tests_to_execute = get_tests_list(patterns, list(globals().keys()))
```
rwrzesien: As discussed before, please don't use `globals()`. What you are trying to do here is implement your own test runner, something which is already available if you use python default `unittest` or nose tests by rewritting test functions into test class.

As discussed, we don't want to change it now, but we should in the future.

concent_api/api-integration-force-accept-or-reject-results-test.py
```diff
@@ -109,181 +103,147 @@ def report_computed_task(timestamp = None, task_to_compute = None):
...
                     )
                 )
             )
         ),
-        headers = {
-            'Content-Type':                     'application/octet-stream',
-            'temporary-account-funds':          'True'
+        headers={
+            'Content-Type': 'application/octet-stream',
+            'temporary-account-funds': 'True'
```
rwrzesien: Are you aware that this is part you said is not needed in https://github.com/golemfactory/concent/pull/213/files ?

dybi: it is still needed, because @kbeker 's changes are not yet merged. It should be removed as a part of @kbeker 's PR

Issue #305, opened 2018-04-19
Concent crashes if requestor checks for ForceGetTaskResultDownload too late


dybi:
SCENARIO:

>  Requestor -> Concent:     ForceGetTaskResult
>         Concent   -> Requestor:   AckForceGetTaskResult
>         Concent   -> Provider:    ForceGetTaskResultUpload
>         Provider  -> Concent:     Upload good files
>         Concent   -> Requestor:    #ForceGetTaskResultDownload

Concent should always have not outdated toke, when checks if file transefer is finished.
Client should be able to receive message with outdated token (when requests `receive/` endpoint too late) - it's client's problem what to do with such a token.

cameel:
Crash report in Sentry: https://talkback.golem.network/sentry/concent/issues/9083

PR #306, opened 2018-04-19
Fix FileTransferToken creation for concent

dybi:
resolves: https://github.com/golemfactory/concent/issues/305


concent_api/core/transfer_operations.py
```diff
@@ -102,33 +99,64 @@ def store_pending_message(
...
+
 ) -> FileTransferToken:
     """
     Function to create FileTransferToken from ReportComputedTask message
     """
+    def calculate_token_expiration_deadline(
+        operation: FileTransferToken.Operation,
+        deadline: Optional[int],
+    ) -> int:
+        if deadline is not None:
```
rwrzesien: Extra assert that `operation in FileTransferToken.Operation` won't hurt, if that condition won't be met the code will fail.

concent_api/core/transfer_operations.py
```diff
@@ -102,33 +99,64 @@ def store_pending_message(
...
 
 
-def create_file_transfer_token(
+def create_file_transfer_token_for_concent(
+    report_computed_task: message.tasks.ReportComputedTask,
+    authorized_client_public_key: bytes,
+    operation: FileTransferToken.Operation,
+) -> FileTransferToken:
+    ten_minutes = 600
+    return _create_file_transfer_token(report_computed_task, authorized_client_public_key, operation, ten_minutes)
```
rwrzesien: Maybe split into separate lines for readability.

concent_api/core/transfer_operations.py
```diff
@@ -102,33 +99,64 @@ def store_pending_message(
...
+
+
+def create_file_transfer_token_for_golem_client(
     report_computed_task:       message.tasks.ReportComputedTask,
     authorized_client_public_key: bytes,
     operation:                  FileTransferToken.Operation,
+) -> FileTransferToken:
+    return _create_file_transfer_token(report_computed_task, authorized_client_public_key, operation)
```
rwrzesien: Same here.

concent_api/core/transfer_operations.py
```diff
@@ -102,33 +99,64 @@ def store_pending_message(
...
+                        settings.SUBTASK_VERIFICATION_TIME
+                )
+        return token_expiration_deadline
+
     task_id         = report_computed_task.task_to_compute.compute_task_def['task_id']
     subtask_id      = report_computed_task.task_to_compute.compute_task_def['subtask_id']
     file_path       = get_storage_file_path(task_id, subtask_id)
 
     assert isinstance(authorized_client_public_key, bytes)
+    assert isinstance(deadline, int) or deadline is None
```
rwrzesien: Check this out:
```
>>> deadline = False
>>> isinstance(deadline, int) or deadline is None
True
```

concent_api/core/tests/test_unit_transfer_operation.py
```diff
@@ -1,13 +1,19 @@
+import datetime
 from django.test import override_settings
+from django.test import TestCase
 import mock
+from freezegun import freeze_time
 
+from golem_messages.factories.tasks import ReportComputedTaskFactory
```
kbeker: :+1: for getting messages also from golem_message factory

concent_api/core/tests/test_unit_transfer_operation.py
```diff
@@ -47,31 +53,74 @@ def test_properly_work_of_request_upload_status_function(self):
...
-                request_upload_status(
-                    file_transfer_token,
-                    report_computed_task,
+                request_upload_status(report_computed_task)
+
+
+@override_settings(
+    CONCENT_PRIVATE_KEY=CONCENT_PRIVATE_KEY,
+    CONCENT_PUBLIC_KEY=CONCENT_PUBLIC_KEY,
+)
```
kbeker: If you are using settings e.g `MAXIMUM_DOWNLOAD_TIME` like you doing below you should override them to be sure that they working exactly how you want. Second thing is that if there won't be this setting this test may crash

concent_api/core/tests/test_unit_transfer_operation.py
```diff
@@ -47,31 +53,74 @@ def test_properly_work_of_request_upload_status_function(self):
...
+                upload_token = create_file_transfer_token_for_golem_client(
+                    self.report_computed_task,
+                    self.authorized_client_public_key,
+                    FileTransferToken.Operation.download
+                )
+                self.assertTrue(upload_token.timestamp > upload_token.token_expiration_deadline)
+
+    def _get_deadline_exceeded_time_for_upload_token(self):
+        return self.time + datetime.timedelta(
+            seconds=settings.SUBTASK_VERIFICATION_TIME + self.deadline + 1
```
kbeker: Please add to override settings at the top

concent_api/core/tests/test_unit_transfer_operation.py
```diff
@@ -47,31 +53,74 @@ def test_properly_work_of_request_upload_status_function(self):
...
+        )
+
+    def test_that_file_transfer_token_for_concent_is_never_out_of_date(self):
+        report_computed_task = ReportComputedTaskFactory()
+
+        token = create_file_transfer_token_for_concent(
+            report_computed_task,
+            self.authorized_client_public_key,
+            FileTransferToken.Operation.download
+
```
rwrzesien: Please remove blank line.

concent_api/core/tests/test_unit_transfer_operation.py
```diff
@@ -47,31 +53,74 @@ def test_properly_work_of_request_upload_status_function(self):
...
+@override_settings(
+    CONCENT_PRIVATE_KEY=CONCENT_PRIVATE_KEY,
+    CONCENT_PUBLIC_KEY=CONCENT_PUBLIC_KEY,
+)
+class FileTransferTokenCreationTest(TestCase):
+    def setUp(self):
+        self.time = datetime.datetime.strptime("2017-11-17 10:00:00", "%Y-%m-%d %H:%M:%S")
+        self.deadline = 10
+        self.authorized_client_public_key = b'7' * 64
```
rwrzesien: Should it be bytes or string ?

cameel: `bytes`. `authorized_client_public_key` is not base64-encoded. Just like the public keys in `TaskToCompute`.

concent_api/core/transfer_operations.py
dybi: normal thing for python - it treats `bool` values as `int` values. It's not a bug, it's a feature ;)

dybi: done

concent_api/core/tests/test_unit_transfer_operation.py
dybi: done

concent_api/core/tests/test_unit_transfer_operation.py
dybi: done

concent_api/core/tests/test_unit_transfer_operation.py
dybi: done

concent_api/core/transfer_operations.py
cameel: The situation when a condition cannot be met is exactly the case where an assert is appropriate. Otherwise you would use an `if`.

I would not call it useless duplication in the same sense the type annotation are not useless duplication.

rwrzesien: Basically, without assert if you can this function with `deadline = 1` and `operation = 'something'` the code will fail on line 157.

concent_api/core/transfer_operations.py
rwrzesien: I know, but this value will pass this assert.

concent_api/core/transfer_operations.py
dybi: how one should call this function directly when it is inner function of `_create_file_transfer_token` which already has this `assert`?

rwrzesien: It is now, but you never know how it is going to be used in the future. For example, I am going to use this function in verification too.

cameel: @rwrzesien I just talked about this with @dybi
I missed that it's an inner function. In that case I'd say that skipping it is fine. No one is going to call it from the outside and the outer function already has an assert for that.

rwrzesien: :+1:

PR #308, opened 2018-04-20
logs: Change golem_messages int TYPES to string names

### LINKED ISSUE #296: Refer to Golem message types by names in error messages and logs
Currently in some places we're generating error messages that contain numeric identifiers of Golem messages rather than names. For example this error message:
```
There was an exception when validating if golem_message 2002 is signed with public key
```
would be more useful and self-explanatory if it looked like this:
```
TaskToCompute is not signed with requestor's public key
```

Find and fix all places where we're generating error or log messages with numeric identifiers.

#### Getting message name
You can get message name with `type(message).__name__` if you already have it deserialized. If you only have the numeric ID you can get the name this way:

```
from golem_messages.message import registered_message_types

registered_message_types.get(msg_type)
```

kbeker:
Change golem messages types from ints to strings in logs
Resolves #296

### DIFF
--- a/concent_api/utils/logging.py
+++ b/concent_api/utils/logging.py
@@ -12,7 +12,7 @@ def log_message_received(message: Message, client_public_key: str):
     task_id = get_task_id_for_logging(message)
     subtask_id = get_subtask_id_for_logging(message)
     logger.info('A message has been received in `send/` -- MESSAGE_TYPE: {} -- TASK_ID: {} -- SUBTASK_ID: {} -- CLIENT PUBLIC KEY: {}'.format(
-        message.TYPE,
+        type(message).__name__,
         task_id,
         subtask_id,
         client_public_key,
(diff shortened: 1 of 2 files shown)


concent_api/utils/logging.py
```diff
@@ -11,35 +11,40 @@
...
+    logger.info('A message has been received in `send/` -- MESSAGE: {} --TYPE: {} -- TASK_ID: {} -- SUBTASK_ID: {} -- CLIENT PUBLIC KEY: {}'.format(
+        type(message).__name__,
         message.TYPE,
         task_id,
         subtask_id,
         client_public_key,
     ))
-    logger.debug('A message has been received in `send/` -- MESSAGE: {}'.format(
-        message,
+    logger.debug('A message has been received in `send/` -- MESSAGE: {} --TYPE: {}'.format(
```
rwrzesien: Here we wanted to print whole message, please revoke this change.

concent_api/utils/logging.py
```diff
@@ -11,35 +11,40 @@
...
+        type(message).__name__,
```
rwrzesien: Same here.

concent_api/utils/logging.py
kbeker: Done

concent_api/utils/logging.py
kbeker: Done

Issue #241, opened 2018-04-04
Verifier: uploading rendering result to the storage server


cameel:
This is a part of the implementation of the #193 blueprint. You will need settings defined in #221.

When the verifier is done with rendering, it should perform the following steps:
- Generate a `FileTransferToken` valid for an upload of the image generated by blender.
    - The image should be stored under `blender/verifier-output/{task_id}/{subtask_id}.{extension}` on the storage server.
    - It does not have to be compressed or encrypted
        - We will consider compression for image formats that do not have compression built-in later. This is just an optimization - there is always a trade-off betwen the space taken and the extra processing and memory that's needed to do the compression.
- Upload the image.
    - If upload fails, report/log an error but do not end the verification.
- Delete the image from local storage (only if upload did not fail)
    - If the operation fails, report/log an error but do not end the verification.

#### Storage server
You'll need a local server configured for file upload. Use the script provided in https://github.com/golemfactory/concent-deployment/issues/94.

PR #318, opened 2018-04-24
[Verification 8] Feature verification verifier uploading rendering result to storage server

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/241.


concent_api/verifier/utils.py
```diff
@@ -96,3 +97,46 @@ def delete_file(file_path):
...
     except OSError as exception:
         logger.warning(f'File with path {file_path} was not deleted, exception: {exception}')
+
+
+def upload_file_to_storage_cluster(file_content, file_path, upload_token):
+    dumped_upload_token = dump(upload_token, None, settings.CONCENT_PUBLIC_KEY)
+    b64_encoded_token = b64encode(dumped_upload_token).decode()
+    headers = {
+        'Authorization': 'Golem ' + b64_encoded_token,
+        'Concent-Auth': b64encode(
```
dybi: why `b64encode(<SOMETHING>).decode()`?

concent_api/verifier/utils.py
```diff
@@ -96,3 +97,46 @@ def delete_file(file_path):
...
+def generate_blender_output_file_name(scene_file):
+    return f'{settings.VERIFIER_STORAGE_PATH}/{scene_file}_out'
+
+
+def generate_upload_file_name(subtask_id):
+    return f'blender/verifier-output/{subtask_id}/{subtask_id}'
+
+
+def get_file_size_and_checksum(file_path):
+    with open(file_path, 'r') as file:
```
dybi: if `file` fails to be read, `file_content` will be udefined... Maybe this function should do something more explicit in case that reading the file fails?

concent_api/verifier/tasks.py
```diff
@@ -179,6 +183,33 @@ def blender_verification_order(
     for file_path in source_files_list + [source_package_path]:
         delete_file(file_path)
 
+    blender_output_file_name = generate_blender_output_file_name(scene_file)
```
dybi: could this new block be extracted to a new function? (i.e. `upload_image()`)

concent_api/verifier/utils.py
rwrzesien: We do this in the same way in `request_upload_status` in `transer_operations.py` and in cluster tests, so I think it is proper way to do it.

concent_api/verifier/utils.py
rwrzesien: Right, how about wrapping the usage of it in try/except OsError ?

concent_api/verifier/tasks.py
rwrzesien: It is used only once now, so I don't see reason for this.

dybi: just for improved clarity/readability ;)

concent_api/verifier/utils.py
rwrzesien: :+1: done

concent_api/verifier/tasks.py
rwrzesien: I would like to keep most of the verifier process in single place, if it is not a part that is currently reused, and then maybe take a better look on the whole task.

concent_api/verifier/utils.py
cameel: @dybi It's Ok because it's `bytes.decode()`. It's just a conversion from `bytes` to `str`.

It's equivalent to `bytes.decode('utf-8')`. i.e. it takes a sequence of bytes, assumes it has UTF-8 encoding and returns a `str`. UTF-8 is not really relevant here. We could do `bytes.decode('ascii')` just as well and it would work too, because `b64encode()` is garanteed to include only characters from a narrow subset of ASCII.

As an aside, I'd prefer if we always passed a parameter to `decode()` explicitly. It would make it easier to see what we are doing and that we decode as 'utf-8'. Also, `decode('ascii')` would really be more appropriate here because we don't expect any unicode characters anyway.

concent_api/verifier/utils.py
cameel: WTF. Why are we computing the size and checksum here? You're supposed to get this information in `blender_verification_order`.

cameel: Oh, this is a file we generate ourselves. In that case it's OK.

cameel: > Maybe this function should do something more explicit in case that reading the file fails?

In normal circumstances I would say it should just fail. For example if you can't read it due to permissions, something must be broken in container configuration. The whole verification should fail.

But since this step is not critical (it's only for auditing), I'd say that cathing the exception, silencing it and sending a crash report to Sentry would be fine too.

concent_api/verifier/tasks.py
```diff
@@ -179,6 +183,33 @@ def blender_verification_order(
...
+        with open(os.path.join(settings.VERIFIER_STORAGE_PATH, blender_output_file_name)) as f:
+            upload_file_to_storage_cluster(
+                f.read(),
+                upload_file_name,
+                upload_file_transfer_token,
+            )
+
+        # Delete the image from local storage.
+        delete_file(blender_output_file_name)
+    except OSError as e:
```
cameel: Please use full names: `except OSError as exception`. It's a small thing but, at least for me, it makes it much easier to read. When it's called `exception` I can read it right away without having to translate `e` to `exception` in my head.

concent_api/verifier/tasks.py
```diff
@@ -179,6 +183,33 @@ def blender_verification_order(
...
+        upload_file_transfer_token = create_file_transfer_token_for_concent(
+            subtask_id=subtask_id,
+            result_package_path=upload_file_name,
+            result_size=file_size,
+            result_package_hash=file_check_sum,
+            operation=message.FileTransferToken.Operation.upload,
+        )
+
+        # Upload the image.
+        with open(os.path.join(settings.VERIFIER_STORAGE_PATH, blender_output_file_name)) as f:
```
cameel: Please rename `f` to `blender_output_file`.

concent_api/verifier/tasks.py
```diff
@@ -179,6 +183,33 @@ def blender_verification_order(
     for file_path in source_files_list + [source_package_path]:
         delete_file(file_path)
 
+    blender_output_file_name = generate_blender_output_file_name(scene_file)
+    upload_file_name = generate_upload_file_name(subtask_id)
+    (file_size, file_check_sum) = get_file_size_and_checksum(blender_output_file_name)
```
cameel: Let's keep consistently calling it `checksum`, not `check_sum`. Keeping it consistent makes it easier to find stuff in the code when grepping or searching on github.

concent_api/verifier/tasks.py
```diff
@@ -179,6 +183,33 @@ def blender_verification_order(
...
+
+    try:
+        # Generate a FileTransferToken valid for an upload of the image generated by blender.
+        upload_file_transfer_token = create_file_transfer_token_for_concent(
```
cameel: This should be outside of the `try ... except` block. An exception handler should cover only the code where you expect the exception.

concent_api/verifier/tasks.py
```diff
@@ -179,6 +183,33 @@ def blender_verification_order(
...
+            upload_file_to_storage_cluster(
+                f.read(),
+                upload_file_name,
+                upload_file_transfer_token,
+            )
+
+        # Delete the image from local storage.
+        delete_file(blender_output_file_name)
+    except OSError as e:
+        logger.info(f'Uploading blender generated file to storage fails with: {e}')
```
cameel: `fails` -> `failed`

cameel: This is an error not info. You should also log the exception. And report it to Sentry.

concent_api/verifier/utils.py
rwrzesien: This code was moved directly to `task.py` and is now handled differently.

concent_api/verifier/utils.py
```diff
@@ -96,3 +97,46 @@ def delete_file(file_path):
...
+    return requests.post(
+        "{}upload/".format(settings.STORAGE_CLUSTER_ADDRESS),
+        headers=headers,
+        data=file_content,
+        verify=False
+    )
+
+
+def generate_blender_output_file_name(scene_file):
+    return f'{settings.VERIFIER_STORAGE_PATH}/{scene_file}_out'
```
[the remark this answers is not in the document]
rwrzesien: :+1:

concent_api/verifier/utils.py
```diff
@@ -96,3 +97,46 @@ def delete_file(file_path):
...
+
+
+def generate_upload_file_name(subtask_id):
+    return f'blender/verifier-output/{subtask_id}/{subtask_id}'
```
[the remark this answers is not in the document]
rwrzesien: :+1:

concent_api/verifier/utils.py
```diff
@@ -96,3 +97,46 @@ def delete_file(file_path):
             os.unlink(file_path)
     except OSError as exception:
         logger.warning(f'File with path {file_path} was not deleted, exception: {exception}')
+
+
+def upload_file_to_storage_cluster(file_content, file_path, upload_token):
+    dumped_upload_token = dump(upload_token, None, settings.CONCENT_PUBLIC_KEY)
+    b64_encoded_token = b64encode(dumped_upload_token).decode()
```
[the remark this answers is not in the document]
rwrzesien: :+1:

concent_api/verifier/utils.py
```diff
@@ -96,3 +97,46 @@ def delete_file(file_path):
...
+        'Concent-Auth': b64encode(
+            dump(
+                message.concents.ClientAuthorization(
+                    client_public_key=settings.CONCENT_PUBLIC_KEY,
+                ),
+                settings.CONCENT_PRIVATE_KEY,
+                settings.CONCENT_PUBLIC_KEY
+            ),
+        ).decode(),
+        'Concent-upload-path': file_path,
```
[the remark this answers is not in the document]
rwrzesien: :+1:

concent_api/verifier/tasks.py
rwrzesien: :+1:

concent_api/verifier/tasks.py
rwrzesien: :+1: It was cleaned in next branch but I will add it here too.

concent_api/verifier/tasks.py
rwrzesien: :+1: Already done

concent_api/verifier/tasks.py
rwrzesien: :+1:

concent_api/verifier/tasks.py
rwrzesien: :+1:

concent_api/verifier/tasks.py
rwrzesien: :+1: done but please confirm that I have used it correctly.

concent_api/verifier/utils.py
```diff
@@ -96,3 +96,36 @@ def delete_file(file_path):
             os.unlink(file_path)
     except OSError as exception:
         logger.warning(f'File with path {file_path} was not deleted, exception: {exception}')
+
+
+def upload_file_to_storage_cluster(file_content, file_path, upload_token):
```
dybi: this function is almost identical to the one defined in: `api-integration-force-get-task-result-test`. Please extract it to `utils` module and reuse here and in `api-integration-force-get-task-result-test`

rwrzesien: :+1: done

concent_api/verifier/tasks.py
```diff
@@ -164,6 +169,41 @@ def blender_verification_order(
...
+    upload_file_name = generate_upload_file_name(subtask_id, output_format)
+
+    upload_file_content = None
+
+    try:
+        with open(os.path.join(settings.VERIFIER_STORAGE_PATH, blender_output_file_name), 'r') as upload_file:
+            upload_file_content = upload_file.read()
+    except OSError as exception:
+        logger.error(f'Uploading blender generated file to storage failed with: {exception}')
+        sentry_logger.error(str(exception))
```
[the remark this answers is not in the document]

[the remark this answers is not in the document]

Issue #295, opened 2018-04-18
Don't generate RejectReportComputedTask when provider starts the REPORT use case after deadline


cameel:
If the provider sends `ForceReportComputedTask` to Concent after subtask deadline, Concent immediately returns `ForceReportComputedTaskResponse` with `reason` set to `SubtaskTimeout`. Currently it also generates an instance of `RejectReportComputedTask` and puts it inside. The spec is not clear on whether this is necessary.

@lukasz-glen Says that it's not - so change the code so that Concent just puts `None` in that field.

lukasz-glen:
`ForceReportComputedTaskResponse` does not have the reason `TaskTimeLimitExceeded`. Did you mean `SubtaskTimeout`?

cameel:
Looks like I mixed up the constants, `TaskTimeLimitExceeded` is a reason in `RejectReportComputedTask` and `SubtaskTimeout` is a reason in `ForceReportComputedTaskResponse`. So yes, I meant `SubtaskTimeout`. Description updated.

PR #319, opened 2018-04-24
Implement don't generate RejectReportComputedTask when provider start…

rwrzesien:
…s the REPORT use case after deadline

Resolves https://github.com/golemfactory/concent/issues/295.


dybi:
@rwrzesien, it seems that this PR contains some other changes, not only related to the BUG in the description?

PR #322, opened 2018-04-24
Fix replace report computed task when requestor sends diff than provider

### LINKED ISSUE #174: Replace ReportComputedTask when requestor sends different one than provider
In `FORCING REPORT` state if Concent gets `AckReportComputedTask` from requestor, it may contain `ReportComputedTask` instance different than the one the provider sent in `ForceReportComputedTask` (for example if the provider has sent a different `ReportComputedTask` to Concent and a different one to requestor). In that case it's the `ReportComputedTask` from requestor that counts. It's the one that should be stored in the Subtask table. In case there was already `ReportComputedTask` assigned to Subtask which was sent by provider, it should be replaced by the one sent by requestor.

kbeker:
Resolves #174

### DIFF
--- a/concent_api/core/message_handlers.py
+++ b/concent_api/core/message_handlers.py
@@ -28,10 +28,10 @@ from core.queue_operations import send_blender_verification_request
 from core.subtask_helpers import verify_message_subtask_results_accepted
 from core.transfer_operations import store_pending_message
 from core.transfer_operations import create_file_transfer_token_for_golem_client
+from core.validation import validate_all_messages_identical
 from core.validation import validate_ethereum_addresses
 from core.validation import validate_golem_message_signed_with_key
 from core.validation import validate_golem_message_subtask_results_rejected
-from core.validation import validate_list_of_identical_task_to_compute
 from core.validation import validate_report_computed_task_time_window
 from core.validation import validate_task_to_compute
 from utils import logging
@@ -93,6 +93,7 @@ def handle_send_force_report_computed_task(client_message):
 
 def handle_send_ack_report_computed_task(client_message):
     task_to_compute = client_message.report_computed_task.task_to_compute
+    report_computed_task = client_message.report_computed_task
     provider_public_key = task_to_compute.provider_public_key
     requestor_public_key = task_to_compute.requestor_public_key
 
@@ -126,11 +127,18 @@ def handle_send_ack_report_computed_task(client_message):
                 "Received AckReportComputedTask but RejectReportComputedTask "
                 "or another AckReportComputedTask for this task has already been submitted."
             )
-
-        validate_list_of_identical_task_to_compute([
+        validate_all_messages_identical([
             task_to_compute,
             deserialize_message(subtask.task_to_compute.data.tobytes()),
         ])
+        new_report_computed_task = None
+        try:
+            validate_all_messages_identical([
+                report_computed_task,
+                deserialize_message(subtask.report_computed_task.data.tobytes()),
+            ])
+        except Http400:
+            new_report_computed_task = report_computed_task
 
         subtask = update_subtask(
             subtask                     = subtask,
@@ -138,6 +146,7 @@ def handle_send_ack_report_computed_task(client_message):
             next_deadline               = None,
             set_next_deadline           = True,
             ack_report_computed_task    = client_message,
+            report_computed_task        = new_report_computed_task,
         )
         store_pending_message(
             response_type       = PendingResponse.ResponseType.ForceReportComputedTaskResponse,
@@ -231,7 +240,7 @@ def handle_send_reject_report_computed_task(client_message):
         raise Http400("Subtask requestor key does not match current client key. Can't accept your 'RejectReportComputedTask'.")
 
     tasks_to_compute.append(deserialize_message(subtask.task_to_compute.data.tobytes()))
-    validate_list_of_identical_task_to_compute(tasks_to_compute)
+    validate_all_messages_identical(tasks_to_compute)
 
     if (
         client_message.reason in
@@ -482,7 +491,7 @@ def handle_send_force_subtask_results_response(client_message):
     if subtask.subtask_results_accepted_id is not None or subtask.subtask_results_rejected_id is not None:
         raise Http400("This subtask has been resolved already.")
 
-    validate_list_of_identical_task_to_compute([
+    validate_all_messages_identical([
         task_to_compute,
         deserialize_message(subtask.task_to_compute.data.tobytes()),
     ])
@@ -1022,7 +1031,13 @@ def set_subtask_messages(
 
     for message_name, message_type in Subtask.MESSAGE_FOR_FIELD.items():
         message_to_store = subtask_messages_to_set.get(message_name)
-        if message_to_store is not None and getattr(subtask, message_name) is None:
+        if (
+            message_to_store is not None and
+            (
+                getattr(subtask, message_name) is None or
+                message_to_store.__class__ in Subtask.MESSAGE_REPLACEMENT_FOR_STATE[subtask.state_enum]
+            )
+        ):
             assert isinstance(message_to_store, message_type)
             stored_message = store_message(
                 message_to_store,
@@ -1061,7 +1076,7 @@ def store_or_update_subtask(
 
     if subtask is not None:
         if task_to_compute is not None and subtask.task_to_compute is not None:
-            validate_list_of_identical_task_to_compute([
+            validate_all_messages_identical([
                 task_to_compute,
                 deserialize_message(subtask.task_to_compute.data.tobytes()),
             ])
--- a/concent_api/core/models.py
+++ b/concent_api/core/models.py
@@ -231,6 +231,42 @@ class Subtask(Model):
         for state in states
     }.issubset(set(SubtaskState))
 
+    # Defines in which states exist possibility to replace message nested in Subtask.
+    MESSAGE_REPLACEMENT_FOR_STATE = {
+        SubtaskState.FORCING_REPORT: {
+            message.ReportComputedTask,
+        },
+        SubtaskState.REPORTED: {
+            None,
+        },
+        SubtaskState.FORCING_RESULT_TRANSFER: {
+            None,
+        },
+        SubtaskState.RESULT_UPLOADED: {
+            None,
+        },
+        SubtaskState.FORCING_ACCEPTANCE: {
+            None,
+        },
+        SubtaskState.REJECTED: {
+            None,
+        },
+        SubtaskState.VERIFICATION_FILE_TRANSFER: {
+            None,
+        },
+        SubtaskState.ADDITIONAL_VERIFICATION: {
+            None,
+        },
+        SubtaskState.ACCEPTED: {
+            None,
+        },
+        SubtaskState.FAILED: {
+            None,
+        },
+    }
+
+    assert set(MESSAGE_REPLACEMENT_FOR_STATE) == set(SubtaskState)
+
     class Meta:
         unique_together = (
             ('requestor', 'task_id'),
--- a/concent_api/core/tests/test_integration_report_computed_task.py
+++ b/concent_api/core/tests/test_integration_report_computed_task.py
@@ -164,13 +164,15 @@ class ReportComputedTaskIntegrationTest(ConcentIntegrationTestCase):
         self._assert_stored_message_counter_not_increased()
 
         # STEP 3: Requestor accepts computed task via Concent
-
+        report_computed_task_from_requestor = self._get_deserialized_report_computed_task(
+            timestamp = "2017-12-01 10:55:00",
+            task_to_compute = task_to_compute,
+        )
         serialized_ack_report_computed_task = self._get_serialized_ack_report_computed_task(
             timestamp = "2017-12-01 11:00:05",
             ack_report_computed_task = self._get_deserialized_ack_report_computed_task(
                 timestamp = "2017-12-01 11:00:05",
-                subtask_id = '8',
-                task_to_compute = task_to_compute
+                report_computed_task = report_computed_task_from_requestor,
             ),
             requestor_private_key = self.REQUESTOR_PRIVATE_KEY
         )
@@ -181,10 +183,10 @@ class ReportComputedTaskIntegrationTest(ConcentIntegrationTestCase):
                 data                           = serialized_ack_report_computed_task,
                 content_type                   = 'application/octet-stream',
             )
-
+        self._test_report_computed_task_in_database(report_computed_task_from_requestor)
         self.assertEqual(response_3.status_code,  202)
         self.assertEqual(len(response_3.content), 0)
-        self._assert_stored_message_counter_increased(increased_by = 1)
+        self._assert_stored_message_counter_increased(increased_by = 2)
         self._test_subtask_state(
             task_id                  = '1',
             subtask_id               = '8',
@@ -1034,8 +1036,7 @@ class ReportComputedTaskIntegrationTest(ConcentIntegrationTestCase):
             timestamp = "2017-12-01 11:00:05",
             ack_report_computed_task = self._get_deserialized_ack_report_computed_task(
                 timestamp = "2017-12-01 11:00:05",
-                subtask_id = '1',
-                task_to_compute = task_to_compute
+                report_computed_task = report_computed_task,
             ),
             requestor_private_key = self.REQUESTOR_PRIVATE_KEY
         )
@@ -2076,8 +2077,7 @@ class ReportComputedTaskIntegrationTest(ConcentIntegrationTestCase):
             timestamp = "2017-12-01 11:00:05",
             ack_report_computed_task = self._get_deserialized_ack_report_computed_task(
                 timestamp       = "2017-12-01 11:00:05",
-                subtask_id      = '8',
-                task_to_compute = task_to_compute
+                report_computed_task = report_computed_task,
             ),
             requestor_private_key = self.REQUESTOR_PRIVATE_KEY
         )
@@ -2245,8 +2245,7 @@ class ReportComputedTaskIntegrationTest(ConcentIntegrationTestCase):
             timestamp = "2017-12-01 11:00:05",
             ack_report_computed_task = self._get_deserialized_ack_report_computed_task(
                 timestamp       = "2017-12-01 11:00:05",
-                subtask_id      = '1',
-                task_to_compute = task_to_compute
+                report_computed_task = report_computed_task,
             ),
             requestor_private_key = self.REQUESTOR_PRIVATE_KEY
         )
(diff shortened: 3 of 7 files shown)


concent_api/core/validation.py
```diff
@@ -106,6 +106,23 @@ def validate_list_of_identical_task_to_compute(list_of_task_to_compute: List[mes
     return True
 
 
+def validate_list_of_identical_report_computed_task(list_of_report_computed_task: List[message.tasks.ReportComputedTask]):
```
rwrzesien: How about making `validate_list_of_identical_task_to_compute` generic instead of introducing new one ? You have to only:
- pass `message_type` to use it in lines 85, 93 and 96
- change names to more generic (`golem_message` instead `task_to_compute`)

concent_api/core/message_handlers.py
```diff
@@ -1032,7 +1044,7 @@ def set_subtask_messages(
 
     for message_name, message_type in Subtask.MESSAGE_FOR_FIELD.items():
         message_to_store = subtask_messages_to_set.get(message_name)
-        if message_to_store is not None and getattr(subtask, message_name) is None:
+        if message_to_store is not None and (getattr(subtask, message_name) is None or replace == message_to_store):
```
rwrzesien: I see some downfalls of this mechanism, for example:
- you can replace only message at a time,
- you have to have original message which you are going to replace for comparison in this condition.

How about doing this:
- remove the `replace` parameter and related stuff.
- introduce a dict of lists in `Subtask` model which will define in which state which messages can be replaced (for validation). Lets call it `MESSAGE_REPLACEMENT_FOR_STATE`. For now it can only contain values for key `FORCING_REPORT`. Please remember about some extra asserts for this dict, see how they are done in other dicts in `Subtask` model.
- in the `if` above, instead of checking `replace == message_to_store` check if message type is in `MESSAGE_REPLACEMENT_FOR_STATE[subtask.state]`.

@cameel Should we keep in database old `report_computed_task` ? If not @kbeker please remember to remove it.

If not, what will happen if after assigning second `report_computed_task` you will call `subtask.report_computed_task` ? Which one will be called ?

concent_api/core/validation.py
dybi: I would add one more general comment to this.
For me `validate_...` functions should not return anything. In case validation passes, nothing happens. If it doesn't, function should raise proper Exception.
If you need boolean returning functions, these should be in form of a question, i.e.
`are_report_computed_task_messageges_identical` or sth similar.
Otherwise, it is very misleading.

concent_api/core/tests/test_unit_validation.py
```diff
@@ -62,3 +64,27 @@ def test_that_function_raise_http400_when_price_is_negative(self):
 
     def test_that_function_will_not_return_anything_when_price_is_zero(self):  # pylint: disable=no-self-use
         validate_subtask_price_task_to_compute(mocked_message_with_price(0))
+
+
+class TestValidatorReportComputedTaskList(ConcentIntegrationTestCase):
+
```
dybi: I like how neat and precise the tests are :+1: 
One problem though is naming of the tested function (please see: my comment below)

concent_api/core/message_handlers.py
cameel: > Should we keep in database old `report_computed_task` ?

Won't hurt anything to keep it. We're keeping everything right now, right? It won't be used for processing but may help us see what actually happened when something goes wrong. So basically just for audit or debugging.

But it would be fine to remove it if keeping it is a problem.

> If not, what will happen if after assigning second `report_computed_task` you will call `subtask.report_computed_task` ? Which one will be called ?

The one the foreign key in the subtask points at. And that's what we want. We must never pull messages directly from `StoredMessage` because they don't have any inherent meaning. `StoredMessage` is meant to be just dumb storage. Only when a message is associated with something via a foreign key you can know it's the right message.

> introduce a dict of lists in Subtask model which will define in which state which messages can be replaced (for validation).

And here add an `assert` that does not let you change the foreign key if it's not `None` and is not listed in that dict.

concent_api/core/validation.py
```diff
@@ -106,6 +106,23 @@ def validate_list_of_identical_task_to_compute(list_of_task_to_compute: List[mes
...
+    assert all([isinstance(report_computed_task, message.tasks.ReportComputedTask) for report_computed_task in list_of_report_computed_task])
+
+    if len(list_of_report_computed_task) <= 1:
+        return False
+
+    base_report_computed_task = list_of_report_computed_task[0]
+
+    for report_computed_task in list_of_report_computed_task:
+        for slot in message.tasks.ReportComputedTask.__slots__:
+            if getattr(base_report_computed_task, slot) != getattr(report_computed_task, slot):
```
cameel: Doesn't Golem's `Message` have an equality operator? If not, maybe it would make sense to add it there and make a pull request?

concent_api/core/validation.py
```diff
@@ -106,6 +106,23 @@ def validate_list_of_identical_task_to_compute(list_of_task_to_compute: List[mes
     return True
 
 
+def validate_list_of_identical_report_computed_task(list_of_report_computed_task: List[message.tasks.ReportComputedTask]):
+    assert isinstance(list_of_report_computed_task, list)
+    assert all([isinstance(report_computed_task, message.tasks.ReportComputedTask) for report_computed_task in list_of_report_computed_task])
+
+    if len(list_of_report_computed_task) <= 1:
+        return False
```
cameel: Why aren't all messages on a list containing one message considered identical? Same for zero messages. I think `True` would make more sense here.

concent_api/core/validation.py
```diff
@@ -106,6 +106,23 @@ def validate_list_of_identical_task_to_compute(list_of_task_to_compute: List[mes
...
+
+    base_report_computed_task = list_of_report_computed_task[0]
+
+    for report_computed_task in list_of_report_computed_task:
```
cameel: There's no need to start from the first element because then you're comparing it with itself. This would work just as fine if you started from the second one:
``` python
for report_computed_task in list_of_report_computed_task[1:]:
```
as a bonus this works even if there's just one message on the list.

There's also a very easy way to create a version that does not need any special cases, even for zero:
```
for message1, message2 in zip(list_of_report_computed_task, list_of_report_computed_task[1:]):
    # Compare message1 and message 2
```

concent_api/core/validation.py
cameel: By adding `*` in front of the list you can easily make a function that takes variable number of arguments (not sure if the type annotation is still allowed in that case):
``` python
def validate_list_of_identical_report_computed_task(*list_of_report_computed_task: List[message.tasks.ReportComputedTask]):
```

Then you won't have to use a list when you call it:
``` python
identical_report_computed_task = validate_list_of_identical_report_computed_task(
    report_computed_task,
    deserialize_message(subtask.report_computed_task.data.tobytes()),
)
```

concent_api/core/validation.py
rwrzesien: @cameel Function `validate_list_of_identical_task_to_compute` resolves most of this issues, that's why I have suggested to use it and making it generic.

cameel: OK. It's fine. But please make sure that issues that are not resolved there are addressed in this pull request.

concent_api/core/validation.py
cameel: OK. Makes sense.

concent_api/core/message_handlers.py

concent_api/core/validation.py
kbeker: It's working exactly as you wrote but Concent code is already adjusted to this kind parametrs forwarding to this function (already function merged with function to comparing TaskToCompute messages) and I think might be too much changes to do.

concent_api/core/validation.py
kbeker: We can't do it this way becuase sometimes we comparing more than 2 messages.

concent_api/core/validation.py
kbeker: Done

kbeker: Done

concent_api/core/validation.py
cameel: Try it. It works on more than two messages just fine.

`zip()` combines two lists into a list of pairs. Say, we have messages A, B, C, D, E. This code will give you pairs (A, B), (B, C), (C, D), (D, E). If all of these pairs contain identical messages, it's mathematically guaranteed that all the messages on the list are identical.

rwrzesien: @cameel I am thinking about this one, isn't it also guaranteed if we take element A and compare it with all following elements ? Also wouldn't, it be quicker ?

concent_api/core/validation.py
rwrzesien: I agree with @kbeker , I am not a fan of using `*` or `**` unless it is giving a big advantage and can save code. We use it nicely in cluster tests in one place, but here it would not improve the code much in my opinion.

concent_api/core/validation.py
```diff
@@ -80,26 +80,28 @@ def validate_golem_message_client_authorization(golem_message: message.concents.
     validate_public_key(golem_message.client_public_key, 'client_public_key')
 
 
-def validate_list_of_identical_task_to_compute(list_of_task_to_compute: List[message.TaskToCompute]):
-    assert isinstance(list_of_task_to_compute, list)
-    assert all([isinstance(task_to_compute, message.TaskToCompute) for task_to_compute in list_of_task_to_compute])
+def all_messages_identical(golem_messages_list: List[message.Message]):
```
rwrzesien: I would be great if we could keep `validate_` prefix in this function name, as we have in other validation functions.

dybi: @rwrzesien, in my opinion if it returns boolean it should be in a form of a question; my thoughts on that topic: https://github.com/golemfactory/concent/pull/322#discussion_r183947116
@kbeker: I would prefer `are_all_messages_identical`

concent_api/core/tests/test_integration_report_computed_task.py
```diff
@@ -2612,3 +2611,168 @@ def test_provider_forces_computed_task_report_and_requestor_sends_rejection_due_
...
+        # STEP 1: Provider forces computed task report via Concent
+
+        task_to_compute = self._get_deserialized_task_to_compute(
+            timestamp = "2017-12-01 10:00:00",
+            task_id = '1',
+            subtask_id = '8',
+            deadline = "2017-12-01 11:00:00"
+        )
+
+        report_computed_task_for_requestor = self._get_deserialized_report_computed_task(
```
dybi: I think it should be called `report_computed_task_from_requestor`

concent_api/core/tests/test_integration_report_computed_task.py
```diff
@@ -2612,3 +2611,168 @@ def test_provider_forces_computed_task_report_and_requestor_sends_rejection_due_
...
+            timestamp = "2017-12-01 10:58:00",
+            task_to_compute = task_to_compute,
+        )
+        report_computed_task_for_concent = self._get_deserialized_report_computed_task(
```
dybi: and this one (if I understand correctly) is `report_computed_task_from_provider`

concent_api/core/tests/test_unit_authentication_utilities.py
```diff
@@ -136,45 +136,45 @@ def test_validate_golem_message_signed_with_key_should_raise_error_if_incorrect_
...
+            all_messages_identical(
                 [
                     message.TaskToCompute(),
                     message.TaskToCompute(),
                     message.Ping(),
                 ]
             )
 
-    def test_validate_list_of_identical_task_to_compute_should_return_true_if_list_has_0_or_one_element(self):
+    def test_all_messages_identical_should_return_true_if_list_has_0_or_one_element(self):
```
dybi: Why not two separate tests?

concent_api/core/tests/test_unit_authentication_utilities.py
```diff
@@ -136,45 +136,45 @@ def test_validate_golem_message_signed_with_key_should_raise_error_if_incorrect_
 
 class ValidateListOfIdenticalTaskToComputeUnitTest(TestCase):
```
dybi: BTW: I think ot should inherit from `unittest.TestCase`. Inheriting from `django.test.TestCase` seems a bit of overkill for simple unit test ;)

concent_api/core/tests/test_unit_validation.py
```diff
@@ -62,3 +64,30 @@ def test_that_function_raise_http400_when_price_is_negative(self):
...
+
+class TestAllMessagesIdentical(ConcentIntegrationTestCase):
+
+    def setUp(self):
+        super().setUp()
+        self.report_computed_task = self._get_deserialized_report_computed_task()
+
+    def test_that_function_returns_true_when_in_list_is_less_than_one_item(self):
+        response = all_messages_identical([])
```
dybi: I think it is `result`, same below ;)

concent_api/core/message_handlers.py
```diff
@@ -31,7 +31,7 @@
 from core.validation import validate_ethereum_addresses
 from core.validation import validate_golem_message_signed_with_key
 from core.validation import validate_golem_message_subtask_results_rejected
-from core.validation import validate_list_of_identical_task_to_compute
+from core.validation import all_messages_identical
```
Jakub89: Alphabetical order ;)

concent_api/core/validation.py
cameel: Yes, it's guaranteed too. I wasn't saying that the current method does not work. I only pointed out that you can do this without the extra `if`s and without singling out the first element.

Also, I'm not saying that this method is overall the best and that we have to use it. That's for your consideration. It's shorter but that's not the only criterion. I'm only responding to a comment that it would not work. For me the weak point of my approach is that it may not obvious what `zip()` does when lists are not of the same length (it ignores the extra items).

As for which one executes faster - I think that your method might be faster but not significantly. It's a difference of a few assignments (times the number of items). And given that we're doing that for very short lists of constant length, readability matters more. Please choose the one you think is easier to read and understand, not the one that's faster.

Jakub89:
Don't want to be picky, but I've notice You don't stick everywhere to "PEP8" style that we agree on.

concent_api/core/validation.py
cameel: Sure. It was just a suggestion. I pointed it out because I would probably do it myself in this particular case and I wasn't sure if @kbeker considered this feature here.

But I agree that it should be used sparingly. For me it's a nice feature of cases where I know that a function will most of the time be used with short lists of constant length where each element is created separately by the caller.

concent_api/core/validation.py
cameel: I'm also for dropping `validate_` prefix. That's because it's not validation in the same sense the other functions are. It's not raising validation errors but just returning the result of a comparison. It's a generic function for comparing messages. I would even of as far as to move it to utils or helpers.

cameel: Oh, I see it's raising exceptions now. But I'm with @dybi on that. You should make up your mind on whether you want a validation (raising exceptions) or a utility function (returning booleans) and stick to it.

rwrzesien: So lets drop the condition which returns True and add `validation_` prefix, and assert that list has at least one element. The rest of the function should work just fine with those changes.

cameel: Sounds fine for me.

kbeker: Done

concent_api/core/models.py
```diff
@@ -231,6 +231,46 @@ class SubtaskState(ChoiceEnum):
...
+        SubtaskState.ACCEPTED: {
+            None,
+        },
+        SubtaskState.FAILED: {
+            None,
+        },
+    }
+
+    assert set(MESSAGE_REPLACEMENT_FOR_STATE) == set(SubtaskState)
+    assert {
```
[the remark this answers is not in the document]

concent_api/core/validation.py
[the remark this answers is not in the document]
kbeker: Done

concent_api/core/tests/test_unit_authentication_utilities.py
kbeker: Done

concent_api/core/tests/test_unit_validation.py
kbeker: Done

concent_api/core/message_handlers.py

kbeker:
@rwrzesien @cameel @dybi @Jakub89 
Fixed all review notes

concent_api/core/tests/test_unit_validation.py
kbeker: Done

concent_api/core/tests/test_unit_authentication_utilities.py
kbeker: Done

concent_api/core/tests/test_unit_authentication_utilities.py
kbeker: Done

concent_api/core/tests/test_integration_report_computed_task.py
kbeker: Done

concent_api/core/tests/test_integration_report_computed_task.py
kbeker: Done

concent_api/core/tests/test_unit_authentication_utilities.py
```diff
@@ -136,45 +136,42 @@ def test_validate_golem_message_signed_with_key_should_raise_error_if_incorrect_
...
 
-    def test_validate_list_of_identical_task_to_compute_should_raise_error_when_not_list_is_used(self):
+    def test_validate_all_messages_identical_should_raise_error_when_not_list_is_used(self):
```
dybi: `test_that` ;)

concent_api/core/tests/test_unit_authentication_utilities.py
```diff
@@ -136,45 +136,42 @@ def test_validate_golem_message_signed_with_key_should_raise_error_if_incorrect_
...
         with self.assertRaises(AssertionError):
-            validate_list_of_identical_task_to_compute('str')
+            validate_all_messages_identical('str')
 
         with self.assertRaises(AssertionError):
-            validate_list_of_identical_task_to_compute(True)
+            validate_all_messages_identical(True)
 
-    def test_validate_list_of_identical_task_to_compute_should_raise_error_when_not_all_list_items_are_instances_of_task_to_compute(self):
+    def test_validate_all_messages_identical_should_raise_error_when_not_all_list_items_are_instances_of_task_to_compute(self):
```
dybi: 'test_that' ;)

concent_api/core/tests/test_unit_authentication_utilities.py
```diff
@@ -136,45 +136,42 @@ def test_validate_golem_message_signed_with_key_should_raise_error_if_incorrect_
...
                     message.TaskToCompute(),
                     message.Ping(),
                 ]
             )
 
-    def test_validate_list_of_identical_task_to_compute_should_return_true_if_list_has_0_or_one_element(self):
-        self.assertTrue(
-            validate_list_of_identical_task_to_compute([])
-        )
+    def test_that_function_raise_assertion_error_when_list_has_0_elements(self):
```
dybi: `raise' -> 'raises'

concent_api/core/tests/test_unit_authentication_utilities.py
```diff
@@ -136,45 +136,42 @@ def test_validate_golem_message_signed_with_key_should_raise_error_if_incorrect_
...
+            validate_all_messages_identical([])
 
-        self.assertTrue(
-            validate_list_of_identical_task_to_compute(
-                [
-                    message.TaskToCompute(),
-                ]
-            )
-        )
+    def test_that_function_pass_when_list_has_1_elemt(self):
```
dybi: 'pass' -> 'passes' or better: `doesnt_rasie_exception`

concent_api/core/tests/test_unit_authentication_utilities.py
```diff
@@ -136,45 +136,42 @@ def test_validate_golem_message_signed_with_key_should_raise_error_if_incorrect_
...
+        try:
+            validate_all_messages_identical([message.TaskToCompute()])
+        except Exception:  # pylint: disable=broad-except
+            self.fail()
 
-    def test_validate_list_of_identical_task_to_compute_should_raise_http400_when_not_all_task_to_compute_are_identical(self):
+    def test_validate_all_messages_identical_should_raise_http400_when_not_all_task_to_compute_are_identical(self):
```
dybi: `test_that` ;)

concent_api/core/tests/test_unit_authentication_utilities.py
```diff
@@ -188,11 +185,11 @@ def test_validate_list_of_identical_task_to_compute_should_raise_http400_when_no
         ]
 
         with self.assertRaises(Http400):
-            validate_list_of_identical_task_to_compute(
+            validate_all_messages_identical(
                 list_of_not_identical_task_to_compute
             )
 
-    def test_validate_list_of_identical_task_to_compute_should_not_raise_http400_when_all_task_to_compute_are_identical(self):
+    def test_validate_all_messages_identical_should_not_raise_http400_when_all_task_to_compute_are_identical(self):
```
dybi: `test_that` ;)

concent_api/core/tests/test_unit_validation.py
```diff
@@ -62,3 +64,28 @@ def test_that_function_raise_http400_when_price_is_negative(self):
 
     def test_that_function_will_not_return_anything_when_price_is_zero(self):  # pylint: disable=no-self-use
         validate_subtask_price_task_to_compute(mocked_message_with_price(0))
+
+
+class TestAllMessagesIdentical(ConcentIntegrationTestCase):
```
dybi: `TestValidateAll...`

concent_api/core/validation.py
```diff
@@ -80,31 +80,29 @@ def validate_golem_message_client_authorization(golem_message: message.concents.
     validate_public_key(golem_message.client_public_key, 'client_public_key')
 
 
-def validate_list_of_identical_task_to_compute(list_of_task_to_compute: List[message.TaskToCompute]):
-    assert isinstance(list_of_task_to_compute, list)
-    assert all([isinstance(task_to_compute, message.TaskToCompute) for task_to_compute in list_of_task_to_compute])
+def validate_all_messages_identical(golem_messages_list: List[message.Message]):
+    assert isinstance(golem_messages_list, list)
+    assert len(golem_messages_list) >= 1
+    assert all(isinstance(nested_message, message.Message) for nested_message in golem_messages_list)
```
dybi: it's just `msg` or sth like this, but definitely not nested ;)

Issue #155, opened 2018-02-27
Add more checks to system_check


rwrzesien:
Currently we are only checking few things in system_check. We should extend it and check more things with more generic checks. We should check:

- If necessary settings are defined.
- If necessary settings have correct values.
- If RELEASE-VERSION exists.

More points might be added to this list.

rwrzesien:
Old description as a reference:

Currently we are only checking few things in system_check. We should extend it and check more things with more generic checks. We should check:

- If necessary settings are defined.
- If necessary settings have correct values.
- If RELEASE-VERSION exists.
- If `CREATION_NEW_CHAIN_SEGMENT_TIME` is different than 0.
- If CONCENT_PUBLIC_KEY and CONCENT_PRIVATE_KEY are correct (length, type etc.).
- If all databases have `ATOMIC_REQUESTS` enabled (#279).

More points might be added to this list.

cameel:
I like the idea. This is something repetitive that certainly can be improved.

1) Maybe use something like `check.Type`, `check.Required` instead of `SystemCheckType` and `SystemCheckRequired`?

2) Instead of `(GreaterThanOrEqualSystemCheck, [1])` you could do `(check.GreaterThanOrEqual, 1)` by assuming that the first element is always the check class and the rest of parameters are passed directly to `__init__()`.

    But maybe it would be better not to do anything complicated in `__init__()` and just create instances right away? They should not really have any local state except for their parameters.

3) `RequiredForConcentFeatureSystemCheck`: should support multiple features. And forcing the setting not to be defined when feature is disabled.

4) > One note - I don't like current way of setting system check error id like concent.E002. We only have few of those, but we already have some of them repeated, and there is no way to automatically prevent this. Typical case when this may happen is when two or more developers are defining system check erros in separate branches. Maybe we should rethink how this is done ?

    This is the default Django convention but we could change it for our own checks. For example to names similar to error codes.

5) > Do you see possibility to create tests for system checks in the same way ?

    We could have a test case class and special assertions for stuff that repeats in system checks.

rwrzesien:
> 1. Maybe use something like check.Type, check.Required instead of SystemCheckType and SystemCheckRequired?

:+1: Sounds good, just to be sure you are saying that there should be module `check.py` which should be imported, and then we should use classes like `Type` from it ?

[...]

PR #324, opened 2018-04-24
Add defaults times to settings based on constants from golem messages

### LINKED ISSUE #282: Defaults for timing settings based on constants from golem-messages
The timing settings in `base.py` should get their default values from constants defined in golem-messages.

This task depends on https://github.com/golemfactory/golem-messages/issues/196.

kbeker:
Resolves #282 
Ready to merge after https://github.com/golemfactory/concent/pull/323

### DIFF
--- a/concent_api/concent_api/settings/base.py
+++ b/concent_api/concent_api/settings/base.py
@@ -223,11 +223,11 @@ LOGGING = {
 
 # A global constant defining the length of the time window within which a requestor or a provider is supposed to
 # contact concent and send or receive a message as defined in the protocol.
-CONCENT_MESSAGING_TIME = 3600  # seconds
+CONCENT_MESSAGING_TIME = int(constants.CMT.total_seconds())
 
 # A global constant defining the length of the time window within which a requestor can
 # contact concent for forced getting task results.
-FORCE_ACCEPTANCE_TIME = 4 * 3600  # 4 hours
+FORCE_ACCEPTANCE_TIME = int(constants.FAT.total_seconds())
 
 # A global constant defining the assumed default resource download rate.
 MINIMUM_UPLOAD_RATE = constants.DEFAULT_UPLOAD_RATE
@@ -242,7 +242,7 @@ SUBTASK_VERIFICATION_TIME = 4 * 3600  # 4 hours
 ADDITIONAL_VERIFICATION_CALL_TIME = 4 * 3600  # 4 hours
 
 # A global constant defining the lenght of the time window within which a requestor must pay
-PAYMENT_DUE_TIME = 12 * 3600  # 12 hours
+PAYMENT_DUE_TIME = int(constants.PDT.total_seconds())
 
 # A global constant defining currently used payment backend.
 PAYMENT_BACKEND = 'core.payments.mock'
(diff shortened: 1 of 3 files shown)


concent_api/concent_api/settings/base.py
```diff
@@ -223,11 +223,11 @@
 
 # A global constant defining the length of the time window within which a requestor or a provider is supposed to
 # contact concent and send or receive a message as defined in the protocol.
-CONCENT_MESSAGING_TIME = 3600  # seconds
+CONCENT_MESSAGING_TIME = int(constants.CMT.total_seconds())  # seconds
```
rwrzesien: Maybe it would be good moment to add system checks and their tests for those settings ?

Checks:
- must be set if `concent_api` is in `CONCENT_FEATURES`.
- must be int.
- must be non-negative.

As those checks will be the same for all 3, maybe try to create generic check to not reproduce code ? Same for tests. If it is not possible or will make the code unreadable then just do them separately. 
However, if you chose the generic check, please see if it is applicable to replace any already existing check.
I know it is a lot of extra work in comparison to what was done in this pull request so if you have something more important to do then maybe just update https://github.com/golemfactory/concent/issues/155 with that info above.

@cameel @dybi What do you guys think about this ? Can you think of other checks for those settings ?

kbeker: I saw this task in our IceBox and at first I decided to do it later as new PR, but if you guys decide to do it now I will expand this task and do tests to all settings.

dybi: @rwrzesien, I agree, it's good idea.

cameel: You could check also that SVT > 4 CMT + 3 MDT.

kbeker: @cameel We don't have now setting MAXIMUM_DOWNLOAD_TIME. There is neither here in Concent nor in golem_messages constans. With which should I compare it then?

concent_api/concent_api/system_check.py
```diff
@@ -99,6 +99,14 @@ def create_error_22_if_download_leadin_time_has_wrong_value():
     )
 
 
+def create_error_23_if_concent_time_settings_have_wrong_value(concent_setting_name):
```
dybi: I would consider splittimng this into two:
1) Not defined
2) Defined but wrong value
In `system_check` you make distinction between those two cases and it should be reflected in errors created

concent_api/concent_api/tests/test_check_concents_time_settings.py
```diff
@@ -0,0 +1,92 @@
+from django.test                import override_settings
+from django.test                import TestCase
+
+from concent_api.system_check   import check_concents_time_settings
+from concent_api.system_check   import create_error_23_if_concent_time_settings_have_wrong_value
+
+
+class TestConcentsTimeSettingsCheck(TestCase):
```
dybi: I like the structure of this test class :+1:

concent_api/concent_api/settings/base.py
cameel: OK, so ignore it. When I wrote it yesterday I for some reason assumed we're replacing MDT and SVT with a constant expression that that's obviously not true. We could check that the condition still holds minimum time/deadline, but well, it's not that important.

Issue #57, opened 2018-01-18
Returning JSON on HTTP 5xx errors


cameel:
Currently in case of an exception Concent API and gatekeeper return a HTML page that just says "Internal Server Error". When `DEBUG=True` the page contains more information but I don't think we want to run with `DEBUG` enabled on the cluster.

1) We want a JSON response rather than HTML
2) The response should contain an error message and error code
3) In production mode we want just a generic message. In development we also want to include stack trace and exception message.
    - Add a separate setting to switch between these modes (`DEBUG_INFO_IN_ERROR_RESPONSES` or something like that). If the setting is not defined its value should be the same as `DEBUG`. But it should be possible to set it independently.

You'll probably have to catch all exceptions in views to achieve this. Or create a middleware. In any case, keep in mind that in case of an error you have to roll back the current database transaction.

Please remember that this middleware should be turned off when running tests.

rwrzesien:
@dybi Let me put some thoughts to it:
`handle_errors_and_responses` - yes, that would be a good place to put it, however you won't be able to reuse it in other views, as `gatekeeper` and `conductor` handle their errors in different way.
So if you want to extend this error handling on all views, I would suggest either middleware as you have proposed, or a new decorator, which will be put on top of all others, only for catching and handling 500 of everything below.

cameel:
After this is merged, we should enable the extra info on our development clusters: https://github.com/golemfactory/concent-deployment/issues/120

cameel:
@rwrzesien Actually `handle_errors_and_responses` is pretty generic now. I think it would be a good idea to use it on all API views. Some views don't need all of its features (for example support for encoding Golem messages is not necessary in gatekeeper) but having them there does not hurt.

As for decorator vs middleware: there's also a third option - [handler500](https://docs.djangoproject.com/en/2.0/topics/http/views/#customizing-error-views). We could override it to provide custom error 500 "page" that's just a piece of JSON. If possible, we want to use this because it extends Django's mechanism rather than just replace it.

The goal here is never to return HTML for HTTP 500 in our API views because the client will simply not be able to get any useful information out of that. Does not really matter where the error comes from, only that if the client is accessing an API URL we should try hard never to give it HTML. Middleware seems more comprehensive in that regard because we can handle stuff coming from the middleware too.

There's also a case that I forgot to mention in the description: the admin panel. It's would be best to keep HTML error pages in the admin if possible. So we should actually look at the `Accept` header to decide what we should return.

[...]

PR #328, opened 2018-04-27
Add error handling middleware

dybi:
- Exceptions not caught in views are intercepted and transformed to JsonResponse with status=500 and error details
resolves: https://github.com/golemfactory/concent/issues/57

### DIFF
--- a/concent_api/concent_api/middleware.py
+++ b/concent_api/concent_api/middleware.py
@@ -1,8 +1,13 @@
 import os
+import traceback
 
 from django.conf    import settings
+from django.db import transaction
+from django.http import JsonResponse
 
 from golem_messages import __version__
+from concent_api.constants import DEFAULT_ERROR_MESSAGE
+from utils.constants import ErrorCode
 
 
 class GolemMessagesVersionMiddleware(object):
@@ -39,3 +44,33 @@ class ConcentVersionMiddleware(object):
         response = self.get_response(request)
         response['Concent-Version'] = self._concent_version
         return response
+
+
+class HandleServerErrorMiddleware(object):
+    """
+    Used to catch all unhandled exceptions and return JSON response containing error information.
+    """
+    def __init__(self, get_response):
+        self.get_response = get_response
+        self.sid = None
+
+    def __call__(self, request):
+        self.sid = transaction.savepoint()
+        response = self.get_response(request)
+        transaction.savepoint_commit(self.sid)
+        return response
+
+    def process_exception(self, request, exception):  # pylint: disable=unused-argument
+        transaction.savepoint_rollback(self.sid)
+        return self._build_json_response(exception)
+
+    @staticmethod
+    def _build_json_response(exception):
+        message = {
+            "error_message": str(exception) or DEFAULT_ERROR_MESSAGE,
+            "error_code": getattr(exception, "error_code", ErrorCode.CONCENT_APPLICATION_CRASH.value),
+        }
+        debug_info = getattr(settings, 'DEBUG_INFO_IN_ERROR_RESPONSES', settings.DEBUG)
+        if debug_info:
+            message["stack_trace"] = traceback.format_exc()
+        return JsonResponse(message, status=500)
--- a/concent_api/concent_api/settings/base.py
+++ b/concent_api/concent_api/settings/base.py
@@ -29,6 +29,7 @@ INSTALLED_APPS = [
 ]
 
 MIDDLEWARE = [
+    'concent_api.middleware.HandleServerErrorMiddleware',
     'django.middleware.security.SecurityMiddleware',
     'django.contrib.sessions.middleware.SessionMiddleware',
     'django.middleware.common.CommonMiddleware',
@@ -291,3 +292,6 @@ CONCENT_FEATURES = []  # type: ignore
 
 # URL format: 'protocol://<user>:<password>@<hostname>:<port>/<virtual host>'
 # CELERY_BROKER_URL = ''
+
+# Debug setting for adding stack traces in HTTP500 responses
+# DEBUG_INFO_IN_ERROR_RESPONSES = ''
--- a/concent_api/concent_api/constants.py
+++ b/concent_api/concent_api/constants.py
@@ -56,3 +56,4 @@ APP_LABEL_TO_DATABASE = {
     'database':     'default',
     'sessions':     'default',
 }
+DEFAULT_ERROR_MESSAGE = "Something went wrong, sorry"
--- a/concent_api/concent_api/settings/development.py
+++ b/concent_api/concent_api/settings/development.py
@@ -4,6 +4,7 @@ from .base import *  # NOQA  # pylint: disable=wildcard-import
 SECRET_KEY = 'testkey'
 
 DEBUG = True
+DEBUG_INFO_IN_ERROR_RESPONSES = True
 
 DATABASES['default']['USER']     = 'postgres'
 DATABASES['default']['PASSWORD'] = ''
--- a/concent_api/concent_api/tests/test_middleware.py
+++ b/concent_api/concent_api/tests/test_middleware.py
@@ -1,17 +1,39 @@
+import json
+
 import mock
 
 from django.test            import override_settings
 from django.test            import TestCase
 from django.urls            import reverse
+
 from golem_messages         import dump
 from golem_messages         import message
 from golem_messages         import __version__
 
-from utils.testing_helpers  import generate_ecc_key_pair
+from concent_api.constants import DEFAULT_ERROR_MESSAGE
+from core.tests.utils import ConcentIntegrationTestCase
+from utils.constants import ErrorCode
+from utils.testing_helpers import generate_ecc_key_pair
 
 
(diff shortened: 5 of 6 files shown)


concent_api/concent_api/middleware.py
```diff
@@ -39,3 +43,33 @@ def __call__(self, request):
...
+            sid = transaction.savepoint()
+            response = self.get_response(request)
+            transaction.savepoint_commit(sid)
+            return response
+        except Exception as exc:  # pylint: disable=broad-except
+            transaction.savepoint_rollback(sid)
+            logging.logger.exception(exc)
+            return self._build_json_response(exc)
+
+    @staticmethod
```
rwrzesien: Will this ever be used as a static method ?

dybi: I do not think so... Would you prefer to have me disable pylin warning? ;)

rwrzesien: I didn't notice that, then we can keep it.

concent_api/concent_api/middleware.py
```diff
@@ -39,3 +43,33 @@ def __call__(self, request):
...
+        transaction.savepoint_commit(self.sid)
+        return response
+
+    def process_exception(self, request, exception):  # pylint: disable=unused-argument
+        transaction.savepoint_rollback(self.sid)
+        logging.logger.exception(exception)
+        return self._build_json_response(exception)
+
+    def _build_json_response(self, exception):  # pylint: disable=no-self-use
+        message = {"error": str(exception)}
```
cameel: Please call the field `error_message` for consistency with #131.

concent_api/concent_api/middleware.py
```diff
@@ -39,3 +43,33 @@ def __call__(self, request):
...
+        if hasattr(settings, "DEBUG_INFO_IN_ERROR_RESPONSES"):
+            debug_info = settings.DEBUG_INFO_IN_ERROR_RESPONSES
```
cameel: Please include every custom setting and its description in `base.py`. Commented out if there is no good default for it. You can say in a comment that it's by default set to `DEBUG`.

Please consider including it also in `development.py`. It won't change the functionality (`DEBUG` is already enabled in development) but it might make the setting easier to discover for someone looking for debug options.

cameel: This could be shortened to:
``` python
debug_info = getattr(settings, 'DEBUG_INFO_IN_ERROR_RESPONSES', settings.DEBUG)
```

concent_api/concent_api/middleware.py
cameel: Please include an `error_code` field and set it to `'crash'`.

Also, you might check if the exception has an `error_code` field and it so, use this code instead (@rwrzesien included it in some exception classes in #133 and I think we'll be adding it to the new ones too). Maybe even it would be a good idea to add a base exception class with this field as a part of your pull request.

concent_api/concent_api/middleware.py
dybi: ok, so I will revert it to `@staticmethod` :)

concent_api/concent_api/settings/base.py
```diff
@@ -36,6 +36,7 @@
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
     'concent_api.middleware.GolemMessagesVersionMiddleware',
     'concent_api.middleware.ConcentVersionMiddleware',
+    'concent_api.middleware.HandleServerErrorMiddleware',
```
cameel: Middleware is executed from top to bottom. I think it would be test to put ours on top so that it can catch all exceptions that escape, not only those from the view.

Well, not sure about `SecurityMiddleware`. Maybe it's better if ours is below it. All the HTTP 5xx responses should be in JSON but not at the cost of exposing a potential hole.

Please test it to make sure it really works that way - for example this middleware should be able to catch an exception raised by `GolemMessagesVersionMiddleware`.

concent_api/concent_api/middleware.py
```diff
@@ -39,3 +45,33 @@ def __call__(self, request):
...
+
+    def __call__(self, request):
+        self.sid = transaction.savepoint()
+        response = self.get_response(request)
+        transaction.savepoint_commit(self.sid)
+        return response
+
+    def process_exception(self, request, exception):  # pylint: disable=unused-argument
+        transaction.savepoint_rollback(self.sid)
+        logging.logger.exception(exception)
```
cameel: Does your middleware bypass the built-in mechanism that reports HTTP 500 errors?
- If not then this logging is not necessary.
- If yes then you're missing the part that sends an error notification (by logging using the `mail_admins` handler). In effect we'll stop receiving error notifications via Sentry an e-mail.

Please compare the logging output when your middleware is enabled and when it's disabled and make sure that we're not losing any messages.

concent_api/concent_api/settings/base.py
dybi: when returning from view, middleware is executed in reversed order. If the middleware itself raises and nothing above it catches it, exception will be automatically converted to HttpResponse thanks to django `convert_exception_to_response` mechanism that kicks in during loading of the middleware

rwrzesien: Exactly, so if you put in on top, it would be executed as last one when processing response.

concent_api/concent_api/middleware.py
dybi: done

concent_api/concent_api/middleware.py
dybi: done

concent_api/concent_api/middleware.py
dybi: done

dybi: done

concent_api/concent_api/settings/base.py

concent_api/concent_api/middleware.py
dybi: ok, from my analysis, it seems that having my middleware put on top we should be safe with all logging handlers.
I will remove logging from my middleware

dybi: done

dybi:
@cameel, @rwrzesien changes implemented, rebased on `master`, if you do not find any more issues, let's merge it ;)

concent_api/concent_api/middleware.py
```diff
@@ -39,3 +44,34 @@ def __call__(self, request):
...
+        transaction.savepoint_commit(self.sid)
+        return response
+
+    def process_exception(self, request, exception):  # pylint: disable=unused-argument
+        transaction.savepoint_rollback(self.sid)
+        return self._build_json_response(exception)
+
+    @staticmethod
+    def _build_json_response(exception):
+
```
rwrzesien: Please remove blank line.

concent_api/concent_api/middleware.py
```diff
@@ -39,3 +44,34 @@ def __call__(self, request):
...
+        message = {
+            "error_message": str(exception) or DEFAULT_ERROR_MESSAGE,
+            "error_code": getattr(exception, "error_code", "crash"),
```
rwrzesien: `crash` error_code should be defined along with other error codes and imported from there. Btw. I think crash is not best name here, how about something like `unhandled_error` ?

concent_api/concent_api/middleware.py
```diff
@@ -1,9 +1,14 @@
 import os
+import traceback
 
 from django.conf    import settings
+from django.db import transaction
+from django.http import JsonResponse
 
 from golem_messages import __version__
 
+DEFAULT_ERROR_MESSAGE = "Something went wrong, sorry"
```
rwrzesien: Please move to `constants.py`

concent_api/concent_api/middleware.py
dybi: ok, I will introduce additional `error_code`. "crash" is taken from @cameel 's comment https://github.com/golemfactory/concent/pull/328#discussion_r184674933

concent_api/concent_api/middleware.py
dybi: done

concent_api/concent_api/middleware.py
dybi: done

concent_api/concent_api/middleware.py
rwrzesien: Ok, I see you have added the code but you need to use it here too. I mean, `from utils.constants import ErrorCode` and replace `"crash"` with `ErrorCode.CONCENT_APPLICATION_CRASH.value`.

rwrzesien:
@dybi One last thing to do https://github.com/golemfactory/concent/pull/328#discussion_r186565433

concent_api/concent_api/middleware.py
dybi: sorry, missed that.

done!

concent_api/concent_api/middleware.py
cameel: Looks like it does not work that way. The middleware is silencing the error notifications.

concent_api/concent_api/middleware.py
cameel: I meant the base exception class that has the `error_code` field. There was one in @rwrzesien's pull request for JSON error codes.

Issue #330, opened 2018-04-27
Rename the 'default' database to 'control'


cameel:
There's a problem in our deployment for conductor: Django always requires the `'default'` database to be defined. It can have empty config but it must be there (see [Multiple databases > Defining your databases](https://docs.djangoproject.com/en/1.11/topics/db/multi-db/#defining-your-databases)).

Each of our apps needs only one database and we were just going to delete the config of the other one during deployment but this makes it impossible. We can work around it by setting `'default'` to `{}` but it would be better to just get rid of the concept of the derault database in our project. The default one would have to change depending on what we're deploying to make sense.

Steps to do:
- Rename `DATABASES['default']` to `DATABASES['control']` in `base.py`, `testing.py` and any other place in the code where it's used (other configs, database router, etc.).
- Add a new entry called `'default'` and set it to an empty dict.
- Modify our system checks so that a configuration consisting of just an empty dict is allowed. Right now we require `ATOMIC_REQUESTS` in every configuration dict (#279).
- Update the migration command in the `README` so that it properly migrates both databases. If plain `migrate` without extra options does not migrate both databases, replace it with two separate commands.
- (Optional) Since #279 is done, you can also refactor our database config so that there are no defaults for production:
    - Move `DATABASES` from `base.py` to `development.py`. But at least leave a comment in `base.py` explaning how it should be configured (i.e. that `'default'` should be left empty and which app needs which database).

cameel:
If #152/#233 gets merged before this, remember to update the database used for live config: https://github.com/golemfactory/concent/pull/233#discussion_r184784795

cameel:
1) What about [README](https://github.com/golemfactory/concent/blob/master/README.md)? Does it migrate both databases without extra options?

2) If some parts of the issue are marked as optional, please remember to drop a comment in the issue on whether they were actually implemented or not. @bartoszbetka will need to adjust his deployment changes accordingly and it would be easier for him if he could easily find that out without digging into the pull request.

rwrzesien:
The optional part was not done.

PR #334, opened 2018-04-27
Set default database empty

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/330.

### DIFF
--- a/concent_api/concent_api/settings/base.py
+++ b/concent_api/concent_api/settings/base.py
@@ -1,5 +1,7 @@
 import os
 
+from mypy.types import Dict  # noqa # pylint: disable=unused-import
+
 from golem_messages import constants
 
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
@@ -66,7 +68,10 @@ WSGI_APPLICATION = 'concent_api.wsgi.application'
 # https://docs.djangoproject.com/en/1.11/ref/settings/#databases
 
 DATABASES = {
-    'default': {
+    # We set 'default' database to empty config explicitly, as we want to control very carefully which models are in
+    # which database. However, django requires 'default' database to exists in settings.
+    'default': {},
+    'control': {
         'ENGINE':     'django.db.backends.postgresql_psycopg2',
         'NAME':       'concent_api',
         # 'USER':     'postgres',
@@ -88,12 +93,12 @@ DATABASES = {
         # Wrap each request in a transactions and rolled back on failure by default
         'ATOMIC_REQUESTS': True,
     }
-}
+}  # type: Dict[Dict]
 
 DATABASE_ROUTERS = ['concent_api.database_router.DatabaseRouter']
 
 # Defines database used by Constance app.
-CONSTANCE_DBS = ['default']
+CONSTANCE_DBS = ['control']
 
 # Password validation
 # https://docs.djangoproject.com/en/1.11/ref/settings/#auth-password-validators
--- a/concent_api/concent_api/settings/development.py
+++ b/concent_api/concent_api/settings/development.py
@@ -5,8 +5,8 @@ SECRET_KEY = 'testkey'
 
 DEBUG = True
 
-DATABASES['default']['USER']     = 'postgres'
-DATABASES['default']['PASSWORD'] = ''
+DATABASES['control']['USER']     = 'postgres'
+DATABASES['control']['PASSWORD'] = ''
 
 DATABASES['storage']['USER']     = 'postgres'
 DATABASES['storage']['PASSWORD'] = ''
--- a/concent_api/concent_api/settings/testing.py
+++ b/concent_api/concent_api/settings/testing.py
@@ -6,14 +6,23 @@ SECRET_KEY = 'testkey'
 DEBUG = True
 
 DATABASES = {
+    # 'default' must be set for testing, otherwise tests won't work
     'default': {
+        'ENGINE': 'django.db.backends.postgresql_psycopg2',
+        'USER': 'postgres',
+        'ATOMIC_REQUESTS': True,
+        'TEST': {
+            'NAME': 'test_default'
+        }
+    },
+    'control': {
         # NAME intentionally left unset, in case tests try to access the production database.
         # This could happen e.g. if someone executes a query at module level.
         'ENGINE':          'django.db.backends.postgresql_psycopg2',
         'USER':            'postgres',
         'ATOMIC_REQUESTS': True,
         'TEST': {
-            'NAME': 'test_concent_api'
+            'NAME': 'test_control'
         }
     },
     'storage': {
--- a/concent_api/concent_api/system_check.py
+++ b/concent_api/concent_api/system_check.py
@@ -42,11 +42,10 @@ def create_error_17_if_geth_container_address_has_wrong_value():
     )
 
 
-def create_error_18_atomic_requests_not_set_for_database(database_name):
+def create_error_18_invalid_setting_type(setting, value):
     return Error(
-        f"ATOMIC_REQUESTS for database {database_name} is not set to True",
-        hint="ATOMIC_REQUESTS must be set to True for all databases because our views rely on the fact that on errors "
-             "all database changes are rolled back",
+        f"Setting {setting} has incorrect value {value}",
+        hint=f"Set correct value on setting {setting}",
         id="concent.E018",
     )
 
@@ -96,6 +95,15 @@ def create_error_24_if_concent_time_settings_have_wrong_value(concent_setting_na
         f"{concent_setting_name} has wrong value",
         hint=f"{concent_setting_name} must be set to non-negative integer",
         id="concent.E024",
+    )
+
+
+def create_error_25_atomic_requests_not_set_for_database(database_name):
+    return Error(
+        f"ATOMIC_REQUESTS for database {database_name} is not set to True",
+        hint="ATOMIC_REQUESTS must be set to True for all databases because our views rely on the fact that on errors "
+             "all database changes are rolled back",
+        id="concent.E019",
     )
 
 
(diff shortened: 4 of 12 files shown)


concent_api/utils/decorators.py
```diff
@@ -96,21 +96,21 @@ def wrapper(request, *args, **kwargs):
 def handle_errors_and_responses(view):
     """
     Decorator for handling responses from Concent
-    for golem clients
+    for golem clients and handling transactions on 'control' database.
```
cameel: We talked about it on Slack and I said that it's fine to hard-code this decorator for the `'control'` database but now, after looking closer at the implementation, I think we should not do this. I thought the implementation was more specific to `core` but it's actually pretty generic. Enough that we could use it just fine in other apps. For example in gatekeeper. Actually I think we should start using it there and returning JSON with status 200 as simple dicts.

I think that it would be best to add a parameter for selecting the database here and always use the decorator with that parameter. Also, support the case where there's no database (parameter set to `None`).

concent_api/concent_api/system_check.py
```diff
@@ -252,7 +252,10 @@ def check_atomic_requests(app_configs = None, **kwargs):  # pylint: disable=unus
     errors = []
     if hasattr(settings, 'DATABASES') and isinstance(settings.DATABASES, dict):
         for database_name, database_config in settings.DATABASES.items():
-            if 'ATOMIC_REQUESTS' not in database_config or database_config['ATOMIC_REQUESTS'] is not True:
+            if (
+                database_config.get('ENGINE') != 'django.db.backends.dummy' and
+                ('ATOMIC_REQUESTS' not in database_config or database_config['ATOMIC_REQUESTS'] is not True)
```
cameel: You could simplify this line to: `database_config.get('ATOMIC_REQUESTS', False)`

concent_api/concent_api/settings/base.py
```diff
@@ -64,9 +64,10 @@
 # https://docs.djangoproject.com/en/1.11/ref/settings/#databases
 
 DATABASES = {
-    'default': {
+    'default': {},
```
cameel: We need a comment explaining how `'default'` works now and why it's still here. Actually a comment explaning the whole issue with `DATABASES` and our assumptions about it would be best.

concent_api/concent_api/settings/testing.py
```diff
@@ -7,13 +7,21 @@
 
 DATABASES = {
     'default': {
```
cameel: Do we really need non-empty `default` here?

concent_api/concent_api/settings/base.py
```diff
@@ -64,9 +64,10 @@
...
+    'control': {
         'ENGINE':     'django.db.backends.postgresql_psycopg2',
-        'NAME':       'concent_api',
+        'NAME':       'control',
```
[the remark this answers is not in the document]
rwrzesien: :+1:  I must have missed this when committing, I will change it back, there should be no problem with it.

concent_api/utils/decorators.py
rwrzesien: Ok, I will add it.

concent_api/concent_api/system_check.py
rwrzesien: Actually your way is not working exactly the same, because it returns True for `database_config['ATOMIC_REQUESTS']` values like `'a'`, `1` etc. I think we should check explicitly for `True`.

concent_api/concent_api/settings/development.py
```diff
@@ -5,11 +5,11 @@
...
 
-DATABASES['default']['USER']     = 'postgres'
-DATABASES['default']['PASSWORD'] = ''
+DATABASES['control']['USER']     = 'postgres'  # type: ignore
+DATABASES['control']['PASSWORD'] = ''  # type: ignore
 
-DATABASES['storage']['USER']     = 'postgres'
-DATABASES['storage']['PASSWORD'] = ''
+DATABASES['storage']['USER']     = 'postgres'  # type: ignore
+DATABASES['storage']['PASSWORD'] = ''  # type: ignore
```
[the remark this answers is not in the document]
rwrzesien: It is an open issue in mypy https://github.com/python/mypy/issues/2900.

concent_api/concent_api/settings/base.py
rwrzesien: :+1: DONE

concent_api/utils/decorators.py
```diff
@@ -93,63 +93,70 @@ def wrapper(request, *args, **kwargs):
...
-                view.__name__,
-                client_public_key,
-            )
-            return HttpResponse("", status = 204)
-        elif isinstance(response_from_view, bytes):
-            return HttpResponse(response_from_view)
-
-        assert False, "Invalid response type"
-        raise Exception("Invalid response type")
+    assert database_name in settings.DATABASES
```
dybi: Actually, due to this assert, `None` will not be accepted as `database_name`, thus if chescs below seem to be unnecessary. I reckon we should either enhance `settings.DATABASES` with `None` (does it make sense?) or get rid of checks.

concent_api/concent_api/settings/development.py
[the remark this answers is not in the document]
cameel: Fine.

Still, are we going to have to use this on every dict that contains something that can be indexed with `[]`? Is there a workaround?

concent_api/concent_api/system_check.py
cameel: Yeah, but that's exactly why I personally like to treat booleans differently in `if`s.

In `if variable is None:` vs `if variable:` I prefer the former because the latter is more prone to errors. It's likely that you don't want to treat zero as `None` and if it's not stated explicitly you have to always stop and think about it when reading the code.

In `if variable is True:` vs `if variable:` I prefer the latter because there's no such problem here. If I know the value is either `False` or `True` then it's fine. If I'm worried that it's something other than that this means that there's an `assert` or a validation missing somewhere. And even if for whatever reason it turns out that the value is not boolean, the rules are such that the result will still most likely be sensible. Treating `'a'`, `1` as `True` is more likely to be correct than treating them as `False`.

So the fact that in my version these values are treated differently is a feature, not a bug :)

cameel: Simply, if you're worried about that I think you should just add an explicit check for `isinstance(database_config['ATOMIC_REQUESTS'], bool)`.

concent_api/core/views.py
```diff
@@ -50,7 +50,7 @@ def receive(_request, message, _client_public_key):
 @csrf_exempt
 @require_POST
 @require_golem_auth_message
-@handle_errors_and_responses
+@handle_errors_and_responses(database_name='control')
```
cameel: Please remove the default value. It's dangerous to have defaults like that.

concent_api/utils/decorators.py
cameel: Yes, it does not handle `None`, but please do not assume that `None` will always be present in `settings.DATABASES`. It's not meant to be part of that dict.

`assert database_name in [settings.DATABASES, None]` would be just fine as long as there are special cases for `None` below.

concent_api/concent_api/settings/base.py
```diff
@@ -64,7 +64,10 @@
 # https://docs.djangoproject.com/en/1.11/ref/settings/#databases
 
 DATABASES = {
-    'default': {
+    # We set 'default' database to empty config explicitly, as we want to control very carefully what data use which
```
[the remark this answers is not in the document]
rwrzesien: :+1:

concent_api/concent_api/settings/development.py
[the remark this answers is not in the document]
rwrzesien: There is, but it also requires additional workarounds :) We can do `  # type: Dict[str, Dict]` after `DATABASES` definition in base.py, add import for `from mypy.types import Dict` but it also requires ignoring `W:  3, 0: Unused Dict imported from mypy.types (unused-import)`. So I think the current way would be good, and there are not many places where we need this.

concent_api/concent_api/system_check.py
rwrzesien: But in your version the condition will be met when `ATOMIC_REQUESTS` will be set in `database_config`, which will result in raising error, and we want the opposite.

rwrzesien:
@cameel @dybi Updated.

concent_api/concent_api/settings/development.py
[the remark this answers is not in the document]
cameel: Even though the annotation adds more code, I think it's a better solution than just ignoring it.

And maybe there's a way to help pylint notice imports used in annotations?

concent_api/concent_api/system_check.py
cameel: Yeah, sorry, I missed that it's `is not True`, rather than `is True`. The point still stands.

I'd write it like this:
``` python
if database_config.get('ENGINE') != 'django.db.backends.dummy' and not database_config.get('ATOMIC_REQUESTS', False):
     errors.append(create_error_19_atomic_requests_not_set_for_database(database_name))
```

Or, if I wanted to be sure that the value is definitely a boolean:
``` python
if database_config.get('ENGINE') != 'django.db.backends.dummy':
    atomic_requests = database_config.get('ATOMIC_REQUESTS', False)

    if not isinstance(atomic_requests, bool):
         errors.append(create_error_18_invalid_setting_type(f"DATABASES[{database_name}]['ATOMIC_REQUESTS']", atomic_requests))

    if not atomic_requests:
         errors.append(create_error_19_atomic_requests_not_set_for_database(database_name))
```

concent_api/concent_api/settings/development.py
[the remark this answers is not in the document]
rwrzesien: > Even though the annotation adds more code, I think it's a better solution than just ignoring it.

Ok, then I will go this way.

> And maybe there's a way to help pylint notice imports used in annotations?

There is ongoing bug for this:
https://github.com/PyCQA/pylint/issues/1063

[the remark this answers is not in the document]
rwrzesien: :+1: done

concent_api/concent_api/system_check.py

PR #267, opened 2018-04-12
Add machine codes and their handling in gatekeeper

rwrzesien:
Partially resolves https://github.com/golemfactory/concent/issues/131.

@cameel According to your comment in https://github.com/golemfactory/concent/pull/133, this is part related to gatekeeper, rebased on latest master. We can discuss here.

PR #338, opened 2018-04-30
Machine-readable error codes in Concent code

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/131.

This is second part, first was https://github.com/golemfactory/concent/pull/267.


concent_api/core/exceptions.py
```diff
@@ -1,7 +1,18 @@
+from utils.constants import ErrorCode
+
 
 class UnexpectedResponse(Exception):
     pass
 
 
 class Http400(Exception):
-    pass
+    error_code = None
```
dybi: I have some doubts about `__init__` of this exception. The way it is constructed enforces that, `error_message` must be provided as positional argument and `error_code` must be provided as keyword argument. This seems to overcomplicate things...
I would propse to change the method to sth like this:
```
    def __init__(self, error_message, error_code):
        super().__init__(error_message)
        assert isinstance(error_code, ErrorCode)
        self.error_code = error_code
```
It simplifies constructor and eliminates the need of class fields: `error_code=None`, and `error_message=''`

cameel: In general it's a good point but for this particular exception I think that mandatory error code is a good idea. I can't think of a situation when we would want to return it without a code. So I'm for leaving it as is, without a default.

cameel: Sorry, my comment was not clear enough. I also commented in a hurry just before the meeting and did not look at the code close enough.

Arguments are indeed mandatory and @dybi's suggestion preserves that so please ignore that remark.

Upon looking at it closely I completely agree with @dybi. The constructor would be better and simpler having normal, named arguments rather than relying on `args` and `kwargs`.

rwrzesien: One note - `error_message` must be set as an instance attribute, otherwise there is no other way to get pure message in decorators.

rwrzesien: :+1: done

concent_api/core/validation.py
```diff
@@ -100,7 +134,8 @@ def validate_all_messages_identical(golem_messages_list: List[message.Message]):
                         slot,
                         getattr(base_golem_message, slot),
                         getattr(golem_message, slot),
-                    )
+                    ),
+                    error_code=ErrorCode.MESSAGE_WRONG_FIELDS,
```
cameel: This is technically correct but misleading in my opinion. The code should indicate that it's not a problem with data format. It's that the messages are expected to be identical but they are not.

Also, I think it's the wrong place to construct this error because this function does not know enough context about why there's a problem and why it's a problem. It should just return a value indicating that messages are different and the caller should check it and construct the error message himself using his broader knowledge about the situation. By trying to be this generic you make the error incomprehensible to the client. The client may not be aware that you're putting messages on some list, where those messages come from and which one has which index. This is an implementation detail. The client should just be told which field does not match which other field.

concent_api/core/validation.py
```diff
@@ -119,35 +154,57 @@ def validate_golem_message_signed_with_key(
             'There was an exception when validating if golem_message {} is signed with public key {}'.format(
                 golem_message.TYPE,
                 public_key
-            )
+            ),
+            error_code=ErrorCode.MESSAGE_SIGNATURE_WRONG,
```
cameel: Same as above. This should be outside of the function so that you can report which specific piece of data has this problem.

concent_api/core/message_handlers.py
```diff
@@ -479,21 +529,35 @@ def handle_send_force_subtask_results_response(client_message):
...
+                client_message.TYPE
+            ),
+            error_code=ErrorCode.QUEUE_REQUESTOR_PUBLIC_KEY_MISMATCH,
+        )
 
     if subtask.subtask_results_accepted_id is not None or subtask.subtask_results_rejected_id is not None:
-        raise Http400("This subtask has been resolved already.")
+        raise Http400(
+            "This subtask has been resolved already.",
+            error_code=ErrorCode.QUEUE_MESSAGE_ALREADY_PROCESSED,
```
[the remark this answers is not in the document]
rwrzesien: :+1:

concent_api/core/message_handlers.py
```diff
@@ -60,7 +61,10 @@ def handle_send_force_report_computed_task(client_message):
     if Subtask.objects.filter(
         subtask_id = client_message.report_computed_task.task_to_compute.compute_task_def['subtask_id'],
     ).exists():
-        raise Http400("{} is already being processed for this task.".format(client_message.__class__.__name__))
+        raise Http400(
+            "{} is already being processed for this task.".format(client_message.__class__.__name__),
+            error_code=ErrorCode.QUEUE_MESSAGE_ALREADY_PROCESSED,
```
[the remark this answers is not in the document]
rwrzesien: :+1:

concent_api/core/message_handlers.py
```diff
@@ -280,7 +324,10 @@ def handle_send_reject_report_computed_task(client_message):
 
     if get_current_utc_timestamp() <= deserialized_message.compute_task_def['deadline'] + settings.CONCENT_MESSAGING_TIME:
         if subtask.ack_report_computed_task_id is not None or subtask.ack_report_computed_task_id is not None:
-            raise Http400("Received RejectReportComputedTask but AckReportComputedTask or another RejectReportComputedTask for this task has already been submitted.")
+            raise Http400(
+                "Received RejectReportComputedTask but AckReportComputedTask or another RejectReportComputedTask for this task has already been submitted.",
+                error_code=ErrorCode.QUEUE_RESPONSE_ALREADY_SUBMITTED,
```
rwrzesien: @cameel Should it be `SUBTASK_DUPLICATE_REQUEST` too ?

concent_api/core/message_handlers.py
```diff
@@ -113,23 +117,36 @@ def handle_send_ack_report_computed_task(client_message):
...
+                "Subtask requestor key does not match current client key. Can't accept your 'AckReportComputedTask'.",
+                error_code=ErrorCode.QUEUE_REQUESTOR_PUBLIC_KEY_MISMATCH,
+            )
 
         if subtask.ack_report_computed_task_id is not None or subtask.reject_report_computed_task_id is not None:
             raise Http400(
                 "Received AckReportComputedTask but RejectReportComputedTask "
-                "or another AckReportComputedTask for this task has already been submitted."
+                "or another AckReportComputedTask for this task has already been submitted.",
+                error_code=ErrorCode.QUEUE_RESPONSE_ALREADY_SUBMITTED,
```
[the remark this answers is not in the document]
rwrzesien: :+1:

concent_api/core/validation.py
```diff
@@ -19,43 +20,70 @@ def validate_int_value(value):
...
-            raise Http400("Wrong type, expected a value that can be converted to an integer.")
+            raise Http400(
+                "Wrong type, expected a value that can be converted to an integer.",
+                error_code=ErrorCode.MESSAGE_VALUE_NOT_INTEGER,
+            )
     if value < 0:
-        raise Http400("Wrong type, expected non-negative integer but negative integer provided.")
+        raise Http400(
+            "Wrong type, expected non-negative integer but negative integer provided.",
+            error_code=ErrorCode.MESSAGE_VALUE_NOT_INTEGER,
```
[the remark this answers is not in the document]
rwrzesien: :+1:

concent_api/core/validation.py
```diff
@@ -19,43 +20,70 @@ def validate_int_value(value):
...
+        )
     return value
 
 
 def validate_id_value(value, field_name):
     if not isinstance(value, str):
-        raise Http400("{} must be string.".format(field_name))
+        raise Http400(
+            "{} must be string.".format(field_name),
+            error_code=ErrorCode.MESSAGE_VALUE_NOT_STRING,
```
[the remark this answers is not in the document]
rwrzesien: :+1:

concent_api/core/validation.py
```diff
@@ -19,43 +20,70 @@ def validate_int_value(value):
...
 
 
 def validate_public_key(value, field_name):
     assert isinstance(field_name, str)
 
     if not isinstance(value, bytes):
-        raise Http400("{} must be bytes.".format(field_name))
+        raise Http400(
+            "{} must be bytes.".format(field_name),
+            error_code=ErrorCode.MESSAGE_VALUE_NOT_BYTES,
```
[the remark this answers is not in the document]
rwrzesien: :+1:

concent_api/core/validation.py
```diff
@@ -19,43 +20,70 @@ def validate_int_value(value):
...
+        raise Http400(
+            "{} cannot be blank.".format(field_name),
+            error_code=ErrorCode.MESSAGE_VALUE_BLANK,
+        )
 
     if len(value) > MESSAGE_TASK_ID_MAX_LENGTH:
-        raise Http400("{} cannot be longer than {} chars.".format(field_name, MESSAGE_TASK_ID_MAX_LENGTH))
+        raise Http400(
+            "{} cannot be longer than {} chars.".format(field_name, MESSAGE_TASK_ID_MAX_LENGTH),
+            error_code=ErrorCode.MESSAGE_VALUE_WRONG_LENGTH,
```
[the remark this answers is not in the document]
rwrzesien: There is a problem, because this function validates currently both fields `task_id` and `subtask_id` but it can do any other field that requires this set of validators. Same for `validate_public_key` - we use it for `provider_public_key`, `requestor_public_key`, `client_public_key` and I guess it few others too.

I don't want to make error codes formatted, because I think that is what error messages are for. I also don't want to add extra if-ology only to determine error code in a function that can accept any field value as a parameter.

It would be possible to add field name ie. to `validate_subtask_price_task_to_compute` but then we will loose some kind of consistency, so I would like to avoid that.

concent_api/core/validation.py
```diff
@@ -19,43 +20,70 @@ def validate_int_value(value):
...
+            error_code = ErrorCode.MESSAGE_VALUE_WRONG_LENGTH,
+        )
 
 
 def validate_task_to_compute(task_to_compute: message.TaskToCompute):
     if not isinstance(task_to_compute, message.TaskToCompute):
-        raise Http400("Expected TaskToCompute.")
+        raise Http400(
+            "Expected TaskToCompute.",
+            error_code=ErrorCode.MESSAGE_UNEXPECTED,
```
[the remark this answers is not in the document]
rwrzesien: :+1:

concent_api/core/validation.py
```diff
@@ -70,12 +98,18 @@ def validate_report_computed_task_time_window(report_computed_task):
     assert isinstance(report_computed_task, message.ReportComputedTask)
 
     if report_computed_task.timestamp < report_computed_task.task_to_compute.timestamp:
-        raise Http400("ReportComputedTask timestamp is older then nested TaskToCompute.")
+        raise Http400(
+            "ReportComputedTask timestamp is older then nested TaskToCompute.",
+            error_code=ErrorCode.MESSAGE_TIMESTAMP_ERROR,
```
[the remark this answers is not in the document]
rwrzesien: :+1:

concent_api/core/validation.py
rwrzesien: :+1: I will change to `MESSAGES_NOT_IDENTICAL`.

About the second part, lets create the separate issue out of it for two reasons; first is that this pull requests is going to be big even without those changes (tests updates are still to come), second is that it is not related to adding error code. If you agree I will add it.

concent_api/core/validation.py
rwrzesien: Same here, lets add it to same separate issue as mentioned above.

concent_api/core/exceptions.py
```diff
@@ -1,11 +1,19 @@
...
+class ConcentInSoftShutdownMode(Exception):
     pass
 
 
-class ConcentInSoftShutdownMode(Exception):
-    pass
+class Http400(Exception):
+
+    def __init__(self, error_message: str, error_code: ErrorCode, *args, **kwargs) -> None:
+        assert isinstance(error_message, (str, None))
```
dybi: if we want to allow `error_message` to be `None`, then in the method's signature it should be -> error_message: **Optional[str]**

concent_api/core/exceptions.py
```diff
@@ -1,11 +1,19 @@
...
+        assert isinstance(error_code, ErrorCode)
+        self.error_code = error_code
+        self.error_message = '' if error_message is None else error_message
+        super().__init__(error_message, args, kwargs)
```
dybi: what do we need `args` and `kwargs` for?

concent_api/core/exceptions.py
```diff
@@ -1,11 +1,19 @@
...
 
-class Http400(Exception):
+class ConcentInSoftShutdownMode(Exception):
     pass
 
 
-class ConcentInSoftShutdownMode(Exception):
-    pass
+class Http400(Exception):
+
```
dybi: this line could be removed

concent_api/core/tests/test_integration_accept_or_reject_results.py
```diff
@@ -1684,22 +1698,20 @@ def test_requestor_send_again_subtask_results_accepted_or_rejected_when_message_
...
-            timestamp = "2018-02-05 11:00:00",
-            subtask_results_accepted = self._get_deserialized_subtask_results_accepted(
+        subtask = store_or_update_subtask(
+            task_id=deserialized_force_subtask_results.ack_report_computed_task.report_computed_task.task_to_compute.compute_task_def['task_id'],  # pylint: disable=no-member,
+            subtask_id=deserialized_force_subtask_results.ack_report_computed_task.report_computed_task.task_to_compute.compute_task_def['subtask_id'],  # pylint: disable=no-member,
+            provider_public_key=self.PROVIDER_PUBLIC_KEY,
+            requestor_public_key=self.REQUESTOR_PUBLIC_KEY,
+            state=Subtask.SubtaskState.ACCEPTED,
+            report_computed_task=deserialized_force_subtask_results.ack_report_computed_task.report_computed_task,  # pylint: disable=no-member,
+            task_to_compute=deserialized_force_subtask_results.ack_report_computed_task.report_computed_task.task_to_compute,  # pylint: disable=no-member,
```
dybi: please extract `task_to_compute` and use it also above to obtain `task_id` and `subtask_id`

concent_api/core/tests/test_integration_accept_or_reject_results.py
```diff
@@ -1794,43 +1808,27 @@ def test_requestor_send_again_subtask_results_accepted_or_rejected_when_message_
...
         )
 
+        subtask = store_or_update_subtask(
+            task_id=deserialized_force_subtask_results.ack_report_computed_task.report_computed_task.task_to_compute.compute_task_def['task_id'],  # pylint: disable=no-member,
+            subtask_id=deserialized_force_subtask_results.ack_report_computed_task.report_computed_task.task_to_compute.compute_task_def['subtask_id'],  # pylint: disable=no-member,
+            provider_public_key=self.PROVIDER_PUBLIC_KEY,
+            requestor_public_key=self.REQUESTOR_PUBLIC_KEY,
+            state=Subtask.SubtaskState.ACCEPTED,
+            report_computed_task=deserialized_force_subtask_results.ack_report_computed_task.report_computed_task,  # pylint: disable=no-member,
+            task_to_compute=deserialized_force_subtask_results.ack_report_computed_task.report_computed_task.task_to_compute,  # pylint: disable=no-member,
```
dybi: please extract `task_to_compute` and use it also above to obtain `task_id` and `subtask_id`

concent_api/core/exceptions.py
rwrzesien: :+1:

concent_api/core/exceptions.py
rwrzesien: `Exception` class accepts it in its `__init__` method, so the class which inherits from it should accept it too.

concent_api/core/exceptions.py
rwrzesien: From pep8:

> Method definitions inside a class are surrounded by a single blank line.

concent_api/core/tests/test_integration_accept_or_reject_results.py
rwrzesien: :+1: done

concent_api/core/tests/test_integration_accept_or_reject_results.py
rwrzesien: :+1: done

concent_api/core/exceptions.py
dybi: Why? You can call `super().__init__(error_message)` ;) You do not make any use of `args` and `kwargs` in your class and you don't expect anyone to `raise Http400` with any other arguments than `error_code` and `error_message`. 
For me having `args` and `kwargs` make thing unnecessarily confusing.
(https://docs.python.org/3/tutorial/errors.html#user-defined-exceptions)

concent_api/core/tests/test_core_views.py
```diff
@@ -218,8 +221,7 @@ def test_send_should_return_http_400_if_data_is_incorrect(self):
             ),
             content_type                        = 'application/octet-stream',
         )
-        self.assertEqual(response.status_code, 400)
-        self.assertIn('error', response.json().keys())
+        self._test_400_response(response)
```
kbeker: Why here is no error code?

concent_api/core/tests/test_core_views.py
```diff
@@ -391,8 +398,7 @@ def test_send_should_reject_message_when_timestamp_too_far_in_future(self):
                 content_type                   = 'application/octet-stream',
             )
 
-        self.assertEqual(response.status_code, 400)
-        self.assertIn('error', response.json().keys())
+        self._test_400_response(response)
```
kbeker: Also here

concent_api/core/tests/test_core_views.py
```diff
@@ -439,8 +445,7 @@ def test_send_should_return_http_400_if_task_to_compute_deadline_is_not_an_integ
                     content_type = 'application/octet-stream',
                 )
 
-            self.assertEqual(response_400.status_code, 400)
-            self.assertIn('error', response_400.json().keys())
+            self._test_400_response(response_400)
```

concent_api/core/tests/test_core_views.py
rwrzesien: No, because here the response is received directly from decorator handling the errors.

concent_api/core/tests/test_core_views.py
rwrzesien: Same here.

concent_api/core/tests/test_core_views.py
rwrzesien: Same here.

concent_api/core/exceptions.py
rwrzesien: I know, but if the upper class `__init__` accepts `args` and `kwargs`, so I think that inheriting class and overriding a particular method should keep signature as close to the overridden a possible, just in case someone would try to use it. It is rather a matter of good practice, not this particular use case.

dybi: @rwrzesien, in staticaly typed languages like Java or C++ I wouldn't argue with that. However, in python it adds no value, because you do not have strict polymorphism, i. e., everything is checked at run-time (because there is no compile-time). 
I would really like not to have `args` and `kwargs` here, but I leave up to you ;)

rwrzesien:
> HEADER_AUTHORIZATION_MISSING vs HEADER_AUTHORIZATION_MISSING_TOKEN ?
In first case there is no HTTP_AUTHORIZATION in request.META, 
in second there is HTTP_AUTHORIZATION in request.META but it doesn't contain token.

> HEADER_AUTHORIZATION_TOKEN_INVALID_MESSAGE vs HEADER_AUTHORIZATION_UNRECOGNIZED_SCHEME?
In first case HTTP_AUTHORIZATION in request.META is not golem message, 
in second HTTP_AUTHORIZATION doesn't start with 'Golem ' prefix.

> HEADER_CONTENT_TYPE_NOT_SUPPORTED vs HEADER_AUTHORIZATION_UNRECOGNIZED_SCHEME?
Those are two different cases, first one is about Content-Type header, second is about HTTP_AUTHORIZATION header.

> MESSAGE_FILES_CHECKSUM_WRONG_FORMAT vs MESSAGE_FILES_CHECKSUM_WRONG_TYPE?
MESSAGE_FILES_CHECKSUM_WRONG_TYPE => int, None, list etc.
MESSAGE_FILES_CHECKSUM_WRONG_FORMAT => doesn't start with 'sha:'.

> MESSAGE_UNEXPECTED vs MESSAGE_UNKNOWN?
MESSAGE_UNEXPECTED => is a Golem Message, but not the one expected in this context.
MESSAGE_UNKNOWN => is not a Golem Message, but should be.

> MESSAGE_INVALID vs MESSAGE_WRONG_FIELDS?
MESSAGE_INVALID => Wrong message when a specific one is expected.
MESSAGE_WRONG_FIELDS => when message doesn't contain fields required by Concent.

> MESSAGE_VALUE_WRONG_TYPE vs MESSAGE_VALUE_NOT_INTEGER, MESSAGE_VALUE_NOT_STRING (wrong type should be sufficient?)
MESSAGE_VALUE_NOT_INTEGER, MESSAGE_VALUE_NOT_STRING => This actually gives more info what was the expected type.

> MESSAGE_UNABLE_TO_DESERIALIZE vs MESSAGE_INVALID ?
Two completely different cases:
MESSAGE_INVALID => Wrong message when a specific one is expected.
MESSAGE_UNABLE_TO_DESERIALIZE => something went wrong with 'load()' on serialized message.

> MESSAGE_FILES_SIZE_EMPTY vs MESSAGE_VALUE_BLANK?
MESSAGE_FILES_SIZE_EMPTY is for gatekeeper, when validation is less generic, it gives more detailed info.

concent_api/core/validation.py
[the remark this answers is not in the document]
cameel: Oh I see that this function is just for validating the ID and nothing else. When I made this comment must have I assumed that it was something that validates a whole message.

Well, in that case there's probably not much that we can do here.

Though I'm more and more convinced that we should add `field` to the response in the future so that we can indicate which specific field has a problem. Then we could have more generic errors (like @dybi wanted) but still let the client watch for very specific errors.

concent_api/core/validation.py
cameel: OK. Let's do it separately. Could you create the issue?

concent_api/core/validation.py
cameel: OK. Please create an issue.

concent_api/core/message_handlers.py
cameel: Yeah. I generally tried not to add a comment in every place I found because there would be a lot of repetition. If I mentioned something in one place it's very possible that there are other places that are like this.

concent_api/core/exceptions.py
cameel: I think it's fine to have `args` and `kwargs` in case someone wants to add extra info. `BaseException.__init__()` stores anything passed to it on `self` so I see how someone might find it useful (though I've rarely had an actual use for it). We don't have to break this feature, though I share the sentiment that the extra args make the code uglier :)

But if you're passing the arguments down to the inherited init, isn't this superfluous?
``` python
        self.error_code = error_code
        self.error_message = '' if error_message is None else error_message
```

concent_api/core/exceptions.py
cameel: I don't like these extra lines too, but if PEP8 says so then we have no choice :)

concent_api/core/validation.py
rwrzesien: Yes, I will add it.

rwrzesien: @cameel https://github.com/golemfactory/concent/issues/394

concent_api/core/exceptions.py
rwrzesien: > But if you're passing the arguments down to the inherited init, isn't this superfluous?

The point is to allow inheriting class to use features (by passing arguments) to parent class method. But in this case there are no such features, so I have done it rather automatically, as a good practice. You are both right that in this case this is not necessary, so I have removed it.

concent_api/core/exceptions.py
cameel: Yeah... I kill them on sight :)

concent_api/core/exceptions.py
cameel: > But in this case there are no such features

There are. I said it my comment. `BaseException` stores the parameters you give it on `self`. So you don't have to do it yourself.

Issue #340, opened 2018-05-08
Add non-NULL database constraint for Subtask.task_to_compute


cameel:
`TaskToCompute` is by design necessary to process messages related to a subtask. The #139 blueprint calls for it to be non-NULLable because there's no situation where we should allow adding a subtask to the `Subtask` table without having access to that message.

Make this field non-NULLable.

PR #342, opened 2018-05-08
Change task_to_compute in Subtask model to be not nullable

dybi:
resolves https://github.com/golemfactory/concent/issues/340


dybi:
migration added

PR #343, opened 2018-05-08
More precise estimation of payment block

kbeker:


concent_api/core/payments/sci_backend.py
```diff
@@ -24,8 +22,7 @@ def get_list_of_payments(
     assert isinstance(current_time,             int) and current_time   > 0
     assert isinstance(transaction_type,         Enum) and transaction_type in TransactionType
 
-    block_distance = math.ceil((current_time - payment_ts) / settings.AVERAGE_BLOCK_TIME)
-    last_block_before_payment = ConcentRPC().get_block_number() - block_distance  # type: ignore  # pylint: disable=no-member
+    last_block_before_payment = BlocksHelper(ConcentRPC()).get_first_block_after(payment_ts).number
```
dybi: please extract `ConCentRPC()` to a variable and reuse it throughout the function. I know it's singleton, but having it as variable conctructed in one place makes it more clear we do not create separate instances every time we use it ;)

PR #345, opened 2018-05-09
Chore add both task and subtask id when logging

pawelkisielewicz:


concent_api/utils/logging.py
```diff
@@ -1,112 +1,121 @@
-from logging                        import getLogger
+from logging import getLogger
```
kbeker: Same here

Issue #312, opened 2018-04-23
SUBTASK_VERIFICATION_TIME must depend on task timeout


cameel:
Just like `MAXIMUM_DOWNLOAD_TIME` (#281), `SUBTASK_VERIFICATION_TIME` is currently a constant but should really depend on how hard the task is to compute. Golem is going to change that on their side soon. We should too.

- Remove the `SUBTASK_VIERIFICATION_TIME` setting.
- Create a function that computes subtask verification time based on settings and the maximum computation time agreed between the provider and the requestor.
    - The formula is `4 * CMT + 3 * MDT + task timeout / 2` but you won't need to compute it yourself. A function that computes this should be added to golem-messages by the time you start implementing this task (EDIT: It's done: https://github.com/golemfactory/golem-messages/pull/202).
    - The only thing your function will have to do is to pass the right values in parameters.
        - CMT comes from settings
        - MDT comes from the function implemented in #281
        - task timeout is `TaskToCompute.deadline - TaskToCompute.timestamp`.
- All timeouts based on `SUBTASK_VERIFICATION_TIME` should now use our function instead.
- Remove `subtask_verification_time` from `protocol-constants`.

rwrzesien:
@cameel Function defined in https://github.com/golemfactory/golem-messages/pull/202 can't be used directly, because it uses values from golem-messages constants, and we need option to override it in our settings ie. for tests. So this function need to be rewritten in Concent code.

PR #347, opened 2018-05-09
Feature subtask verification time must depend on task timeout

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/312


concent_api/core/tests/test_auth_accept_or_reject_results.py
```diff
@@ -20,11 +20,22 @@
...
     CONCENT_MESSAGING_TIME    = 10,  # seconds
     FORCE_ACCEPTANCE_TIME     = 10,  # seconds
-    SUBTASK_VERIFICATION_TIME = 10,  # seconds
     CONCENT_ETHEREUM_ADDRESS  = 'x' * ETHEREUM_ADDRESS_LENGTH
 )
 class AuthAcceptOrRejectIntegrationTest(ConcentIntegrationTestCase):
 
+    def setUp(self):
+        super().setUp()
+        self.patcher_1 = mock.patch('core.message_handlers.calculate_subtask_verification_time', return_value=10)
```
dybi: Actually there is one pitfall patching this way. If for some reason a test encounters an error (not fails, but breaks/crashes) `tearDown` method will not be called for it, thus patch will remian 'unstopped'. To overcame this, you could rewrite the setup to sth like:
```
def setUp(self):
        super().setUp()
        patcher_in_message_handlers = mock.patch('core.message_handlers.calculate_subtask_verification_time', return_value=1800)
        patcher_in_transfer_operations = mock.patch('core.transfer_operations.calculate_subtask_verification_time', return_value=1800)
        patcher_in_message_handlers.start()
        patcher_in_transfer_operations.start()
        self.addCleanup(patcher_in_message_handlers.stop)
        self.addCleanup(patcher_in_transfer_operations.stop)
```
and then you do not have to overwrite `tearDown` and you don't need `self.moc_foo`.
Reference: https://docs.python.org/3/library/unittest.html#unittest.TestCase.addCleanup
Please apply this approach to another test classes as well

concent_api/utils/helpers.py
```diff
@@ -159,3 +160,13 @@ def calculate_maximum_download_time(size: int) -> int:
             settings.MINIMUM_UPLOAD_RATE,
         ).total_seconds()
     )
+
+
+def calculate_subtask_verification_time(report_computed_task: message.ReportComputedTask) -> int:
```
dybi: I am wondering if we could add some unit tests for this function. What it does is basically taking a `datetime.timedelta` object and extracts seconds out of it as `int` value. Not much, but I guess one unit test won't hurt ;)

Or, alternatively, you could change `setUp` in `AuthAcceptOrRejectIntegrationTest`, `AuthGetTaskResultIntegrationTest`, `AcceptOrRejectIntegrationTest`, `GetTaskResultIntegrationTest` to mock `subtask_verification_time` (to return i. e. `timedelta(seconds=X)` instead of mocking `calculate_subtask_verification_time`.

concent_api/core/tests/test_auth_accept_or_reject_results.py
cameel: This might be a good addition to our style guide. It's meant to list common pitfalls like this.

dybi: > Cleanup items are called even if setUp fails (unlike tearDown)
my mistake, I will edit first comment

rwrzesien: :+1: done

concent_api/utils/helpers.py
rwrzesien: :+1: done

concent_api/core/tests/test_integration_accept_or_reject_results.py
```diff
@@ -33,11 +34,16 @@ def _get_requestor_account_status(_provider, _requestor):
...
     CONCENT_MESSAGING_TIME    = 10,  # seconds
     FORCE_ACCEPTANCE_TIME     = 10,  # seconds
-    SUBTASK_VERIFICATION_TIME = 10,  # seconds
     CONCENT_ETHEREUM_ADDRESS  = 'x' * ETHEREUM_ADDRESS_LENGTH
 )
 class AcceptOrRejectIntegrationTest(ConcentIntegrationTestCase):
 
+    def setUp(self):
+        super().setUp()
+        self.patcher = mock.patch('utils.helpers.subtask_verification_time', return_value=datetime.timedelta(seconds=10))
```
dybi: `patcher` doesn't have to be instance's field ;)

Issue #235, opened 2018-03-29
Use real sample of TaskToCompute in tests


rwrzesien:
We have received a real-world example of `TaskToCompute`. We might use it in our tests to avoid surprises in the future. While working on this issue please do:

* Format data to a nice, human-readable form.
* Add a function in test utils (probably `ConcentIntegrationTestCase` should be ok) or update existing one, which will return `TaskToCompute` build from this default data, but with possibility to overwrite each attribute by parameter.
* Make sure all our tests use this function when they use `TaskToCompute`.

Sample `TaskToCompute` fully filled:

```
TaskToCompute(
    header=MessageHeader(
        type_=2002,
        timestamp=1522323925,
        encrypted=False),
    sig=b'T\xbb;\x8e`\xfb6\x84\xeaO)\xa4\x1b\x84\x87@\xa8\xe8\xdeZWO \xea\xd5oQ\xce\x1d\x8f\x15;L\xc0\xd0.t\xe9\xddz\xe9\x90\xf6\xd7\x8f\xec@\x1f\xcb\xbf\xf0\x94\xcb\xc0\x96\xa9\xfe\x9b\xf0\xe5\xe7f\xdd\x9c\x00',
    slots=[
        ['requestor_id', '3fa1fe13c1631e157a692317939ce0fe82b39df65336ec1ff17f0e7376aa374858c980cf54261caa5af9dc5939abd7e3455cdaf0888f05e3f4ae964ef8a217ab'],
        ['requestor_public_key', '3fa1fe13c1631e157a692317939ce0fe82b39df65336ec1ff17f0e7376aa374858c980cf54261caa5af9dc5939abd7e3455cdaf0888f05e3f4ae964ef8a217ab'],
        ['requestor_ethereum_public_key', '3fa1fe13c1631e157a692317939ce0fe82b39df65336ec1ff17f0e7376aa374858c980cf54261caa5af9dc5939abd7e3455cdaf0888f05e3f4ae964ef8a217ab'],
        ['provider_id', 'b7da70a8bbb439e8f1cdf7494ce163294a884ccaf9f117023e3cad5e5d4cb60655aec697c859fdeb9880ef94e3705ee0a2dd44eba7d68cabcd90e7972b48d00f'],
[...]

PR #348, opened 2018-05-09
Task to compute real sample

Jakub89:
Resolves: #235


Jakub89:
Mypy is not happy with default dict and lists:
```
W:497, 4: Dangerous default value {} as argument (dangerous-default-value)
W:497, 4: Dangerous default value [] as argument (dangerous-default-value)
```
How to solve this?

Issue #349, opened 2018-05-09
Errors caused by exceptions raised by golem-messages are not always reported with the original message


cameel:
Concent ignores the original exception messages from `FieldError` `MessageFromFutureError` `MessageTooOldError` raised by golem-messages:

https://github.com/golemfactory/concent/blob/7935a627a24bf84782db5ad7c54f1b6b86a27136/concent_api/utils/decorators.py#L46-L55

This is because at some point in the past we were getting exceptions with an empty message. But now they have proper messages and we should include them. Otherwise it's hard for the client to say what exactly is wrong.

Please change the code so that the original message is always included if it's non-empty.

PR #356, opened 2018-05-10
Add full message to golem-message error responses

Jakub89:
Resolves: #349


concent_api/utils/decorators.py
```diff
@@ -43,16 +43,16 @@ def wrapper(request, *args, **kwargs):
...
             except Http400 as exception:
                 return JsonResponse({'error': f'{exception}'}, status = 400)
-            except FieldError:
-                return JsonResponse({'error': "Golem Message contains wrong fields."}, status = 400)
-            except MessageFromFutureError:
-                return JsonResponse({'error': 'Message timestamp too far in the future.'}, status = 400)
-            except MessageTooOldError:
-                return JsonResponse({'error': 'Message is too old.'}, status = 400)
+            except FieldError as exception:
+                return JsonResponse({'error': f'Golem Message contains wrong fields.{exception}'}, status = 400)
```
cameel: It would be nice if there was a space between our message and the exception message.

Jakub89: Done.

cameel: Well, that will work but it's a lazy solution ;)

Maybe you could create a very simple utility function that will join N strings, add spaces between them and ignore the empty ones so that there's no separator if not needed?
``` python
join_messages('Golem Message contains wrong fields.', str(exception))
```

I wanted a suggest a simple one-liner but I think it would be too verbose to just paste it everywhere so putting it into a utility function would be better:

``` python
' '.join(m for m in messages if m not in ['', None])
```

Jakub89: Okey, I've added join_messages() function, I'm not sure if it's easier to read but maybe we could use it more in the future.

concent_api/utils/helpers.py
```diff
@@ -159,3 +159,10 @@ def calculate_maximum_download_time(size: int) -> int:
             settings.MINIMUM_UPLOAD_RATE,
         ).total_seconds()
     )
+
+
+def join_messages(*msg):
```
cameel: 1) Please avoid using shortcuts. `message` whould be a better name.
2) This is a collection so it should be `messages`.

concent_api/utils/helpers.py
```diff
@@ -159,3 +159,10 @@ def calculate_maximum_download_time(size: int) -> int:
...
+    list_of_messages = []
+    for msg_part in msg:
```
cameel: `msg_part` -> `messsage`

concent_api/utils/helpers.py
```diff
@@ -159,3 +159,10 @@ def calculate_maximum_download_time(size: int) -> int:
...
+        list_of_messages.append(msg_part.strip())
+    return ' '.join(m for m in list_of_messages if m not in ['', None])
```
cameel: This could be shortened to:

``` python
def join_messages(*messages):
    return ' '.join(m.strip() for m in messages if m not in ['', None])
```

concent_api/utils/helpers.py
Jakub89: Like previous comment, `message` would be confusing with golem-messages "message".

concent_api/utils/helpers.py
Jakub89: `.strip()` is just a safety so there can't be more then one whitespace, but okey.

PR #364, opened 2018-05-11
Chore return json on http 5xx errors (second time ;))

dybi:
resolves: https://github.com/golemfactory/concent/issues/57

### DIFF
--- a/concent_api/concent_api/middleware.py
+++ b/concent_api/concent_api/middleware.py
@@ -1,8 +1,19 @@
+import logging
 import os
+import sys
+import traceback
 
 from django.conf    import settings
+from django.db import transaction
+from django.http import HttpResponse
+from django.http import JsonResponse
+from django.views.debug import technical_500_response
+from django.views.defaults import server_error
+from mimeparse import best_match
 
 from golem_messages import __version__
+from concent_api.constants import DEFAULT_ERROR_MESSAGE
+from utils.constants import ErrorCode
 
 
 class GolemMessagesVersionMiddleware(object):
@@ -39,3 +50,64 @@ class ConcentVersionMiddleware(object):
         response = self.get_response(request)
         response['Concent-Version'] = self._concent_version
         return response
+
+
+def determine_return_type(request_meta):
+    try:
+        preferred_types = ['application/json', 'text/html']
+        if "HTTP_ACCEPT" in request_meta:
+            return best_match(preferred_types, request_meta['HTTP_ACCEPT'])
+        return "application/json"
+    except Exception:  # pylint: disable=broad-except
+        # This is a bit of a hack - as there is no specific exception that `best_match` function raises in case of
+        # invalid/broken accept header, we just catch every possible exception that comes from it and treat such
+        # a situation as a broken header
+        return ""
+
+
+class HandleServerErrorMiddleware(object):
+    """
+    Used to catch all unhandled exceptions and return JSON response containing error information.
+    """
+    def __init__(self, get_response):
+        self.get_response = get_response
+        self.sid = None
+
+    def __call__(self, request):
+        self.sid = transaction.savepoint()
+        response = self.get_response(request)
+        transaction.savepoint_commit(self.sid)
+        return response
+
+    def process_exception(self, request, exception):  # pylint: disable=unused-argument
+        transaction.savepoint_rollback(self.sid)
+        request_logger = logging.getLogger('django.request')
+        request_logger.error(
+            'Internal Server Error: %s', request.path,
+            exc_info=sys.exc_info(),
+            extra={'status_code': 500, 'request': request},
+        )
+        debug_info_enabled = getattr(settings, 'DEBUG_INFO_IN_ERROR_RESPONSES', settings.DEBUG)
+        return_type = determine_return_type(request.META)
+        if return_type == "application/json":
+            return self._build_json_response(exception, debug_info_enabled)
+        elif return_type == "text/html":
+            return self._build_html_response(request, debug_info_enabled)
+        else:
+            return HttpResponse(status=406)
+
+    @staticmethod
+    def _build_html_response(request, debug_info_enabled):
+        if debug_info_enabled:
+            return technical_500_response(request, *sys.exc_info())
+        return server_error(request)
+
+    @staticmethod
+    def _build_json_response(exception, debug_info_enabled):
+        message = {
+            "error_message": str(exception) or DEFAULT_ERROR_MESSAGE,
+            "error_code": getattr(exception, "error_code", ErrorCode.CONCENT_APPLICATION_CRASH).value,
+        }
+        if debug_info_enabled:
+            message["stack_trace"] = traceback.format_exc()
+        return JsonResponse(message, status=500)
--- a/concent_api/concent_api/settings/base.py
+++ b/concent_api/concent_api/settings/base.py
@@ -32,6 +32,7 @@ INSTALLED_APPS = [
 ]
 
 MIDDLEWARE = [
+    'concent_api.middleware.HandleServerErrorMiddleware',
     'django.middleware.security.SecurityMiddleware',
     'django.contrib.sessions.middleware.SessionMiddleware',
     'django.middleware.common.CommonMiddleware',
@@ -299,3 +300,6 @@ CONCENT_FEATURES = []  # type: ignore
 
 # URL format: 'protocol://<user>:<password>@<hostname>:<port>/<virtual host>'
 # CELERY_BROKER_URL = ''
+
+# Debug setting for adding stack traces in HTTP500 responses
+#DEBUG_INFO_IN_ERROR_RESPONSES =
--- a/concent_api/concent_api/settings/base.py
+++ b/concent_api/concent_api/settings/base.py
@@ -32,7 +32,7 @@ INSTALLED_APPS = [
 ]
 
 MIDDLEWARE = [
-    'concent_api.middleware.HandleServerErrorMiddleware',
+    'concent_api.middleware.HandleServerErrorMiddleware',  # this middleware is disabled in tests - check testing.py
     'django.middleware.security.SecurityMiddleware',
     'django.contrib.sessions.middleware.SessionMiddleware',
     'django.middleware.common.CommonMiddleware',
--- a/concent_api/concent_api/settings/testing.py
+++ b/concent_api/concent_api/settings/testing.py
@@ -75,3 +75,7 @@ MINIMUM_UPLOAD_RATE       = 1
 DOWNLOAD_LEADIN_TIME      = 3
 
 CUSTOM_PROTOCOL_TIMES = True
+
+# disable HandleServerErrorMiddleware in tests
+if MIDDLEWARE.index('concent_api.middleware.HandleServerErrorMiddleware') is not None:
+    MIDDLEWARE.remove('concent_api.middleware.HandleServerErrorMiddleware')
--- a/concent_api/concent_api/tests/test_middleware.py
+++ b/concent_api/concent_api/tests/test_middleware.py
@@ -1,17 +1,43 @@
+import json
+import unittest
+
 import mock
+from django.conf import settings
 
 from django.test            import override_settings
 from django.test            import TestCase
 from django.urls            import reverse
+
 from golem_messages         import dump
 from golem_messages         import message
 from golem_messages         import __version__
 
-from utils.testing_helpers  import generate_ecc_key_pair
+from concent_api.constants import DEFAULT_ERROR_MESSAGE
+from concent_api.middleware import determine_return_type
+from core.tests.utils import ConcentIntegrationTestCase
+from utils.constants import ErrorCode
+from utils.testing_helpers import generate_ecc_key_pair
 
 
 (CONCENT_PRIVATE_KEY,   CONCENT_PUBLIC_KEY)   = generate_ecc_key_pair()
 (PROVIDER_PRIVATE_KEY,  PROVIDER_PUBLIC_KEY)  = generate_ecc_key_pair()
+
+
+class CustomException(Exception):
+    pass
+
+
+class CustomExceptionWithStringRepr(Exception):
+    def __init__(self, error_message, error_code):
+        super().__init__()
+        self.message = error_message
+        self.error_code = error_code
+
+    def __repr__(self):
+        return self.message
+
+    def __str__(self):
+        return self.message
 
 
 @override_settings(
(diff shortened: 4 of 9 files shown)


concent_api/concent_api/middleware.py
```diff
@@ -39,3 +49,80 @@ def __call__(self, request):
...
+        self.sid = None
+
+    def __call__(self, request):
+        self.sid = transaction.savepoint()
+        response = self.get_response(request)
+        transaction.savepoint_commit(self.sid)
+        return response
+
+    def process_exception(self, request, exception):  # pylint: disable=unused-argument
+        transaction.savepoint_rollback(self.sid)
```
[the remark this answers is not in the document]
dybi: `logger` I am using here (`'django.request'`) is logging to sentry and mail_admins ;)

concent_api/concent_api/middleware.py
```diff
@@ -1,8 +1,18 @@
+import logging
 import os
+import sys
+import traceback
 
 from django.conf    import settings
+from django.db import transaction
+from django.http import JsonResponse, HttpResponseServerError
```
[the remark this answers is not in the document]
dybi: ok, done

concent_api/concent_api/middleware.py
```diff
@@ -1,8 +1,19 @@
...
+from django.db import transaction
+from django.http import HttpResponseServerError
+from django.http import JsonResponse
 
 from golem_messages import __version__
+from concent_api.constants import DEFAULT_ERROR_MESSAGE
+from utils.constants import ErrorCode
+
+
+logger = logging.getLogger('django.request')
```
cameel: Please call it something else, like `request_logger`.

It's typical in our code to do `logger = logging.getLogger(__name__)` at the top of the module. So the natural assumption is that a variable called `logger` refers to the default logger in a module. If someone comes here later and changes something, he'll likely assume that it's just such a logger without even checking it.

Since `django.request` behaves in a very special way, we must make sure it's not overused. For that reason it should not be confused with a normal logger. Maybe even it would be better to call `getLogger()` only in the function where you're using it instead of putting it here, at the module level.

concent_api/concent_api/middleware.py
```diff
@@ -39,3 +50,80 @@ def __call__(self, request):
...
+
+
+def should_use_html(request):
+    """
+    Function inspects request headers and determines is response should be in HTML format.
+    1) By default it HTML should not be used
+    2) If `text/html` is present in 'Accept' header and 'application/json' is not, HTML should be used
+    3) If both are present, HTML will be used if it has greater weight than JSON
+    """
+    def get_probability(selected_method, accepted_methods):
```
cameel: I'd call it `format` or `content_type` rather than `method`.

And `weight` rather than `probability`.

concent_api/concent_api/middleware.py
```diff
@@ -39,3 +50,80 @@ def __call__(self, request):
...
+        for method in accepted_methods:
+            if selected_method in method:
+                if ":q=" in method:
```
cameel: It should be `;q=` (with a semicolon), not `:q=`.

concent_api/concent_api/middleware.py
```diff
@@ -39,3 +50,80 @@ def __call__(self, request):
...
+                    return float(method.split(":q=")[1])
```
cameel: Could you maybe make a regex for this whole part. I think it's a bit brittle. For example you're not checking where the `selected_method` is in the string. The current code would probably accept something like `WTF;q=text/html` and fail trying to convert `'text/html'` to `float`.

When constructing the regex you can just assume that `selected_method` contains only alphanumeric characters and `/` (but please `assert` that).

cameel: This will result in HTTP 500 when the weight is not something that can be converted to `float`. You should return some kind of 4xx error in that situation instead.

Basically, assume that the format is `<content_type>(;q=<number>)?`. Maybe allow white space between components but anything other than that should be treated as an error. For example weight that is not a number or repeating the `;q=` part twice.

concent_api/concent_api/middleware.py
```diff
@@ -39,3 +50,80 @@ def __call__(self, request):
         response = self.get_response(request)
         response['Concent-Version'] = self._concent_version
         return response
+
+
+def should_use_html(request):
+    """
+    Function inspects request headers and determines is response should be in HTML format.
+    1) By default it HTML should not be used
```
cameel: Maybe it would be clearer to say explicitly that JSON is the default?

concent_api/concent_api/middleware.py
```diff
@@ -39,3 +50,80 @@ def __call__(self, request):
...
+        if 'headers' in request.META and 'Accept' in request.META['headers'] and "application/json" in request.META['headers']['Accept']:
+            return get_probability("application/json", request.META['headers']['Accept'])
+        return 0
+
+    html_probability = get_html_probability()
+    json_probability = get_json_probability()
+    if html_probability > 0:
+        if json_probability > 0:
+            return html_probability > json_probability
+        return True
```
cameel: You could shorten this to just:

``` python
return html_probability > json_probability
```

concent_api/concent_api/middleware.py
```diff
@@ -39,3 +50,80 @@ def __call__(self, request):
...
+        if should_use_html(request):
+            return self._build_html_response(debug_info_enabled)
+        return self._build_json_response(exception, debug_info_enabled)
+
+    @staticmethod
+    def _build_html_response(debug_info_enabled):
+        message = '<h1>Server Error (500)</h1>'
+        if debug_info_enabled:
+            message += f"\n\nSTACK TRACE:\n{traceback.format_exc()}"
+        return HttpResponseServerError(message, content_type='text/html')
```
cameel: Will this give us the standard Django HTML error page? If not, is there a way to get it? For example by reraising the exception (if still possible but since we can use `exc_info()`, I guess we're still inside some `except` block) or raising a new one?

concent_api/concent_api/middleware.py
```diff
@@ -39,3 +50,80 @@ def __call__(self, request):
...
+
+    @staticmethod
+    def _build_json_response(exception, debug_info_enabled):
+        message = {
+            "error_message": str(exception) or DEFAULT_ERROR_MESSAGE,
+            "error_code": getattr(exception, "error_code", ErrorCode.CONCENT_APPLICATION_CRASH.value),
```
cameel: I remember that I pointed out to @rwrzesien in the pull request that added an exception with error code that we should use the enum items (rather than names or values) wherever possible. And I think that the `error_code` field contains an item of `ErrorCode` type rather than it's value. And if it does not, it probably should. Please check, because if so, you need to call `.value` on it too.

concent_api/concent_api/settings/base.py
```diff
@@ -290,3 +291,6 @@
 
 # URL format: 'protocol://<user>:<password>@<hostname>:<port>/<virtual host>'
 # CELERY_BROKER_URL = ''
+
+# Debug setting for adding stack traces in HTTP500 responses
+# DEBUG_INFO_IN_ERROR_RESPONSES = ''
```
cameel: This is not a string. Please remove the misleading "default". It's fine to leave it as:

``` python
# Debug setting for adding stack traces in HTTP500 responses
#DEBUG_INFO_IN_ERROR_RESPONSES =
```

concent_api/concent_api/settings/testing.py
```diff
@@ -65,3 +65,8 @@
 MINIMUM_UPLOAD_RATE       = 1
 
 DOWNLOAD_LEADIN_TIME      = 3
+
+# disable HandleServerErrorMiddleware in tests
+index = MIDDLEWARE.index('concent_api.middleware.HandleServerErrorMiddleware')
+if index is not None:
+    MIDDLEWARE.remove('concent_api.middleware.HandleServerErrorMiddleware')
```
cameel: Please avoid defining variables at module level in settings (even test settings) because it's polluting the namespace. This would be better as:

``` python
if MIDDLEWARE.index('concent_api.middleware.HandleServerErrorMiddleware') is not None:
    MIDDLEWARE.remove('concent_api.middleware.HandleServerErrorMiddleware')
```

And if you absolutely have to, please give the variable a much more specific name.

concent_api/concent_api/settings/base.py
```diff
@@ -32,6 +32,7 @@
 ]
 
 MIDDLEWARE = [
+    'concent_api.middleware.HandleServerErrorMiddleware',
```
cameel: Could you add a comment explaining that we're disabling this middleware in `testing.py`? It's not necessarily something that belongs here but I don't know of a better place to put it and it's something that someone tinkering with our middleware settings should be aware of.

concent_api/concent_api/tests/test_middleware.py
```diff
@@ -81,3 +106,229 @@ def test_golem_messages_version_middleware_should_attach_http_header_to_response
...
+
+        is_html_used = should_use_html(request)
+
+        self.assertFalse(is_html_used)
+
+    def test_that_with_acceptance_of_html_with_greater_weight_than_json_in_request_html_is_used(self):
+        request = mock.Mock()
+        request.META = {
+            'headers': {
+                'Accept': ["text/html;q=0.9", "application/json;q=0.8"]
```
cameel: This test must be broken. The implementation has a bug because it checks for `:q=`, not `;q=`. If the test passes then it's too weak.

concent_api/concent_api/middleware.py
cameel: HTML/JSON should be used only if it's present in `Accept` or the `Accept` header is missing. Otherwise HTTP 406 should be returned to the client.

concent_api/concent_api/middleware.py
dybi: done

concent_api/concent_api/middleware.py
dybi: done

concent_api/concent_api/middleware.py
dybi: ok, I will add that

concent_api/concent_api/middleware.py
dybi: good point :) done

concent_api/concent_api/settings/base.py

concent_api/concent_api/settings/testing.py
dybi: done

concent_api/concent_api/settings/base.py

concent_api/concent_api/tests/test_middleware.py
dybi: there was a bug in the tests. Fixed ;)

concent_api/concent_api/middleware.py
dybi: @cameel , are you sure?
1) It is overcomplicating thing from my PoV. What is the benefit here? Do you expect client to handle HTTP406?
2) 
> The HyperText Transfer Protocol (HTTP) `406 Not Acceptable` client error response code indicates that a response matching the list of acceptable values defined in `Accept-Charset` and `Accept-Language` cannot be served.

So in my opinion we would be using `HTTP406` not according to its purpose.

concent_api/concent_api/middleware.py
dybi: done

concent_api/concent_api/middleware.py
dybi: I think it will give standard Django HTML error page

concent_api/concent_api/middleware.py
dybi: there is no point in asserting that `selected_method` (now `selected_content_type`) contains only alphanumeric characters - `get_probability` (`get_weight`)

dybi: This is bascially checkin the format of request header. It should not be done in InternalServerErrorHandlingMiddleware. It should be validated on `process_request` step by different entity (middleware)

concent_api/concent_api/middleware.py
cameel: > It is overcomplicating thing from my PoV. What is the benefit here? Do you expect client to handle HTTP406?

If you're implementing a client that can process XML and you tell the server explicitly that you want XML and nothing else, you want it to either give you XML or give you an error. Just giving it JSON is not helpful (even though it's allowed by the spec). If the client did not care about the format, then why didn't it just skip `Accept` header or add `*/*` to it?

I mean, it's not like we'll be running into this case constantly but for me the default thing to do is to be strict about input because it makes it easier to spot your mistakes. I think that the opposite requires justification. It would be a bit like just ignoring invalid input. The difference between returning this error and not doing it is a single `if` so I don't see any reason not to do it.

> So in my opinion we would be using HTTP406 not according to its purpose.

I've seen this on MDN too but the RFC says something different. It lists `Accept` too.

https://tools.ietf.org/html/rfc7231#section-6.5.6
> The 406 (Not Acceptable) status code indicates that the target
>    resource does not have a current representation that would be
>    acceptable to the user agent, according to the proactive negotiation
>    header fields received in the request (Section 5.3), and the server
>    is unwilling to supply a default representation.

https://tools.ietf.org/html/rfc7231#section-5.3
> ```
>    +-------------------+---------------+
>    | Header Field Name | Defined in... |
>    +-------------------+---------------+
>    | Accept            | Section 5.3.2 |
>    | Accept-Charset    | Section 5.3.3 |
>    | Accept-Encoding   | Section 5.3.4 |
>    | Accept-Language   | Section 5.3.5 |
>    +-------------------+---------------+
> ```

concent_api/concent_api/middleware.py
cameel: Could you please check that? I think it won't. Sorry for not doing it myself, but I think Radek will prefer if I take care of the error code review first ;)

concent_api/concent_api/middleware.py
cameel: Then there should be at least an `assert` for this. This middleware basically depends on that other middleware doing this. A dependency like this is something that will break very easily and no one will notice if we don't have a sanity check enforcing this.

If someone removes the other middleware in settings I think we would much prefer to get `AssertionError` than just for it to stop working silently. An `assert` is a quick and simple solution.

cameel: Of course there is - if you implement my suggestion and create a regex.

If I see a regex with an embedded variable like this, it looks like a big, gaping hole in your code. You can force me to analyze the code around it and waste time finding out that it's not possible for anything bad to happen here. Or you can just add a comment explaining that there's nothing to worry about because the only possible values can't have any special characters. Or - even better - you can add an `assert` which won't easily go out of date like a normal comment.

concent_api/concent_api/middleware.py
```diff
@@ -39,3 +49,81 @@ def __call__(self, request):
...
+    Function inspects request headers and determines is response should be in HTML format.
+    1) By default HTML should not be used (JSON is the default format)
+    2) If `text/html` is present in 'Accept' header and 'application/json' is not, HTML should be used
+    3) If both are present, HTML will be used if it has greater weight than JSON
+    """
+    def get_weight(selected_content_type, accepted_content_types):
+        for accepted_content_type in accepted_content_types.split(','):
+            if selected_content_type in accepted_content_type:
+                match = re.search(f"(?:{selected_content_type})(;q=(?P<weight>0.\d))?", accepted_content_type)
+                if match is not None and match.group('weight') is not None:
```
cameel: I don't think `(?P<weight>0.\d)` is correct:
- `.` will match any character, not just a dot. Should be `\.`.
- It won't match `q=1` and `q=0`.

BTW, here's the spec for the weight: https://tools.ietf.org/html/rfc7231#section-5.3.1

> ```
>      weight = OWS ";" OWS "q=" qvalue
>      qvalue = ( "0" [ "." 0*3DIGIT ] )
>             / ( "1" [ "." 0*3("0") ] )
> ```

concent_api/concent_api/middleware.py
```diff
@@ -39,3 +49,81 @@ def __call__(self, request):
...
+                    return float(match.group('weight'))
+        return 1
+
+    def get_html_weight():
+        if 'headers' in request.META and 'Accept' in request.META['headers'] and \
+                any("text/html" in accepted for accepted in request.META['headers']['Accept'].split(',')):
+            return get_weight("text/html", request.META['headers']['Accept'])
+        return 0
```
cameel: I think the `any("text/html" in accepted for accepted in request.META['headers']['Accept'].split(','))` part is superflous here. It may (insignificantly) speed things up when `text/html` is not present. It may slow things up when it is. I think this micro optimization is not really worth it.

concent_api/concent_api/middleware.py
```diff
@@ -39,3 +49,81 @@ def __call__(self, request):
...
+    2) If `text/html` is present in 'Accept' header and 'application/json' is not, HTML should be used
+    3) If both are present, HTML will be used if it has greater weight than JSON
+    """
+    def get_weight(selected_content_type, accepted_content_types):
+        for accepted_content_type in accepted_content_types.split(','):
+            if selected_content_type in accepted_content_type:
+                match = re.search(f"(?:{selected_content_type})(;q=(?P<weight>0.\d))?", accepted_content_type)
+                if match is not None and match.group('weight') is not None:
+                    return float(match.group('weight'))
+        return 1
```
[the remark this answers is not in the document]
dybi: when selected content type is not present, `get_weight()` function will not be called; please check `get_html_weight` and `get_json_weight` implementations

concent_api/concent_api/settings/base.py
```diff
@@ -290,3 +291,6 @@
 
 # URL format: 'protocol://<user>:<password>@<hostname>:<port>/<virtual host>'
 # CELERY_BROKER_URL = ''
+
+# Debug setting for adding stack traces in HTTP500 responses
+# DEBUG_INFO_IN_ERROR_RESPONSES =
```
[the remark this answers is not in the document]
dybi: It is not inline with `pep8` and the rest of `base.py`

concent_api/concent_api/middleware.py
```diff
@@ -39,3 +49,62 @@ def __call__(self, request):
         response = self.get_response(request)
         response['Concent-Version'] = self._concent_version
         return response
+
+
+def determine_return_type(request_meta):
+    try:
+        preffered_types = ['application/json', 'text/html']
```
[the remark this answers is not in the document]
dybi: done

concent_api/concent_api/middleware.py
```diff
@@ -39,3 +49,62 @@ def __call__(self, request):
...
+        if "HTTP_ACCEPT" in request_meta:
+            return best_match(preffered_types, request_meta['HTTP_ACCEPT'])
+        return "application/json"
+    except Exception:  # pylint: disable=broad-except
+        return ""
```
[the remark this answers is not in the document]
dybi: No. The only thing that can fail is the call to `best_match`. What would you like to fix if it fails? It can fail only if there is something wrong with `HTTP_ACCEPT` and in such case we return `HTTP406` (see `HandleServerErrorMiddleware.process_exception`).

Imagine the processing:
1) Some kind of unhandled internal server error occurs.
2) It is intercepted by the middleware
3) Best return type is being determined
4) `HTTP_ACCEPT` header is somehow broken
5) We return `HTTP406`

In your case:
5") We fail in unhandled error handling middleware.
Is it what we really want? If so, I don't get the point of havin this new middleware.

concent_api/concent_api/middleware.py
dybi: Ok, I finally managed to preserve tha behaviour. Done.

concent_api/concent_api/tests/test_middleware.py
```diff
@@ -1,19 +1,45 @@
...
 
+class CustomException(Exception):
+    pass
+
+
+class CustomExceptionWithStringRepr(Exception):
+    def __init__(self, error_message, error_code):
+        super().__init__()
+        self.message = error_message
+        self.error_code = error_code
```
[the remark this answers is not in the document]
dybi: It is a dummy `Exception` creted for tests only. You are nitpicking.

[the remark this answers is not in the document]
cameel: Oh, you're right. Sorry. I wanted to get this reviewed for you for today and in a hurry I must have taken it for the exception type we're using to pass around the error code.

concent_api/concent_api/settings/base.py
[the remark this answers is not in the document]
cameel: Well, it used to be that way in `base.py` and every where in our code but now I see that newly added comments do not follow that.

I'm going to propose a change to our style then.

concent_api/concent_api/middleware.py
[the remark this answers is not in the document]
cameel: But does `best_match()` raise just plain `Exception`? Doesn't it have any specific error type? You should be catching only that type.

Also, the handler - especially such a broad handler - should have the smallest scope possible to avoid accidentally catching other errors. So if the only thing you think is going to fail is `best_match()` the handler should cover it and nothing else. Not even the `if` around it.

This handler might make sense if you just logged the error (including sending a report to Sentry) and re-raised immediately after because a crash here will not go through the usual error handling Django does for a request.

> Is it what we really want? If so, I don't get the point of havin this new middleware.

Yes. Exactly. Well, I see that the real problem here is that we have fundamentally different approaches to error handling. So let me explain mine:

There are two types of errors: failures and bugs.
1. A **failure** is dependent on external conditions and is inherent in the design of the system. For example running out of disk space, connection error due to server being down, timeout, etc. Failure is inevitable. The final product will always have to deal with failures. We want to detect them and handle them gracefully, though most of the time all we can do is to catch the error and display a message. You cannot predict a failure but you can predict the types of failures that the design allows for and be prepared to handle them.
2. A **bug** is a defect in the code or design. It can be corrected. It should be caught as soon as possible and ideally the final product should contain no bugs. That's rarely possible but the number of bugs left is inversely correlated with the amount of testing the product goes through you do so you can at least work towards eliminating them. But for the testing to be effective it must be easy to detect errors. That's why we want the product to fail loudly whenever anything seems out of ordinary. That's also why I'm so keen on `assert` - you can use it to turn some discrepancy that would go unnoticed into a nice crash that can't be ignored.

My approach is simple: every crash is a bug unless proven otherwise. If you get an exception it's either one of a few specific types indicating a failure that you know you have to handle or something unexpected that should be let through.

During development and Q&A the code goes through exhaustive testing and this approach means that the most egregious bugs are quickly identified and fixed.

Yeah, this means that if we don't find a bug in that phase then some unfortunate client might run into it and get a crash instead of a useful response. But while this may not be good for that client, I think it's best for all clients overall. The damage is limited because we're notified about the crash immediately and fix it in the next release. The alternative is that we silence the problem for the sake of that one client even though we don't really know what what's wrong and other clients keep running into it time and time again without us knowing anything about it.

That's why for me `except Exception:` or `except:` are pretty much a bad smell in the code. They immediately stand out in the review. There are rare cases when their use is warranted but most of the time when I see it it's someone trying to handle failures in a way that unnecessarily silences exceptions caused by bugs. That's why pylint even has the `broad-except` warning. If it wasn't overused it would not be warning about it :)

Sorry for stopping this pull request but I'm quite adamant about this approach. If you don't agree we'll have to talk it over.

concent_api/concent_api/middleware.py
cameel: Great :)

concent_api/concent_api/middleware.py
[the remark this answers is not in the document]
dybi: @cameel , well, I agree to a point. Yes, I can admit, we can divide the perfect world to:
1) bugs
2) failures,
but in my opinion there is some middle ground. An example from our world: we are trying to to get Subtask out of DB at point of the flow when it should be there. Biut for whatever reason it is not. Is it a **bug**? Might be. Or maybe some has been tinkering with the DB (and that caused a **failure**)?

My approach is to deliver the best service to the customers. If we cannot do anything sensible with some unexpected situation, we should fail. But if we can, why should we? Even if there is some bug, we have logging and can use it to find the root cause and provide the fix. And the customer doesn't have to be affected.

This discussion should take place in some meeting with the broader audience.

Getting back to this particular situation:
1) As I wrote before, the only entity that can raise exeption is `best_match` function
2) Having this `try... except` block as it is, is IMHO the most readable way I can think of.
3) As `best_match` is third party function, I try to handle all exceptions it may raise.
4) Example with `SyntaxError` doesn't seem to be adequate here - with modern tools and our `lint`, `fullcheck`, etc. script it is **very** unlikely to get it and not realise it.
5) Function, by the design, should be able to determine which return type should be used (in `Java` I would call this function `throw-safe`)

To sum up. If my arguments do not convince you, then this task should remain opened until the more in-depth discussion can take place. Or someone can take it up from here and implement the desired way (what would it look like, btw?). At the end of the day, you are the architect, so you are the oracle.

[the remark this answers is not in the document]
cameel: So we'll have to talk about it because I unfortunately cannot fully agree.

I think it's safer to err on the side of crashing too much. From the code I've seen in various projects I think people have a tendency to catch too much and to treat bugs like failures rather than the other way around.

> but in my opinion there is some middle ground. An example from our world: we are trying to to get Subtask out of DB at point of the flow when it should be there. Biut for whatever reason it is not. Is it a bug? Might be. Or maybe some has been tinkering with the DB (and that caused a failure)?

Definitely a bug as long as the database is under the control of our app. I would agree if there was another client indepently accessing the database but in Django apps that's pretty clear cut for me. They're so tightly tied to their database structure that I consider it fully under their control.

There is some middle ground and that's why there are rare cases where a handler for `Exception` is appropriate. But they are really rare. Like, there might be one or two cases in our whole application (probably in error handling decorators).

> My approach is to deliver the best service to the customers. If we cannot do anything sensible with some unexpected situation, we should fail. But if we can, why should we? 

Because if you catch `Exception` you don't know what you're dealing with. In many situations the safest thing to do is actually to fail and terminate rather than go on and do something wrong. That's my approach - try to make the best of the situation when you can. But don't try to guess when you don't have a clue what's happening.

Simply, it's fine to catch specific exceptions from `best_match()`. Not `Exception`.

dybi:
merging, as discussed (at last!) ;)

PR #369, opened 2018-05-13
Fix calculate svt manually

rwrzesien:
Temporary solution before https://github.com/golemfactory/golem-messages/issues/217 is implemented.

Solves calculation of `subtask_verification_time` using Concent settings instead of `golem-messages` constants.


concent_api/api-integration-force-accept-or-reject-results-test.py
```diff
@@ -69,20 +72,37 @@ def report_computed_task(timestamp = None, task_to_compute = None):
     with freeze_time(timestamp):
         return message.tasks.ReportComputedTask(
             task_to_compute = task_to_compute,
-            size=10,
+            size=REPORT_COMPUTED_TASK_SIZE,
         )
 
 
+def calculate_timestamp(current_time, cluster_consts):
```
dybi: acatually, you need just `concent_messaging_time`, please change function signature (for all three functions)

concent_api/api-integration-force-accept-or-reject-results-test.py
```diff
@@ -232,13 +252,13 @@ def test_case_2c_wrong_timestamps(cluster_consts, cluster_url, test_id):
...
-                timestamp=timestamp_to_isoformat(current_time - (SUBTASK_VERIFICATION_TIME * 1.4)),
+                timestamp=timestamp_to_isoformat(current_time),
                 report_computed_task=report_computed_task(
                     task_to_compute=create_signed_task_to_compute(
-                        timestamp=timestamp_to_isoformat(current_time - SUBTASK_VERIFICATION_TIME * 1.5),
+                        timestamp=calculate_timestamp(current_time, cluster_consts),
                         task_id=task_id,
                         subtask_id=subtask_id,
-                        deadline=current_time - (SUBTASK_VERIFICATION_TIME * 20),
+                        deadline=current_time - (1 + ((20 * cluster_consts.concent_messaging_time) + 183)),
```
dybi: Could you please use a new function here? It is hard do figure out why such numbers...

concent_api/api-integration-force-accept-or-reject-results-test.py
```diff
@@ -266,13 +286,13 @@ def test_case_2b_not_enough_funds(cluster_consts, cluster_url, test_id):
...
-                timestamp=timestamp_to_isoformat(current_time - (SUBTASK_VERIFICATION_TIME * 1.4)),
+                timestamp=timestamp_to_isoformat(current_time),
                 report_computed_task=report_computed_task(
                     task_to_compute=create_signed_task_to_compute(
-                        timestamp=timestamp_to_isoformat(current_time - SUBTASK_VERIFICATION_TIME * 1.5),
+                        timestamp=calculate_timestamp(current_time, cluster_consts),
                         task_id=task_id,
                         subtask_id=subtask_id,
-                        deadline=current_time - (SUBTASK_VERIFICATION_TIME),
+                        deadline=current_time - (1 + ((4 * cluster_consts.concent_messaging_time) + 183)),  # svt 40183
```
dybi: Same here - please try to express this via new `calculate_deadline`

concent_api/api-integration-force-accept-or-reject-results-test.py
rwrzesien: This one calculates a different value. If you look at test description and this calculation it becomes clear.

concent_api/api-integration-force-accept-or-reject-results-test.py
rwrzesien: :+1: done

concent_api/api-integration-force-accept-or-reject-results-test.py
rwrzesien: :+1: done

concent_api/api-integration-force-accept-or-reject-results-test.py
dybi: would it be possible to have it like ~:
`deadline=calculate_deadline(current_time, cluster_consts.concent_messaging_time) + SOME_OFFSET`? Or even extract it to function: `calculate_wrong_timestamp` or sth similar?
"readibility counts" ;) (https://www.python.org/dev/peps/pep-0020/)

concent_api/utils/helpers.py
```diff
@@ -165,10 +164,21 @@ def calculate_maximum_download_time(size: int) -> int:
...
+    )
+    subtask_timeout = subtask_dt - ttc_dt
+
     return int(
-        subtask_verification_time(
-            report_computed_task
-        ).total_seconds()
+        (4 * settings.CONCENT_MESSAGING_TIME) +
+        (3 * mdt.total_seconds()) +
+        (0.5 * subtask_timeout.total_seconds())
```
cameel: Could you add a comment explaining that we're not using `subtask_verification_time()` from golem-messages only because it has hard-coded values and we can't make it use values from our settings? And that it's only a temporary solution. Adding a link to the issue in golem-messages would be a good idea too.

Issue #215, opened 2018-03-26
Concent: Handling verification_result


cameel:
Create a handler for `verification_result` Celery task.

### Sequence of operation
This is a part of the sequence described in #193 blueprint. All steps below happen inside the Concent worker - a celery worker processing a task from work queue.

1) `verification_result` (see #193) is received from work queue.
2) Worker locks database row corresponding to the subtask in the subtask table
    - If the row is already locked, task fails so that Celery can retry later.
3) Worker checks subtask state
    - A) If subtask is already in `ACCEPTED` state, worker logs a warning and returns. This means that the verification has timed out and a client has already asked about the result. Or that `verification_result` for this subtask must have been already processed.
    - B) If subtask is already in `FAILED` state, worker logs an error and returns. This means that `verification_result` for this subtask must have been already processed.
    - C) If subtask is any state other than `ADDITIONAL VERIFICATION` state, worker sends an error report and returns. This means that something is seriously wrong.
4) Worker checks the current time. If the time is already past `next_deadline` for the subtask (`SubtaskResultsRejected.timestamp + AVCT`) worker ignores worker's message and processes the timeout as in #228.
5) Worker checks verification result
    - A) If `result == MISMATCH`
        - Worker adds `SubtaskResultsRejected` to provider's and requestor's receive queues (both out-of-band)
        - Worker changes subtask state to `FAILED`
    - B) If `result == MATCH` or `result == ERROR`
        - If `result == ERROR`, worker logs the error code and message
        - Worker makes a payment from requestor's deposit just like in the forced acceptance use case.
            - Simulate it for now. Leave actual payment implementation for #229 because that's a part of the code that requires special attention.
        - Worker adds `SubtaskResultsSettled` to provider's and requestor's receive queues (both out-of-band)
        - Worker changes subtask state to `ACCEPTED`
6) Worker releases the database lock

### Task separation
We want to be able to connect three separate workers types to the queue: Concent, Conductor and Verifier. Each of them should have its own task handlers. When you start one of them with `celery worker` command it must not start processing tasks that belong to other worker types.

We'll achieve this by adding a new value to `settings.CONCENT_FEATURES`. Worker's tasks handlers should be registered only if that setting contains `'concent-worker'` string.

### Setup for local development
You can assume that Celery is already configured (#197).

[...]

PR #380, opened 2018-05-15
Feature handling verification result

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/215.

### DIFF
--- a/concent_api/verifier/tasks.py
+++ b/concent_api/verifier/tasks.py
@@ -2,8 +2,23 @@ from enum import Enum
 import logging
 
 from celery import shared_task
+from mypy.types import Optional
 
+from django.db import DatabaseError
+from django.db import transaction
+
+from core.message_handlers import store_pending_message
+from core.models import PendingResponse
+from core.models import Subtask
+from utils.constants import ErrorCode
+from utils.helpers import get_current_utc_timestamp
+from utils.helpers import parse_timestamp_to_utc_datetime
 from .constants import VerificationResult
+from .constants import CELERY_LOCKED_SUBTASK_DELAY
+from .constants import MAXIMUM_VERIFICATION_RESULT_TASK_RETRIES
+from .constants import VERIFICATION_RESULT_SUBTASK_STATE_ACCEPTED_LOG_MESSAGE
+from .constants import VERIFICATION_RESULT_SUBTASK_STATE_FAILED_LOG_MESSAGE
+from .constants import VERIFICATION_RESULT_SUBTASK_STATE_UNEXPECTED_LOG_MESSAGE
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


concent_api/verifier/tasks.py
```diff
@@ -24,28 +35,95 @@ def blender_verification_order(
...
+
+    if subtask.state_enum == Subtask.SubtaskState.ACCEPTED:
+        logger.warning(
+            'This means that the verification has timed out and a client has already asked about the result. '
+            'Or that verification_result for this subtask must have been already processed.'
+        )
+        return
+
+    elif subtask.state_enum == Subtask.SubtaskState.FAILED:
+        logger.error('This means that verification_result for this subtask must have been already processed.')
```
Jakub89: Shouldn't this be `logger.warning()` as well? FAILED SubtaskState can happen because of timeout, which is not really an error.

concent_api/verifier/tasks.py
```diff
@@ -24,28 +35,95 @@ def blender_verification_order(
...
+        return
+
+    elif subtask.state_enum != Subtask.SubtaskState.ADDITIONAL_VERIFICATION:
+        logger.error('This means that something is seriously wrong.')
```
Jakub89: I'm not sure if @cameel meant for those error messages to be so straight forward :P but at least print that wrong state_enum in the message.

concent_api/verifier/tests/test_verification_result_task.py
```diff
@@ -0,0 +1,170 @@
...
+        self.assertEqual(
+            logging_info_mock.call_args_list[1][0][0],
+            f'verification_result_task processing error result with: '
+            f'SUBTASK_ID {self.subtask.subtask_id} -- RESULT {VerificationResult.ERROR} -- ERROR MESSAGE test -- ERROR CODE {ErrorCode.REQUEST_BODY_NOT_EMPTY}'
+        )
+
+    def test_that_verification_result_match_should_add_pending_messages_subtask_results_settled_and_change_subtask_state_to_accepted(self):
+        verification_result(
+            self.subtask.subtask_id,
+            VerificationResult.ERROR,
```
dybi: should be `VerificationResult.MATCH` ;)

concent_api/verifier/tasks.py
dybi: similarily here ;)

concent_api/verifier/tasks.py
```diff
@@ -24,28 +35,95 @@ def blender_verification_order(
...
 
-    logger.info(
-        'verification_result_task ends with: SUBTASK_ID {} -- RESULT {}'.format(
-            subtask_id,
-            result,
+    # Worker locks database row corresponding to the subtask in the subtask table.
+    try:
+        subtask = Subtask.objects.select_for_update(nowait=True).get(subtask_id=subtask_id)
+    except DatabaseError:
+        # If the row is already locked, task fails so that Celery can retry later.
```
dybi: Just an idea: shouldn't we limit the number of retries and/or log the attempts?

cameel: Definitely log it.

As for limiting retries - yeah, we should. It's not stated explicitly in the issue but having a way out of an infinite loop is a good idea :) But don't set the limit too low. It's much, much more likely that some other process is just holding the lock.

concent_api/verifier/tasks.py
cameel: No. Timeout means that provider wins by default. So it goes into `ACCEPTED`.

concent_api/verifier/tasks.py
cameel: It's really one of those "this can never happen but the message comes from a separate component that could have bugs that you never imagined so better safe than sorry" errors :)

concent_api/verifier/tasks.py
rwrzesien: In issue

> B) If subtask is already in FAILED state, worker logs an error and returns. This means that verification_result for this subtask must have been already processed.

So I understood that I should 'log error'. But as you guys think, should it be changed ? Same question to 

```
    elif subtask.state_enum != Subtask.SubtaskState.ADDITIONAL_VERIFICATION:
        logger.error('This means that something is seriously wrong.')
```

concent_api/verifier/tasks.py
rwrzesien: I have copied those messages directly from issue description :) As a placeholders, but yes, they can be changed. I will add wrong state_enum too.

concent_api/verifier/tasks.py
rwrzesien: After thinking through I will change to log.warning if state is FAILED and keep log.error if state is unexpected.

concent_api/verifier/tasks.py
rwrzesien: How would you like to store it ? The only way I see is to add a field on `Subtask`. Is that ok ?

concent_api/verifier/tasks.py
```diff
@@ -24,28 +35,95 @@ def blender_verification_order(
...
+            error_message=error_message,
+            error_code=error_code,
+            countdown=CELERY_LOCKED_SUBTASK_DELAY
         )
-    )
+        return
+
+    if subtask.state_enum == Subtask.SubtaskState.ACCEPTED:
+        logger.warning(
+            'This means that the verification has timed out and a client has already asked about the result. '
```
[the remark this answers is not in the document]
rwrzesien: :+1:  done

concent_api/verifier/tasks.py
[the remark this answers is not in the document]
rwrzesien: :+1:  done

concent_api/verifier/tests/test_verification_result_task.py
rwrzesien: :+1:  done

concent_api/verifier/tasks.py
rwrzesien: :+1:  done

rwrzesien: :+1:  done

concent_api/verifier/tasks.py
rwrzesien: :+1:  done

concent_api/verifier/tasks.py
dybi: another argument passed to `verification_result` (the number of attempt), maybe with default value 0?

rwrzesien: Good idea, lets go this way.

cameel: Celery already stores the number of retries:

[Task.max_retries](http://docs.celeryproject.org/en/latest/reference/celery.app.task.html#celery.app.task.Task.max_retries)
[Task.retry()](http://docs.celeryproject.org/en/latest/reference/celery.app.task.html#celery.app.task.Task.retry)
[Task.apply()](http://docs.celeryproject.org/en/latest/reference/celery.app.task.html#celery.app.task.Task.apply) has a parameter called `retries`. This values is also accessible in some way when you get the task.

rwrzesien: `apply` is the blocking call, so we don't want it, but `retry` and `max_retries` looks good for this case and should be enough.

cameel: There are also retry policies that let you e.g. use increasing intervals between retries. See [Message Sending Retry](http://docs.celeryproject.org/en/latest/userguide/calling.html#message-sending-retry)

cameel: > `apply` is the blocking call, so we don't want it

I was just an example. `apply_async()` has it too.

rwrzesien:
@Jakub89 @cameel 

> There's a comment in this issue:
> "Update: Celery tasks handlers should be registered only if settings.CONCENT_FEATURES contains 'concent-worker' string."
> What about it?

I have missed this one, but it can't be done this way if this app will contain other tasks used by other `AVAILABLE_CONCENT_FEATURES` (currently it has `blender_verification_order` loaded if `verifier` is in `CONCENT_FEATURES`). If we want it to be loaded separately we have to:
1. Make it a separate django app and available concent feature, or
2. define exact list of tasks to be loaded for each available feature. But then our custom loader would the have to go through all the apps, and check all `tasks.py` files for those names. It can be problematic, lead to unexpected results if there will be same names, and difficult to maintain. So I suggest option 1.

What do you guys think ?

dybi:
@rwrzesien, for me 1) seems better. Seems like easier/cheaper to maintain option.

PR #383, opened 2018-05-15
[WIP][M] Add tables for each golem message

kbeker:
[WIP] conceptual PR to make sure that everything goes in good way


concent_api/core/models.py
```diff
@@ -485,3 +493,478 @@ def clean(self):
...
+        'concent_enabled',
+        'price',
+        'header',
+        'sig',
+    ]
+
+    requestor_id = BinaryField()
+    requestor_public_key = BinaryField()
+    requestor_ethereum_public_key = BinaryField()
+    provider_id = BinaryField()
```
cameel: Please make sure that we can do queries like finding `TaskToCompute` with a specific requestor key. I'm not sure if `BinaryField` can be used in queries. If not, we'll have to find a workaround.

concent_api/core/models.py
```diff
@@ -485,3 +493,478 @@ def clean(self):
...
+    provider_public_key = BinaryField()
+    provider_ethereum_public_key = BinaryField()
+    compute_task_def = ForeignKey(StoredComputeTaskDef)
+    package_hash = CharField(max_length=HASH_LENGTH)
```
cameel: You should include the prefix length here.

If it's already included in `HASH_LENGTH`, please make it a separate constant.

concent_api/core/models.py
```diff
@@ -485,3 +493,478 @@ def clean(self):
...
+    concent_enabled = BooleanField(default=False)
+    price = PositiveIntegerField()
+
+    if not len(ITEMS) == len(message.tasks.TaskToCompute().__slots__):
+        raise ValidationError({
+            'StoredTaskToCompute': 'StoredTaskToCompute model has not same fields as golem_messages.message.tasks.TaskToCompute()'
+        })
```
cameel: This shouuld be an `assert`.

concent_api/core/models.py
```diff
@@ -485,3 +493,478 @@ def clean(self):
...
+
+    for model_item, golem_message_item in zip(ITEMS, message.tasks.TaskToCompute().__slots__):
+        if not model_item == golem_message_item:
+            raise ValidationError({
+                'StoredTaskToCompute': 'StoredTaskToCompute model has not same fields as golem_messages.message.tasks.TaskToCompute()'
```
cameel: This should be an `assert`.

concent_api/core/models.py
```diff
@@ -485,3 +493,478 @@ def clean(self):
...
+            })
+
+    def clean(self):
+        if not len(self.requestor_id) == GOLEM_PUBLIC_KEY_LENGTH or not isinstance(self.requestor_id, bytes):
+            raise ValidationError({
+                'requestor_id': f'requestor_id must be bytes instance and must have length equal to {GOLEM_PUBLIC_KEY_LENGTH}'
+            })
```
cameel: Since it's `BinaryField`, the `not isinstance(self.requestor_id, bytes)` can't be false - it should be an `assert`.

Same in other validations below.

Also, this is a bit repetitive. You could check the lengths of all those fields in a loop.

concent_api/core/models.py
```diff
@@ -485,3 +493,478 @@ def clean(self):
...
+
+        if not len(self.provider_ethereum_public_key) == GOLEM_PUBLIC_KEY_LENGTH or not isinstance(self.provider_ethereum_public_key, bytes):
+            raise ValidationError({
+                'provider_ethereum_public_key': f'provider_ethereum_public_key must be bytes instance and must have length equal to {GOLEM_PUBLIC_KEY_LENGTH}'
+            })
+
+        if not self.package_hash[:4] == HASH_FUNCTION:
+            raise ValidationError({
+                'package_hash': f'package_hash should be hashed with {HASH_FUNCTION} function'
+            })
```
cameel: You're not checking that there's a semicolon between the prefix and the hash.

Please look at the checksum validations in gatekeeper.

concent_api/core/models.py
```diff
@@ -485,3 +493,478 @@ def clean(self):
...
+
+        if not len(self.package_hash) == HASH_LENGTH:
+            raise ValidationError({
+                'package_hash': f'package_hash should has length equal to {HASH_LENGTH}'
+            })
+
+        if not isinstance(self.concent_enabled, bool):
+            raise ValidationError({
+                'concent_enabled': 'concent_enabled should be instance of boolean'
+            })
```
cameel: Checking types is not necessary in model validations. The fields already have types defined and Django/database will enforce them.

concent_api/core/models.py
```diff
@@ -485,3 +493,478 @@ def clean(self):
...
+
+        if not isinstance(self.price, int) or not self.price >= 0:
+            raise ValidationError({
+                'price': 'Price must be an integer and bigger than or equal 0'
+            })
```
cameel: This validation is not necessary. You're using `PositiveIntegerField`.

Actually, the field probably does not allow `0`. Is this the right type then?

concent_api/core/models.py
```diff
@@ -485,3 +493,478 @@ def clean(self):
...
+
+    if not len(ITEMS) == len(message.tasks.ComputeTaskDef().ITEMS.keys()):
+        raise ValidationError({
+            'ComputeTaskDef': 'StoredComputeTaskDef model has not same fields as golem_messages.message.tasks.ComputeTaskDef()'
+        })
+
+    for model_item, golem_message_item in zip(ITEMS, message.tasks.ComputeTaskDef().ITEMS.keys()):
+        if not model_item == golem_message_item:
+            raise ValidationError({
+                'ComputeTaskDef': 'StoredComputeTaskDef model has not same fields as golem_messages.message.tasks.ComputeTaskDef()'
```
cameel: I think it would be better to replace both of those checks with one that creates sets of field names and compares them:

``` python
assert set(ITEMS) == set(message.tasks.ComputeTaskDef().ITEMS):
```

concent_api/core/models.py
```diff
@@ -485,3 +493,478 @@ def clean(self):
...
+        abstract = True
+
+
+class StoredComputeTaskDef(AbstractMessageModel):
+
+    # Defines items list from ComputeTaskDef
+    ITEMS = [
+        'task_id',
+        'subtask_id',
+        'deadline',
```
cameel: You're not checking in any way that this list contains the names of fields defined in the model. Someone can just come here, ignore your list, add a new field to the model below and the checks will still pass.

Instead of defining `ITEMS` here you should be comparing slots in golem-messages with the list of model fields obtained from [get_fields()](https://docs.djangoproject.com/en/2.0/ref/models/meta/#django.db.models.options.Options.get_fields).

concent_api/core/models.py
```diff
@@ -485,3 +493,478 @@ def clean(self):
...
+    for model_item, golem_message_item in zip(ITEMS, message.tasks.CannotComputeTask().__slots__):
+        if not model_item == golem_message_item:
+            raise ValidationError({
+                'StoredCannotComputeTask': 'StoredCannotComputeTask model has not same fields as golem_messages.message.tasks.CannotComputeTask()'
+            })
+
+    class CannotComputeTaskReason(ChoiceEnum):
+        WrongCTD = 'WrongCTD'
+        WrongKey = 'WrongKey'
+        WrongAddress = 'WrongAddress'
```
cameel: Isn't this enum already defined in golem-messages? We should not be copying it here.

If you just want to use `.choices()` on it, you should just make a separate function (or a one-liner) that prepares the list of choices for the model instead of repeating the whole enum.

concent_api/core/models.py
```diff
@@ -485,3 +493,478 @@ def clean(self):
...
+        })
+
+    for model_item, golem_message_item in zip(CannotComputeTaskReason.__members__, message.tasks.CannotComputeTask.REASON.__members__):
+        if not model_item == golem_message_item:
+            raise ValidationError({
+                'CannotComputeTaskReason': 'CannotComputeTaskReason Enum has not same fields as golem_messages.message.tasks.CannotComputeTask.REASON'
+            })
+
+    task_to_compute = ForeignKey(StoredTaskToCompute)
+    reason = CharField(max_length=32, choices=CannotComputeTaskReason.choices())
```
cameel: Why 32? You should create a function that computes the maximum possible length for this enum and call it here.

concent_api/core/models.py
```diff
@@ -485,3 +493,478 @@ def clean(self):
...
+
+    for model_item, golem_message_item in zip(ITEMS, message.tasks.ReportComputedTask().__slots__):
+        if not model_item == golem_message_item:
+            raise ValidationError({
+                'StoredReportComputedTask': 'StoredReportComputedTask model has not same fields as golem_messages.message.tasks.ReportComputedTask()'
+            })
+
+    result_type = PositiveSmallIntegerField()
+    node_name = CharField(max_length=32)
+    address = CharField(max_length=15)
```
cameel: There should be constants for these two values. Where do they come from?

concent_api/core/models.py
```diff
@@ -485,3 +493,478 @@ def clean(self):
...
+    node_info = JSONField()
+    port = PositiveIntegerField()
+    key_id = BinaryField()
+    extra_data = JSONField()
+    eth_account = CharField(max_length=32)
+    task_to_compute = ForeignKey(StoredTaskToCompute)
+    size = PositiveIntegerField()
+    package_hash = CharField(max_length=HASH_LENGTH)
+    multihash = CharField(max_length=HASH_LENGTH)
+    secret = TextField()
```
cameel: What's `secret`? How secret is it?

My assumption so far is that there's nothing in these messages that's really a secret (e.g. like a password). If there is, this affects stuff like logging (we'll have to censor it), so please check what exactly does this field store.

concent_api/core/models.py
```diff
@@ -485,3 +493,478 @@ def clean(self):
...
+
+    def clean(self):
+        '''
+        VALIDATION FOR result_type
+        VALIDATION FOR node_name
+        '''
+        try:
+            validate_ipv4_address(self.address)
+        except ValidationError:
+            return ValidationError({
```
cameel: You should use [GenericIPAddressField](https://docs.djangoproject.com/en/2.0/ref/models/fields/#genericipaddressfield). It automatically validates this.

concent_api/core/constants.py
```diff
@@ -10,3 +10,11 @@
 
 # Defines length of Clients ids, public keys or ethereum public keys
 TASK_OWNER_KEY_LENGTH = 64
+
+# Defines hash function used by Concent for file hashing
+HASH_FUNCTION = 'sha1'
```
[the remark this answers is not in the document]
kbeker: At this moment I asked @bartoszbetka which function we're using and just added that one what we have now. But you're right that it should be prepared for extending list of hash function.

concent_api/core/constants.py
```diff
@@ -10,3 +10,11 @@
...
+
+# Defines length of hash
+HASH_LENGTH = 45
```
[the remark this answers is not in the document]
kbeker: Should I then do a dict with a name of hash function, and length to be sure that this is always connected with each other?

concent_api/core/models.py
```diff
@@ -485,3 +493,478 @@ def clean(self):
...
+        'extra_data',
+        'short_description',
+        'working_directory',
+        'performance',
+        'docker_images',
+    ]
+
+    task_id = CharField(max_length=MESSAGE_TASK_ID_MAX_LENGTH)
+    subtask_id = CharField(max_length=MESSAGE_TASK_ID_MAX_LENGTH)
+    deadline = DateTimeField(blank = True, null = True)
```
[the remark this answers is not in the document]
kbeker: For this moment it is still rough draft. It of course won't be null :)

concent_api/core/models.py
```diff
@@ -485,3 +493,478 @@ def clean(self):
...
+    src_code = TextField()
+    extra_data = JSONField()
+    short_description = TextField()
+    working_directory = TextField()
```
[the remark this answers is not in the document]
kbeker: Because in CharField we have to specify how long this field will be. With TextField() we don't have this problem, becuase we don't know how long it can be we can just assume some number

concent_api/core/models.py
```diff
@@ -485,3 +493,478 @@ def clean(self):
...
+    performance = DecimalField(decimal_places=2, max_digits=3)
```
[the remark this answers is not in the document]
kbeker: Not yet. I haven't fixed it after answer from Darek.

concent_api/core/models.py
```diff
@@ -485,3 +493,478 @@ def clean(self):
...
+    docker_images = JSONField()
```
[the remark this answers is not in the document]
kbeker: So every list should have separate model where we will store every item from this list?

concent_api/core/models.py
```diff
@@ -485,3 +493,478 @@ def clean(self):
...
+
+    if not len(ITEMS) == len(message.tasks.ComputeTaskDef().ITEMS.keys()):
+        raise ValidationError({
+            'ComputeTaskDef': 'StoredComputeTaskDef model has not same fields as golem_messages.message.tasks.ComputeTaskDef()'
+        })
```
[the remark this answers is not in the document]
kbeker: This validation starts when we start django server. If any message would change their structure in golem_message repo we will know it from beginning. It can be also assert and as I saw below we can do it with `set() == set()`

concent_api/core/models.py
kbeker: 15 is maximum length of IPv4 address -> 192.168.128.100 -> This gives exactly 15 chars. I'll add it to constants. Node name is not precised yet

concent_api/core/models.py
kbeker: It will be BinaryField with max length equals 32.

concent_api/core/models.py
rwrzesien: But it will require new migration each time a new entry is added to enum extending current max value. We already discuss it, it can't be done this way.

concent_api/core/models.py
```diff
@@ -485,3 +493,478 @@ def clean(self):
...
+        'sig',
+    ]
+
+    requestor_id = BinaryField()
+    requestor_public_key = BinaryField()
+    requestor_ethereum_public_key = BinaryField()
+    provider_id = BinaryField()
+    provider_public_key = BinaryField()
+    provider_ethereum_public_key = BinaryField()
+    compute_task_def = ForeignKey(StoredComputeTaskDef)
```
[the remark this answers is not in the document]
rwrzesien: > But I think you should just add all the fields from ComputeTaskDef directly to this model. There isn't any gain in keeping them separate.

You can attach one ComputeTaskDef to many TaskToCompute :)

concent_api/core/models.py
```diff
@@ -485,3 +493,478 @@ def clean(self):
             raise ValidationError({
                 'pending_response': 'PaymentInfo should be related with Pending Response'
             })
+
+
+class AbstractMessageModel(Model):
+
+    sig = BinaryField()
+    timestamp = IntegerField()
```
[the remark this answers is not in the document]
rwrzesien: We can store type in extra field, and store value as a string. Better way would be to create a custom field holding those two information together and automatically handling casting (similar to our Base64Field)

concent_api/core/models.py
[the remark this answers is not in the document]
rwrzesien: So actually it should be function containing assert(s), I think, as this validation is repeated in every class, it should be extracted to function.

concent_api/core/models.py
rwrzesien: No, you can't query `BinaryField` but you can use super awesome feature called `Base64Field` ;)

concent_api/core/models.py
rwrzesien: Yes, lets make it similar to what we wanted to do with system checks, reusable atomic generic validators.

concent_api/core/models.py
rwrzesien: I think the reason for that is the same as in Karol's answer:

> This validation starts when we start django server. If any message would change their structure in golem_message repo we will know it from beginning.

kbeker:
Closing because it is now outdated and waits for response from Golem Team

Issue #228, opened 2018-03-28
Concent: verification timeout


cameel:
This is a part of the implementation of the #193 blueprint. 

### Sequence of operation
All steps below happen inside Concent views:

1) Provider or requestor contacts Concent (any view)
2) Concent runs that code that checks for timeouts
    - If there's a task in `ADDITIONAL VERIFICATION` state and it's already past the deadline:
        - Concent makes a payment from requestor's deposit just like in the forced acceptance use case.
            - Simulate it for now. Leave actual payment implementation for #229 because that's a part of the code that requires special attention.
        - Concent adds `SubtaskResultsSettled` to provider's and requestor's receive queues (both out-of-band)
        - Concent changes its state to `ACCEPTED`

Use database locks on the `Subtask` row corresponding to the task being processed to avoid race conditions between simultaneous requests or a request handler and a worker.

PR #385, opened 2018-05-16
Feature verification timeout

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/228.

### DIFF
--- a/concent_api/core/subtask_helpers.py
+++ b/concent_api/core/subtask_helpers.py
@@ -66,7 +66,25 @@ def update_timed_out_subtasks(
                 queue               = PendingResponse.Queue.ReceiveOutOfBand,
                 subtask             = subtask,
             )
-        assert subtask.state is not Subtask.SubtaskState.ADDITIONAL_VERIFICATION.name  # pylint: disable=no-member
+        elif subtask.state == Subtask.SubtaskState.ADDITIONAL_VERIFICATION.name:  # pylint: disable=no-member
+            # TODO: Make payment from requestor's deposit just like in the forced acceptance use case.
+            locked_subtask = Subtask.objects.select_for_update().get(pk=subtask.pk)
+            update_subtask_state(
+                subtask                 = locked_subtask,
+                state                   = Subtask.SubtaskState.ACCEPTED.name,  # pylint: disable=no-member
+            )
+            store_pending_message(
+                response_type       = PendingResponse.ResponseType.SubtaskResultsSettled,
+                client_public_key   = locked_subtask.provider.public_key_bytes,
+                queue               = PendingResponse.Queue.ReceiveOutOfBand,
+                subtask             = locked_subtask,
+            )
+            store_pending_message(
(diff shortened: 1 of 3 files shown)


concent_api/core/subtask_helpers.py
```diff
@@ -66,7 +66,25 @@ def update_timed_out_subtasks(
                 queue               = PendingResponse.Queue.ReceiveOutOfBand,
                 subtask             = subtask,
             )
-        assert subtask.state is not Subtask.SubtaskState.ADDITIONAL_VERIFICATION.name  # pylint: disable=no-member
+        elif subtask.state == Subtask.SubtaskState.ADDITIONAL_VERIFICATION.name:  # pylint: disable=no-member
+            # TODO: Make payment from requestor's deposit just like in the forced acceptance use case.
+            locked_subtask = Subtask.objects.select_for_update().get(pk=subtask.pk)
```
dybi: I am not sure if I understand why we use `select_for_update` just here. Could you please explain?

rwrzesien: @dybi According to issue description

> Use database locks on the Subtask row corresponding to the task being processed to avoid race conditions between simultaneous requests or a request handler and a worker.

cameel: We do need locks in more places. That's one I stated explicitly in the issue but I have more issues about that in the pipeline. If you see another place here were we would need a lock, it can be added too.

concent_api/core/tests/test_integration_subtask_results_verify.py
```diff
@@ -337,6 +340,65 @@ def test_that_concent_accepts_valid_request_and_sends_verification_order_to_work
...
+        )
+        self._test_undelivered_pending_responses(
+            subtask_id=subtask.subtask_id,
+            client_public_key=self._get_encoded_requestor_public_key(),
+            client_public_key_out_of_band=self._get_encoded_requestor_public_key(),
+            expected_pending_responses_receive_out_of_band=[
+                PendingResponse.ResponseType.SubtaskResultsSettled,
+            ]
+        )
+
```
Jakub89: I think we should add `receive_out_of_band` responses for both provider and requestor to this test. Since those messages are in database undelivered?

concent_api/core/tests/test_integration_subtask_results_verify.py
```diff
@@ -337,6 +340,97 @@ def test_that_concent_accepts_valid_request_and_sends_verification_order_to_work
...
+            response_2,
+            status          = 200,
+            key             = self.REQUESTOR_PRIVATE_KEY,
+            message_type    = message.concents.SubtaskResultsSettled,
+            fields          = {
+                'task_to_compute': self.report_computed_task.task_to_compute,
+            }
+        )
+
+        response_2 = self.client.post(
```
dybi: Maybe `response_3`? Just not to ovverride existing one.

Issue #271, opened 2018-04-13
Implement sending blender_verification_request


dybi:
Extracted part from: https://github.com/golemfactory/concent/pull/251#discussion_r179715707

Jakub89:
@cameel where does `source_package_path` and `result_package_path` initially should come from, when we want to call  `blender_verification_request`? From `compute_task_def['extra_data']` like `scene_file` and `output_format`?

PR #391, opened 2018-05-16
Add real blender_verification_request in message_handlers

Jakub89:
Resolves: #271

### DIFF
--- a/concent_api/core/queue_operations.py
+++ b/concent_api/core/queue_operations.py
@@ -1,2 +1,19 @@
-def send_blender_verification_request():  # TODO implement sending real verivication_request to Work Queue
-    pass
+from conductor.tasks import blender_verification_request
+from utils.helpers import get_storage_result_file_path
+from utils.helpers import get_storage_source_file_path
+
+
+def send_blender_verification_request(compute_task_def):
+    task_id = compute_task_def['task_id']
+    subtask_id = compute_task_def['subtask_id']
+    source_package_path = get_storage_source_file_path(subtask_id, task_id)
+    result_package_path = get_storage_result_file_path(subtask_id, task_id)
+    output_format = compute_task_def['extra_data']['output_format']
+    scene_file = compute_task_def['extra_data']['scene_file']
+    blender_verification_request.delay(
+        subtask_id=subtask_id,
+        source_package_path=source_package_path,
+        result_package_path=result_package_path,
+        output_format=output_format,
+        scene_file=scene_file,
+    )
--- a/concent_api/core/tests/test_integration_subtask_results_verify.py
+++ b/concent_api/core/tests/test_integration_subtask_results_verify.py
@@ -8,11 +8,15 @@ from freezegun import freeze_time
 
 from golem_messages import message, load
 
+from conductor.models import BlenderSubtaskDefinition
 from core.message_handlers import store_or_update_subtask
 from core.models import Subtask
 from core.tests.utils import ConcentIntegrationTestCase
 from core.transfer_operations import create_file_transfer_token_for_golem_client
 from utils.constants import ErrorCode
+from utils.helpers import get_storage_result_file_path
+from utils.helpers import get_storage_scene_file_path
+from utils.helpers import get_storage_source_file_path
 from utils.helpers import sign_message
 from utils.testing_helpers import generate_ecc_key_pair
 
@@ -30,6 +34,8 @@ class SubtaskResultsVerifyIntegrationTest(ConcentIntegrationTestCase):
         super().setUp()
         self.task_id = "task1"
         self.subtask_id = "subtask1"
+        self.source_package_path = get_storage_source_file_path(self.subtask_id, self.task_id)
+        self.result_package_path = get_storage_result_file_path(self.subtask_id, self.task_id)
         self.report_computed_task = self._create_report_computed_task()
 
     def test_that_concent_responds_with_service_refused_when_verification_for_this_subtask_is_duplicated(self):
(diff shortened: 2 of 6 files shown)


concent_api/utils/helpers.py
```diff
@@ -154,6 +154,10 @@ def get_storage_file_path(subtask_id, task_id):
     return f'blender/result/{task_id}/{task_id}.{subtask_id}.zip'
 
 
+def get_source_package_path(subtask_id, task_id):
+    return f'blender/source/{task_id}/{task_id}.{subtask_id}.zip'
```
cameel: This is not something that should be in `utils`. `utils` is for generic stuff not related specifically to Concent. This is part of the protocol or at least business logic. It has no place here. Please move it to one of the more specific apps.

concent_api/core/queue_operations.py
```diff
@@ -1,2 +1,27 @@
...
+from utils.helpers import get_storage_file_path
+
+
+def send_blender_verification_request(compute_task_def):
+    task_id = compute_task_def['task_id']
+    subtask_id = compute_task_def['subtask_id']
+    source_package_path = get_source_package_path(subtask_id, task_id)
+    result_package_path = get_storage_file_path(subtask_id, task_id)
+    output_format = get_output_format_from_package(result_package_path)  # pylint: disable=assignment-from-no-return
+    scene_file = get_scene_file_from_package(result_package_path)  # pylint: disable=assignment-from-no-return
```
cameel: Please remove those `pylint: disable`s. Avoid disabling warnings unless absolutely necessary.

In this case pylint is 100% right. Your functions do not return anything. If the code is unfinished it's a good thing that it has warnings indicating that. Why would we want to hide this information?

concent_api/core/queue_operations.py
```diff
@@ -1,2 +1,19 @@
...
+
+
+def send_blender_verification_request(compute_task_def):
+    task_id = compute_task_def['task_id']
+    subtask_id = compute_task_def['subtask_id']
+    source_package_path = get_storage_source_file_path(subtask_id, task_id)
+    result_package_path = get_storage_result_file_path(subtask_id, task_id)
+    output_format = compute_task_def['extra_data']['output_format']
+    scene_file = compute_task_def['extra_data']['scene_file']
+    blender_verification_request(
```
dybi: sholdn't `apply` or `delay` be used here?

concent_api/core/tests/test_integration_subtask_results_verify.py
```diff
@@ -297,15 +298,14 @@ def test_that_concent_accepts_valid_request_and_sends_verification_order_to_work
 
         # when
         with mock.patch("core.message_handlers.base.is_account_status_positive", return_value=True) as is_account_status_positive_mock:
-            with mock.patch("core.message_handlers.send_blender_verification_request") as send_verification_request_mock:
-                with freeze_time(subtask_results_verify_time_str):
```
dybi: why this mock has been removed?

concent_api/core/queue_operations.py
rwrzesien: `delay`, `apply` is for synchronous call.

concent_api/core/tests/test_integration_subtask_results_verify.py
```diff
@@ -375,11 +374,25 @@ def _create_serialized_subtask_results_verify(
...
             task_id=self.task_id,
-            subtask_id=self.subtask_id
+            subtask_id=self.subtask_id,
+            extra_data={
+                'end_task': 6,
+                'frames': [1],
+                'outfilebasename': 'Heli-cycles(3)',
+                'output_format': BlenderSubtaskDefinition.OutputFormat.JPG.name,  # pylint: disable=no-member
+                'path_root': '/home/dariusz/Documents/tasks/resources',
+                'scene_file': 'blender/scene/ef0dc1/ef0dc1.zzz523',
```
rwrzesien: Please add helper for creating this path from `task_id` and `subtask_id`, as we be probably reusing it.

concent_api/core/queue_operations.py
dybi: ok, from what I read (http://docs.celeryproject.org/en/latest/reference/celery.app.task.html#celery.app.task.Task.delay), `apply_async` and `delay` are for asynchronous calls, `apply` is for synchronous. Here we don't call any of them. What is called by default?

dybi: Ok, I just read (http://docs.celeryproject.org/en/latest/userguide/calling.html#basics) that _calling_ the task will execute it locally (in the current process), withou sending the message to the queue. Is this what we want here?

rwrzesien: I think that we want `delay` here.

Jakub89: I've added it before you post this, just to see if it works. You can check if everything is implemented correctly ;)

concent_api/core/tests/test_integration_subtask_results_verify.py
```diff
@@ -286,6 +288,7 @@ def test_that_concent_responds_with_service_refused_when_subtask_results_rejecte
         )
         self._assert_stored_message_counter_not_increased()
 
+    @override_settings(CELERY_TASK_ALWAYS_EAGER=True)
```
rwrzesien: Instead of this please create mock for `blender_verification_request.delay` and add assert that it was called with exact values (use `assert_called_with`). This is actually quite important to have this call tested.

Jakub89: Done.

concent_api/core/tests/test_integration_subtask_results_verify.py
Jakub89: Okey, I've change it to more specific `mock.patch("core.queue_operations.blender_verification_request.delay")`

concent_api/core/tests/test_integration_subtask_results_verify.py
```diff
@@ -310,6 +316,13 @@ def test_that_concent_accepts_valid_request_and_sends_verification_order_to_work
         is_account_status_positive_mock.assert_called_with(
             client_eth_address=self.report_computed_task.task_to_compute.requestor_ethereum_address
         )
+        send_verification_request_mock.assert_called_with(
```
dybi: better use `assert_called_once_with`

Issue #388, opened 2018-05-16
Bad handling in handle_send_force_subtask_results_response


dybi:
`report_computed_task` should not be passed to `update_subtask` function

dybi:
updated the description, old one for reference:

1) there is no check that client_message contains `subtask_results_rejected` if it doesn't contain `subtask_results_accepted`
2) in case contained message is `subtask_results_rejected`, `report_computed_task` should not be passed to `update_subtask` function

dybi:
Changed the description again - the check *is* done in `handle_messages` and in `handle_send_force_subtask_results_response` one of the `subtask_results_rejected`, `subtask_results_accepted` must be present

PR #392, opened 2018-05-17
Fix bad handling in send force subtask results response

dybi:
resolves: https://github.com/golemfactory/concent/issues/388


dybi:
@rwrzesien, updated the description, after speaking it was misleading - we should not pass `report_computed_task` in neither of those cases (and that's why it should be removed)

rwrzesien:
@dybi Ok, I just wanted to clarify those differences. Btw. I see there is still `acceptance_deadline` check in your branch, which is not in master and was removed some time ago. Please make sure you rebase on latest master before merge and that this check is not there.

Issue #379, opened 2018-05-15
Add non-NULL database constraint for Subtask.report_computed_task


dybi:
According to https://github.com/golemfactory/concent/issues/139, `ReportComputedTask` should be not-nullable, but in code it is nullable.

PR #397, opened 2018-05-18
[M] Make report_computed_task field not-nullable in Subtask model

dybi:
resolves https://github.com/golemfactory/concent/issues/379


concent_api/core/models.py
```diff
@@ -296,7 +296,7 @@ class Meta:
 
     # Related messages
     task_to_compute             = OneToOneField(StoredMessage, blank = True, null = False, related_name = 'subtasks_for_task_to_compute')
-    report_computed_task        = OneToOneField(StoredMessage, blank = True, null = True, related_name = 'subtasks_for_report_computed_task')
+    report_computed_task        = OneToOneField(StoredMessage, blank = True, null = False, related_name = 'subtasks_for_report_computed_task')
```
cameel: If it's non-nullable then there's no reason to allow `blank` here.

dybi: corrected for both fields

PR #417, opened 2018-05-23
Add '_' and '-' to ids validation

Jakub89:
We revert previous bugfix for this one.

### DIFF
--- a/concent_api/core/tests/test_unit_validation.py
+++ b/concent_api/core/tests/test_unit_validation.py
@@ -7,6 +7,7 @@ from golem_messages.message.tasks import SubtaskResultsRejected
 from golem_messages.message.tasks import TaskToCompute
 
 from core.constants import ETHEREUM_ADDRESS_LENGTH
+from core.constants import MESSAGE_TASK_ID_MAX_LENGTH
 from core.tests.utils import ConcentIntegrationTestCase
 from core.exceptions import Http400
 from core.validation import validate_all_messages_identical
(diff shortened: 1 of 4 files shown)


concent_api/core/validation.py
```diff
@@ -50,7 +52,8 @@ def validate_id_value(value, field_name):
             error_code=ErrorCode.MESSAGE_VALUE_WRONG_LENGTH,
         )
 
-    if not value.isalnum():
+    valid_id_regex = re.compile(r'^[a-zA-Z0-9_-]*')
```
rwrzesien: Please move this to constants. There is no need to compile this on each validation.

concent_api/core/tests/test_unit_validation.py
```diff
@@ -112,8 +113,8 @@ def test_that_function_should_pass_when_value_is_alphanumeric(self):
     def test_that_function_should_raise_exception_when_value_is_not_alphanumeric(self):
```
dybi: should be renamed: `_` and `-` are accepted as well
EDIT:
please rename the first test too

dybi: BTW: Could you please make this class inherit from `unittest.TestCase`? There is no need for heavy `ConcentIntegrationTestCase` base.

concent_api/core/validation.py
dybi: If you are using `fullmatch`, than I guess `^` is redundant

concent_api/core/tests/test_unit_validation.py
Jakub89: Done

Jakub89: Done

concent_api/core/validation.py

Issue #377, opened 2018-05-14
Fix overriding MAXIMUM_DOWNLOAD_TIME and SUBTASK_VERIFICATION_TIME


cameel:
Golem messages provides `maximum_download_time()` and `subtask_verification_time()` helpers and we have recently started using them (#312, #281). These helpers use constants from golem-messages in the calculation and it's not possible to change them directly. This means that we're no longer able to deploy Concent with reduced protocol times for use with cluster tests.

We need to work around this. The helpers (at least for the time being) won't be parameterized (see https://github.com/golemfactory/golem-messages/issues/217). We could work around that by using `mock.patch()` but since protocol times in Concent are settings, it does not seem right. `mock.patch()` breaks easily when imports change and using it outside of strictly test code is asking for trouble so please don't do this.

Instead:
- Provide an alternative implementation of these functions. They should work exactly the same way but just take all the timing constants as parameters.
- Add a boolean setting called `CUSTOM_PROTOCOL_TIMES`.
    - It should be `False` by default.
    - When it's `True`, use the alternative implementation that takes into account our timing settings.
    - When `False` and our settings do not match constants in golem-messages, `manage.py check` should report an error.
- Add unit tests that ensure that both implementations give the same results when the constants match the settings.
    - Maybe also use `mock.patch()` to test that both implementations give the same results in other cases.
- Remove any workarounds in our cluster tests that were made to keep them working when the helpers are used. All cluster tests should work with both normal and reduced times. The only thing that may be different is the time they take to execute.

PR #419, opened 2018-05-24
Fix  overriding maximum download time and subtask verification time

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/377


concent_api/concent_api/system_check.py
```diff
@@ -147,6 +149,33 @@ def create_error_30_verifier_storage_path_is_not_accessible():
...
+    return Error(
+        'CUSTOM_PROTOCOL_TIMES setting is not set',
+        hint='Set CUSTOM_PROTOCOL_TIMES in your local_settings.py to the boolean value.',
+        id='concent.E031',
+    )
+
+
+def create_error_32_custom_protocol_times_has_wrong_value():
+    return Error(
+        'CUSTOM_PROTOCOL_TIMES setting has_wrong_value',
```
cameel: too_many_underscores

concent_api/concent_api/system_check.py
```diff
@@ -147,6 +149,33 @@ def create_error_30_verifier_storage_path_is_not_accessible():
...
+        hint='Set CUSTOM_PROTOCOL_TIMES in your local_settings.py to the boolean value.',
+        id='concent.E032',
+    )
+
+
+def create_error_33_custom_protocol_times_is_false_and_settings_does_not_match_golem_messages_constants(
+    concent_setting_name,
+):
+    return Error(
+        f'CUSTOM_PROTOCOL_TIMES setting is False and Concent settings {concent_setting_name} does not match golem '
```
cameel: `settings` -> `setting`

concent_api/concent_api/system_check.py
```diff
@@ -147,6 +149,33 @@ def create_error_30_verifier_storage_path_is_not_accessible():
...
+        'messages constant',
+        hint='If CUSTOM_PROTOCOL_TIMES is False, Concent times settings must match golem messages constants.',
```
cameel: `times` -> `time`

concent_api/utils/helpers.py
```diff
@@ -166,15 +169,27 @@ def get_storage_source_file_path(subtask_id, task_id):
     return get_storage_file_path('source', subtask_id, task_id)
 
 
-def calculate_maximum_download_time(size: int) -> int:
+def calculate_maximum_download_time(size: int, rate: int = constants.DEFAULT_UPLOAD_RATE) -> int:
```
cameel: Please move it out of `utils`. This app is meant for much more generic stuff. This is part of our business logic, not a general purpose utility and belongs in `core`.

concent_api/utils/helpers.py
```diff
@@ -2,16 +2,19 @@
 import base64
```
cameel: Commit name: `Fix for implementation of maximum_download_time and subtask_verification_time`

Please be more specific in commit names. This should really tell us what you're fixing without having to look into the code. What was broken before? How was it fixed?

It's not really a fix to be honest. You're just adding support for `settings.CUSTOM_PROTOCOL_TIMES`. And maybe you're changing something in the implementation but it's hard to tell at a glance.

concent_api/utils/helpers.py
cameel: Please add assertions for invalid values. Does the function return anything sensible for negative `size` or `rate`? Does it work with negative `DOWNLOAD_LEADIN_TIME`? If not, this should be guarded with `assert`. Otherwise you have to test and support these parameter combinations (which in my opinion is a waste of time).

concent_api/utils/helpers.py
```diff
@@ -185,22 +200,29 @@ def calculate_subtask_verification_time(report_computed_task: message.ReportComp
...
+        return int(
+            (4 * settings.CONCENT_MESSAGING_TIME) +
+            (3 * mdt.total_seconds()) +
+            (0.5 * subtask_timeout.total_seconds())
+        )
+    else:
+        return int(
+            subtask_verification_time(
+                report_computed_task
+            ).total_seconds()
```
cameel: I think this is short enough to be on a single line. It's acceptable like this but it does not have multiple parameters, long mathematical expressions or really anything that would require it to be split like that.

concent_api/utils/tests/test_helpers.py
```diff
@@ -81,3 +87,23 @@ def test_join_messages_should_return_joined_string_separeted_with_whitespace(sel
 
     def test_join_messages_with_single_argument_should_return_single_string(self):
         self.assertEqual(join_messages('Error in Golem Message.'), 'Error in Golem Message.')
+
+    @override_settings(
+        DOWNLOAD_LEADIN_TIME=constants.DOWNLOAD_LEADIN_TIME
+    )
```
cameel: These two tests are not enough to give us confidence that we will notice when the implementations go out of sync. Please check all the important cases:
- `maximum_download_time()`
    - `CUSTOM_PROTOCOL_TIMES` on or off
    - `rate` given explicitly or left out
    - `DOWNLOAD_LEADIN_TIME` and `MINIMUM_UPLOAD_RATE` matching the values from golem-messages or not.
    - A few different combinations of `size` and `rate`
- `subtask_verification_time()`
    - `CUSTOM_PROTOCOL_TIMES` on or off
    - A few different computation times in TTC
    - A few different results from `maximum_download_time()`

Also, please make these tests more compact. We're testing two extremely simple functions that just accept numbers and spit out the result of a short mathematical expression. I think it's bad if testing a new combination requires you to add ten lines of tests. With simple mathematical functions like that we should make it cheap to test a dozen sets of inputs and expectations in one go. You could use a loop or any other mechanism suggested by @dybi. Just make it easy to cover many cases - we badly need it here.

concent_api/concent_api/system_check.py

concent_api/concent_api/system_check.py

concent_api/concent_api/system_check.py

concent_api/utils/helpers.py
rwrzesien: :+1:

concent_api/utils/helpers.py
rwrzesien: :+1: Ok, I will pay more attention to this next time.

concent_api/utils/helpers.py
rwrzesien: :+1:

concent_api/utils/helpers.py
rwrzesien: :+1:

concent_api/concent_api/system_check.py
```diff
@@ -346,3 +375,27 @@ def check_concents_time_settings(app_configs=None, **kwargs):  # pylint: disable
...
+    if settings.CUSTOM_PROTOCOL_TIMES is False:
+        for concent_setting_name, is_equal in {
+            'CONCENT_MESSAGING_TIME':   settings.CONCENT_MESSAGING_TIME == int(constants.CMT.total_seconds()),
+            'FORCE_ACCEPTANCE_TIME':    settings.FORCE_ACCEPTANCE_TIME  == int(constants.FAT.total_seconds()),
+            'PAYMENT_DUE_TIME':         settings.PAYMENT_DUE_TIME       == int(constants.PDT.total_seconds()),
+            'DOWNLOAD_LEADIN_TIME':     settings.DOWNLOAD_LEADIN_TIME   == int(constants.DOWNLOAD_LEADIN_TIME.total_seconds()),
+            'MINIMUM_UPLOAD_RATE':      settings.MINIMUM_UPLOAD_RATE    == constants.DEFAULT_UPLOAD_RATE,
+        }.items():
+            if not is_equal:
+                return [
```
dybi: Maybe it would be better to iterate over all settings (and perhaps append errors to the list) than just returning on the first; that way we would have more information.

concent_api/concent_api/tests/test_check_custom_protocol_times.py
```diff
@@ -0,0 +1,56 @@
...
+
+            errors = check_custom_protocol_times()
+
+            self.assertEqual([create_error_32_custom_protocol_times_has_wrong_value()], errors)
+
+    @override_settings(
+        CONCENT_MESSAGING_TIME=1,
+        CUSTOM_PROTOCOL_TIMES=False,
+    )  # pylint: disable=no-self-use
+    def test_that_check_custom_protocol_times_should_return_error_if_custom_protocol_times_setting_does_not_match_golem_messages_constant(self):
```
dybi: As we switch to gathering errors, instead of returning on first one (as suggested above), we would need some more tests like this

concent_api/core/tests/test_utils.py
```diff
@@ -0,0 +1,107 @@
...
+
+        mock_maximum_download_time.assert_called_once_with(
+            1,
+            settings.MINIMUM_UPLOAD_RATE,
+        )
+
+    @override_settings(
+        CUSTOM_PROTOCOL_TIMES=True
+    )  # pylint: disable=no-self-use
+    def test_that_calculate_maximum_download_time_will_call_maximum_download_time_when_custom_protocol_times_is_on(self):
```
dybi: "will __not__ call" ;)

concent_api/core/tests/test_utils.py
```diff
@@ -0,0 +1,107 @@
...
+        with mock.patch('core.utils.maximum_download_time', return_value=datetime.timedelta()) as mock_maximum_download_time:
+            calculate_maximum_download_time(size=1)
+
+        mock_maximum_download_time.assert_not_called()
+
+    @override_settings(
+        DOWNLOAD_LEADIN_TIME=int(constants.DOWNLOAD_LEADIN_TIME.total_seconds())
+    )
+    def test_that_both_maximum_download_time_implementation_should_return_same_result_when_golem_messages_constants_match_concent_settings(self):
```
dybi: I would add some more "not so round" values, e. 7, 19, 331, etc.

concent_api/core/tests/test_utils.py
```diff
@@ -0,0 +1,107 @@
...
+    def test_that_calculate_subtask_verification_time_will_call_subtask_verification_time_when_custom_protocol_times_is_off(self):
+        with mock.patch('core.utils.subtask_verification_time', return_value=datetime.timedelta()) as mock_subtask_verification_time:
+            calculate_subtask_verification_time(self.report_computed_task)
+
+        mock_subtask_verification_time.assert_called_once_with(self.report_computed_task)
+
+    @override_settings(
+        CUSTOM_PROTOCOL_TIMES=True
+    )
+    def test_that_calculate_subtask_verification_time_will_call_subtask_verification_time_when_custom_protocol_times_is_on(self):
```
dybi: "will __not__ call" ;)

concent_api/core/tests/test_utils.py
```diff
@@ -0,0 +1,107 @@
...
+            calculate_subtask_verification_time(self.report_computed_task)
+
+        mock_subtask_verification_time.assert_not_called()
+
+    @override_settings(
+        CONCENT_MESSAGING_TIME=int(constants.CMT.total_seconds())
+    )
+    def test_that_both_subtask_verification_time_implementation_should_return_same_result_when_golem_messages_constants_match_concent_settings(self):
+        for size, timestamp, deadline in [
+            (10,    get_current_utc_timestamp(),            get_current_utc_timestamp() + 10),
```
dybi: please extract `get_current_utc_timestamp()` to a variable and reuse it

concent_api/core/tests/test_utils.py
```diff
@@ -0,0 +1,107 @@
...
+        with mock.patch('core.utils.subtask_verification_time', return_value=datetime.timedelta()) as mock_subtask_verification_time:
+            calculate_subtask_verification_time(self.report_computed_task)
+
+        mock_subtask_verification_time.assert_not_called()
+
+    @override_settings(
+        CONCENT_MESSAGING_TIME=int(constants.CMT.total_seconds())
+    )
+    def test_that_both_subtask_verification_time_implementation_should_return_same_result_when_golem_messages_constants_match_concent_settings(self):
+        for size, timestamp, deadline in [
```
dybi: again, I would suggest adding some more "not so round" values, eg. 7, 19, 331

concent_api/concent_api/system_check.py

concent_api/core/tests/test_utils.py
rwrzesien: :+1:

concent_api/core/tests/test_utils.py
rwrzesien: :+1:

concent_api/core/tests/test_utils.py
rwrzesien: :+1:

concent_api/core/tests/test_utils.py
rwrzesien: :+1:

concent_api/core/tests/test_utils.py
rwrzesien: :+1:

concent_api/concent_api/tests/test_check_custom_protocol_times.py
rwrzesien: :+1: done 

I have added one more check, and I consider this to be enough to check the simple loop and comparison.

concent_api/utils/tests/test_helpers.py
cameel: Not if `CUSTOM_PROTOCOL_TIMES` off. I really meant to check with various values of these two settings. But the two most important cases are: when the match and when they do not match the built-in values.

concent_api/concent_api/system_check.py
cameel: I think it would be even better to just compare the whole dicts. Then, if they don't match, the Django prints a diff and you see exactly which values do not match.

concent_api/core/utils.py
```diff
@@ -0,0 +1,65 @@
+import datetime
```
cameel: File name: `core/utils.py`

Could you maybe call it `timing_helpers.py`? I think we should try to avoid putting all "utils" on one big heap. Instead, try to group them in some logical way.

concent_api/core/utils.py
```diff
@@ -0,0 +1,65 @@
...
+
+from django.conf                    import settings
+
+from golem_messages                 import constants
+from golem_messages                 import message
+from golem_messages.helpers         import maximum_download_time
+from golem_messages.helpers         import subtask_verification_time
+
+
+def calculate_maximum_download_time(size: int, rate: int = constants.DEFAULT_UPLOAD_RATE) -> int:
```
cameel: I glossed over it earlier, but now I think we should remove this default. We have a setting for it (`MINIMUM_UPLOAD_RATE`) and it's too easy to forget to set it.

concent_api/core/utils.py
```diff
@@ -0,0 +1,65 @@
...
+        bytes_per_sec = rate << 10
+        download_time = int(
+            datetime.timedelta(
+                seconds=int(math.ceil(size / bytes_per_sec))
+            ).total_seconds()
+        )
+
+        return settings.DOWNLOAD_LEADIN_TIME + download_time
+    else:
+        return int(maximum_download_time(size, settings.MINIMUM_UPLOAD_RATE).total_seconds())
```
cameel: Please don't hard-code `settings.MINIMUM_UPLOAD_RATE`. You should use the `rate` parameter here.

concent_api/core/utils.py
rwrzesien: Then I would put them in `utils\timing_helper` :) I think that we should follow django convention on naming files inside apps. Currently in `core` we have `message_handlers`, `subtask_helper` and `transfers_operations` and I never remember where to look for a given function, because current distribution of them is not intuitive. I would really like to avoid that. If I see `utils.py` I know what to expect inside. If they grow to much, we can split them into `core\utils\type_1.py`, `core\utils\type_2.py` etc.

concent_api/core/utils.py
rwrzesien: :+1:

concent_api/core/utils.py
rwrzesien: :+1:

concent_api/concent_api/settings/base.py
```diff
@@ -285,6 +285,8 @@
 # rendering results and any intermediate files.
 # VERIFIER_STORAGE_PATH = ''
 
+CUSTOM_PROTOCOL_TIMES = False
```
kbeker: Should it has default value?

rwrzesien: According to issue description, yes.

> It should be False by default.

concent_api/core/utils.py
cameel: So maybe `core/helpers/timing.py`? The `utils` part is the one I have issues with. To me `utils` suggests general-purpose utilities, not something intimately tied to the business logic. `helpers` is a tad better though I'd really prefer a more specific name. I think that the best organization is when every class has its own file though when they're small it sometimes makes sense to group several together.

But in general, if you don't like the current organization then I think you should meet with @dybi and figure out a better one. There was a moment where we started splitting files and I was really assuming that you both have it under control. That you have some idea and we're heading towards it. I think that it's something that we need to plan ahead rather than assume that it will just happen spontaneously.

Issue #407, opened 2018-05-22
Routing work queue tasks


cameel:
Currently our Celery/RabbitMQ setup is running with a single queue. All tasks are routed to that queue regardless of their type and size.

Please read this first: [Celery Documentation > Routing Tasks](http://docs.celeryproject.org/en/latest/userguide/routing.html). It has information about queues, routing, `celery amqp` console and starting workers in such a way that they attach to a specific queue.

- Define three queues: `concent`, `conductor` and `verifier`.
- Configure Celery so that new queues are not created automatically.
- Route tasks to queues in the following way:
    - `blender_verification_request` -> `conductor`
    - `blender_verification_order` - > `verifier`
    - `verification_result` -> `concent`
    - In general, tasks from each app should go to the respective queue.
- Disable the default queue if possible. If not, make `concent` the default one. It would be best if someone adding a new task had to explicitly decide which queue it's routed to. Or if the default was decided automatically based on which app the task comes from.

Please test this by running RabbitMQ in a container and three separate worker instances, each one configured to attach to a different queue. Then use `celery amqp` console to issue tasks. Make sure that each worker processes only its own tasks. Also that when `verifier` is processing a big `blender_verification_order` task, the smaller tasks meant for other workers do not get stuck in queue behind it.

PR #423, opened 2018-05-24
Feature conductor notifying the control cluster about finished uploads for verification subtasks

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/411

PR #424, opened 2018-05-25
Feature routing work queue tasks

dybi:
Resolves: https://github.com/golemfactory/concent/issues/407

### DIFF
--- a/concent_api/concent_api/celery.py
+++ b/concent_api/concent_api/celery.py
@@ -24,7 +24,16 @@ from concent_api.constants  import AVAILABLE_CONCENT_FEATURES  # noqa: E402 pyli
 for feature_name in settings.CONCENT_FEATURES:
 
     app.conf.imports += tuple(
-        ['{}.tasks'.format(app)
+        [f'{app}.tasks'
          for app in AVAILABLE_CONCENT_FEATURES[feature_name]['required_django_apps']
-         if importlib.util.find_spec('{}.tasks'.format(app)) is not None]
+         if importlib.util.find_spec(f'{app}.tasks') is not None]
     )
+
+app.conf.task_create_missing_queues = True
+app.conf.task_routes = ([
+    ('verifier.tasks.verification_result', {'queue': 'concent'}),
+    ('conductor.tasks.blender_verification_request', {'queue': 'conductor'}),
+    ('verifier.tasks.blender_verification_order', {'queue': 'verifier'}),
+],)
+app.conf.task_default_queue = 'non_existing'
+app.conf.task_create_missing_queues = False
--- a/concent_api/core/models.py
+++ b/concent_api/core/models.py
@@ -306,6 +306,9 @@ class Subtask(Model):
         super().__init__(*args, **kwargs)
         self._current_state_name = None
 
+    def __repr__(self):
+        return f"Subtask: task_id={self.task_id}, subtak_id={self.subtask_id}, state={self.state_enum}"
+
     @classmethod
     def from_db(cls, db, field_names, values):
         new = super().from_db(db, field_names, values)
--- a/concent_api/verifier/tasks.py
+++ b/concent_api/verifier/tasks.py
@@ -45,11 +45,12 @@ def blender_verification_order(
     subtask_id: str,
     source_package_path: str,
     result_package_path: str,
-    output_format: BlenderSubtaskDefinition.OutputFormat,  # pylint: disable=unused-argument
+    output_format: str,
     scene_file: str,  # pylint: disable=unused-argument
-    report_computed_task: message.ReportComputedTask,
+    report_computed_task:   message.ReportComputedTask,
 ):
     assert source_package_path != result_package_path
+    assert output_format in BlenderSubtaskDefinition.OutputFormat.__members__.keys()
 
     # Generate a FileTransferToken valid for a download of any file listed in the order.
     file_transfer_token = create_file_transfer_token_for_concent(
@@ -82,18 +83,18 @@ def blender_verification_order(
             logger.info('blender_verification_order for SUBTASK_ID {subtask_id} failed with error {exception}.')
             verification_result.delay(
                 subtask_id,
-                VerificationResult.ERROR,
+                VerificationResult.ERROR.name,
                 str(exception),
-                ErrorCode.VERIFIIER_FILE_DOWNLOAD_FAILED
+                ErrorCode.VERIFIIER_FILE_DOWNLOAD_FAILED.name
             )
             return
         except Exception as exception:
             logger.info('blender_verification_order for SUBTASK_ID {subtask_id} failed with error {exception}.')
             verification_result.delay(
                 subtask_id,
-                VerificationResult.ERROR,
+                VerificationResult.ERROR.name,
                 str(exception),
-                ErrorCode.VERIFIIER_FILE_DOWNLOAD_FAILED
+                ErrorCode.VERIFIIER_FILE_DOWNLOAD_FAILED.name
             )
             raise
 
@@ -115,7 +116,7 @@ def blender_verification_order(
 
     verification_result.delay(
         subtask_id,
-        VerificationResult.MATCH,
+        VerificationResult.MATCH.name,
     )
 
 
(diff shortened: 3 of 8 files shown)


concent_api/core/models.py
```diff
@@ -306,6 +306,9 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._current_state_name = None
 
+    def __repr__(self):
+        return f"Subtask: task_id={self.task_id}, subtaks_id={self.subtask_id}, state={self.state_enum}"
```
cameel: `subtaks_id` -> `subtask_id`

concent_api/verifier/tasks.py
```diff
@@ -132,6 +132,9 @@ def verification_result(
             throw=False,
         )
         return
+    except Exception as e:  # pylint: disable=broad-except
+        logging.warning(f"Exception: {e}, Type {type(e)}")
+        return
```
cameel: Please remove this overly broad exception handler. I agree with pylint here. If there's a problem, the task should crash. The rule is to crash early and loudly so that we can identify the problems and fix them while still in development and testing. Leave the crash for Celery or Kubernetes to handle. Only catch specific exceptions in specifc situations that you know need to be handled in a special way (like the `DatabaseError` above; though I think that catching all `DatabaseError`s is also a bit too broad there - we only want to handle timeouts not for example errors in the query or in the schema).

You can log it but it's definitely not just a `warning`. But I think the exception will be logged anyway if it escapes from here so it's probably not necessary.

concent_api/core/models.py
cameel: This should be `__str__()`, not `__repr__()`. [repr()](https://docs.python.org/3/library/functions.html#repr) is supposed to return a reprsentation that can be evaluated to get the original object back. `str()` can be more lax.

concent_api/concent_api/celery.py
```diff
@@ -24,7 +24,15 @@
...
     app.conf.imports += tuple(
-        ['{}.tasks'.format(app)
+        [f'{app}.tasks'
          for app in AVAILABLE_CONCENT_FEATURES[feature_name]['required_django_apps']
-         if importlib.util.find_spec('{}.tasks'.format(app)) is not None]
+         if importlib.util.find_spec(f'{app}.tasks') is not None]
     )
+
+app.conf.task_create_missing_queues = True
+app.conf.task_routes = ([
```
rwrzesien: What do you think about moving this to settings or constants ?

concent_api/concent_api/celery.py
```diff
@@ -24,7 +24,15 @@
...
+    ('verifier.tasks.verification_result', {'queue': 'concent'}),
+    ('conductor.tasks.blender_verification_request', {'queue': 'conductor'}),
+    ('verifier.tasks.blender_verification_order', {'queue': 'verifier'}),
+],)
```
rwrzesien: This has to be tuple ?

concent_api/concent_api/celery.py
```diff
@@ -24,7 +24,15 @@
...
+app.conf.task_default_queue = 'concent'
```
rwrzesien: This might also be configurable.

concent_api/verifier/tasks.py
rwrzesien: I think that the only thing that can go wrong here is `Subtask.DoesNotExists` error.

concent_api/concent_api/celery.py
cameel: I wanted to suggest this myself but actually I think I like @dybi's solution more. This is something that we never expect the admin to tweak. Keeping it here as a part of code rather than settings makes it pretty clear.

concent_api/concent_api/celery.py
cameel: It really shouldn't. At least not by someone who's setting up Concent on the cluster.

If there are no bugs in the code this default should never come into play. Each task is explicitly routed to the right queue. If we have a bug and someone forgets to route a task I'd like the code to crash so that we notice it an fix it.

Actually, now that I think about it - maybe we could use a name of a non-existent queue. Would it crash then?

cameel:
One more thing. If #423 goes into master before this pull request please add routing for tasks defined in it.

concent_api/verifier/tasks.py
dybi: I will remove it.

concent_api/core/models.py
dybi: Not really. 
1) `__repr__` :
> If at all possible, this should look like a valid Python expression that could be used to recreate an object with the same value (given an appropriate environment).
In this case it is not. 
2) The most important trait of `__repr__` is that it should be unambiguous. Printing information 
about `task_id` and `subtask_id`  is enough to identify a Subtask object.
3) `__str__` of collection (like `QuerySet`) uses `__repr__` of objects it contains (and not their `__str__`).
4) If an object doesn't have `__str__` defined, it will use its `__repr__`, but not vice versa.

**TL;DR**
The use of `__repr__` was intentional.

https://stackoverflow.com/a/2626364

concent_api/concent_api/celery.py
dybi: This seems a good idea.
@cameel ?

concent_api/concent_api/celery.py
dybi: yes.
http://docs.celeryproject.org/en/latest/userguide/routing.html#automatic-routing (see: _Note_)

concent_api/core/models.py
dybi: done

concent_api/verifier/tasks.py
dybi: done

concent_api/core/models.py
cameel: Fair enough. The part about it only having to be unambiguous seems convincing.

But maybe let's tweak the representation a bit so it looks like `Subtask(task_id=1, subtask_id=2, state=3)`? Otherwise it may be hard to read if you print a list of subtasks.

concent_api/concent_api/celery.py
cameel: As I wrote above, I like this approach and I think we should leave it as is. `celery.py` seems like a very good place for something like this.

concent_api/concent_api/celery.py
cameel: `task_create_missing_queues` definitely should be disabled. I even stated it in the issue:

> Configure Celery so that new queues are not created automatically.

Think about the use case. When would we ever want a queue to be created automatically? We have a small and constant number of consumer types: concent, conductor and verifier. We know at development time which one should process each task. If won't ever change at runtime. By enabling automatic queue creation the only result is that if you make a typo, your task will go to a new queue that is not being processed by anything. I think it's much, much better if Celery can notify us that we're trying to use a queue that does not exist. If we need a new queue, we'll define it.

concent_api/verifier/tasks.py
```diff
@@ -105,17 +106,21 @@ def blender_verification_order(
...
+    assert isinstance(result, str)
+    assert result in VerificationResult.__members__.keys()
     assert isinstance(error_message, (str, type(None)))
-    assert isinstance(error_code, (ErrorCode, type(None)))
-    assert all([error_message, error_code]) if result == VerificationResult.ERROR else True
+    assert isinstance(error_code, (str, type(None)))
+
+    result_enum = VerificationResult[result]
+
+    assert all([error_message, error_code]) if result_enum == VerificationResult.ERROR else True
```
cameel: This could be shortened to:

``` python
assert result_enum != VerificationResult.ERROR or all([error_message, error_code])
```

cameel: `[error_message, error_code]` is not a boolean expression. Since it's a non-empty array, it's always true. What is it supposed to check?

concent_api/verifier/tasks.py
```diff
@@ -186,13 +191,13 @@ def verification_result(
...
+    elif result_enum in (VerificationResult.MATCH, VerificationResult.ERROR):
         # Worker logs the error code and message
-        if result == VerificationResult.ERROR:
+        if result_enum == VerificationResult.ERROR:
             logger.info(
                 f'verification_result_task processing error result with: '
-                f'SUBTASK_ID {subtask_id} -- RESULT {result} -- ERROR MESSAGE {error_message} -- ERROR CODE {error_code}'
-            )
+                f'SUBTASK_ID {subtask_id} -- RESULT {result_enum.name} -- ERROR MESSAGE {error_message} -- ERROR CODE {error_code}'
+            )  # pylint: disable=no-member
```
cameel: Could you please disable this warning for this enum to `pylintrc`? I see that we're still using `pylint: disable` for enums and it's a really bad practice because it makes us used to disabling warnings rather than treating them as something that has to be fixed.

concent_api/verifier/tasks.py
cameel: Sorry, I was wrong. It's fine. It does expect a list. I'm too used to the `all(condition for x in collection)` form :)

Issue #413, opened 2018-05-23
Concent: Handling notifications about finished uploads for verification subtasks


cameel:
When #411 is implemented, Conductor will start sending notifications to the control cluster to indicate that all files are ready for verification. Control cluster should handle this by acknowledging the upload provided that the notification comes before the time is up.

- This issue depends on #412. Use the same method to ensure that the flow described below never runs in parallel with the code that enforces timeouts or changes subtask state.
- Add a Celery handler for the `upload_finished` task. Upon receiving it, Concent should:
    - Check subtask state:
        - If it's one of the states that can precede verification, report an error. Processing ends here. This indicates that verification has never been requested and initiated - this makes no sense and should never really happen as long as the system is working as designed
        - If it's `ADDITIONAL VERIFICATION`, `ACCEPTED` or `FAILED`, log a warning and ignore the notification. Processing ends here. This means that it's a duplicate notification. This shouldn't happen in normal circumstances but is possible if some operation fails and and is retried. We should know about it and investigate but it's not critical.
        - If it's `VERIFICATION FILE TRANSFER`, proceed with the task.
    - Check `Subtask.next_deadline`.
        - If we're past the deadline, trigger the code that processes the timeout - change subtask state, add responses to the receive queue issue payments, etc. This will solve our problem with payments only happening when the client asks (at least in this use case).
        - Otherwise proceed with the task.
    - Change subtask state to `ADDITIONAL VERIFICATION`.
    - Add `upload_acknowledged` task to the work queue.

dybi:
@cameel, there is no information to which work queue `upload_finished` should be put. I assume, it is `conductor`, isn't it?

dybi:
REMARKS:

- enforcing that handling never runs in parallel has not been implemented yed

cameel:
> there is no information to which work queue `upload_finished` should be put. I assume, it is `conductor`, isn't it?

No.
- `upload_finished`: `concent` queue (queued by Conductor, processed by Concent)
- `upload_acknowledged` -> `conductor` queue (queued by Concent, processed by Conductor)

The the description says that Conductor sends a notification (that's `upload_finished`), Concent handles it and sends back an acknowledgement (`upload_acknowledged`) to the Conductor.

dybi:
@cameel, sorry I ment sth different than I wrote ;) Anyway, it is cristal clear now, thanks :)

PR #429, opened 2018-05-28
Feature concent handling notifications about finished uploads for verification subtasks

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/413.


concent_api/core/tests/test_upload_finished_task.py
```diff
@@ -0,0 +1,82 @@
...
+            self.subtask.refresh_from_db()
+
+            with mock.patch('core.tasks.logging.warning') as logging_warning_mock:
+                upload_finished(self.subtask.subtask_id)  # pylint: disable=no-value-for-parameter
+
+            logging_warning_mock.assert_called_once_with(
+                f'Subtask with ID {self.subtask.subtask_id} is expected to be in `VERIFICATION_FILE_TRANSFER` state, but was in {self.subtask.state}.'
+            )
+
+    def test_that_scheduling_task_for_subtask_with_unexpected_state_should_log_warning_and_finish_task(self):
```
dybi: should `...log_error...`

concent_api/core/tests/test_upload_finished_task.py
```diff
@@ -0,0 +1,82 @@
...
+    def test_that_scheduling_task_for_subtask_before_deadline_should_change_subtask_state_and_schedule_upload_acknowledged_task(self):
+        with freeze_time(parse_timestamp_to_utc_datetime(self.subtask.next_deadline.timestamp() - 1)):
+            with mock.patch('core.tasks.upload_acknowledged.delay') as upload_acknowledged_delay_mock:
+                upload_finished(self.subtask.subtask_id)  # pylint: disable=no-value-for-parameter
+
+        self.subtask.refresh_from_db()
+        self.assertEqual(self.subtask.state_enum, Subtask.SubtaskState.ADDITIONAL_VERIFICATION)
+        upload_acknowledged_delay_mock.assert_called_once_with(self.subtask.subtask_id)
+
+    def test_that_scheduling_task_for_subtask_after_deadline_should_process_timeout(self):
```
dybi: shouldn't this check payment as well?

dybi: `mock` backend is used. I suggest patching payment function, so call to to it can be asserted.

concent_api/core/tasks.py
```diff
@@ -2,10 +2,83 @@
...
+from utils.helpers import get_current_utc_timestamp
+
 
 logger = logging.getLogger(__name__)
 
 
 @shared_task
-def upload_finished(_subtask_id: str):
-    pass
+def upload_finished(subtask_id: str):
```
dybi: too many reruns in this function... it makes it a bit hard to read. Please refactor

dybi: done

dybi: done

concent_api/core/tests/test_upload_finished_task.py
dybi: done

concent_api/core/tests/test_upload_finished_task.py
dybi: done

concent_api/core/tests/test_upload_finished_task.py
```diff
@@ -0,0 +1,89 @@
...
+        for state in [
+            Subtask.SubtaskState.ACCEPTED,
+            Subtask.SubtaskState.FAILED,
+            Subtask.SubtaskState.ADDITIONAL_VERIFICATION,
+        ]:
+            Subtask.objects.filter(subtask_id=self.subtask.subtask_id).update(state=state.name)  # pylint: disable=no-member
+            self.subtask.refresh_from_db()
+
+            with mock.patch('core.tasks.logging.warning') as logging_warning_mock:
+                upload_finished(self.subtask.subtask_id)  # pylint: disable=no-value-for-parameter
```
Jakub89: My pylint does't need this disabled command.

concent_api/core/tests/test_upload_finished_task.py
```diff
@@ -0,0 +1,89 @@
...
+            logging_warning_mock.assert_called_once_with(
+                f'Subtask with ID {self.subtask.subtask_id} is expected to be in `VERIFICATION_FILE_TRANSFER` state, but was in {self.subtask.state}.'
+            )
+
+    def test_that_scheduling_task_for_subtask_with_unexpected_state_should_log_error_and_finish_task(self):
+        Subtask.objects.filter(subtask_id=self.subtask.subtask_id).update(state=Subtask.SubtaskState.REPORTED.name)  # pylint: disable=no-member
+        self.subtask.refresh_from_db()
+
+        with mock.patch('core.tasks.logging.error') as logging_error_mock:
+            upload_finished(self.subtask.subtask_id)  # pylint: disable=no-value-for-parameter
```
Jakub89: or this one.

concent_api/core/tests/test_upload_finished_task.py
```diff
@@ -0,0 +1,89 @@
...
+
+        logging_error_mock.assert_called_once_with(
+            f'Subtask with ID {self.subtask.subtask_id} is expected to be in `VERIFICATION_FILE_TRANSFER` state, but was in {self.subtask.state}.'
+        )
+
+    def test_that_scheduling_task_for_subtask_before_deadline_should_change_subtask_state_and_schedule_upload_acknowledged_task(self):
+        with freeze_time(parse_timestamp_to_utc_datetime(self.subtask.next_deadline.timestamp() - 1)):
+            with mock.patch('core.tasks.upload_acknowledged.delay') as upload_acknowledged_delay_mock:
+                upload_finished(self.subtask.subtask_id)  # pylint: disable=no-value-for-parameter
```
Jakub89: or this.

concent_api/core/tests/test_upload_finished_task.py
```diff
@@ -0,0 +1,89 @@
...
+
+        self.subtask.refresh_from_db()
+        self.assertEqual(self.subtask.state_enum, Subtask.SubtaskState.ADDITIONAL_VERIFICATION)
+        upload_acknowledged_delay_mock.assert_called_once_with(self.subtask.subtask_id)
+
+    def test_that_scheduling_task_for_subtask_after_deadline_should_process_timeout(self):
+        datetime = parse_timestamp_to_utc_datetime(get_current_utc_timestamp() + settings.CONCENT_MESSAGING_TIME + 1)
+        with freeze_time(datetime):
+            with mock.patch('core.tasks.base.make_force_payment_to_provider', autospec=True) as payment_function_mock:
+                upload_finished(self.subtask.subtask_id)  # pylint: disable=no-value-for-parameter
```
Jakub89: or even this one.

concent_api/core/tasks.py
rwrzesien: Actually it was done on purpose for two reasons: 
1. To make the task easier to read. It was showing clearly where and when the task should end. 
2. To avoid future errors. Now, if somebody put some code at the end of the task and forget to add returns everywhere where they were before, the task will not work correctly.

Please add them back.
I even think we should add this as a good practice for celery tasks.

concent_api/core/tests/test_upload_finished_task.py
rwrzesien: I thought we agreed that when the mock backend is used we don't have to mock those calls. But actually I prefer that we do this. So lets agree on one way and stick to it. And make it as our good practice. And add it where it is missing.

PR #431, opened 2018-05-28
Timing diagram

cameel:
This is the timing diagram from #165 blueprint. The pull request contains the diagram source (.graphml format, file can be opened with [yEd](https://www.yworks.com/products/yed)) and an SVG export so that I can later include it in the docs. There should not be any differences between the two. The .svg file is always created and updated by overwriting the old file with a fresh export in yEd.

When reviewing please focus on discrepancies between the diagram and the code in master. The diagram is supposed to show what the current implementation does (or will be doing when verification is finished).

Typos and bad English are fair game too. Please point them out if you find any.

Remarks on how to improve the diagram are welcome, though I won't necessarily be doing them all in this pull request. I want to get it merged and then work on improvements.


cameel:
@rwrzesien Answers to questions from https://github.com/golemfactory/concent/pull/432#issuecomment-392873778

> 1. Why in `Report` case there is `CMT` used twice ?

The second one is provider's time window for receiving the response. It was present in Łukasz's specification but our implementation does not impose this limitation. Provider can get the response even if he asks much later.

So it's not very relevant but you can think about it as the minimum time the response is guaranteed to be stored. After that Concent is not required to have the response.

> 2. In code in `Report` FRCTR and VRCT becomes available after `TTC.deadline + CMT`, not after `TTC.deadline`.

On the diagram it's not `TTC.deadline`. If you look closer you'll see that it's immediately after the requestor has submitted ARCT or RRCT. Which might be anything between `TTC.deadline` and `TTC.deadline + CMT`.

I think that's also what we have in the code, right? To make it available exactly after `TTC.deadline + CMT` we'd have to have some mechanism that prevents a pending message from being returned before a specific time.

> 3. In code in `ResultTranfer` we FGTRF becomes available after `RCT.timestamp + FAT + CMT`. This is not visible on this schema.

It's a bug in the code. The code should have been changed in #166. Could you check if this is still the case? #452 might have already fixed it. If not, please report a bug.

> 4. In code in `Acceptance`, time window is based on `CTD.deadline` instead of `TTC.deadline`.

By CTD do you mean `ComputeTaskDef`? `ComputeTaskDef` is an implementation detail of  `TaskToCompute` and I'm writing `TaskToCompute.deadline` instead of `TaskToCompute.compute_task_def.dedaline` just for brevity.

> 5. Why in `Acceptance` case there is `CMT` used twice ?

Same as in report use case. It's provider's window.

> 6. Maybe it would be possible somehow to show that in `Payment` `SubtaskResultAccepted.payment_ts` is the oldest one from collection ?

I'd like to keep the diagram focused on the lifecycle of a single subtask. This diagram simply does not illustrate the case of `ForcePayment` containing multiple acceptances. I can't think of any way to show multiple subtasks without making the diagram overly complex.

It's mentioned in the text that will accompany this diagram (#165) which will be submitted in a separate pull request. I wanted to get the diagrams through first because they make sense on their own and already contain a lot of information.

> 7. Why in `AdditionalVerification` case there is `SVT` used ?

That's what Łukasz's docs say. There's a diagram on page 21 that shows an interval described as "verification, not longer than svt".

I talked with Łukasz about it at some point and he explained that he considers it an implementation detail anyway. He said that the time allotted for verification should be pretty long but sensible, so that verification does not time out on typical computations.

So at the time of your question SVT had a purpose. But later we did #556 which made AVCT longer and dependent on task deadline. I have just realized that we did it wrong ;) Or actually I did it wrong... Now we accept `SubtaskResultsVerify` in the whole extended AVCT window (which may be pretty long). I should have extended the SVT, not AVCT. I'm going to create a task to fix that.

BTW, I see that there's now an extra MDT between AVCT and SVT in Łukasz's spec. There will be a separate issue for adding it. For now the diagram and the code are consistent in this regard.

Anyway, right now there are timing issues in verification that need fixes but the point of this pull request is to have the diagram match the implementation ;) We'll be updating it along with future fixes.

rwrzesien:
@cameel 
> The second one is provider's time window for receiving the response. It was present in Łukasz's specification but our implementation does not impose this limitation. Provider can get the response even if he asks much later.

Why not then just remove it from both blueprint and Łukasz's specification? I think the consistency here matters. Otherwise someone can read it backwards - ie. check the blueprint, check the code, then report the bug. Actually I saw this few times in our project when some time passed since some agreements were made and we started to forget about them.

> On the diagram it's not TTC.deadline. If you look closer you'll see that it's immediately after the requestor has submitted ARCT or RRCT. Which might be anything between TTC.deadline and TTC.deadline + CMT.
> I think that's also what we have in the code, right? To make it available exactly after TTC.deadline + CMT we'd have to have some mechanism that prevents a pending message from being returned before a specific time.

Yes, that is what is in the code. But if it would be possible to distinct it somehow on the diagram it would be great, currently those lines are close to each other. Maybe just extend the leftmost part a little bit. It is a little bit compact when compared to the rest. Off course if it is not too much effort.

> It's a bug in the code. The code should have been changed in #166. Could you check if this is still the case? #452 might have already fixed it. If not, please report a bug.

It is fixed already, I must have placed that comment earlier.

> By CTD do you mean ComputeTaskDef? ComputeTaskDef is an implementation detail of TaskToCompute and I'm writing TaskToCompute.deadline instead of TaskToCompute.compute_task_def.dedaline just for brevity.

Yes, it is clear now.

> Same as in report use case. It's provider's window.

Same here, maybe it should be just removed then?

> I'd like to keep the diagram focused on the lifecycle of a single subtask. This diagram simply does not illustrate the case of ForcePayment containing multiple acceptances. I can't think of any way to show multiple subtasks without making the diagram overly complex.

Maybe just add `min([payment_ts in SubtaskResultAccepted])`. But it is up to you, maybe it would be really to much detail.

> So at the time of your question SVT had a purpose. But later we did #556 which made AVCT longer and dependent on task deadline. I have just realized that we did it wrong ;) Or actually I did it wrong... Now we accept SubtaskResultsVerify in the whole extended AVCT window (which may be pretty long). I should have extended the SVT, not AVCT. I'm going to create a task to fix that.

Shouldn't then SVT be removed too for now, as mentioned CMT in REPORT and ACCEPTANCE?

cameel:
> Why not then just remove it from both blueprint and Łukasz's specification? I think the consistency here matters. Otherwise someone can read it backwards - ie. check the blueprint, check the code, then report the bug. Actually I saw this few times in our project when some time passed since some agreements were made and we started to forget about them.

I'll use a different color and add it to the legend.
I'd still leave it in the spec because in the P2P version where messages are not stored this does matter.

I think I'll also explain it in the timing blueprint.

> Yes, that is what is in the code. But if it would be possible to distinct it somehow on the diagram it would be great, currently those lines are close to each other. Maybe just extend the leftmost part a little bit. It is a little bit compact when compared to the rest. Off course if it is not too much effort.

I'll see what I can do. One of the reasons that it's so tiny and hard to see is that I tried to keep longer time intervals longer on the diagram too. Not always successfully (see SVT). CMT is one of the shortest unfortunately.

Maybe I'll manage to express it with an extra line...

> Maybe just add `min([payment_ts in SubtaskResultAccepted])`. But it is up to you, maybe it would be really to much detail.

I'd rather not, because as I said, it's supposed to show only one Subtask. But I'll try to explain that in some way on the diagram or in the text.

> Shouldn't then SVT be removed too for now, as mentioned CMT in REPORT and ACCEPTANCE?

I think it should stay in this pull request because it matches what was in my specifications (i.e. in the issues). We will fix both the diagram and the code in the upcoming issues.

cameel:
I have updated the diagram and #165.

> I'll use a different color and add it to the legend.
> I'd still leave it in the spec because in the P2P version where messages are not stored this does matter.

I have decided to remove the second CMTs after all because they do not correspond to anything.

The time windows for fetching a messages from Concent are not even on the diagram. There are only time windows for sending (vivid green). The pale green boxes just show you how long a specific time interval is supposed to be without attaching any meaning to it.

> > Yes, that is what is in the code. But if it would be possible to distinct it somehow on the diagram it would be great, currently those lines are close to each other. Maybe just extend the leftmost part a little bit. It is a little bit compact when compared to the rest. Off course if it is not too much effort.
>
> (...)
> Maybe I'll manage to express it with an extra line...

I looked at it closer and I think it's actually clear enough given the limited space. The bullets indicating the `ARCT/RRCT` message and the corresponding pending responses are clearly within the CMT, not at the beginning or end of the interval.

The interval is small and there's not much wiggle room. I'd have to redo interval lengths to make it any clearer than that. I think that redoing the diagram is out of scope of this pull request. I do want to tweak it in the future but here the goal is just to get what we already have into the repo and ensure that it matches the code.

> > Maybe just add `min([payment_ts in SubtaskResultAccepted])`. But it is up to you, maybe it would be really to much detail.
>
> I'd rather not, because as I said, it's supposed to show only one Subtask. But I'll try to explain that in some way on the diagram or in the text.

I have added a clarification in the blueprint:

<i>Note: the diagram above shows the biggest possible set of use cases one subtask could go through. It's not the only possible one though since all use cases are optional. Also keep in mind that the diagram only shows one subtask but the `PAYMENT` use case can receive more than one in a single message. In that case the time window starts when acceptances of **all** those subtasks are older than PDT.</i>

cameel:
> > Shouldn't then SVT be removed too for now, as mentioned CMT in REPORT and ACCEPTANCE?
>
> I think it should stay in this pull request because it matches what was in my specifications (i.e. in the issues). We will fix both the diagram and the code in the upcoming issues.

I have added #828 to fix his problem. If that gets merged before this pull request, I'll update the diagram here.

PR #436, opened 2018-05-29
Fix concent verification fixes

rwrzesien:
Resolves issues related to verification on concent-api side.

Issued fixed:
- Change provider to request public key usage in AckSubtaskResultsVerify response.
- Set SubtaskResultsVerify as requestor message.
- Add pending_value to payment backend call in additional verification use case.

### DIFF
--- a/concent_api/core/message_handlers.py
+++ b/concent_api/core/message_handlers.py
@@ -1287,12 +1287,11 @@ def handle_send_subtask_results_verify(
 
     send_blender_verification_request(compute_task_def)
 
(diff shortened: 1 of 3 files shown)


cameel:
Since this pull request does not have a corresponding issue, please explain what you're fixing in the description.

cameel:
A general remark: could you do `import core.payments.base` instead of `from core.payments import base`? `base` is such a generic name that without using the full name it's hard to tell what it really is.

PR #437, opened 2018-05-31
Change way in which payments are imported

rwrzesien:
Changes way in which payments are imported and updates related tests.

This issue appeared while working on https://github.com/golemfactory/concent/pull/436 and is discussed in comment https://github.com/golemfactory/concent/pull/436#issuecomment-393111894.


cameel:
> I would prefer to import each function speartely, than have such a long path in function usage, but I can live with it ;)

Yeah, I would prefer that too (that's why I suggested it in my comment https://github.com/golemfactory/concent/pull/436#issuecomment-393111894). Still, using full path is a big improvement over just importing something called `base`.

Please refactor it further when you get the chance.

PR #439, opened 2018-06-01
Change ForcePayment validation only for unique subtask_id

### LINKED ISSUE #156: Validating that ForcePayment does not list the same subtask twice
When client sends `ForcePayment` we should check if all `ComputeTaskDef` in `TaskToCompute` in `SubtaskResultsAccepted` in `subtask_results_accepted_list` contain unique `subtask_id`. If not, Concent should return an error. Please also add tests for this case.

kbeker:
Resolves #156

### DIFF
--- a/concent_api/core/message_handlers.py
+++ b/concent_api/core/message_handlers.py
@@ -646,7 +646,7 @@ def handle_send_force_payment(client_message: message.concents.ForcePayment) -> 
             reason = message.concents.ServiceRefused.REASON.InvalidRequest
         )
 
-    if not are_ids_unique_in_subtask_results_accepted_list(client_message.subtask_results_accepted_list):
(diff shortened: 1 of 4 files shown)


concent_api/core/message_handlers.py
```diff
@@ -1362,8 +1362,8 @@ def is_message_recieved_in_wrong_state(subtask_id, forbidden_states):
     ).exists()
 
 
-def are_ids_unique_in_subtask_results_accepted_list(subtask_results_accepted_list):
+def are_subtask_ids_unique_in_subtask_results_accepted_list(subtask_results_accepted_list):
```
dybi: Personally, I would change this to sth more general, i.e.:
```python
def are_items_unique(items: list):
    return len(items) == len(set(items))
```
and then in usage:

```python
are_items_unique([subtask_results_accepted.subtask_id for subtask_results_accepted in subtask_results_accepted_list])
```
in this way function has one single responsibility, is general and can be reused ;)

Also, it would made unit-testing much easier ;)

Issue #243, opened 2018-04-04
Verifier: Comparing rendering results


cameel:
This is a part of the implementation of the #193 blueprint. This task depends on #225 and #226.

When the rendering is done and verifier has the output both from the provider and from its own Blender run it's time to compare the images.

### New settings
- `VERIFIER_MIN_SSIM`: The minimum acceptable value of the SSIM metric. Values below this threshold for an image pair will result in a failed verification (i.e. the images are considered different).
    - Floating-point value.
    - Must be between -1 and 1.
    - Required if `'verifier'` is in `CONCENT_FEATURES`. Otherwise must be `None` or undefined.
    - Define a sensible default in `base.py`. This should be based on the tests we did for Golem before (ask @bartoszbetka, @Jakub89 and @kbeker).
- `VERIFIER_MAX_IMAGE_SIZE_IN_MEMORY`: The maximum size (in bytes) of an image when uncompressed and loaded into memory for the purpose of image comparison.
    - Integer value.
    - Cannot be negative or zero.
    - Required if `'verifier'` is in `CONCENT_FEATURES`. Otherwise must be `None` or undefined.
    - Set default to 10000 * 10000 * 4 (10k RGBA image with 8-bits per channel.
    - Add this setting only if necessary. See below for details.

### Sequence of operation
- Load both images into memory
    - The code must support all the image formats that we allow as output from Blender.
- If loading fails because because of insufficient memory, verification ends here. Send `verification_result` with `result == ERROR` to the work queue and stop processing the task.
- Compute [SSIM](https://en.wikipedia.org/wiki/Structural_similarity) for the image pair.
    - Use [measure.compare_ssim()](http://scikit-image.org/docs/dev/api/skimage.measure.html#skimage.measure.compare_ssim) from scikit. That's the library Golem is using as of 0.15.1 (see [img_metrics_calculator.compare_images()](https://github.com/golemfactory/golem/blob/0.15.1/apps/rendering/resources/scripts/img_metrics_calculator.py#L88-L122)).
- Compare SSIM with `VERIFIER_MIN_SSIM`.
    - If the computed SSIM value is above the threshold, add `verification_result` with `result == MATCH` to the queue.
    - Otherwise add `verification_result` with `result == MISMATCH` to the queue.
    - Note that the range of SSIM is [-1, 1] and 1 means that the images are identical.
- The Celery task should end here. You may optionally delete all files from local storage directory to save some time during the next run.

[...]

PR #441, opened 2018-06-01
[Verification 9] Feature verification verifier comparing rendering results

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/243.
Resolves https://github.com/golemfactory/concent/issues/527.

### DIFF
--- a/concent_api/concent_api/settings/base.py
+++ b/concent_api/concent_api/settings/base.py
@@ -297,6 +297,10 @@ BLENDER_MAX_RENDERING_TIME = 60 * 60 * 24 * 7
 
 CUSTOM_PROTOCOL_TIMES = False
 
+# The minimum acceptable value of the SSIM metric. Values below this threshold for an image pair will result in a
+# failed verification (i.e. the images are considered different).
+VERIFIER_MIN_SSIM = 0.94
+
 # Which components of this Django application should be enabled in this particular server instance.
 # The application is basically a bunch of services with totally different responsibilites that share a lot of code.
 # In a typical setup each instance has only one or two features enabled. Some of them provide public APIs, others are
--- a/concent_api/concent_api/tests/test_check_verifier_min_ssim.py
+++ b/concent_api/concent_api/tests/test_check_verifier_min_ssim.py
@@ -1,0 +1,78 @@
+from django.test                import override_settings
+from django.conf                import settings
+from django.test                import TestCase
+
+from concent_api.system_check   import create_error_31_verifier_min_ssim_is_not_set
+from concent_api.system_check   import create_error_32_verifier_min_ssim_is_set
+from concent_api.system_check   import create_error_33_verifier_min_ssim_has_wrong_type
+from concent_api.system_check   import create_error_34_verifier_min_ssim_has_wrong_value
+from concent_api.system_check   import check_verifier_min_ssim
+
+
+class TestStorageServerInternalAddressCheck(TestCase):
+
+    @override_settings(
+        CONCENT_FEATURES=[
+            'verifier'
+        ]
+    )
+    def test_that_verifier_min_ssim_not_set_will_produce_error_when_verifier_is_in_available_concent_features(self):
+        del settings.VERIFIER_MIN_SSIM
+
+        errors = check_verifier_min_ssim()
+
+        self.assertEqual(len(errors), 1)
+        self.assertEqual(errors[0], create_error_31_verifier_min_ssim_is_not_set())
+
+    @override_settings(
+        CONCENT_FEATURES=[]
+    )
+    def test_that_verifier_min_ssim_not_set_will_not_produce_error_when_verifier_is_not_in_available_concent_features(self):
+        del settings.VERIFIER_MIN_SSIM
+
+        errors = check_verifier_min_ssim()
+
+        self.assertEqual(errors, [])
+
+    @override_settings(
+        CONCENT_FEATURES=[],
+        VERIFIER_MIN_SSIM=0,
+    )
+    def test_that_verifier_min_ssim_set_will_produce_error_when_verifier_is_not_in_available_concent_features(self):
+        errors = check_verifier_min_ssim()
+
+        self.assertEqual(len(errors), 1)
+        self.assertEqual(errors[0], create_error_32_verifier_min_ssim_is_set())
+
+    @override_settings(
+        CONCENT_FEATURES=[],
+        VERIFIER_MIN_SSIM=None,
+    )
+    def test_that_verifier_min_ssim_set_will_not_produce_error_when_verifier_is_not_in_available_concent_features(self):
+        errors = check_verifier_min_ssim()
+
+        self.assertEqual(errors, [])
+
+    @override_settings(
+        CONCENT_FEATURES=[
+            'verifier'
+        ],
+        VERIFIER_MIN_SSIM='test'
+    )
+    def test_that_verifier_min_ssim_set_to_non_float_will_produce_error(self):
+        errors = check_verifier_min_ssim()
+
+        self.assertEqual(len(errors), 1)
+        self.assertEqual(errors[0], create_error_33_verifier_min_ssim_has_wrong_type())
+
+    @override_settings(
+        CONCENT_FEATURES=[
+            'verifier'
+        ],
+        VERIFIER_MIN_SSIM=-1.01
+    )
+    def test_that_verifier_min_ssim_set_to_wrong_value_will_produce_error(self):
+        errors = check_verifier_min_ssim()
+
+        self.assertEqual(len(errors), 1)
+        self.assertEqual(errors[0], create_error_34_verifier_min_ssim_has_wrong_value(settings.VERIFIER_MIN_SSIM))
--- a/concent_api/utils/constants.py
+++ b/concent_api/utils/constants.py
@@ -58,8 +58,12 @@ class ErrorCode(enum.Enum):
     QUEUE_WRONG_STATE                                                   = 'queue.wrong_state'
     REQUEST_BODY_NOT_EMPTY                                              = 'request_body.not_empty'
     SUBTASK_DUPLICATE_REQUEST                                           = 'subtask.duplicate_request'
-    VERIFIIER_FILE_DOWNLOAD_FAILED                                      = 'verifier.file_download_failed'
-    VERIFIIER_UNPACKING_ARCHIVE_FAILED                                  = 'verifier.unpacking_archive_failed'
+    VERIFIER_COMPUTING_SSIM_FAILED                                     = 'verifier.computing_ssim_failed'
+    VERIFIER_FILE_DOWNLOAD_FAILED                                      = 'verifier.file_download_failed'
+    VERIFIER_LOADING_FILES_INTO_MEMORY_FAILED                          = 'verifier.loading_files_into_memory_failed'
+    VERIFIER_LOADING_FILES_WITH_OPENCV_FAILED                          = 'verifier.loading_files_with_opencv_failed'
+    VERIFIER_RUNNING_BLENDER_FAILED                                    = 'verifier.running_blender_failed'
+    VERIFIER_UNPACKING_ARCHIVE_FAILED                                  = 'verifier.unpacking_archive_failed'
 
 
 class MessageIdField(enum.Enum):
--- a/concent_api/verifier/tasks.py
+++ b/concent_api/verifier/tasks.py
@@ -7,6 +7,8 @@ import os
 from celery import shared_task
 from golem_messages import message
 from requests import HTTPError
+from skimage.measure import compare_ssim
+import cv2
 
 from django.conf import settings
 
@@ -21,10 +23,12 @@ from utils.decorators import log_task_errors
 from utils.decorators import provides_concent_feature
 from utils.logging import log_string_message
 from utils.helpers import upload_file_to_storage_cluster
+from .exceptions import VerificationError
 from .utils import clean_directory
 from .utils import delete_file
 from .utils import generate_blender_output_file_name
 from .utils import generate_upload_file_name
+from .utils import generate_verifier_storage_file_path
 from .utils import get_files_list_from_archive
 from .utils import prepare_storage_request_headers
 from .utils import run_blender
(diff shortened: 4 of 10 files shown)


concent_api/verifier/tests/test_integration_verification.py
```diff
@@ -240,5 +272,200 @@ def test_blender_verification_order_should_call_verification_result_with_result_
...
+        self.assertEqual(mock_get_files_list_from_archive.call_count, 1)
+        self.assertEqual(mock_delete_file.call_count, 2)
+        mock_verification_result.assert_called_once_with(
+            self.compute_task_def['subtask_id'],
+            VerificationResult.ERROR,
+            'error',
+            ErrorCode.VERIFIIER_LOADING_FILES_INTO_MEMORY_FAILED.name,
+        )
+
+    def test_blender_verification_order_should_continue_verification_if_opening_first_file_raise_os_error(self):
```
dybi: BTW: I am not sure if this is proper behaviour - if reading the file fails and it cannot be uploaded, why should verification continue?

concent_api/verifier/tasks.py
```diff
@@ -144,27 +152,125 @@ def blender_verification_order(
...
+        logger.error(f'Loading result files into memory failed with: {exception}')
+        verification_result.delay(
+            subtask_id,
+            VerificationResult.ERROR,
+            str(exception),
+            ErrorCode.VERIFIIER_LOADING_FILES_INTO_MEMORY_FAILED.name
+        )
+        return
+
+    if upload_file_content is not None:
```
dybi: I would prefer not to have `None` based checks... This whole `if` block should be moved below line `182`

concent_api/verifier/tasks.py
```diff
@@ -40,7 +48,7 @@ def blender_verification_order(
     result_size: int,
     result_package_hash: str,
     output_format: str,
-    scene_file: str,  # pylint: disable=unused-argument
+    scene_file: str,
```
dybi: this whole task has grown dramatically. There's a lot of logic and for me it is very hard to follow what's going on. I would really instist on having it refactored in a way that small functions are extracted and the task consits of calls to such small functions.

concent_api/verifier/tasks.py
```diff
@@ -144,27 +152,125 @@ def blender_verification_order(
...
-    )
+    blender_output_file_name = generate_blender_output_file_name(scene_file)
+    upload_file_name = generate_upload_file_name(subtask_id, output_format)
+
+    upload_file_content = None
+
+    try:
+        with open(generate_verifier_storage_file_path(blender_output_file_name), 'r') as upload_file:
+            upload_file_content = upload_file.read()
+    except OSError as exception:
```
dybi: why processing is not stopped on `OSError`, while it is on `MemoryError`?

concent_api/utils/constants.py
```diff
@@ -57,7 +57,11 @@ class ErrorCode(enum.Enum):
     QUEUE_WRONG_STATE                                                   = 'queue.wrong_state'
     REQUEST_BODY_NOT_EMPTY                                              = 'request_body.not_empty'
     SUBTASK_DUPLICATE_REQUEST                                           = 'subtask.duplicate_request'
+    VERIFIIER_COMPUTING_SSIM_FAILED                                     = 'verifier.computing_ssim_failed'
     VERIFIIER_FILE_DOWNLOAD_FAILED                                      = 'verifier.file_download_failed'
+    VERIFIIER_LOADING_FILES_INTO_MEMORY_FAILED                          = 'verifier.loading_files_into_memory_failed'
```
dybi: I am mot sure if we need to distinguish between these two:
`VERIFIIER_LOADING_FILES_INTO_MEMORY_FAILED` vs `VERIFIIER_LOADING_FILES_WITH_OPENCV_FAILED`

concent_api/verifier/tests/test_integration_verification.py
cameel: I agree. The verification should crash if the file cannot be opened.

The only case where we want to ignore problems with a file is when we try to upload our rendering result to the storage server. It's only for auditing purposes so we can ignore it and continue verification. Any other file is critical for us and a problem should result either in a crash (if we're sure this could have happened only due to a bug) or a verification error reported to the client (if it's client's fault, e.g. invalid file type or an empty file).

BTW, I see a typo in the code above: `VERIFIIER` -> `VERIFIER`.

concent_api/verifier/tasks.py
cameel: Yeah. I wouldn't catch `OSError` here. Make sure that the file exists. If it does than anything else going wrong is probably due to container misconfiguration, lack of disk space, etc.

concent_api/utils/constants.py
cameel: You have double `I` here.

concent_api/concent_api/tests/test_check_verifier_min_ssim.py
```diff
@@ -0,0 +1,78 @@
...
+        errors = check_verifier_min_ssim()
+
+        self.assertEqual(len(errors), 1)
+        self.assertEqual(errors[0], create_error_32_verifier_min_ssim_is_set())
+
+    @override_settings(
+        CONCENT_FEATURES=[],
+        VERIFIER_MIN_SSIM=None,
+    )
+    def test_that_verifier_min_ssim_set_will_not_produce_error_when_verifier_is_none_in_available_concent_features(self):
```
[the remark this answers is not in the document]
rwrzesien: Actually it was true too, but your version sounds much better :)

concent_api/verifier/tasks.py
```diff
@@ -144,27 +152,125 @@ def blender_verification_order(
...
+            VerificationResult.ERROR,
+            str(exception),
+            ErrorCode.VERIFIIER_COMPUTING_SSIM_FAILED.name
+        )
+        return
+
+    assert isinstance(ssim, float)
+
+    # Compare SSIM with VERIFIER_MIN_SSIM.
+    if ssim > settings.VERIFIER_MIN_SSIM:
```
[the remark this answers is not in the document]
rwrzesien: It is ok, from issue description:
`If the computed SSIM value is above the threshold, add verification_result with result == MATCH to the queue.`

Btw. I have reversed the condition to match our styling.

concent_api/verifier/tests/test_integration_verification.py
rwrzesien: It is working ok, just the test name was misleading, `first_file` meant blender output file in this context, I have changed the test name.

concent_api/verifier/tasks.py
rwrzesien: @dybi Lets discuss this tomorrow.

concent_api/utils/constants.py
rwrzesien: Right, I will get rid of second one.

concent_api/verifier/tasks.py
rwrzesien: @dybi I have followed @cameel comment in https://github.com/golemfactory/concent/pull/318#discussion_r195136617

```
This should be outside of the try ... except block. An exception handler should cover only the code where you expect the exception.
```

Personally I think both of you are right in this case, but using `None` is less evil. I have moved place where variable is created to `expect` block

concent_api/verifier/tests/test_integration_verification.py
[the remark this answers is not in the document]
rwrzesien: :+1: done

concent_api/utils/constants.py
rwrzesien: :+1: done

concent_api/verifier/tasks.py
```diff
@@ -119,12 +127,12 @@ def blender_verification_order(
             unpack_archive(
                 os.path.basename(file_path)
             )
-        except (OSError, BadZipFile) as e:
+        except (OSError, BadZipFile) as exception:
```
rwrzesien: @cameel comment from https://github.com/golemfactory/concent/pull/514#discussion_r195779279
> What kind of OSError are you expecting here? I think you should crash on OSError.

Any problem that is related with reading file from disk. I have followed info from issue that if archive file can't be extracted (in this case, it can't too) verification should end here.

concent_api/verifier/tasks.py
```diff
@@ -119,12 +127,12 @@ def blender_verification_order(
             unpack_archive(
```
rwrzesien: @cameel comment from https://github.com/golemfactory/concent/pull/514#discussion_r195780046

> Will it crash if any of the unpacked files already exists? It should. We prefer to crash than to overwrite anything.

concent_api/verifier/tasks.py
```diff
@@ -100,7 +108,7 @@ def blender_verification_order(
                 subtask_id,
                 VerificationResult.ERROR.name,
                 str(exception),
-                ErrorCode.VERIFIIER_FILE_DOWNLOAD_FAILED.name
```
rwrzesien: @cameel comment from https://github.com/golemfactory/concent/pull/514#discussion_r195780493

> Will it crash if the file being downloaded already exists? It should.
> Any file operation where the file being created is not supposed to exist but it does should crash the verifier so that we can detect and fix the issue.

concent_api/concent_api/settings/base.py
```diff
@@ -223,6 +223,12 @@
             'level':     'DEBUG',
             'propagate': True,
         },
+        'concent.crash': {
```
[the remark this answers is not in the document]
rwrzesien: Please check my answer here https://github.com/golemfactory/concent/pull/518#discussion_r196187130

concent_api/verifier/tasks.py
```diff
@@ -144,27 +152,126 @@ def blender_verification_order(
...
-        VerificationResult.MATCH.name,
-    )
+    blender_output_file_name = generate_blender_output_file_name(scene_file)
+    upload_file_name = generate_upload_file_name(subtask_id, output_format)
+
+    # Read Blender output file.
+    try:
+        with open(generate_verifier_storage_file_path(blender_output_file_name), 'r') as upload_file:
+            upload_file_content = upload_file.read()
+    except OSError as exception:
```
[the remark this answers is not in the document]
rwrzesien: Ok, however the issue description says that we should continue.

concent_api/verifier/tasks.py
dybi: as discussed it will be done in separate issue

rwrzesien: Issue https://github.com/golemfactory/concent/issues/524

concent_api/utils/constants.py
```diff
@@ -58,8 +58,12 @@ class ErrorCode(enum.Enum):
...
     REQUEST_BODY_NOT_EMPTY                                              = 'request_body.not_empty'
     SUBTASK_DUPLICATE_REQUEST                                           = 'subtask.duplicate_request'
-    VERIFIIER_FILE_DOWNLOAD_FAILED                                      = 'verifier.file_download_failed'
-    VERIFIIER_UNPACKING_ARCHIVE_FAILED                                  = 'verifier.unpacking_archive_failed'
+    VERIFIER_COMPUTING_SSIM_FAILED                                     = 'verifier.computing_ssim_failed'
+    VERIFIER_FILE_DOWNLOAD_FAILED                                      = 'verifier.file_download_failed'
+    VERIFIER_LOADING_FILES_INTO_MEMORY_FAILED                          = 'verifier.loading_files_into_memory_failed'
+    VERIFIER_LOADING_FILES_WITH_OPENCV_FAILED                          = 'verifier.loading_files_with_opencv_failed'
+    VERIFIER_RUNNING_BLENDER_FAILED                                    = 'verifier.running_blender_failed'
+    VERIFIER_UNPACKING_ARCHIVE_FAILED                                  = 'verifier.unpacking_archive_failed'
```
kbeker: Please delete spaces:)

rwrzesien: If you really want I will remove them, but isn't it a good example when it increases readability ? @dybi what do you think ?

concent_api/verifier/tasks.py
rwrzesien: No, currently it will overwrite.

concent_api/verifier/tasks.py
rwrzesien: I don't see an option to set it in unpacking function, so I would have to check manually if the any file already exists. But just a line before I am doing `clean_directory`, so there should not be any files there. Do you think we need this extra check ?

rwrzesien: Added, it will now crash with raising exception, is that what you expected ? Or should I catch it first and log, or handle somehow ?

concent_api/verifier/tasks.py
rwrzesien: I don't see an option to set it in unpacking function, so I would have to manually extract files list and check if any file already exists. But just few lines before I am doing `clean_directory`, so there should not be any files there. Do you think we need this extra check ?

cameel: It's just a sanity check. It would be better to have this but we can skip it if it adds complexity or requires a lot of extra work in this particular case.

concent_api/verifier/tasks.py
cameel: The handler on the task is supposed to catch it, report to sentry and then reraise. You do not have to do anything else.

concent_api/verifier/tasks.py
```diff
@@ -17,15 +20,20 @@
...
+from .utils import generate_upload_file_name
+from .utils import generate_verifier_storage_file_path
 from .utils import get_files_list_from_archive
 from .utils import prepare_storage_request_headers
 from .utils import run_blender
 from .utils import store_file_from_response_in_chunks
 from .utils import unpack_archive
 
 
+crash_logger = logging.getLogger('concent.crash')
```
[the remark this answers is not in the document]
cameel: One is for reporting to sentry, the other one is for writing to the log.

concent_api/verifier/tasks.py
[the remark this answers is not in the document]
cameel: Like I said elsewhere, this case is for uploading the rendering result for audit purposes. If it fails, well, tough luck. We should not stop verification because of that.

concent_api/verifier/tasks.py
cameel: By "archive can't be extracted" I mean a damaged archive. I think `BadZipFile` covers that. Not stuff like permission errors, missing files, etc. Those things should crash the verifier.

cameel:
I have added answers for @rwrzesien's questions about error handling. But honestly, it's pretty hard to give good, consistent answers when reviewing each case in isolation. I'm afraid that my answers might not be consistent. Please apply the requested changes or answer if it's still unclear or if my suggestions seem inconsistent.

But I think I'll need to look at the task as a whole at some point and review the error handling from that perspective if you want consistent answers :)

concent_api/verifier/tasks.py
rwrzesien: Should it just raise exception or schedule `verification_result` task with error ?

cameel: Both :)

Ideally you would raise an exception, that exception would be caught just before escaping the task and it would result in the error being reported to Sentry and `verification_result` being sent to the control cluster. One important thing is that if you schedule a `verification_result`, the task must not be retried any more.

Thats a bit like the decorator catching and reporting exceptions but not quite because it's specific to verification (it reports the result) and it does not reraise the exception.

rwrzesien: :+ done

Issue #446, opened 2018-06-05
The conductor accepts only request with the  `.zip` file extension


bartoszbetka:
The conductor `/report-upload/` endpoint accepts request with the `.zip` file extension. When uploaded file have diffrent extension like: `.rar`, `.blend`, conductor send response with HTTP 404 status. We want to conductor accepts all file, even without extension.

PR #448, opened 2018-06-05
Fix regex in conductor/urls to match all files

dybi:
resolves: https://github.com/golemfactory/concent/issues/446
- Previously only file with ".zip" extension were supported

### DIFF
--- a/concent_api/conductor/urls.py
+++ b/concent_api/conductor/urls.py
@@ -3,5 +3,5 @@ from django.conf.urls import url
 from .views import report_upload
 
 urlpatterns = [
-    url(r'^report-upload/(?P<file_path>.+\.zip)$', report_upload, name='report-upload'),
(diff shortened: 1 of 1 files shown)


concent_api/conductor/urls.py
```diff
@@ -3,5 +3,5 @@
 from .views import report_upload
 
 urlpatterns = [
-    url(r'^report-upload/(?P<file_path>.+\.zip)$', report_upload, name='report-upload'),
+    url(r'^report-upload/(?P<file_path>.*)$', report_upload, name='report-upload'),
```
cameel: Please use `.+` instead of `.*` so that empty names are not allowed.

Unless there's something in the view that will reject them.

dybi: Hm... I don't think that view would reject it. Changing.

Issue #382, opened 2018-05-15
Concent API does not have the same checksum validations as gatekeeper


cameel:
Gatekeeper currently has extensive validations for `FileTransferToken`, including file checksums. Concent API does not have the same validations. In effect it's possible for the client to send us `ReportComputedTask` with a bad checksum and crash the server because we'll use that checksum in a token and use it in a request to the storage server.

To fix this:
- Validate `package_hash` in every `TaskToCompute` and `ReportComputedTask` the same way gatekeeper does.
- When creating tokens, `assert` that the created token passes those validations.
- Use the new `is_valid` hook from golem-messages.

Please also do #79 first.

Jakub89:
`is_valid()` method is possible to use only for `message.tasks.TaskMessage` so we can't use it with `ForcePayment` message. Don't know if that's a bug, just saying that I won't be able to do it as a part of that issue.

cameel:
OK. That's something we should raise with @shadeofblue. I think `ForcePayment` should have validations too.

cameel:
@Jakub89 could you explain why you needed to validate `ForcePayment` in this task?

I'm reopening this just so that it does not get buried because @Jakub89 is now on vacation.

Jakub89:
@cameel I don't really need to validate it, It's just something I've notice while adding `is_valid()` method. We use it for every message that go through `send()` endpoint, and `ForcePayment` is the only one that is not instance of `message.tasks.TaskMessage`, so we need to have special case for it.
I've created issue on golem-messages repo, but still now reply:
https://github.com/golemfactory/golem-messages/issues/233

dybi:
a separate issue has been created in proper repository - closing

PR #485, opened 2018-06-12
Fix concent api checksum validation

Jakub89:
Resolves: #382

### DIFF
--- a/concent_api/core/validation.py
+++ b/concent_api/core/validation.py
@@ -14,9 +14,10 @@ from core.constants                 import GOLEM_PUBLIC_KEY_HEX_LENGTH
 from core.constants                 import MESSAGE_TASK_ID_MAX_LENGTH
 from core.constants import VALID_ID_REGEX
 from core.constants import VALID_SHA1_HASH_REGEX
-from core.exceptions import HashingAlgorithmError
 from core.enums import HashingAlgorithm
 from core.exceptions import FileTransferTokenError
+from core.exceptions import GolemMessageValidationError
+from core.exceptions import HashingAlgorithmError
 from core.exceptions import Http400
 from core.utils import hex_to_bytes_convert
 from utils.helpers import join_messages
@@ -263,6 +264,11 @@ def get_validated_client_public_key_from_client_message(golem_message: message.b
             )
 
     elif isinstance(golem_message, message.tasks.TaskMessage):
+        if not golem_message.is_valid():
+            raise GolemMessageValidationError(
+                "Golem message invalid",
+                error_code=ErrorCode.MESSAGE_INVALID
+            )
         task_to_compute = golem_message.task_to_compute
     else:
         raise Http400(
--- a/concent_api/core/validation.py
+++ b/concent_api/core/validation.py
@@ -15,10 +15,10 @@ from core.constants                 import MESSAGE_TASK_ID_MAX_LENGTH
 from core.constants import VALID_ID_REGEX
 from core.constants import VALID_SHA1_HASH_REGEX
 from core.exceptions import HashingAlgorithmError
+from core.enums import HashingAlgorithm
 from core.exceptions import FileTransferTokenError
 from core.exceptions import Http400
 from core.utils import hex_to_bytes_convert
-from gatekeeper.enums import HashingAlgorithm
 from utils.helpers import join_messages
 from utils.constants                import ErrorCode
 
--- a/concent_api/core/validation.py
+++ b/concent_api/core/validation.py
@@ -14,6 +14,7 @@ from core.constants                 import GOLEM_PUBLIC_KEY_HEX_LENGTH
 from core.constants                 import MESSAGE_TASK_ID_MAX_LENGTH
 from core.constants import VALID_ID_REGEX
 from core.constants import VALID_SHA1_HASH_REGEX
+from core.exceptions import HashingAlgorithmError
 from core.exceptions import FileTransferTokenError
 from core.exceptions import Http400
 from core.utils import hex_to_bytes_convert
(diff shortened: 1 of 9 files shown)


concent_api/core/validation.py
```diff
@@ -121,6 +122,9 @@ def validate_task_to_compute(task_to_compute: message.TaskToCompute):
 
     validate_hex_public_key(task_to_compute.provider_public_key, 'provider_public_key')
     validate_hex_public_key(task_to_compute.requestor_public_key, 'requestor_public_key')
+
+    validate_secure_hash_algorithm(task_to_compute.package_hash)
```
rwrzesien: I think extra lines above and below are not needed.

concent_api/core/validation.py
```diff
@@ -373,3 +371,21 @@ def validate_file_transfer_token(file_transfer_token: message.concents.FileTrans
     categories = [file_info['category'] for file_info in file_transfer_token.files]
     if len(set(categories)) != len(categories):
         raise FileTransferTokenError("'category' field must be unique across FileInfo list.", ErrorCode.MESSAGE_FILES_CATEGORY_NOT_UNIQUE)
+
+
+def validate_secure_hash_algorithm(checksum: str):
+    if not isinstance(checksum, str):
+        raise HashingAlgorithmError("'checksum' must be a string.", ErrorCode.MESSAGE_FILES_CHECKSUM_WRONG_TYPE)
```
rwrzesien: Please break lines in those raises here.

concent_api/core/validation.py

concent_api/core/validation.py

concent_api/core/validation.py
```diff
@@ -373,3 +369,36 @@ def validate_file_transfer_token(file_transfer_token: message.concents.FileTrans
     categories = [file_info['category'] for file_info in file_transfer_token.files]
     if len(set(categories)) != len(categories):
         raise FileTransferTokenError("'category' field must be unique across FileInfo list.", ErrorCode.MESSAGE_FILES_CATEGORY_NOT_UNIQUE)
+
+
+def validate_secure_hash_algorithm(checksum: str):
```
dybi: please add unit tests for the validation

concent_api/core/validation.py
```diff
@@ -373,3 +369,36 @@ def validate_file_transfer_token(file_transfer_token: message.concents.FileTrans
...
+    if not isinstance(checksum, str):
+        raise HashingAlgorithmError(
+            "'checksum' must be a string.",
+            ErrorCode.MESSAGE_FILES_CHECKSUM_WRONG_TYPE
+        )
+
+    if len(checksum) == 0 or checksum.isspace():
```
dybi: why blending those two together? Wouldn't it be better if we had two separete check with two separate errors?

concent_api/core/validation.py
```diff
@@ -373,3 +369,36 @@ def validate_file_transfer_token(file_transfer_token: message.concents.FileTrans
...
+        raise HashingAlgorithmError(
+            "'checksum' cannot be blank or contain only whitespace.",
+            ErrorCode.MESSAGE_FILES_CHECKSUM_EMPTY
+        )
+
+    if ":" not in checksum:
+        raise HashingAlgorithmError(
+            "'checksum' must consist of two parts separated with a semicolon.",
```
dybi: I think I would suggest the form of checksum here:
`"checksum must be in format of: '<ALGORITHM>:<HASH>'"` or sth similar

concent_api/core/validation.py
```diff
@@ -262,6 +264,11 @@ def get_validated_client_public_key_from_client_message(golem_message: message.b
             )
 
     elif isinstance(golem_message, message.tasks.TaskMessage):
+        if not golem_message.is_valid():
+            raise Http400(
```
dybi: Please do not use `Http400` in validation functions. As discussed earlier, validation layer should not be tied to any particulat transportation layer. If there is no suitable exception defined, please define a new one and map it to proper HttpResponse/JsonResponse in decorator layer.

concent_api/core/validation.py
```diff
@@ -14,6 +14,7 @@
 from core.constants                 import MESSAGE_TASK_ID_MAX_LENGTH
 from core.constants import VALID_ID_REGEX
 from core.constants import VALID_SHA1_HASH_REGEX
+from core.exceptions import HashingAlgorithmError
```
dybi: since `HashingAlgorithmError` is defined in `core`, and `HashingAlgorithm` is used in `validation` module only, its defition should be moved from `gatekeeper` to `core`

concent_api/core/validation.py
```diff
@@ -373,3 +369,36 @@ def validate_file_transfer_token(file_transfer_token: message.concents.FileTrans
...
+            ErrorCode.MESSAGE_FILES_CHECKSUM_EMPTY
+        )
+
+    if ":" not in checksum:
+        raise HashingAlgorithmError(
+            "'checksum' must consist of two parts separated with a semicolon.",
+            ErrorCode.MESSAGE_FILES_CHECKSUM_WRONG_FORMAT
+        )
+
+    if not checksum.split(":")[0] == set(HashingAlgorithm._value2member_map_).pop():  # type: ignore
```
dybi: this is really bad way of checking if algorithm written in part of the checksum is supported. Why not just sth like:
`if checksum.split(":")[0] not in [e.value for e in HashingAlgorithm]`?
Or you can even add a static method to `HashingAlgorithm`:
```python
@unique
class HashingAlgorithm(Enum):
    SHA1 = 'sha1'

   def values():
       return [a.value for a in HashingAlgorithm]
```
and then check it like: ``if checksum.split(":")[0] not in HashingAlgorithm.values()`

concent_api/core/validation.py
```diff
@@ -373,3 +369,36 @@ def validate_file_transfer_token(file_transfer_token: message.concents.FileTrans
...
+        raise HashingAlgorithmError(
+            "One of the checksums is from an unsupported hashing algorithm.",
```
dybi: I do not like the message. Validation function checks a checksum and is not aware that there are many of them. It should be more like `f"Checksum {checksum} comes from an unsupported hashing algorithm"`

concent_api/core/validation.py

concent_api/core/validation.py
Jakub89: Not sure it's pretty simple check :P

concent_api/core/validation.py
Jakub89: I've made custom `GolemMessageValidationError` for this. Should I apply it to all `Http400` exceptions in this function?

concent_api/core/validation.py

concent_api/core/validation.py

concent_api/core/validation.py

concent_api/core/validation.py

PR #494, opened 2018-06-13
Fix static files collecting

MichalW-CP:


concent_api/concent_api/settings/base.py
```diff
@@ -136,7 +136,7 @@
 
 # Static files (CSS, JavaScript, Images)
 # https://docs.djangoproject.com/en/1.11/howto/static-files/
-STATICFILES_DIRS = [os.path.join(BASE_DIR, "static")]
+STATICFILES_DIRS = [os.path.join(BASE_DIR, "core", "static")]
```
cameel: That's not what we agreed on yesterday. It was supposed to be at the same level as `core` - i.e. inside `BASE_DIR`. The version before this change was correct.

PR #498, opened 2018-06-13
Bugfix transistion to additional verification state doesent add svt time to next deadline

### LINKED ISSUE #497: Missing addition SVT to next_deadline during transition change from VERIFICATION FILE TRANSFER TO ADDITIONAL VERIFICATION
Concent must add `Subtask_verification_time` to next_deadline when it is passing to `ADDITIONAL VERIFICATION` state. It's abbreviating time for `Provider` which he needs to get verification results. In other way it may cause problems because during `Blender` rendering it may pass deadline and Provider won't get his founds and appropriate message.

kbeker:
First needs to be merged: https://github.com/golemfactory/concent/pull/495

Resolves #497

### DIFF
--- a/concent_api/core/subtask_helpers.py
+++ b/concent_api/core/subtask_helpers.py
@@ -1,5 +1,6 @@
 from base64                     import b64encode
 from logging import getLogger
+from typing import Optional
 
 from django.db.models           import Q
 from django.utils               import timezone
@@ -11,6 +12,7 @@ from core.transfer_operations   import store_pending_message
 from core.transfer_operations   import verify_file_status
 from utils.helpers              import deserialize_message
 from utils.helpers              import get_current_utc_timestamp
+from utils.helpers import parse_timestamp_to_utc_datetime
 from utils                      import logging
 
 logger = getLogger(__name__)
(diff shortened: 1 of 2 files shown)


concent_api/core/tasks.py
```diff
@@ -67,9 +68,11 @@ def upload_finished(subtask_id: str):
             return
 
         # Change subtask state to ADDITIONAL VERIFICATION.
-        subtask.state = Subtask.SubtaskState.ADDITIONAL_VERIFICATION.name  # pylint: disable=no-member
-        subtask.full_clean()
-        subtask.save()
+        update_subtask_state(
+            subtask=subtask,
+            state=Subtask.SubtaskState.ADDITIONAL_VERIFICATION.name,  # pylint: disable=no-member
+            next_deadline=subtask.next_deadline.timestamp() + settings.SUBTASK_VERIFICATION_TIME
```
dybi: as the function accepts `Optional[int]` and here we are adding `float + int`, I think I would like to have casting to `int`

concent_api/core/subtask_helpers.py
```diff
@@ -112,14 +114,15 @@ def update_timed_out_subtasks(
...
 ):
     logging.log_change_subtask_state_name(
         logger,
         subtask.state,
         state,
     )
-    subtask.state    = state
-    subtask.next_deadline = None
+    subtask.state = state
+    subtask.next_deadline = next_deadline if next_deadline is None else parse_timestamp_to_utc_datetime(next_deadline)
```
dybi: I think that `None if next_deadline is None else parse_timestamp_to_utc_datetime(next_deadline)` would be more readable

dybi: done

concent_api/core/tasks.py
dybi: done

Issue #473, opened 2018-06-11
Change the way in which files are downloaded and stored by verifier


rwrzesien:
There is a problem with downloading files in verifier's `blender_verification_order` task. Due to the way in which we store and build file paths, we have two files for each subtask, ie:
`blender/source/16356/16356.sub_task_id.zip`
`blender/result/16356/16356.sub_task_id.zip`
So, actually the file names are the same and currently are first one is overwritten by second one. This should be avoided by adding a prefix to each file.

PR #514, opened 2018-06-15
Fix for overwritting files which are downloaded and stored by verifier

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/473

### DIFF
--- a/concent_api/verifier/tasks.py
+++ b/concent_api/verifier/tasks.py
@@ -103,8 +103,13 @@ def blender_verification_order(
     # Remove any files from VERIFIER_STORAGE_PATH.
     clean_directory(settings.VERIFIER_STORAGE_PATH)
 
+    package_paths_to_downloaded_file_names = {
+        source_package_path: f'source_{os.path.basename(source_package_path)}',
+        result_package_path: f'result_{os.path.basename(result_package_path)}',
+    }
+
     # Download all the files listed in the message from the storage server to local storage.
-    for file_path in (source_package_path, result_package_path):
+    for file_path, download_file_name in package_paths_to_downloaded_file_names.items():
         try:
             file_transfer_token.sig = None
             cluster_response = send_request_to_storage_cluster(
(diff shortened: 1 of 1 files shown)


concent_api/verifier/tasks.py
```diff
@@ -114,11 +119,9 @@ def blender_verification_order(
...
 
     # Verifier unpacks the archive with project source.
-    for file_path in (source_package_path, result_package_path):
+    for file_path, download_file_name in package_paths_to_downloaded_file_names.items():
         try:
-            unpack_archive(
-                os.path.basename(file_path)
-            )
+            unpack_archive(download_file_name)
         except (OSError, BadZipFile) as e:
```
cameel: What kind of `OSError` are you expecting here? I think you should crash on `OSError`.

concent_api/verifier/tasks.py
```diff
@@ -114,11 +119,9 @@ def blender_verification_order(
             raise
 
     # Verifier unpacks the archive with project source.
-    for file_path in (source_package_path, result_package_path):
+    for file_path, download_file_name in package_paths_to_downloaded_file_names.items():
         try:
-            unpack_archive(
-                os.path.basename(file_path)
-            )
+            unpack_archive(download_file_name)
```
cameel: Will it crash if any of the unpacked files already exists? It should. We prefer to crash than to overwrite anything.

concent_api/verifier/tasks.py
```diff
@@ -90,7 +95,7 @@ def blender_verification_order(
                 cluster_response,
```
cameel: Will it crash if the file being downloaded already exists? It should.

Any file operation where the file being created is not supposed to exist but it does should crash the verifier so that we can detect and fix the issue.

rwrzesien:
@cameel Those changes on which you have placed comments are not related with this PR/issue, please move them to https://github.com/golemfactory/concent/pull/441.

concent_api/verifier/tasks.py
rwrzesien: Answered in https://github.com/golemfactory/concent/pull/441

concent_api/verifier/tasks.py
rwrzesien: Answered in #441

concent_api/verifier/tasks.py
rwrzesien: Answered in #441

Issue #488, opened 2018-06-12
5xx errors are returned as HTML by default


cameel:
In #57 we implemented middleware that replaces HTML pages for 5xx errors with JSON equivalents. JSON should be returned by default as can be seen in [determine_return_type()](https://github.com/golemfactory/concent/blob/v0.7.1/concent_api/concent_api/middleware.py#L55-L65). @shadeofblue reports that he still gets the HTML pages on staging even though he's not sending the `Accept` header.

cameel:
Actually, @shadeofblue could you verify that you're not sending the `Accept` header? Maybe requests is including one by default?

rwrzesien:
@cameel `requests` does not but `curl` is, and this header has value `*/*`, and `best_match` returns `text/html` for it. I can add condition which will return `application/json` for this value. What are your thoughts on that ?

cameel:
It should choose `application/json` in any situation where both it and `test/html` have the same `q` value. `*/*` is only a special case of this.

Also, aside from the `q` value there's also specificity. `*/* is less specific than `text/*` which is less specific than `text/html`. The value with higher specificity should be chosen.

>  Media ranges can be overridden by more specific media ranges or specific media types. If more than one media range applies to a given type, the most specific reference has precedence. For example,
> 
>        Accept: text/*, text/html, text/html;level=1, */*
> 
> have the following precedence:
> 
>        1) text/html;level=1
>        2) text/html
>        3) text/*
>        4) */*

[RFC 2616 Hypertext Transfer Protocol -- HTTP/1.1 > 14 Header Field Definitions > 14.1 Accept](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.1)

What does it do e.g. when the header is `text/html; */*`? It should chose HTML then because it's more specific. The hack you propose would not cover that.

PR #517, opened 2018-06-18
Fix for returning HTML on 5xx errors when Accept header is */*

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/488

### DIFF
--- a/concent_api/concent_api/middleware.py
+++ b/concent_api/concent_api/middleware.py
@@ -54,7 +54,9 @@ class ConcentVersionMiddleware(object):
 
 def determine_return_type(request_meta):
     try:
-        preferred_types = ['application/json', 'text/html']
+        # The list of preferred mime-types should be sorted in order of increasing desirability,
+        # in case of a situation where there is a tie.
+        preferred_types = ['text/html', 'application/json']
         if "HTTP_ACCEPT" in request_meta:
             return best_match(preferred_types, request_meta['HTTP_ACCEPT'])
         return "application/json"
--- a/concent_api/concent_api/tests/test_middleware.py
+++ b/concent_api/concent_api/tests/test_middleware.py
@@ -291,14 +291,14 @@ class TestDetermineReturnType(unittest.TestCase):
 
         self.assertEqual(return_type, "text/html")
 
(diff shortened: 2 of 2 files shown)


concent_api/concent_api/tests/test_middleware.py
```diff
@@ -317,3 +318,13 @@ def test_that_empty_string_is_returned_when_accept_header_is_malformed(self):
         return_type = determine_return_type(request_headers)
 
         self.assertEqual(return_type, "")
+
+    def test_that_with_value_from_accept_header_values_transited_to_default_list_json_is_used(self):
+        for value in ACCEPT_HEADER_VALUES_TRANSITED_TO_DEFAULT:
+            request_headers = {
+                "HTTP_ACCEPT": value
```
dybi: what would be desired behaviour with `HTTP_ACCEPT` like:
`"text/html, */*"`
`

rwrzesien: @dybi `best_match` returns `text/html` in that case.

concent_api/concent_api/middleware.py
```diff
@@ -55,7 +56,10 @@ def __call__(self, request):
 def determine_return_type(request_meta):
     try:
         preferred_types = ['application/json', 'text/html']
```
cameel: We have a bug here:

From [best_match()](https://github.com/dbtsai/python-mimeparse/blob/master/mimeparse.py#L155-L184) docstring:
> (...) The value of 'supported'
> is a list of mime-types. The list of supported mime-types should be sorted
> in order of increasing desirability, in case of a situation where there is
> a tie.

So `'application/json'` should be last if we want it to be chosen when the quality values are equal but we have it first.

concent_api/concent_api/tests/test_middleware.py
cameel: And that's correct. Spec says that the more specific types should be chosen (the one with less wildcards), which is `text/html` in this case.

concent_api/concent_api/middleware.py
rwrzesien: :+1: done, and yes, the original issue was caused by this.

Issue #511, opened 2018-06-15
Reporting worker errors to Sentry


cameel:
Django catches all exceptions that escape out of views and handles them by returning HTTP 500 errors. It also logs them to the `django.request` logger. We should have a similar mechanism for Celery tasks.

- Create a logger called `concent.crash`. It should have the same handlers as `django.request` so that things logged there get reported to Sentry.
- Create a decorator that:
    - Catches all exceptions.
    - Logs the exception with its stack trace to the `concent.crash` handler.
    - Re-raises the exception so that it can interrupt the task. It's important for Celery to know that the task did not finish successfully.
- Use the decorator on all our celery tasks.

PR #518, opened 2018-06-18
Feature reporting worker errors to sentry

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/511

### DIFF
--- a/concent_api/utils/decorators.py
+++ b/concent_api/utils/decorators.py
@@ -1,5 +1,6 @@
 from functools                      import wraps
 from logging                        import getLogger
+import traceback
 
 from django.db                      import transaction
 from django.http                    import JsonResponse
@@ -33,6 +34,7 @@ from utils.shortcuts                import load_without_public_key
 from utils                          import logging
 
 logger = getLogger(__name__)
+crash_logger = getLogger('concent.crash')
 
 
 def require_golem_auth_message(view):
@@ -275,4 +277,18 @@ def log_communication(view):
         log_json_message(logger, json_message_to_log)
         response_from_view = view(request,  golem_message, client_public_key)
         return response_from_view
+    return wrapper
+
+
+def log_task_errors(task):
+
+    @wraps(task)
+    def wrapper(*args, **kwargs):
+        try:
+            return task(*args, **kwargs)
+        except Exception as exception:
+            crash_logger.error(
+                f'Exception occurred while executing task {task.__name__}: {exception}, Traceback: {traceback.format_exc()}'
+            )
+            raise
     return wrapper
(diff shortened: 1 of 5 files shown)


concent_api/concent_api/settings/base.py
```diff
@@ -223,6 +223,12 @@
             'level':     'DEBUG',
             'propagate': True,
         },
+        'concent.crash': {
```
dybi: Do we need a separate logger here? `'request.django'` is configured in the same way

rwrzesien: @dybi According to description, yes

> Create a logger called concent.crash. It should have the same handlers as django.request so that things logged there get reported to Sentry.

But personally I think we might use 'request.django' as well.

dybi: Ok, missed that. It such case - just merge it ;)

cameel: We don't strictly need it but I think using `django.request` everywhere would be pretty misleading and might lead to people using it the wrong way. The hierarchy of loggers mirrors the code structure - as  the typical usage where the dotted path of the module becomes the name of hte logger indicates. The fact that two loggers have the same handlers does not mean that they have to be merged into a single logger.

I think that we should eventually replace all other uses of `django.request` with the new logger and leave the original logger to Django.

concent_api/utils/decorators.py
```diff
@@ -276,3 +278,18 @@ def wrapper(request, golem_message, client_public_key):
...
+
+    @wraps(task)
+    def wrapper(*args, **kwargs):
+        try:
+            return task(*args, **kwargs)
+        except Exception as exception:
+            if settings.DEBUG is True:
+                crash_logger.error(
+                    f'Exception occurred while executing task {task.__name__}: {exception}, Traceback: {traceback.print_exc()}'
+                )
```
cameel: No!!! ;) We definitely don't want to disable error reports on the clusters (all our clusters are running with `DEBUG = False`). If anything it would be the opposite - don't send mails when running locally but send them when running on the cluster.

But this should not depend on this setting at all. It's up to the handler to suppress the report if you're running in debug mode.

concent_api/utils/tests/test_decorators.py
```diff
@@ -229,3 +230,19 @@ def dummy_view_handle_http_response_not_allowed(_request, _message, _client_publ
 
         self.assertEqual(response.status_code, 405)
         self.assertEqual(len(response.content), 0)
+
+    @override_settings(
+        DEBUG=True
+    )
```
cameel: A test is not effective if you're checking only one condition. You should check both the positive and the negative case in unit tests.

rwrzesien: Right, however after change above it will not be an issue anymore.

Issue #480, opened 2018-06-12
More logging in conductor, verifier and workers


cameel:
We need more logging in the new components added as a part of verification use case. While at it, please also make existing logging in other component consistent with your changes.

- Log the fact of starting and finishing a Celery task.
- Log the fact of receiving a HTTP request and returning a response.
- Log the fact of communication with another component: uploads and downloads from storage server, adding a task to the work queue, etc. 
- You may also log some of the more important choices made by the code. E.g. failed or accepted validation, the fact that conductor has a verification request and all necessary files, issuing a token, etc.
- Every time you log, add some information that will allow us to identify and connect everything logged by one process into a single flow. E.g. subtask ID, client key, etc.

PR #525, opened 2018-06-19
Feature more logging in conductor verifier and workers

pawelkisielewicz:
resolves https://github.com/golemfactory/concent/issues/480


concent_api/conductor/tasks.py
```diff
@@ -79,12 +92,21 @@ def upload_acknowledged(
...
+        f'Source_package_hash: {source_package_hash} ',
+        f'Result_file_size: {result_file_size} ',
+        f'Result_package_hash: {result_package_hash} '
+    )
     assert isinstance(subtask_id, str)
 
     try:
         verification_request = VerificationRequest.objects.get(subtask_id=subtask_id)
     except VerificationRequest.DoesNotExist:
-        logging.error(
```
pawelkisielewicz: Changed level of logging- i think it should not be error. If it should I will change it

cameel: It should be an error.

This situation means that the control cluster is acknowledging an upload for a verification request that does not exist. It should never happen because Conductor is not supposed to send `upload_finished` if it does not exist. It means that something somewhere is seriously wrong.

concent_api/utils/logging.py
```diff
@@ -340,7 +346,7 @@ def log_json_message(logger: Logger, message: JsonResponse):
 
 
 def log_string_message(logger: Logger, *messages_to_log: str):
-    logger.info(join_messages(messages_to_log))
+    logger.info(' '.join(messages_to_log))
```
cameel: Why not `join_messages()`? If you're sure that none of the messages is `None` or `''` please add an assertion checking that. Otherwise you should use `join_messages()`.

concent_api/verifier/tasks.py
```diff
@@ -47,6 +48,19 @@ def blender_verification_order(
     output_format: str,
     scene_file: str,  # pylint: disable=unused-argument
 ):
+    log_string_message(
+        logger,
+        f'Blender_verification_order_starts. SUBTASK_ID: {subtask_id} ',
+        f'Source_package_path {source_package_path} ',
```
cameel: `log_string_message()` already joins the messages with a space in between. Adding a space here at the end will result in messages being separated with two spaces which will look weird.

cameel: You should add some speparator between messages. As it is now it will be hard to tell where one message ends and the other starts.

Also, consider putting the values in quotes.

concent_api/conductor/tasks.py
pawelkisielewicz: Done

concent_api/utils/logging.py
pawelkisielewicz: `join_messages` returns string like `('message1', 'message2')`. It does not look good in logs, so i changed it to simple `join`, instead of use 'join_messages` and format output. Added assertion.

concent_api/verifier/tasks.py
```diff
@@ -65,6 +79,11 @@ def blender_verification_order(
                 subtask_id,
                 VerificationResult.MISMATCH.name,
             )
+        result = str(VerificationResult.MATCH if subtask_id[-1] == 'm' else VerificationResult.MISMATCH)
```
dybi: this check is being made twice. please refactor

concent_api/verifier/tasks.py
pawelkisielewicz: Right... done

pawelkisielewicz: Done. Added colon in messages and dot between messages.

concent_api/verifier/tasks.py
pawelkisielewicz: Done

concent_api/utils/logging.py
pawelkisielewicz: It works good.. I should just use it like `join_messages(*messages_to_log)` instead of `join_messages(messages_to_log)`. It is merged now, will be fixed in separate pull request

PR #530, opened 2018-06-20
Feature rendering image fragment in verifier

### LINKED ISSUE #520: Rendering image fragment in verifier
Currently Verifier tries to render the whole image defined in the `.blend` file. This is not the correct approach because in most cases the goal of a single subtask is to render only a fragment of that image, to be later joined with other fragments rendered by other providers.

[A typical `TaskToCompute` message](https://github.com/golemfactory/concent/issues/235#issuecomment-378525017) contains two scripts:
- `src_code` - a Python script to be executed by Golem inside the docker image. It gets a few parameters from `TaskToCompute` (via a `params.py` module generated by Golem) and runs Blender.
- `script_src` - a Python script to be executed by Blender. It's generated by rendering [blendercrop.py.template](https://github.com/golemfactory/golem/blob/develop/apps/blender/resources/scripts/blendercrop.py.template) with parameters like size of the image, crop window, etc. `src_code` script writes `script_src` to a file called `blenderscript.py` and passes its path to Blender.

To render just the fragment we must pass the `src_code` script to Blender using `-P` option and enable scripting with `-y`.

The purpose of this task is to send the whole script from the control cluster to the verifier and just let Blender execute it. This is not a good design from the security standpoint (the script comes from the client) but will get the job done. There will be a separate task for making this mechanism more secure by allowing only whitelisted scripts.

Steps to do:
- Get the value of `script_src` value from `TaskToCompute` included in `SubtaskResultsVerify` and pass it along to the verifier.
    - Add `blender_crop_script` parameter to `blender_verification_request` and `blender_verification_order` Celery tasks. It should contain the value from `script_src`. 
    - Add `blender_crop_script` field to `BlenderSubtaskDefinition` model.
- In verifier pass the script to Blender.
    - Remove the script after rendering to make sure that the next task does not reuse it.

rwrzesien:
Resolves #520.


concent_api/conductor/models.py
```diff
@@ -77,6 +77,9 @@ class OutputFormat(ChoiceEnum):
     # Relative path to the .blend file inside the source package.
     scene_file = CharField(max_length=MESSAGE_PATH_LENGTH)
 
+    # Relative path to the Python script to be executed by Blender.
```
dybi: it is not a path - it is the content of the script itself

cameel: Exactly. Please look at #520, it's all described in detail there. In particular the sample message linked there (https://github.com/golemfactory/concent/issues/235#issuecomment-378525017) shows the content of the `TaskToCompute.script_src` field which is what is passed to `blender_crop_script`.

rwrzesien: So what does it mean 
> Remove the script after rendering to make sure that the next task does not reuse it.

It means Subtask object field should be cleared ?

cameel: Please look at `src_code` in `TaskToCompute` (that's the other script, not the one we put in `blender_crop_script`). This script creates a file called `WORK_DIR + "/blenderscript.py"` and puts `script_src` in it. Our code for running blender is based on it so I assume you'll be doing something similar. This remark was just to make sure that you do not forget to remove this temporary file when Blender exits.

concent_api/verifier/tasks.py
```diff
@@ -188,7 +189,12 @@ def blender_verification_order(
         completed_process = run_blender(
             scene_file,
             output_format,
+            blender_crop_script,
         )
+
+        # Remove the script after rendering to make sure that the next task does not reuse it.
+        delete_file(blender_crop_script)
```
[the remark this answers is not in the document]
rwrzesien: One is in `try`, other is in `except` block, and `except` block does `return` (`try` is doing to if non-zero exit code is returned from subprocess) so it can't be done after whole `try/except`.

concent_api/conductor/models.py
rwrzesien: :+1: done, but I think it would be good to add this info in a more direct way to the issue.

cameel: I could, but honestly I think that issues are very detailed already. This is a detail that becomes apparent once you try to implement it. One look at the content of `TaskToCompute.script_src` is enough to realize that it's not a path. At worst, it would be discovered once you try to run it after implementing it and realize that it tries to use the script as if it were a path.

Actually, how did it even work for you? It could not have worked as implemented originally.

concent_api/verifier/tasks.py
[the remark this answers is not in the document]
cameel: It can. You could use `try ... finally`. Actually you must use `try ... finally` because otherwise the file is not deleted if an exception other than `SubprocessError` is raised.

concent_api/conductor/models.py
rwrzesien: > One look at the content of TaskToCompute.script_src is enough to realize that it's not a path

I didn't thought it is a path, but that it might be passed directly as a string.

> Actually, how did it even work for you? It could not have worked as implemented originally.

I didn't tested it yet, can it be tested locally ?

cameel: > I didn't thought it is a path, but that it might be passed directly as a string.

It is passed directly as a string. `blender_crop_script` is the whole script, stored directly in a string.

> I didn't tested it yet, can it be tested locally ?

Yes. Just like any other part of verification. For example the blender rendering task is not closed yet because @dybi is (was?) testing it with sample projects.

rwrzesien: > It is passed directly as a string. blender_crop_script is the whole script, stored directly in a string.

Ok, so first I though it is a path to one of the files from archives. I have checked in `TaskToCompute`. It looked like something which is not being set at the moment. Then during review I found out it is a script, so I though the script code might be passed directly to Blender command. As I understand now, it should be first written to a file, and that file name should be passed to Blender. This is the info I consider worth mentioning in description.

@dybi Are you able to test this one too or provide TL:DR how to do this locally ?

concent_api/verifier/tasks.py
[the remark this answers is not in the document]

concent_api/verifier/tests/test_integration_verification.py
```diff
@@ -47,28 +47,17 @@ class VerifierVerificationIntegrationTest(ConcentIntegrationTestCase):
...
-            subtask_id='zzz523',
-        )
-
+        self.task_to_compute = self._get_deserialized_task_to_compute()
+        self.compute_task_def = self.task_to_compute.compute_task_def
         self.source_package_path = get_storage_source_file_path(
-            task_id=self.compute_task_def['task_id'],
-            subtask_id=self.compute_task_def['subtask_id'],
+            self.task_to_compute.subtask_id,
+            self.task_to_compute.task_id,
```
kbeker: Hardly recommend to use parameters to be sure that `subtask_id` is `subtask_id` because we had bug which depends on that `subtask_id` was there `task_id`

Issue #532, opened 2018-06-20
Don't enforce lower time limit in the result transfer use case


cameel:
The clocks of the provider, requestor and Concent might not be perfectly synchronized. Enforcing perfect synchronization is simply not realistic. That's for example why golem-messages, when checking if the timestamp is correct, has such a high tolerance: 10 minutes in the past and 5 minutes in the future (see  [golem_messages/settings.py](https://github.com/golemfactory/golem-messages/blob/v2.10.0/golem_messages/settings.py) and [golem_messages/message/base.py#L38-L49](https://github.com/golemfactory/golem-messages/blob/v2.10.0/golem_messages/message/base.py#L38-L49)).

This means that for example when requestor is creating `ForceGetTaskResult` at 12:00, the `ReportComputedTask` may have a timestamp indicating 12:05.

This problem was raised by Łukasz on Rocket when discussing #519. It has no easy fix if we want to treat the timestamps as exact. The "solution" we use in other cases to simply never treat them as such on the client side. Golem client has some tolerances on message times. For example to be sure that it does not miss a deadline, it can't just send a message just a second before it. On another client or in Concent it may already be past the deadline at that point. To avoid this the client leaves enough room to be sure that even the maximum error coming from clocks being out of sync can't make it miss the deadline.

But in this case that's a bit inconvenient. We could take this approach but that would mean that sometimes the client would have to wait up to 5 mintues before sending FGTR. Fortunately there's a  workaround for this particular situation: we don't really have to check the timestamp on RCT. The only purpose of the lower limit is to make sure that the requestor does not try to start the use case before the provider has finished the computation. If the requestor has already received the RCT (he's including it in his FGTR) there's no reason to prevent him from starting the use case even if the timestamp on RCT is in the future.

The subject of this task is to remove the lower limit from
``` python
ReportComputedTask.timestamp < ForceGetTaskResult.timestamp < TaskToCompute.deadline + MDT + 2 CMT
```
check that we have in Concent Core. Replace it with just
``` python
ForceGetTaskResult.timestamp < TaskToCompute.deadline + MDT + 2 CMT
```

cameel:
Description reworded and extended to make it clearer.

Issue #519, opened 2018-06-18
Concent responds with `ForceGetTaskResultRejected` to a valid `ForceGetTaskResult`


shadeofblue:
When sending a `ForceGetTaskResultRejected` message to the Concent, Golem Client receives:

```
ForceGetTaskResultRejected(header=MessageHeader(type_=4012, timestamp=1529330495, encrypted=True), sig=b'V\x8c\x1d\xc9\x
f4\xcd}\xb4\xa9\xfe\x1c\xc0\x9d\xb3\xccq\x8f\x9f\x15\x05\xdb@\xc0jyD\x9e\x03\x8b\x19`Z$Z\xef\xe6\xa1\x08\xe4\x05\x87\x8a\x05\xce\x9e\x1f\xf88Q\xc4n*\x84\x1b\xe5\xef\x05\xc3\t\xb5L\xb7\xa9?\x01', slots=[['force_g
et_task_result', None], ['reason', 'acceptance_time_limit_exceeded']])
```

The `ForceGetTaskResult` message:

```
ForceGetTaskResult(header=MessageHeader(type_=4009, timestamp=1529330495, encrypted=False), sig=None, slots=
[['report_computed_task', ReportComputedTask(header=MessageHeader(type_=2004, timestamp=1529330495, encrypted=False), sig=b'!\xaf\xf8\xcfn\xa4\xf9\xab\n1\x1b<\x83\xfe\xf0=OO\xfc\xd3\x87}\x11#8\x99#^>\xe6\x11Ggl\
x04o\xe4\x8e\xb1\x00\xbf\x7fnT\x12D\x17\x99w\n\xe5\x1e\xfe$4k\xb9\xfc\xe8\xdcs#\x13\x88\x00', slots=[['result_type', 1], ['node_name', ''], ['address', '10.30.8.141'], ['node_info', {'hyperdrive_prv_port': 3282,
 'hyperdrive_pub_port': 3282, 'key': '627f6a8fdd03bb97dd901c97cd3fb6ff797b17dff8164e0b21a4f1b0e238d97bb120eed3eeffa82b38ca559935be0787172222535a472e1c362706128ae527c6', 'nat_type': 'Symmetric NAT', 'node_name': 
[...]

PR #533, opened 2018-06-21
Feature allow force get task resoult with current timestamp

pawelkisielewicz:
Resolves https://github.com/golemfactory/concent/issues/532 
Resolves https://github.com/golemfactory/concent/issues/538

### DIFF
--- a/concent_api/core/message_handlers.py
+++ b/concent_api/core/message_handlers.py
@@ -402,7 +402,7 @@ def handle_send_force_get_task_result(client_message: message.concents.ForceGetT
         maximum_download_time
     )
 
-    if not client_message.report_computed_task.timestamp < get_current_utc_timestamp() <= force_get_task_result_deadline:
+    if not get_current_utc_timestamp() <= force_get_task_result_deadline:
         logging.log_timeout(
             logger,
             client_message,
@@ -1288,7 +1288,7 @@ def handle_send_subtask_results_verify(
 
     verification_deadline = subtask_results_rejected.timestamp + settings.ADDITIONAL_VERIFICATION_CALL_TIME
 
-    if not subtask_results_rejected.timestamp < get_current_utc_timestamp() <= verification_deadline:
+    if not get_current_utc_timestamp() <= verification_deadline:
         return message.concents.ServiceRefused(
             reason=message.concents.ServiceRefused.REASON.InvalidRequest,
         )
--- a/concent_api/core/tests/test_integration_get_task_result.py
+++ b/concent_api/core/tests/test_integration_get_task_result.py
@@ -81,10 +81,10 @@ class GetTaskResultIntegrationTest(ConcentIntegrationTestCase):
         self.assertEqual(message_from_concent.reason,       message_from_concent.REASON.AcceptanceTimeLimitExceeded)
         self._assert_stored_message_counter_not_increased()
 
-    def test_that_requestor_forces_get_task_result_and_concent_should_immediately_sends_rejection_because_message_is_sent_too_early(self):
+    def test_that_requestor_forces_get_task_result_and_concent_should_immediately_sends_rejection_because_message_is_sent_too_late(self):
         """
         Tests if on requestor ForceGetTaskResult message Concent will return ForceGetTaskResultRejected
-        if message is sent to early.
+        if message is sent to late - after the deadline.
 
         Expected message exchange:
         Requestor -> Concent:    ForceGetTaskResult
@@ -106,7 +106,7 @@ class GetTaskResultIntegrationTest(ConcentIntegrationTestCase):
             timestamp="2017-12-01 11:00:00",
         )
 
-        with freeze_time("2017-12-01 10:59:59"):
+        with freeze_time("2017-12-01 11:05:00"):
             response = self.client.post(
                 reverse('core:send'),
                 data=serialized_force_get_task_result,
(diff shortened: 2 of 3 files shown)


rwrzesien:
@pawelkisielewicz Please write in description which issue this pull request resolves. It will then get closed automatically when this pull request is merged.

concent_api/core/tests/test_integration_get_task_result.py
```diff
@@ -124,6 +124,46 @@ def test_that_requestor_forces_get_task_result_and_concent_should_immediately_se
         )
         self._assert_stored_message_counter_not_increased()
 
+    def test_that_requestor_forces_get_task_result_and_concent_should_answer_AckForceGetTaskResult_if_all_timestamps_are_equal(self):
```
rwrzesien: Test name must be lowercase.

pawelkisielewicz: done

concent_api/core/message_handlers.py
```diff
@@ -400,7 +400,7 @@ def handle_send_force_get_task_result(client_message: message.concents.ForceGetT
         maximum_download_time
     )
 
-    if not client_message.report_computed_task.timestamp < get_current_utc_timestamp() <= force_get_task_result_deadline:
+    if not client_message.report_computed_task.timestamp <= get_current_utc_timestamp() <= force_get_task_result_deadline:
```
dybi: If I understood: https://github.com/golemfactory/concent/issues/532 correctly, `client_message.report_computed_task.timestamp < ` part should be deleted 

@cameel ?

concent_api/core/tests/utils.py
```diff
@@ -134,7 +135,7 @@ def _get_deserialized_report_computed_task(
         task_to_compute = None,
         size: int = 1,
         package_hash: str = 'sha1:4452d71687b6bc2c9389c3349fdc17fbd73b833b',
-        timestamp: int = None,
+        timestamp: Optional[str] = None,
```
dybi: :+1: for finding this

concent_api/core/tests/utils.py
```diff
@@ -152,7 +153,7 @@ def _get_deserialized_report_computed_task(
 
     def _get_deserialized_task_to_compute(
         self,
-        timestamp: int = None,
+        timestamp: Optional[str] = None,
```
dybi: :+1: for finding this

concent_api/core/message_handlers.py
cameel: Yes. Exactly. The task is about removing the check not about making the interval closed on both sides.

rwrzesien: In the way which it is described I have understood it in the way which @pawelkisielewicz did it, so maybe issue should be rewritten and add more info on technical level.

cameel: > Is it possible that `client_message.report_computed_task.timestamp` would be greater then `get_current_utc_timestamp`?

It is and that's where this change comes from. Remember that the clocks of the provider, requestor and Concent might not be perfectly synchronized. Enforcing perfect synchronization is simply not realistic. That's for example why golem-messages, when checking if the timestamp is correct, has such a high tolerance: 10 minutes in the past and 5 minutes in the future: 

See
https://github.com/golemfactory/golem-messages/blob/v2.10.0/golem_messages/message/base.py#L38-L49
https://github.com/golemfactory/golem-messages/blob/v2.10.0/golem_messages/settings.py

``` python
MSG_TTL = datetime.timedelta(minutes=10)
FUTURE_TIME_TOLERANCE = datetime.timedelta(minutes=5)
```

So the timestamps received from another machine can be off by up to 10 minutes!

This means that for example when requestor is creating `ForceGetTaskResult` at 12:00, the `ReportComputedTask` may have a timestamp indicating 12:05.

There was a discussion between me, Przemek and Łukasz on Rocket and in #519 (please read them if you want to know more about why we're making this change) and this problem was raised. It has no easy fix if we want to treat the timestamps as exact. The "solution" is to simply never treat them as such on the client side. Golem client has some tolerances on message times. For example to be sure that it does not miss a deadline, it can't just send a message just a second before it. On another client or in Concent it may already be past the deadline at that point. To avoid this the client leaves enough room to be sure that even the maximum error coming from clocks being out of sync can't make it miss the deadline.

But in this case that's a bit inconvenient. We could take this approach but that would mean that sometimes the client would have to wait up to 5 mintues before sending FGTR. Fortunately there's a  workaround for this particular situation: we don't really have to check the timestamp on RCT. The only purpose of the lower limit is to make sure that the requestor does not try to start the use case before the provider has finished the computation. If the requestor has already received the RCT (he's including it in his FGTR) there's no reason to prevent him from starting the use case even if the timestamp on RCT is in the future.

cameel: > In the way which it is described I have understood it in the way which @pawelkisielewicz did it, so maybe issue should be rewritten and add more info on technical level.

I have rewritten the description to include the information from the comment above.

concent_api/core/message_handlers.py
```diff
@@ -1286,7 +1286,7 @@ def handle_send_subtask_results_verify(
 
     verification_deadline = subtask_results_rejected.timestamp + settings.ADDITIONAL_VERIFICATION_CALL_TIME
 
-    if not subtask_results_rejected.timestamp < get_current_utc_timestamp() <= verification_deadline:
+    if not get_current_utc_timestamp() <= verification_deadline:
```
dybi: is there an issue for this?

concent_api/core/tests/test_integration_get_task_result.py
```diff
@@ -81,10 +81,10 @@ def test_requestor_forces_get_task_result_and_concent_immediately_sends_rejectio
         self.assertEqual(message_from_concent.reason,       message_from_concent.REASON.AcceptanceTimeLimitExceeded)
         self._assert_stored_message_counter_not_increased()
 
-    def test_that_requestor_forces_get_task_result_and_concent_should_immediately_sends_rejection_because_message_is_sent_too_early(self):
+    def test_that_requestor_forces_get_task_result_and_concent_should_immediately_sends_rejection_because_message_is_sent_too_late(self):
         """
         Tests if on requestor ForceGetTaskResult message Concent will return ForceGetTaskResultRejected
-        if message is sent to early.
+        if message is sent to late - after a deadline to do it.
```
dybi: "after the deadline"

concent_api/core/message_handlers.py
rwrzesien: @pawelkisielewicz I think you are mixing two issues here, if yes, add second one to the description.

rwrzesien: @pawelkisielewicz Github will not catch it this way. You need to write
```
Resolves #532
Resolves #538
```

Issue #535, opened 2018-06-21
[VERIFICATION]  Resolution of images to compare should be checked


dybi:
If output image has different resulotion than source image, verification should report `Validation.MISMATCH` immediately (additional check is needed)

PR #541, opened 2018-06-22
Add function to verify sizes of images

pawelkisielewicz:
resolves https://github.com/golemfactory/concent/issues/535

There is only a size of images checked (in pixels). It is also possible to check the number of channels if image is in color. If one image would be in color and second in grayscale it will be easy to check. I'm only not  sure if we want it.

### DIFF
--- a/concent_api/verifier/tasks.py
+++ b/concent_api/verifier/tasks.py
@@ -24,6 +24,7 @@ from common.decorators import provides_concent_feature
 from common.logging import log_string_message
 from common.helpers import upload_file_to_storage_cluster
 from .exceptions import VerificationError
+from .utils import are_image_sizes_and_color_channels_equal
 from .utils import clean_directory
 from .utils import delete_file
 from .utils import generate_blender_output_file_name
@@ -300,6 +301,18 @@ def blender_verification_order(
             VerificationResult.ERROR.name,
             str(exception),
             ErrorCode.VERIFIER_LOADING_FILES_INTO_MEMORY_FAILED.name
+        )
+        return
+
+    if not are_image_sizes_and_color_channels_equal(image_1, image_2):
+        log_string_message(
+            logger,
+            f'Blender verification failed. Sizes in pixels of images are not equal. SUBTASK_ID: {subtask_id}.'
+            f'VerificationResult: {VerificationResult.MISMATCH.name}'
+        )
+        verification_result.delay(
+            subtask_id,
+            VerificationResult.MISMATCH.name,
         )
         return
 
--- a/concent_api/verifier/tests/test_integration_verification.py
+++ b/concent_api/verifier/tests/test_integration_verification.py
@@ -15,7 +15,6 @@ from common.helpers import get_storage_source_file_path
 from common.testing_helpers import generate_ecc_key_pair
 from ..tasks import blender_verification_order
 
-
 (CONCENT_PRIVATE_KEY, CONCENT_PUBLIC_KEY) = generate_ecc_key_pair()
 
 
@@ -83,6 +82,7 @@ class VerifierVerificationIntegrationTest(ConcentIntegrationTestCase):
             mock.patch('verifier.tasks.get_files_list_from_archive', return_value=['file_name']) as mock_get_files_list_from_archive, \
             mock.patch('verifier.tasks.cv2.imread', autospec=True) as mock_imread, \
             mock.patch('verifier.tasks.compare_ssim', return_value=1.0) as mock_compare_ssim, \
+            mock.patch('verifier.tasks.are_image_sizes_and_color_channels_equal', return_value=True), \
             mock.patch('verifier.tasks.delete_file', autospec=True) as mock_delete_file:  # noqa: E125
             blender_verification_order(
                 subtask_id=self.compute_task_def['subtask_id'],
@@ -124,6 +124,7 @@ class VerifierVerificationIntegrationTest(ConcentIntegrationTestCase):
             mock.patch('verifier.tasks.get_files_list_from_archive', autospec=True, return_value=['file_name']) as mock_get_files_list_from_archive, \
             mock.patch('verifier.tasks.cv2.imread', autospec=True) as mock_imread, \
             mock.patch('verifier.tasks.compare_ssim', return_value=(settings.VERIFIER_MIN_SSIM - 0.1)) as mock_compare_ssim, \
+            mock.patch('verifier.tasks.are_image_sizes_and_color_channels_equal', return_value=True), \
             mock.patch('verifier.tasks.delete_file', autospec=True) as mock_delete_file:  # noqa: E125
             blender_verification_order(
                 subtask_id=self.compute_task_def['subtask_id'],
@@ -335,6 +336,7 @@ class VerifierVerificationIntegrationTest(ConcentIntegrationTestCase):
             mock.patch('verifier.tasks.get_files_list_from_archive', return_value=['file_name']) as mock_get_files_list_from_archive, \
             mock.patch('verifier.tasks.cv2.imread', autospec=True) as mock_imread, \
             mock.patch('verifier.tasks.compare_ssim', return_value=1.0) as mock_compare_ssim, \
+            mock.patch('verifier.tasks.are_image_sizes_and_color_channels_equal', return_value=True), \
             mock.patch('verifier.tasks.delete_file') as mock_delete_file:  # noqa: E125
             blender_verification_order(
                 subtask_id=self.compute_task_def['subtask_id'],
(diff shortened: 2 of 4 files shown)


concent_api/verifier/tasks.py
```diff
@@ -40,6 +41,12 @@
 logger = logging.getLogger(__name__)
 
 
+def check_if_sizes_in_pixels_are_equal(image1: ndarray, image2: ndarray) -> bool:
+    shape1 = image1.shape
+    shape2 = image2.shape
+    return True if shape1[0] == shape2[0] and shape1[1] == shape2[1] else False
```
dybi: this could be simplified to: `return image1.shape == image2.shape`

concent_api/verifier/tasks.py
```diff
@@ -40,6 +41,12 @@
 logger = logging.getLogger(__name__)
 
 
+def check_if_sizes_in_pixels_are_equal(image1: ndarray, image2: ndarray) -> bool:
```
dybi: the name of the function does not follow convetion of `bool` returning functions. Please rename it to sth like: `are_pixel_sizes_equal` or `are_image_sizes_equal` or `are_images_resolutioins_equal`

concent_api/verifier/tasks.py
```diff
@@ -293,6 +300,18 @@ def blender_verification_order(
         )
         return
 
+    if not check_if_sizes_in_pixels_are_equal(image_1, image_2):
+        log_string_message(
```
dybi: I don't like the big blob of code here. I think I would prefer having some validation function that raises exception.

concent_api/verifier/tests/test_integration_verification.py
```diff
@@ -478,3 +484,19 @@ def test_blender_verification_order_should_call_verification_result_with_result_
             'error',
             ErrorCode.VERIFIER_COMPUTING_SSIM_FAILED.name,
         )
+
+    def test_that_blender_verification_order_should_return_false_if_sizes_in_pixels_are_not_equal(self):
```
dybi: wrong test name

concent_api/verifier/tests/test_integration_verification.py
```diff
@@ -478,3 +484,19 @@ def test_blender_verification_order_should_call_verification_result_with_result_
...
+        image1 = mock.create_autospec(spec=ndarray, spec_set=True)
+        image2 = mock.create_autospec(spec=ndarray, spec_set=True)
+        image1.shape = (2000, 3000, 3)
+        image2.shape = (3000, 4000, 3)
+        result = check_if_sizes_in_pixels_are_equal(image1, image2)
+        self.assertEqual(result, False)
+
+    def test_that_blender_verification_order_should_return_true_if_sizes_in_pixels_are_equal(self):
```
dybi: wrong test name

dybi:
@cameel , should we check for color as well?

concent_api/verifier/tasks.py
pawelkisielewicz: shape is a tuple with image:
- `width`, `height` and `channels` and has 3 variables if image is in color 
or 
- `width`, `height`and has 2 variables if image is grayscale.

Now i only compare shapes. If color will be as well (i think it should be) it will be replaced with your suggestion.

concent_api/verifier/tasks.py
pawelkisielewicz: done

concent_api/verifier/tasks.py
pawelkisielewicz: If validation 'fails' in this function, the `verification_result` is returning with `VerificationResult.MISMATCH`. Look at the whole block (name already changed ):
```
       if not are_image_sizes_equal(image_1, image_2):
        log_string_message(
            logger,
            f'Blender verification failed. Sizes in pixels of images are not equal. SUBTASK_ID: {subtask_id}.'
            f'VerificationResult: {VerificationResult.MISMATCH.name}'
        )
        verification_result.delay(
            subtask_id,
            VerificationResult.MISMATCH.name,
        )
        return
```
I agree and don't like it as well, but I think setting of `verification_result` should be here. Logging could be moved to `are_image_sizes_equal` and changed to e.g. `are_image_sizes_equal_and_log_if_mismatch`... What do you think?

concent_api/verifier/tasks.py
rwrzesien: Move this function to `verifier/utils.py`.

concent_api/verifier/tests/test_integration_verification.py
rwrzesien: Please write separate test case class if you are testing only this function. Or even better - a separate test file, because you will be really testing `utils.py`.

concent_api/verifier/tasks.py
rwrzesien: I agree that it is written in current style of this task, so I would prefer to keep it as it is, and make flow changes like in refactor task. It might be added there as a comment.

concent_api/verifier/tasks.py
dybi: if color is not to be taken into account, it could be still simplified:
`return shape1[0] == shape2[0] and shape1[1] == shape2[1]`

cameel:
> There is only a size of images checked (in pixels). It is also possible to check the number of channels if image is in color. If one image would be in color and second in grayscale it will be easy to check. I'm only not sure if we want it.

We must check everything that's not checked by SSIM. Basically, if the mismatch results in an exception and crashes the verifier, you should check it before running SSIM.

concent_api/verifier/tasks.py
cameel: > Logging could be moved to are_image_sizes_equal and changed to e.g. are_image_sizes_equal_and_log_if_mismatch... What do you think?

No, definitely not. This looks like a generic utility funtion. It does not and should not know what its result is used for. I.e. that different sizes mean a failure of verification. That's the responsibility of the caller.

cameel: BTW: `Sizes in pixels of images are not equal` would sound less awkward phrased as `Image resolutions differ`.

concent_api/verifier/tasks.py
cameel: Yeah. If anyone ever wonders why we have a rule against function names starting with `check_` - that's why ;-)

concent_api/verifier/tasks.py
pawelkisielewicz: According to @cameel 's comment, color will be check as well. Changed with @dybi suggestion

concent_api/verifier/tasks.py
pawelkisielewicz: yes... done- renamed and moved :)

concent_api/verifier/tasks.py
pawelkisielewicz: I agree, but `resolution` is something different as `size in pixels`. Resolution says, how many `pixels` are in one `inch`. We could have two different images with sizes e.g. 40x50 cm and 30x60 cm, both in resolution e.g. `300 dpi` (dpi - pixels in inch), but sizes in pixels are different.
On the other hand, there can be two images - 30 x 30 cm in 300 dpi and 60 x 60 cm in 150 dpi and they are actually equal.
This function compares tuples with number of pixels. Now it will compare the whole `ndarray.shape` including number of color channels. 
I renamed the name to `are_image_sizes_and_color_channels_equal`, maybe the name is little bit awkward, but it is also very precise. If you want to rename it please let mi know.

concent_api/verifier/tests/test_integration_verification.py
pawelkisielewicz: done

concent_api/verifier/tests/test_integration_verification.py
pawelkisielewicz: done

concent_api/verifier/tasks.py
cameel: > I renamed the name to `are_image_sizes_and_color_channels_equal`, maybe the name is little bit awkward

It's fine. 

> I agree, but `resolution` is something different as `size in pixels`. Resolution says, how many `pixels` are in one `inch`. We could have two different images with sizes e.g. 40x50 cm and 30x60 cm, both in resolution e.g. `300 dpi` (dpi - pixels in inch), but sizes in pixels are different.
On the other hand, there can be two images - 30 x 30 cm in 300 dpi and 60 x 60 cm in 150 dpi and they are actually equal.

Yeah, you're right. Referring to image size as resolution is technically incorrect and pretty informal. "size in pixels" is more precise. But still the message is a bit awkward as is. Maybe `Image sizes in pixels differ.`?

PR #543, opened 2018-06-22
Chore refactor project structure

dybi:
Change some modules naming, move exceptions, etc.


cameel:
I agree with @rwrzesien that you should include a more detailed description. Right now we only see what you're changing but it might be unclear where those changes are heading. We need the see the bigger picture. Please create an issue that describes what you want to do. This also gives @rwrzesien a chance to discuss changes beforehand and makes it less likely that this pull request will be stuck because of endless discussions.

Having said that, I like the fact that finally someone has put some significant effort into refactoring the current structure and I really want it to go through. You have my full support even if it turns out that there are some specific changes that I'm not too fond of. It's fine as long as you have a consistent vision of what you want to achieve and it improves the overall quality of the code.

@rwrzesien Let's make sure this change gets reviewed and merged smoothly unless there are really serious showstopper bugs. I think it might be a good idea to discuss this pull request with @dybi first and hash out the differences. Also remember that you're free to submit issues/pull requests for your own refactoring ideas. But please make sure that you're both going in the same direction :)

rwrzesien:
> After looking through the code I see that you're just moving files around and updating imports.

Exactly, this is just moving code aroung in the directories. Or actually renaming `utils` to `common`. What is the actuall added value of it ? I don't see any. I wouldn't even call this refactoring. What I consider a refactoring is something like:
- you have input code and you create output code which have significly less lines, but you keep same functionality and readability,
- you have a piece of code which does something in not optimized way, so you optimize it,
- you change approach to a given problem by redesigning implementation of solution,
- you modify implementation of class diagram.
etc. Not copy-pasting code from one place to another.

I also think this is not best moment for this type of changes, as we have a lot of more important tasks and deadlines to reach, time spent on this can be spent on more important stuff. Rest of team is hard working on current tasks and every pair of hands is appreciated to help reach our common goal. I think we can think about some real refactoring after that is done.

> Having said that, I like the fact that finally someone has put some significant effort into refactoring the current structure and I really want it to go through.

I agree but I would like to see a significant changes first, I don't consider those to be so.

> Also remember that you're free to submit issues/pull requests for your own refactoring ideas. But please make sure that you're both going in the same direction :)

I think we should move toward more object oriented code, and use more django features. I have a vision of this in my head but it is difficult to find time to write it down while tackling real tasks.

dybi:
First of all, this PR was small enough not to elaborate what is going on in it. The description is pretty accurate - it just:
> Change some modules naming, move exceptions, etc.

This change is small according to our guidelines - to make work in batches, so it can be faster/more easily integreted with the rest of it (merged to `master`). It is a part, or better pre-work of refactoring the verification phase.

> What is the actuall added value of it ? I don't see any. I wouldn't even call this refactoring.

Improving code structure **is** refactoring.

> you have a piece of code which does something in not optimized way, so you optimize it

it is **not** refactoring, it **is** optimization. 


> I have a vision of this in my head but it is difficult to find time to write it down while tackling real tasks.

Refactoring **is** a real task.

>  Rest of team is hard working on current tasks and every pair of hands is appreciated to help reach our common goal. I think we can think about some real refactoring after that is done.

I do not like the accusation that I am not working hard to achieve our common goal. In my understanding, we need to refactor verification before it is delivered, beacause the way it is implemented right now is very error prone and hardo to understand/debug. I have been testing in for the last few days so my statement is gorunded in my experience with it.
And this tiny PR was supposed to be a part or, I you prefer, pre-work of refactoring verification.

dybi:
@rwrzesien oh, BTW: I have just realized, that my first big refactor here: https://github.com/golemfactory/concent/pull/212
was just movin files... Do you think it was unnecessary? Or that it wasn't "real work"?

cameel:
@dybi @rwrzesien Let's stop the discussion here for now. We'll do a hangout after the standup and talk about this issue together with @adrianzamorski.

PR #555, opened 2018-06-25
Bugfix: Change time settings in verification api-test

kbeker:
Small fix to api-test for verification

### DIFF
--- a/concent_api/api-integration-additional-verification-test.py
+++ b/concent_api/api-integration-additional-verification-test.py
@@ -151,7 +151,8 @@ def test_case_1_test_for_positive_case(cluster_consts, cluster_url, test_id):
         source_file_size_2
     ))
 
-    time.sleep(15)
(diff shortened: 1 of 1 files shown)


concent_api/api-integration-additional-verification-test.py
```diff
@@ -151,7 +151,7 @@ def test_case_1_test_for_positive_case(cluster_consts, cluster_url, test_id):
         source_file_size_2
     ))
 
-    time.sleep(10)
+    time.sleep(15)
```
pawelkisielewicz: Ok, but consider instead of hard coded 15 import as variable from settings. If settings will change i don't think we will remember this line :)

cameel: Yeah. We should be using times based on settings in situations like this. There was even a task for refactoring all the code to do this so let's not go back to the state from before.

Issue #550, opened 2018-06-25
Concent tries to use OpenCV outside Verifier - even in `manage.py migrate` or `manage.py collectstatic` commands


bartoszbetka:
Now, when we building container with concent or try migrate database, we need `cv2` lib which require `libSM` system lib. This lib should be require only when  `CONCENT_FEATURES` contains `verifier` app.

``` 
 File "./manage.py", line 10, in <module>
   execute_from_command_line(sys.argv)
 File "/home/builder/build/concent-dev/concent-deployment/containers/build/virtualenv/lib/python3.6/site-packages/django/core/management/__init__.py", line 364, in execute_from_command_line
   utility.execute()
 File "/home/builder/build/concent-dev/concent-deployment/containers/build/virtualenv/lib/python3.6/site-packages/django/core/management/__init__.py", line 338, in execute
   django.setup()
 File "/home/builder/build/concent-dev/concent-deployment/containers/build/virtualenv/lib/python3.6/site-packages/django/__init__.py", line 27, in setup
   apps.populate(settings.INSTALLED_APPS)
 File "/home/builder/build/concent-dev/concent-deployment/containers/build/virtualenv/lib/python3.6/site-packages/django/apps/registry.py", line 116, in populate
   app_config.ready()
 File "/home/builder/build/concent-dev/concent-deployment/containers/build/concent-api-assets/concent_api/concent_api/apps.py", line 8, in ready
   from concent_api import system_check  # noqa, flake8 F401 issue  # pylint: disable=unused-variable
 File "/home/builder/build/concent-dev/concent-deployment/containers/build/concent-api-assets/concent_api/concent_api/system_check.py", line 13, in <module>
   from concent_api.constants  import AVAILABLE_CONCENT_FEATURES
 File "/home/builder/build/concent-dev/concent-deployment/containers/build/concent-api-assets/concent_api/concent_api/constants.py", line 6, in <module>
   import core.urls
 File "/home/builder/build/concent-dev/concent-deployment/containers/build/concent-api-assets/concent_api/core/urls.py", line 3, in <module>
   from .views import send, receive, receive_out_of_band, protocol_constants
 File "/home/builder/build/concent-dev/concent-deployment/containers/build/concent-api-assets/concent_api/core/views.py", line 8, in <module>
   from core.message_handlers          import handle_message
 File "/home/builder/build/concent-dev/concent-deployment/containers/build/concent-api-assets/concent_api/core/message_handlers.py", line 32, in <module>
[...]

PR #558, opened 2018-06-25
Fix concent tries to use opencv outside verifier

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/550

### DIFF
--- a/concent_api/verifier/tasks.py
+++ b/concent_api/verifier/tasks.py
@@ -8,7 +8,6 @@ from celery import shared_task
 from golem_messages import message
 from requests import HTTPError
 from skimage.measure import compare_ssim
-import cv2
 
 from django.conf import settings
 
@@ -39,6 +38,11 @@ from .utils import unpack_archive
 
 crash_logger = logging.getLogger('concent.crash')
 logger = logging.getLogger(__name__)
+
+
+def import_cv2():
+    import cv2
+    return cv2
 
 
 @shared_task
@@ -287,6 +291,12 @@ def blender_verification_order(
         result_files_list = get_files_list_from_archive(
             generate_verifier_storage_file_path(result_package_path)
         )
+
+        # OpenCV is imported here because it requires OpenCV package installed in the system to work,
+        # which is not installed on Concent instances without verifier.
+        # However, this is not imported directly here because we need to mock this in tests, which might be run on
+        # the environment without OpenCV package installed in the system.
+        cv2 = import_cv2()
 
         image_1 = cv2.imread(  # pylint: disable=no-member
             generate_verifier_storage_file_path(blender_output_file_name)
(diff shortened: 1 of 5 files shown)


concent_api/verifier/tasks.py
```diff
@@ -286,6 +290,10 @@ def blender_verification_order(
             generate_verifier_storage_file_path(result_package_path)
         )
 
+        # OpenCV is imported here because it requires OpenCV package installed in the system to work,
+        # which is not installed on Concent instances without verifier.
+        cv2 = import_cv2()
```
dybi: why not just `import cv2` here (local import)?

dybi: ok, I know why - for patching the tests

cameel: @kbeker: If @dybi has to ask - it's a good sign that you should add a comment with an explanation here.

You do have a comment but it's actually @rwrzesien's comment moved from the top of the file without any change. It should be expanded.

cameel:
> I have just added VERIFIER_MIN_SSIM to development.py and testing.py and with @bartoszbetka we tested this on cluster. Everything works :+1: 

It's already in `base.py`. It should no be in `development.py` and `testing.py`. It only adds unnecessary duplication (unless you need a special value in development and tests but that's not the case here).

concent_api/verifier/tests/test_integration_verification.py
```diff
@@ -37,6 +37,29 @@ class CompletedProcessWithError:
     return CompletedProcessWithError()
 
 
+class Image:
```
dybi: Why adding this class? Wouldn't be better to return `ndarray` in `MockedCV2`? Or better, prepare some mock, instead of addding stub classes?

rwrzesien: Could you provide an example ? My goal here is that there are several tests in which using `MockedCV2.imread` should return different results.

concent_api/verifier/tasks.py

concent_api/verifier/tests/test_integration_verification.py
dybi: you could have sth like this utility function/method:
```python
    def _prepare_cv2_mock(self, desired_behaviour='none'):
        mocked_cv2 = mock.Mock()
        mocked_cv2.imread = mock.create_autospec(cv2.imread, spec_set=True)
        if desired_behaviour == 'none':
            mocked_cv2.imread.return_value = None
        elif desired_behaviour =='image':
            mocked_cv2.imread.return_value = ndarray((1024, 768, 3))
        elif desired_behaviour == 'exception':
            mocked_cv2.imread.side_effect = MemoryError('error')
        return mocked_cv2
```
and then use it like:
`mock.patch('verifier.tasks.import_cv2', side_effect=self._prepare_cv2_mock('value')) as mock_import_cv2`

thanks to that, function calls to `imread` are protected and whenever API of the function changes it will be immediately reflected in the tests.

Issue #548, opened 2018-06-22
ForcePaymentRejected.force_payment is empty


jiivan:
`ForcePaymentRejected.force_payment` should be filled by Concent.

PR #559, opened 2018-06-25
Fix force payment rejected force payment is empty

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/548


rwrzesien:
@kbeker Yes, it disappeared after rebase.

PR #561, opened 2018-06-26
Update dependencies to the lastest versions

### LINKED ISSUE #554: Update dependencies to the latest versions
To ensure that everyone is working with the same dependency versions, they're always hard-coded in [requirements.lock](https://github.com/golemfactory/concent/blob/v0.7.2/concent_api/requirements.lock). We haven't updated this file for quite a while though. We should update all packages to the latest versions.

- Create a clean virtualenv.
- Install packages listed in `requirements.txt`.
- Replace `requirements.lock` with the output of `pip freeze`.
- Test on the cluster with newer versions
    - If you find a package where a newer version causes a problem that can't be easily resolved, add a `<=` constraint for that version in `requirements.txt` and repeat the whole procedure. Add an issue for tracking and resolving the problem.

Also:
- Please pin celery version in `requirements.txt` at >= 4.0. Our code won't work with 3.x and earlier versions.
- Django should stay at 1.11.x because it's a LTS release.
- While at it, you can also update the minimal versions listed in `requirements-development.txt` (for those packages where they're present).

kbeker:
Resolves #554

Don't merge it till all tests pass on cluster


cameel:
This does not look like a simple update. There are new dependencies that are quite heavy: numpy, matplotlib, pillow. Where do they come from? Why weren't they there before?

Issue #329, opened 2018-04-27
Admin filter showing tasks that can still have active downloads


rwrzesien:
Extracted from https://github.com/golemfactory/concent/issues/152#issuecomment-384696044. This task depends on #660 and #661.

Add a new filter in the admin panel. It should show subtasks that are either in an active state or are in `RESULT UPLOADED` state while current time <= `TaskToCompute.deadline` + SVT. We'll leave the decision whether to wait for downloads or not up to the person shutting down the service.

To implement this you either need to resolve #141 first or add a `deadline` column to the `Subtask` table. If you do the latter, make sure that it's updated every time `task_to_compute` is updated (like RCT in #174).

cameel:
We can do this even without #141. Just add `deadline` column on the `Subtask` table (and don't forget to update the #139 blueprint).

cameel:
Also, let's split this into two issues. The first point is much more important than the second one.

cameel:
I have created a separate task for the second point (#335) and updated the description.

dybi:
@cameel, as discussed, another filed in `Subtask` model will be needed:
`result_package_size` from `ReportComputedTask`

cameel:
@dybi About a query that finds the tasks that still have active downloads.

I was thinking about something along these lines:
``` python
from django.db.models import F
from django.db.models import Func
from django.db.models import Manager
from django.db.models import Model
from django.db.models import Q


class SubtaskWithTimingColumnsManager(Manager):
    def get_queryset(self):
        bytes_per_sec   = settings.MINIMUM_UPLOAD_RATE << 10
        download_time   = Func(F('result_package_size') / float(bytes_per_sec), function='CEIL'),

        mdt             = F('maximum_download_time')
        ttc_dt          = F('task_to_compute__timestamp')
        subtask_dt      = F('computation_deadline')
        subtask_timeout = Func('second', subtask_dt - ttc_dt, function='DATE_PART')

        return super().get_queryset().annotate(
            maximum_download_time     = settings.DOWNLOAD_LEADIN_TIME + download_time,
            subtask_verification_time = (4 * settings.CONCENT_MESSAGING_TIME) + (3 * mdt) + (0.5 * subtask_timeout),
        )


class Subtask(Model):
    # Current field definitions
    ...
    ...
    objects_with_timing_columns = SubtaskWithTimingColumnsManager()
    ...


subtasks_with_active_downloads = Subtask.objects_with_timing_columns.filter(
[...]

PR #570, opened 2018-06-27
Feature admin filter showing tasks that can still have active downloads

MichalW-CP:
resolves https://github.com/golemfactory/concent/issues/329


concent_api/common/logging.py
```diff
@@ -147,13 +147,17 @@ def log_subtask_stored(
...
     provider_public_key: bytes,
     requestor_public_key: bytes,
+    computation_deadline: int,
     next_deadline: Optional[int] = None,
 ):
     logger.info(
         f"A subtask has been stored -- STATE: {state} -- "
         f"NEXT_DEADLINE: {next_deadline} -- "
+        f"RESULT_PACKAGE_SIZE: {result_package_size} -- "
+        f"COMPUTATION_DEADLINE: {computation_deadline} -- "
```
rwrzesien: I think it should be moved at the end of the log, as least important info.

concent_api/core/admin.py
```diff
@@ -1,20 +1,21 @@
...
+class ActivePassiveDownloadsStateFilter(admin.SimpleListFilter):
+    title = 'Active/passive/downloads state'
     parameter_name = 'active_passive_state'
 
     def lookups(self, request, model_admin):
         return (
-            ('active',  'Active'),
+            ('active', 'Active'),
             ('passive', 'Passive'),
+            ('active_downloads', 'With active Downloads')
```
rwrzesien: `downloads` should start with lowercase.

concent_api/core/admin.py
```diff
@@ -50,12 +71,17 @@ class SubtaskAdmin(ModelAdminReadOnlyMixin, admin.ModelAdmin):
 
     def get_provider_public_key(self, obj):  # pylint: disable=no-self-use
         return obj.provider.public_key
+
```
rwrzesien: There is no need for this extra lines, in Django admin convention it is ok. Is pylint complaining about it ?

concent_api/core/message_handlers.py
```diff
@@ -822,14 +822,20 @@ def store_subtask(
     provider  = Client.objects.get_or_create_full_clean(provider_public_key)
     requestor = Client.objects.get_or_create_full_clean(requestor_public_key)
 
+    computation_deadline = task_to_compute.compute_task_def['deadline']
```
rwrzesien: Blank lines above and below not needed. Beside that, if those are used only once they can be passed directly to logging function.

concent_api/core/message_handlers.py
```diff
@@ -822,14 +822,20 @@ def store_subtask(
...
-        task_to_compute=store_message(task_to_compute, task_id, subtask_id),
+        task_id             = task_id,
+        subtask_id          = subtask_id,
+        provider            = provider,
+        requestor           = requestor,
+        result_package_size = result_package_size,
+        state               = state.name,
+        next_deadline       = parse_timestamp_to_utc_datetime(next_deadline) if next_deadline is not None else None,
+        computation_deadline= parse_timestamp_to_utc_datetime(computation_deadline),
+        task_to_compute     =store_message(task_to_compute, task_id, subtask_id),
```
rwrzesien: Please fix styling.

concent_api/core/migrations/0003_add_and_update_deadline_and_size.py
```diff
@@ -0,0 +1,57 @@
...
+from django.db import models
+from common.helpers import deserialize_message
+from common.helpers import parse_timestamp_to_utc_datetime
+
+
+def populate_subtask_size_field(apps, _schema_editor):
+    Subtask = apps.get_model('core', 'Subtask')
+    StoredMessage = apps.get_model('core', 'StoredMessage')
+    for subtask in Subtask.objects.filter(result_package_size__isnull=True):
+        if StoredMessage.objects.filter(
```
rwrzesien: Please fix styling.

rwrzesien: You can achieve this in single query, using `first()`, then checking if returned value is not None.

concent_api/core/migrations/0003_add_and_update_deadline_and_size.py
```diff
@@ -0,0 +1,57 @@
...
+                StoredMessage.objects.get(pk=subtask.report_computed_task_id).data).size
+            subtask.full_clean()
+            subtask.save()
+
+
+def populate_deadline_size_field(apps, _schema_editor):
+    Subtask = apps.get_model('core', 'Subtask')
+    StoredMessage = apps.get_model('core', 'StoredMessage')
+    for subtask in Subtask.objects.filter(computation_deadline__isnull=True):
+        if StoredMessage.objects.filter(
```
rwrzesien: Same here.

concent_api/core/migrations/0003_add_and_update_deadline_and_size.py
```diff
@@ -0,0 +1,57 @@
...
+            name='computation_deadline',
+            field=models.DateTimeField(blank=True, null=True),
+        ),
+        migrations.AddField(
+            model_name='subtask',
+            name='result_package_size',
+            field=models.IntegerField(blank=True, null=True),
+        ),
+        migrations.RunPython(
+            populate_subtask_size_field,
```
rwrzesien: You should use `RunPython.noop` as the reverse function, otherwise migration can't be reversed.

concent_api/core/migrations/0004_auto_20180620_0712.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
```
rwrzesien: Migrations should be squashed if possible.

concent_api/core/models.py
```diff
@@ -1,17 +1,23 @@
 import base64
 import datetime
 
+from django.conf            import settings
+
```
rwrzesien: Please remove blank line.

concent_api/core/models.py
```diff
@@ -71,6 +96,8 @@ class Subtask(Model):
     """
     Represents subtask states.
     """
+    objects = Manager()
```
rwrzesien: This is by default, please remove.

MichalW-CP: This is not by default and without that Django can't use default manager

concent_api/core/admin.py
MichalW-CP: Done.

concent_api/core/templatetags/admin_tags.py
```diff
@@ -26,7 +26,7 @@ def get_active_subtasks_amount():
 
 @register.assignment_tag
 def get_active_subtasks_status():
-    return get_result_uploaded_subtasks().exists()
+    return get_result_uploaded_subtasks().exists()  # pylint: disable=no-member
```
kbeker: Did you try use newest pylint? Becuase some warnings were fixed in pylint and there is no more need to add disables in some cases

concent_api/core/models.py
```diff
@@ -288,6 +315,10 @@ class Meta:
 
     task_id                     = CharField(max_length = MESSAGE_TASK_ID_MAX_LENGTH)
 
+    computation_deadline = DateTimeField(blank = True, null = True)
```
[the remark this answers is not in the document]
MichalW-CP: Shouldn't they be like in other parts of code in this file - with spaces?

concent_api/core/models.py
```diff
@@ -28,6 +34,25 @@
...
+        task_to_compute_timestamp = Func(Value('epoch'), ttc_dt, function='DATE_PART')
+        subtask_dt = F('computation_deadline')
+        subtask_timeout = Func(Value('epoch'), subtask_dt, function='DATE_PART') - task_to_compute_timestamp
+        svt = (4 * settings.CONCENT_MESSAGING_TIME) + (3 * mdt) + (0.5 * subtask_timeout)
+        sum_of_both_fields = Func(Value('epoch'), subtask_dt, function='DATE_PART') + svt
+        return super().get_queryset().annotate(
+            maximum_download_time=Value(settings.DOWNLOAD_LEADIN_TIME) + download_time).annotate(
+                subtask_verification_time=ExpressionWrapper(svt, output_field=IntegerField()
+            )).annotate(
+                sum_of_both_fields=ExpressionWrapper(sum_of_both_fields, output_field=IntegerField()))
```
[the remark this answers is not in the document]
MichalW-CP: Done.

concent_api/core/models.py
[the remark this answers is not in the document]
MichalW-CP: Done.

concent_api/core/models.py
MichalW-CP: Done.

concent_api/core/migrations/0004_auto_20180620_0712.py
MichalW-CP: Done.

concent_api/core/migrations/0003_add_and_update_deadline_and_size.py
MichalW-CP: Done.

concent_api/core/migrations/0003_add_and_update_deadline_and_size.py
MichalW-CP: Done.

concent_api/core/message_handlers.py
MichalW-CP: Done.

concent_api/core/models.py
```diff
@@ -288,6 +315,10 @@ class Meta:
 
     task_id                     = CharField(max_length = MESSAGE_TASK_ID_MAX_LENGTH)
 
+    computation_deadline = DateTimeField(blank = True, null = True)
+
+    result_package_size = IntegerField(blank = True, null = True)
```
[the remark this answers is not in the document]
MichalW-CP: Done.

concent_api/core/message_handlers.py
MichalW-CP: Those are used twice.

concent_api/core/message_handlers.py
rwrzesien: Still there are some spaces on left side of `=`.

concent_api/core/migrations/0004_auto_20180620_0712.py
rwrzesien: There are still two migration files.

concent_api/core/models.py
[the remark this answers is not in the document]
rwrzesien: Still one `)` can be put into next line.

concent_api/core/models.py
```diff
@@ -28,6 +34,25 @@
 from .constants             import MESSAGE_TASK_ID_MAX_LENGTH
 
 
+class SubtaskWithTimingColumnsManager(Manager):
+    def get_queryset(self):
+        bytes_per_sec = settings.MINIMUM_UPLOAD_RATE << 10
+        download_time = Func(F('result_package_size') / float(bytes_per_sec), function='CEIL')
+
+        mdt = F('maximum_download_time')
```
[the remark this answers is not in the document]
MichalW-CP: Done.

concent_api/core/models.py
```diff
@@ -28,6 +34,25 @@
...
+        ttc_dt = F('task_to_compute__timestamp')
+        task_to_compute_timestamp = Func(Value('epoch'), ttc_dt, function='DATE_PART')
+        subtask_dt = F('computation_deadline')
+        subtask_timeout = Func(Value('epoch'), subtask_dt, function='DATE_PART') - task_to_compute_timestamp
+        svt = (4 * settings.CONCENT_MESSAGING_TIME) + (3 * mdt) + (0.5 * subtask_timeout)
+        sum_of_both_fields = Func(Value('epoch'), subtask_dt, function='DATE_PART') + svt
```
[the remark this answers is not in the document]
MichalW-CP: Done.

concent_api/core/models.py
rwrzesien: I will check when tests will be working.

concent_api/core/migrations/0004_auto_20180620_0712.py
MichalW-CP: There are two migrations in separate commits, because each commit have changes to models

concent_api/core/templatetags/admin_tags.py
MichalW-CP: Yes, but not working

concent_api/core/models.py
```diff
@@ -28,6 +34,25 @@
 from .constants             import MESSAGE_TASK_ID_MAX_LENGTH
 
 
+class SubtaskWithTimingColumnsManager(Manager):
+    def get_queryset(self):
+        bytes_per_sec = settings.MINIMUM_UPLOAD_RATE << 10
```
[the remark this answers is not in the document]
MichalW-CP: @cameel Do we need comments to this function? If we do, can you write some?

concent_api/core/message_handlers.py
```diff
@@ -1102,6 +1108,8 @@ def update_subtask(
 
     if set_next_deadline:
         subtask.next_deadline = next_deadline
+    if report_computed_task:
```
dybi: `if report_computed_task is not None`

concent_api/core/models.py
```diff
@@ -288,6 +314,10 @@ class Meta:
 
     task_id                     = CharField(max_length = MESSAGE_TASK_ID_MAX_LENGTH)
 
+    computation_deadline        = DateTimeField(blank = True, null = True)
```
dybi: should those new fields be nullable?

concent_api/core/admin.py
```diff
@@ -1,20 +1,21 @@
...
+class ActivePassiveDownloadsStateFilter(admin.SimpleListFilter):
+    title = 'Active/passive/downloads state'
     parameter_name = 'active_passive_state'
 
     def lookups(self, request, model_admin):
         return (
-            ('active',  'Active'),
+            ('active', 'Active'),
             ('passive', 'Passive'),
+            ('active_downloads', 'With active downloads')
```
dybi: Do we want it as separate option or we would like to have `active` + `active_downloads`

concent_api/core/migrations/0003_add_and_update_deadline_and_size.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def populate_subtask_size_field(apps, _schema_editor):
+    Subtask = apps.get_model('core', 'Subtask')
+    StoredMessage = apps.get_model('core', 'StoredMessage')
+    for subtask in Subtask.objects.filter(result_package_size__isnull=True):
+        stored_message = StoredMessage.objects.filter(
+            pk=subtask.report_computed_task_id
+        ).first()
+        if stored_message is not None:
```
dybi: Shouldn't this migration fail, if `stored_message` is `None`?

concent_api/core/models.py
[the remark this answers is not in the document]
cameel: Well, indeed, without alignment this looks like a mess. I really hate that our current coding style does not let us make it readable. The original code was quite readable: https://github.com/golemfactory/concent/issues/329#issuecomment-391834635

But this still has a lot of readability problems:
- First, I'd make an exception here and align this part and just add a comment explaining that without alignment this code is unreadable.
- This code is based on `subtask_verification_time()` and `maximum_download_time()` (I even kept the same variable names in my example to make it apparent). A comment saying that is definitely needed. 
    - BTW: you have created tests checking that the results of this query match those functions, right?
- Long mathematical expressions might be more readable broken into multiple lines.
- `time_for_download_file` sounds pretty awkward. This should be called something like `download_deadline`.
    - Actually this variable is not necessary. You know that adding it to `annotate()` creates a new column, right? We don't want any extra columns besides the two that were in the original example.
- Your nesting is off. It's really hard to tell what is nested in what. Is the second `annotate()` nested in the first? I can't tell without counting parentheses. It should be wrapped like this:
    ``` python
    return (super().get_queryset().
        annotate(
            maximum_download_time=Value(settings.DOWNLOAD_LEADIN_TIME) + download_time
        ).
        annotate(
            subtask_verification_time=ExpressionWrapper(subtask_verification_time, output_field=IntegerField())
        ).
        annotate(
            time_for_download_file=ExpressionWrapper(time_for_download_file, output_field=IntegerField())
        )
    )
    ```
    But really, `annotate()` can take multiple parameters. Why don't you do it like in my example?
    ``` python
    return super().get_queryset(). annotate(
        maximum_download_time=Value(settings.DOWNLOAD_LEADIN_TIME) + download_time,
        subtask_verification_time=ExpressionWrapper(subtask_verification_time, output_field=IntegerField()),
        time_for_download_file=ExpressionWrapper(time_for_download_file, output_field=IntegerField()),
    )
    ```
- It may be a good idea to put it in a separate function - that would give it a name. Or maybe even two functions.
    - You could have `add_subtask_verification_time_column()` and `add_maximum_download_time_column()`. The second one would assume that the first one has already been called so that the `subtask_verification_time` column already exists. Then `get_queryset()` would return simply `super().get_queryset().add_subtask_verification_time_column().add_maximum_download_time_column()`
- Finally, you should add a docstring. This docstring should explain what the function does and could also give a  formula for the calculation you're doing.

concent_api/core/templatetags/admin_tags.py
cameel: Please remember about https://github.com/Code-Poets/coding-style/issues/8

concent_api/core/models.py
cameel: Definitely not.

But Przemek mentioned that he's still working on `size` in TTC so maybe Golem is putting `None` there?

concent_api/core/models.py
cameel: @MichalW-CP It's a good idea to refer to docs in cases like this. Saves the reviewer some unnecessary checking. I remember that we went through these docs before:

https://docs.djangoproject.com/en/1.11/topics/db/managers/#default-managers
> If you use custom Manager objects, take note that the first Manager Django encounters (in the order in which they’re defined in the model) has a special status. Django interprets the first Manager defined in a class as the “default” Manager, and several parts of Django (including dumpdata) will use that Manager exclusively for that model. As a result, it’s a good idea to be careful in your choice of default manager in order to avoid a situation where overriding get_queryset() results in an inability to retrieve objects you’d like to work with.

concent_api/core/migrations/0004_auto_20180620_0712.py
cameel: I that case it's fine.

concent_api/core/migrations/0003_add_and_update_deadline_and_size.py
```diff
@@ -0,0 +1,67 @@
...
+    StoredMessage = apps.get_model('core', 'StoredMessage')
+    for subtask in Subtask.objects.filter(computation_deadline__isnull=True):
+        stored_message = StoredMessage.objects.filter(
+            pk=subtask.task_to_compute_id
+        ).first()
+        if stored_message is not None:
+            subtask.computation_deadline = parse_timestamp_to_utc_datetime(
+                deserialize_message(
+                    stored_message.data
+                ).compute_task_def['deadline']
```
cameel: `StoredMessage` can store various types of messages. You'll get not only `TaskToCompute` but all kinds of other messages associated with the task. So the `deadline` field may not be present. It's only in `TaskToCompute`.

Actually, I think that it would be better if `computation_deadline` were `None` on `StoredMessage` instances that do not represent `TaskToCompute`.

concent_api/core/migrations/0003_add_and_update_deadline_and_size.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def populate_deadline_size_field(apps, _schema_editor):
+    Subtask = apps.get_model('core', 'Subtask')
+    StoredMessage = apps.get_model('core', 'StoredMessage')
+    for subtask in Subtask.objects.filter(computation_deadline__isnull=True):
+        stored_message = StoredMessage.objects.filter(
+            pk=subtask.task_to_compute_id
+        ).first()
+        if stored_message is not None:
```
cameel: Can `stored_message` really be `None`?

concent_api/core/migrations/0003_add_and_update_deadline_and_size.py
```diff
@@ -0,0 +1,67 @@
...
+            subtask.save()
+
+
+def populate_deadline_size_field(apps, _schema_editor):
+    Subtask = apps.get_model('core', 'Subtask')
+    StoredMessage = apps.get_model('core', 'StoredMessage')
+    for subtask in Subtask.objects.filter(computation_deadline__isnull=True):
+        stored_message = StoredMessage.objects.filter(
+            pk=subtask.task_to_compute_id
+        ).first()
```
cameel: If you expect only one item to be returned you should either use `get()` or `assert` that there's only one result after the `filter()`. As it is now if you happen to get more than one result you silently ignore it and take the first one.

For me seeing a call to `first()` is an instant red flag. I don't think I've ever had to use `first()` myself. If you use it anywhere else, please make sure you're not making the first error.

concent_api/core/migrations/0003_add_and_update_deadline_and_size.py
cameel: I think so too. There should be an `assert` or `get()` instead of `filter()`.

See https://github.com/golemfactory/concent/pull/570/files#r198907346

concent_api/core/migrations/0003_add_and_update_deadline_and_size.py
cameel: @rwrzesien Maybe you can answer: don't the foreign keys work in migrations too? Is a separate query really necessary here?

concent_api/core/message_handlers.py
```diff
@@ -1117,7 +1125,7 @@ def update_subtask(
     )
 
     # Concent should send e-mail notification when the last active subtask switches to a passive state.
-    if config.SOFT_SHUTDOWN_MODE is True and not Subtask.objects.filter(state__in=Subtask.ACTIVE_STATES).exists():
+    if config.SOFT_SHUTDOWN_MODE is True and not Subtask.objects.filter(state__in=Subtask.ACTIVE_STATES).exists():  # pylint: disable=no-member
```
cameel: You did not even change this line. Why is the ignore necessary? Did someone leave an unfixed warning?

Please make sure you're running the latest version of pylint.

concent_api/core/message_handlers.py
```diff
@@ -1301,7 +1309,7 @@ def handle_send_subtask_results_verify(
             reason=message.concents.ServiceRefused.REASON.InvalidRequest,
         )
 
-    if Subtask.objects.filter(
+    if Subtask.objects.filter(  # pylint: disable=no-member
```
cameel: Having pylint complain about `no-member` on a simple `filter()` call is bizarre. This should not be necessary. There is something really weird going on here. Please investigate.

concent_api/core/message_handlers.py
```diff
@@ -819,15 +819,19 @@ def store_subtask(
...
-        task_to_compute=store_message(task_to_compute, task_id, subtask_id),
+        task_id             =task_id,
+        subtask_id          =subtask_id,
+        provider            =provider,
+        requestor           =requestor,
+        result_package_size =result_package_size,
+        state               =state.name,
+        next_deadline       =parse_timestamp_to_utc_datetime(next_deadline) if next_deadline is not None else None,
+        computation_deadline=parse_timestamp_to_utc_datetime(computation_deadline),
+        task_to_compute     =store_message(task_to_compute, task_id, subtask_id),
```
cameel: This is indented in a very weird way. It follows neither the new nor the old style. Please fix it.

concent_api/core/admin.py
```diff
@@ -24,21 +25,25 @@ def queryset(self, request, queryset):
         elif self.value() == 'passive':
             passive_state_names = [x.name for x in Subtask.PASSIVE_STATES]
             return queryset.filter(state__in=passive_state_names)
+        elif self.value() == 'active_downloads':
+            current_timestamp = get_current_utc_timestamp()
+            return Subtask.objects_with_timing_columns.filter(time_for_download_file__gte=current_timestamp)
         return queryset
```
cameel: What about tasks that are not in the RESULT UPLOADED state? The filter should include only the tasks that may have active downloads.

concent_api/core/admin.py
```diff
@@ -24,21 +25,25 @@ def queryset(self, request, queryset):
         elif self.value() == 'passive':
             passive_state_names = [x.name for x in Subtask.PASSIVE_STATES]
             return queryset.filter(state__in=passive_state_names)
+        elif self.value() == 'active_downloads':
+            current_timestamp = get_current_utc_timestamp()
+            return Subtask.objects_with_timing_columns.filter(time_for_download_file__gte=current_timestamp)
```
cameel: The issue said to add a filter that shows both active tasks and tasks with active downloads at the same time. Your filter shows only one or the other.

concent_api/core/models.py
cameel: Actually, this is `StoredMessage` and it may contain various types of messages, not only `TaskToCompute`. They do not have these fields. So the fields should be nullable after all.

concent_api/core/admin.py
```diff
@@ -24,21 +25,25 @@ def queryset(self, request, queryset):
...
-
     list_display = [
         'subtask_id',
         'task_id',
         'state',
         'get_provider_public_key',
         'get_requestor_public_key',
         'next_deadline',
+        'computation_deadline',
+        'result_package_size',
```
cameel: Please show also the three newly added columns. Especially the one that shows the download deadline. This will allow the admin to see how long each task has until the timeout.

concent_api/core/migrations/0003_add_and_update_deadline_and_size.py
rwrzesien: I have remembered that there was some issues with migrations, and when I saw code ready for review done this way, I thought that it was already taken care off. That's why I didn't put attention to the way its done. So good that you have pointed it out. The foreign keys works fine, so this part should be changed. What did not work is using custom models methods.

concent_api/core/models.py
cameel: This makes the query much more complex and we don't need these columns 99% of the time.

Also, while extra columns added with `annotate()` are nice, they can get in the way if you're doing complex joins. Especially if you're joining again with the table that was already used in column definition.

rwrzesien: > This makes the query much more complex and we don't need these columns 99% of the time.

`get_queryset` could be then changed to be a custom method.

cameel: I meant the underlying raw SQL query. And not just looking more complex. Actually having more joins and conditions.

concent_api/core/migrations/0003_add_and_update_deadline_and_size.py
MichalW-CP: Done.

concent_api/core/migrations/0003_add_and_update_deadline_and_size.py
MichalW-CP: Done.

concent_api/core/migrations/0003_add_and_update_deadline_and_size.py
MichalW-CP: Done.

concent_api/core/message_handlers.py
MichalW-CP: Done.

concent_api/core/message_handlers.py
MichalW-CP: Done.

concent_api/core/migrations/0003_add_and_update_deadline_and_size.py
MichalW-CP: Done.

concent_api/core/migrations/0003_add_and_update_deadline_and_size.py
MichalW-CP: But I make those fields in Subtask model not in StoredMessage. TaskToCompute is always present in Subtask

concent_api/core/admin.py
MichalW-CP: Done.

concent_api/core/models.py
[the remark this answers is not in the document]
MichalW-CP: Done.

concent_api/core/migrations/0003_add_and_update_deadline_and_size.py
MichalW-CP: Done.

concent_api/core/message_handlers.py
MichalW-CP: Done.

concent_api/core/admin.py
```diff
@@ -56,9 +67,14 @@ def get_requestor_public_key(self, obj):  # pylint: disable=no-self-use
         return obj.requestor.public_key
     get_requestor_public_key.short_description = 'Requestor public key'  # type: ignore
 
+    def get_queryset(self, request):
+        return Subtask.objects_with_timing_columns
 
-class PendingResponseAdmin(ModelAdminReadOnlyMixin, admin.ModelAdmin):
+    def download_deadline(self, obj):  # pylint: disable=no-self-use
```
rwrzesien: This should have `short_description` like the methods above.

concent_api/core/message_handlers.py
```diff
@@ -75,8 +75,8 @@ def handle_send_force_report_computed_task(client_message):
         requestor_public_key,
         task_to_compute,
     )
-
-    if Subtask.objects.filter(
+    # QuerySet is inferred as lis; pylint: disable=no-member; https://github.com/PyCQA/pylint-django/issues/165
```
rwrzesien: `lis` -> `list`

concent_api/core/message_handlers.py
```diff
@@ -852,14 +852,18 @@ def store_subtask(
 
     provider  = Client.objects.get_or_create_full_clean(provider_public_key)
     requestor = Client.objects.get_or_create_full_clean(requestor_public_key)
+    computation_deadline = task_to_compute.compute_task_def['deadline']
+    result_package_size = report_computed_task.size
```
rwrzesien: How is this related with the issue which this pull request resolves ?

concent_api/core/migrations/0006_add_and_update_deadline_and_size.py
```diff
@@ -0,0 +1,59 @@
...
+
+from django.db import migrations
+from django.db import models
+from common.helpers import deserialize_message
+from common.helpers import parse_timestamp_to_utc_datetime
+
+
+def populate_subtask_size_field(apps, _schema_editor):
+    Subtask = apps.get_model('core', 'Subtask')
+    for subtask in Subtask.objects.filter(result_package_size__isnull=True):
```
rwrzesien: What if `Subtask` for some reason will not have `report_computed_task` ? The line below will fail then. If that is desired behaviour then it is ok. If not, excluding `report_computed_task=None` should be added to filter.

concent_api/core/migrations/0006_add_and_update_deadline_and_size.py
```diff
@@ -0,0 +1,59 @@
...
+            serialized_report_computed_task.data
+        ).size
+        subtask.full_clean()
+        subtask.save()
+
+
+def populate_subtask_deadline_field(apps, _schema_editor):
+    Subtask = apps.get_model('core', 'Subtask')
+    for subtask in Subtask.objects.filter(computation_deadline__isnull=True):
+        serialized_task_to_compute = subtask.task_to_compute
```
rwrzesien: Same here.

concent_api/core/models.py
```diff
@@ -28,6 +34,29 @@
...
 
 
+class SubtaskWithTimingColumnsManager(Manager):
+    """Creates maximum_download time, subtask_verification_time and download_deadline columns
+    maximum_download_time = DOWNLOAD_LEADIN_TIME + ceil((result_package_size / MINIMUM_UPLOAD_RATE << 10))
+    subtask_verification_time = (4 * CONCENT_MESSAGING_TIME) + (3 * maximum_download_time) + (0.5 * (computation_deadline - task_to_compute_timestamp))
+    download_deadline = computation_deadline + maximum_download_time
+    """
+    def get_queryset(self):
+        bytes_per_sec = settings.MINIMUM_UPLOAD_RATE << 10
```
rwrzesien: I suggest separate with blank lines calculation of all three columns if possible. And add a comment on the start of each calculation.

concent_api/core/models.py
```diff
@@ -28,6 +34,29 @@
...
+        download_time = Func(F('result_package_size') / float(bytes_per_sec), function='CEIL')
+        maximum_download_time = F('maximum_download_time')
+        task_to_compute_timestamp = Func(Value('epoch'), F('task_to_compute__timestamp'), function='DATE_PART')
+        computation_deadline = F('computation_deadline')
+        subtask_timeout = Func(Value('epoch'), computation_deadline, function='DATE_PART') - task_to_compute_timestamp
+        subtask_verification_time = (4 * settings.CONCENT_MESSAGING_TIME) + (3 * maximum_download_time) + (0.5 * subtask_timeout)
+        download_deadline = Func(Value('epoch'), computation_deadline, function='DATE_PART') + subtask_verification_time
+        return super().get_queryset(
```
rwrzesien: Please fix styling.

concent_api/core/models.py
rwrzesien: The raw SQL query would be the same, unless you use custom method.

concent_api/core/admin.py
MichalW-CP: Done.

concent_api/core/message_handlers.py
MichalW-CP: Done.

concent_api/core/migrations/0006_add_and_update_deadline_and_size.py
cameel: `report_computed_task` is not nullable.

concent_api/core/migrations/0003_add_and_update_deadline_and_size.py
cameel: OK. But the name was `stored_message`. That's why you should avoid using misleading names :)

concent_api/core/admin.py
```diff
@@ -1,44 +1,55 @@
...
     def queryset(self, request, queryset):
+        active_state_names = [x.name for x in Subtask.ACTIVE_STATES]
+        passive_state_names = [x.name for x in Subtask.PASSIVE_STATES]
         if self.value() == 'active':
-            active_state_names = [x.name for x in Subtask.ACTIVE_STATES]
             return queryset.filter(state__in=active_state_names)
         elif self.value() == 'passive':
-            passive_state_names = [x.name for x in Subtask.PASSIVE_STATES]
             return queryset.filter(state__in=passive_state_names)
+        elif self.value() == 'active_downloads':
```
cameel: I'd call it `active_or_downloads` to make it clear that it's OR, not AND.

concent_api/core/message_handlers.py
```diff
@@ -75,8 +75,8 @@ def handle_send_force_report_computed_task(client_message):
         requestor_public_key,
         task_to_compute,
     )
-
-    if Subtask.objects.filter(
+    # QuerySet is inferred as list; pylint: disable=no-member; https://github.com/PyCQA/pylint-django/issues/165
+    if Subtask.objects.filter(  # pylint: disable=no-member
```
cameel: Please reword it:

``` python
# We have to disable no-member warning on Subtask queries because pylint thinks the result is a list, not QuerySet.
# See this pylint bug: https://github.com/PyCQA/pylint-django/issues/165
```

concent_api/core/message_handlers.py
cameel: @MichalW-CP What about my question here?

concent_api/core/models.py
cameel: How could it be the same if the extra columns refer for example to `task_to_compute__timestamp`? It must result either in a join or in an extra query.

concent_api/core/models.py
```diff
@@ -291,6 +320,10 @@ class Meta:
 
     task_id = CharField(max_length=MESSAGE_TASK_ID_MAX_LENGTH)
 
+    computation_deadline        = DateTimeField()
+
+    result_package_size         = IntegerField()
```
cameel: Please add a validation (inside the `clean()` method) that these fields always match the values inside `report_computed_task`.

concent_api/core/templatetags/admin_tags.py
cameel: @MichalW-CP Please make a list of disabled warnings as stated above.

concent_api/core/models.py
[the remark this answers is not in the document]
cameel: What about splitting it into two functions as suggested in my comment. I think it would go a long way towards making it more readable.

concent_api/core/admin.py
```diff
@@ -1,44 +1,55 @@
...
+class ActivePassiveDownloadsStateFilter(admin.SimpleListFilter):
+    title = 'Active/passive/downloads state'
     parameter_name = 'active_passive_state'
 
     def lookups(self, request, model_admin):
         return (
-            ('active',  'Active'),
+            ('active', 'Active'),
             ('passive', 'Passive'),
+            ('active_downloads', 'Active with active downloads')
```
dybi: please rename: `'Active or with active downloads'`

concent_api/core/migrations/0006_add_and_update_deadline_and_size.py
```diff
@@ -0,0 +1,57 @@
...
+class Migration(migrations.Migration):
+    dependencies = [
+        ('core', '0005_auto_20180717_0908'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='subtask',
+            name='computation_deadline',
+            field=models.DateTimeField(blank=True, null=True),
```
dybi: I am not sure if I get it right, but in the `Subtask` model this field is neither `blank` nor `null`. Is this setting (`(blank=True, null=True)`) necessary for migration?

concent_api/core/migrations/0006_add_and_update_deadline_and_size.py
```diff
@@ -0,0 +1,57 @@
...
+        ),
+        migrations.AddField(
+            model_name='subtask',
+            name='result_package_size',
+            field=models.IntegerField(blank=True, null=True),
```
dybi: here same as above.

concent_api/core/tests/test_models_custom_manager_queryset.py
```diff
@@ -0,0 +1,59 @@
...
+                next_deadline=get_current_utc_timestamp() + settings.CONCENT_MESSAGING_TIME,
+                task_to_compute=self.task_to_compute,
+                report_computed_task=self.report_computed_task,
+            )
+
+        self.subtask_verification_time = calculate_subtask_verification_time(self.report_computed_task)
+        self.download_deadline = 1512127820
+        self.maximum_download_time = 4
+
+    def test_subtask_verification_time_query(self):
```
dybi: some parametric tests could be used

concent_api/core/tests/test_models_custom_manager_queryset.py
```diff
@@ -0,0 +1,59 @@
...
+from freezegun import freeze_time
+
+from common.helpers import get_current_utc_timestamp
+from core.message_handlers import store_subtask
+from core.models import Subtask
+from core.tests.utils import ConcentIntegrationTestCase
+from core.utils import calculate_subtask_verification_time
+
+
+class SubtaskWithTimingColumnsManagerQuerySetTest(ConcentIntegrationTestCase):
```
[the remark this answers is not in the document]
dybi: @cameel, to be precise, the tests would need to guarantee that @MichalW-CP's implementation won't get out of sync with ** our custom functions that compute MDT and SVT**. Tests for our implementation of MDT and SVT ensure that it is in line with `golem-messages`

concent_api/core/message_handlers.py
MichalW-CP: Pylint version is up to date

concent_api/core/migrations/0006_add_and_update_deadline_and_size.py
MichalW-CP: Migration could not be processed without (blank=True, null=True). This is workaround if we want those fields to be not nullable in model.

concent_api/core/migrations/0006_add_and_update_deadline_and_size.py
MichalW-CP: Migration could not be processed without (blank=True, null=True). This is workaround if we want those fields to be not nullable in model.

concent_api/core/migrations/0006_add_and_update_deadline_and_size.py
cameel: You can do this without the workaround but you need to tweak the migration a bit. It should be non-nullable in the model but you can make it temporarily nullable in the migration:
- Create the field as nullable
- Perform data migration that fills the field
- Modify the field so that it's no longer nullable.

concent_api/core/tests/test_models_custom_manager_queryset.py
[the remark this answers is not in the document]
cameel: Yes, you're right.

concent_api/core/migrations/0006_add_and_update_deadline_and_size.py
rwrzesien: What is this workaround fixing ?

concent_api/core/message_handlers.py
rwrzesien: I did not say it is not related, but asking how.

concent_api/core/migrations/0006_add_and_update_deadline_and_size.py
cameel: I think that the problem is that when you add a new, non-nullable field Django requires a default because otherwise it cannot assign it a value. We don't want to add a default.

rwrzesien: But on the other hand we can't have mismatch between field definition and database structure.

cameel: Yeah but the structure is in the process of being changed when you're inside a migration. So if it's nullable at first and then not, it still matches the database (at a given point in time).

rwrzesien: So the migration for each field should have 3 steps:
1. Create with null and blank.
2. Set desired values.
3. Make non-null and non-blank.

MichalW-CP:
Resolved in PR's:
https://github.com/golemfactory/concent/issues/329
https://github.com/golemfactory/concent/issues/660
https://github.com/golemfactory/concent/issues/661

Issue #549, opened 2018-06-25
Refactor blender_verification_order task


dybi:
There are many "code blobs" (ie. downloading archives, unpacking them, runnnig blender, comparing images, etc) in `blender_verification_order` which have just one reponsibility and should be extracted to small, well-named function.
Also in each of this "code blobs" in case of exception, `VerificationResult(ERROR)` is being scheduled. This is a significant code duplication and should be handled by introduction of decorator, which would schedule `VerificationResult(ERROR)` in case any of new samll functions raises `VerificatioinError` (which should be introduced.
**TL;DR**
1. Extract small well named functions from `blender_verification_order`
2. Introduce new exception `VerificationError` which will be rased in new small functions in case of erroneous situation
3. Introduce new decorator `handle_verification_errors` which would catch all `VerificationError` and schedule `verification_result` witg result `ERROR`

dybi:
Summary:
- mypy annotation is missing, separate task has been created: https://github.com/golemfactory/concent/issues/583
- unit tests for helper functions are missing, separate task has been created: https://github.com/golemfactory/concent/issues/582

Issue #528, opened 2018-06-20
[VERIFICATION] scene file is not related to source_package_path


dybi:
after unpacking `source_package_path`, `scene_file` passed to `blender` process should be taken from unpacked files. This is not the case.

rwrzesien:
@dybi What is the exact problem here ? Which part of the code doesn't work as expected ?

PR #573, opened 2018-06-27
Chore refactor verification

dybi:
It is far form perfect, but it seems much better than it was.
resolves: #549 
resoves #528 

NOTES:
 - there are no unit tests for new functions, they will be added in a separate issue 
 - there are few (probably 2) missing integration tests, they will be added in a separate issue 
 - not yet tested on cluster (will be tomorrow), tested locally and works


concent_api/api-integration-additional-verification-test.py
```diff
@@ -88,12 +88,16 @@ def test_case_1_test_for_positive_case(cluster_consts, cluster_url, test_id):
...
     (subtask_id, task_id) = get_task_id_and_subtask_id(test_id, 'existing_file')
 
-    result_file_content_1 = task_id
-    source_file_content_2 = subtask_id
-    result_file_size_1 = len(result_file_content_1)
-    source_file_size_2 = len(source_file_content_2)
-    result_file_check_sum_1 = 'sha1:' + hashlib.sha1(result_file_content_1.encode()).hexdigest()
-    source_file_check_sum_2 = 'sha1:' + hashlib.sha1(source_file_content_2.encode()).hexdigest()
+    current_dir = os.path.dirname(os.path.abspath(__file__))
+    with open(os.path.join(current_dir, 'tests_resources', 'source.zip'), 'rb') as f:
```
rwrzesien: Please use full variable names -> `f`.

concent_api/verifier/tasks.py
```diff
@@ -106,264 +107,252 @@ def blender_verification_order(
         operation=message.FileTransferToken.Operation.download,
     )
 
-    # Remove any files from VERIFIER_STORAGE_PATH.
-    clean_directory(settings.VERIFIER_STORAGE_PATH)
```
rwrzesien: There was a requirement to clean directory at the beginning of the task. I know that now all exceptions are caught and directory is cleaned there, but I think it is still possible for verification to be stopped without this (ie. machine get powered off). So I think this should stay.

concent_api/verifier/tasks.py
```diff
@@ -106,264 +107,252 @@ def blender_verification_order(
...
-            )
-            return
+    unpack_archives(package_paths_to_downloaded_archive_names.values(), subtask_id)
 
-    # Verifier runs blender process.
-    try:
-        completed_process = run_blender(
-            scene_file,
-            output_format,
+    frame_number = 1
```
rwrzesien: I know this is temporary but you can already add it as a list and render in loop.

concent_api/verifier/tasks.py
```diff
@@ -106,264 +107,252 @@ def blender_verification_order(
...
-        f'Blender finished successfully. SUBTASK_ID: {subtask_id}.'
-        f'VerificationResult: {VerificationResult.MATCH.name}'
-    )
+    compare_images(image_1, image_2, subtask_id)
 
-    blender_output_file_name = generate_blender_output_file_name(scene_file)
-    upload_file_name = generate_upload_file_name(subtask_id, output_format)
 
-    # Read Blender output file.
+def compare_images(image_1, image_2, subtask_id):
```
rwrzesien: All of the functions below should be in utils.

concent_api/verifier/utils.py
```diff
@@ -99,11 +110,18 @@ def delete_file(file_path):
         logger.warning(f'File with path {file_path} was not deleted, exception: {exception}')
 
 
-def generate_blender_output_file_name(scene_file):
-    return f'{settings.VERIFIER_STORAGE_PATH}/out_{scene_file}'
+def generate_full_blender_output_file_name(scene_file, frame_number, output_format):
+    base_blender_output_file_name = generate_base_blender_output_file_name(scene_file)
+    if output_format == 'jpeg':
```
rwrzesien: This `output_format` went out.

concent_api/verifier/utils.py
```diff
@@ -52,30 +52,41 @@ def prepare_storage_request_headers(file_transfer_token: message.FileTransferTok
 
 
 def store_file_from_response_in_chunks(response: requests.Response, file_path: str):
-    with open(file_path, 'x') as f:
+    with open(file_path, 'wb') as f:
```
rwrzesien: Why not 'x' ? It is supposed to fail when file already exists.

rwrzesien:
I have took just a quick view, looks good, added some quick comments/questions. I will try to take to better review tomorrow.

concent_api/api-integration-additional-verification-test.py
dybi: done

concent_api/verifier/tasks.py
dybi: I would prefer not to do this in this PR -> it has a lot of changes as it is ;)
I can do it in separate issue

concent_api/verifier/utils.py
```diff
@@ -52,30 +52,41 @@ def prepare_storage_request_headers(file_transfer_token: message.FileTransferTok
...
+    For desired JPG format, the parameter for blender should be JPEG and the extension of result file is *.jpg.
+    """
+    if output_format.upper() == 'JPG':
+        return 'JPEG'
+    return output_format.upper()
+
+
 def unpack_archive(file_path):
     """ Unpacks archive in chunks. """
     with zipfile.ZipFile(os.path.join(settings.VERIFIER_STORAGE_PATH, file_path), 'r') as zf:
```
kbeker: You can also change this short variable name `zf` :)

concent_api/api-integration-additional-verification-test.py
kbeker: I think if we want to upload these files from our disc and store them on our repository then these files must be as small as possible. Maybe good idea would be to create small `test_resources` with correct and incorrect results?

concent_api/verifier/decorators.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def handle_verification_errors(task):
+    @wraps(task)
+    def wrapper(*args, **kwargs):
+        try:
+            return task(*args, **kwargs)
+        except VerificationError as exception:
+            verification_result.delay(
+                exception.subtask_id,
```
rwrzesien: You can simplify the `VerificationError` and take `subtask_id` from `*args`.

concent_api/verifier/utils.py
```diff
@@ -52,30 +52,41 @@ def prepare_storage_request_headers(file_transfer_token: message.FileTransferTok
...
 
 
-def run_blender(scene_file, output_format, script_file=''):
+def run_blender(scene_file, output_format, frame_number, script_file=''):
+    output_format = adjust_format_name(output_format)
     return subprocess.run(
         [
             "blender",
-            "-b", f"{scene_file}",
+            "-b", f"{settings.VERIFIER_STORAGE_PATH}/{scene_file}",
```
rwrzesien: `generate_verifier_storage_file_path` can be used here.

concent_api/verifier/utils.py
```diff
@@ -87,7 +98,7 @@ def unpack_archive(file_path):
 
 def get_files_list_from_archive(file_path):
     """ Returns list of files from given zip archive. """
-    return zipfile.ZipFile.namelist(file_path)
+    return zipfile.ZipFile(file_path).namelist()
```
rwrzesien: What is the difference between those two ?

concent_api/verifier/utils.py
dybi: actually should be `'xb'`, as it is a binary file, will change it ;)

concent_api/verifier/tasks.py
dybi: ok, I will move it and refactor tests (a loooooot of patches to change;))

rwrzesien: Right, but this can be done with few good `ctrl + R` ;) But I think `tasks.py` should contain only tasks functions.

concent_api/verifier/utils.py
dybi: what do you mean? aren't we supposed to pass this param?

concent_api/verifier/utils.py
dybi: done

concent_api/verifier/tasks.py
dybi: done

concent_api/verifier/utils.py
dybi: the first one doesn't work ;) - raises some Exception

concent_api/verifier/utils.py
dybi: done

concent_api/verifier/decorators.py
dybi: It would be a bit risky or not elegant:
I would need to check both `*args` and `**kwargs` as I have no warranty that `subtaski_id` will be passed this or other way. checking both seems a bit unelegant, so I would be for feeding it explicitly to `VerificationError`

concent_api/verifier/utils.py
dybi: done

concent_api/api-integration-additional-verification-test.py
dybi: I will discuss it wil @cameel

concent_api/verifier/utils.py
rwrzesien: No no, what I have meant is that `jpeg` is no longer value in `OutputFormat` enum, and we validate it at the beginning of the task, so I think this code will never be reached.

concent_api/verifier/utils.py
rwrzesien: Really ? I remember testing this part in context of memory used and it worked fine then. But if you say so and it works after change then it is ok.

concent_api/verifier/decorators.py
rwrzesien: Risky - I agree, not elegant - I think opposite :) But please consider that in case of failure in task you pass the same parameter to `VerificationError`, not knowing if it is really `subtask_id`.

concent_api/verifier/utils.py
dybi: ok, I get it. done (good catch)

cameel: Also, `'jpeg'` should be in some constant or enum, not hard-coded here.

We should have mapping between format names and extensions.

concent_api/api-integration-additional-verification-test.py
dybi: as discussed with @cameel, at the moment we will not store `test_resources` in repo.
I will remove the files

concent_api/verifier/utils.py
rwrzesien: > ok, I get it. done (good catch)

Lets say that I am familiar with how this one works ;)

> We should have mapping between format names and extensions.

Right, but currently it is 1 to 1 I think.

concent_api/api-integration-additional-verification-test.py
rwrzesien: Are we going to use hardcoded subtask_id as in other cluster tests ? Just asking with curiosity.

concent_api/verifier/utils.py
cameel: No, it isn't. There's `jpg` and `jpeg`. Blender expects one on the command line but uses the other as file extension. That's why @dybi added this `if` here.

Issue #486, opened 2018-06-12
store_or_update_subtask() is prone to race conditions and raises ValidationError


cameel:
We have received a crash report indicating that Concent tried to create a `Subtask` instance with an existing (`subtask_id`, `requestor_public_key`) pair. It was caught by the database and resulted in a `ValidationError`. It was probably because two processes tried to insert the same subtask at the same time.

```
ValidationError: {'__all__': ['Subtask with this Requestor and Task id already exists.']}
```

We need to handle in a more graceful way. Concent should detect this situation and simply retry by fetching the object created by the other process from the database.

The tricky part is discerning uniqueness errors from other errors. We want to catch them in `save()`, not during validations. Maybe we need to run validations after `save()`, not before.

### Stack trace
From https://talkback.golem.network/sentry/concent/issues/15340/. 

``` python
Exception(most recent call first)

  File "django/core/handlers/base.py", line 185, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "python3.6/contextlib.py", line 52, in inner
    return func(*args, **kwds)
  File "utils/decorators.py", line 202, in wrapper
    return _function(*args, **kwargs)
  File "django/views/decorators/csrf.py", line 58, in wrapped_view
    return view_func(*args, **kwargs)
  File "django/views/decorators/http.py", line 40, in inner
    return func(request, *args, **kwargs)
  File "utils/decorators.py", line 108, in wrapper
    return view(request, golem_message, client_public_key, *args, *kwargs)
  File "utils/decorators.py", line 124, in wrapper
    response_from_view = view(request, client_message, client_public_key, *args, **kwargs)
  File "core/views.py", line 35, in send
    return handle_message(client_message)
  File "core/message_handlers.py", line 1342, in handle_message
    return handle_send_subtask_results_verify(client_message)
  File "core/message_handlers.py", line 1294, in handle_send_subtask_results_verify
    subtask_results_rejected=subtask_results_rejected,
  File "core/message_handlers.py", line 1198, in store_or_update_subtask
    subtask_results_rejected        = subtask_results_rejected,
  File "core/message_handlers.py", line 824, in store_subtask
    subtask.full_clean()
  File "django/db/models/base.py", line 1250, in full_clean
    raise ValidationError(errors)
```

pawelkisielewicz:
I reproduced this Error in multi-thread tests, talked with @rwrzesien and my idea to solve this problem is to:
- write tests reproducing problem (done)
[...]

Issue #412, opened 2018-05-23
Ensure that the same subtask is never processed by two processes in parallel


cameel:
Ensure that the `send()` view, Celery tasks in `concent-worker` and the function for enforcing timeouts never run in parallel for the same subtask. If one of them is working on a subtask, the others should wait.

Use database locks to achieve this. When one of those pieces of code starts, it should lock the corresponding row in the `Subtask` table. If it's already locked, it should retry a few times with some delay in between. Eventually it should give up and fail. The time after which it gives up should be configurable (but use a sensible default). Also, make sure that Celery tasks that fail are retried. And that we get a Sentry notification when one fails permenently (fails so many times that Celery stops trying).

Make sure that the code releases the lock if it fails. Use a `try ... finally ...` block for that. This does not guarantee releasing it in all cases - for example if the process simply crashes (hard crashes, not just raises an exception) before it gets to the `finally` part. But this should interrupt the session and cause a rollback that releases the lock. Make sure that this is the case.

pawelkisielewicz:
After discuss with @cameel, the best approach to do this task is to use database savepoints and locks. 
Using `select_for_update` for subtasks requires, that subtask exist (this `lock` will not be started, if `subtask` does not exist).

We should prevent errors from database by saving (commiting) `subtask` to `database` just after create (we use `ATOMIC_REQUESTS`, so the changes are commited to `database` at the end of atomic transaction- at the end of whole functions (incl. decorators)). In case there are two the same subtasks processed in one time (and they are not in `database`, will be created with the same id), one of them will be saved (normal, positive scenario), and the second one will fail. In this case we should catch the exception and simply use `get` instead of `get_or_create`.   

After this we can lock database for particular subtask, in this moment we are sure that `subtask` exist in database. We must ensure release of `database lock` and commit changes to database at the end of function. 

In worker case, there will be only `database lock` required. There the subtasks are only updated (not created), so they always exist

rwrzesien:
Where do you want to start and release locks ?

pawelkisielewicz:
In workers the whole functions should be locked. 

In core/send `lock` should start just after `"get"` a `subtask`. `"get"` is a section of getting or creating subtaks (second paragraph) and when we have subtaks in database (commited to database) `lock` starts. In this case `lock` ends at the end of view

pawelkisielewicz:
[...]

PR #586, opened 2018-07-02
Add decorator to avoid race conditions

pawelkisielewicz:
resolves https://github.com/golemfactory/concent/issues/486 and it is a good start to resolve https://github.com/golemfactory/concent/issues/412

### DIFF
--- a/concent_api/core/exceptions.py
+++ b/concent_api/core/exceptions.py
@@ -9,6 +9,10 @@ class Http400(ConcentBaseException):
     pass
 
 
+class Http500(ConcentBaseException):
+    pass
+
+
 class FileTransferTokenError(ConcentBaseException):
     pass
 
--- a/concent_api/core/tests/test_unit_message_handlers.py
+++ b/concent_api/core/tests/test_unit_message_handlers.py
@@ -1,6 +1,13 @@
+import threading
+from threading import Thread
+import time
 from unittest import TestCase
 
+from django.test import TransactionTestCase
+
 from core.message_handlers import are_items_unique
+from core.tests.test_utils import DatabaseHandler
+from core.tests.test_utils import StoreOrUpdateSubtaskTest
 
 
 class TestMessageHandlers(TestCase):
@@ -11,3 +18,36 @@ class TestMessageHandlers(TestCase):
     def test_that_function_returns_false_when_ids_are_the_same(self):
         response = are_items_unique([1, 2, 3, 2, 5])
         self.assertFalse(response)
+
+
+class TestEnsureRetryOfLockedCalls(TransactionTestCase):
+
+    def setUp(self):
+        super().setUp()
+        self.hepler1 = StoreOrUpdateSubtaskTest(task_id='g112d', subtask_id='hrff4')
+        self.hepler2 = StoreOrUpdateSubtaskTest(task_id='h5575', subtask_id='ddfg3')
+        self.hepler3 = StoreOrUpdateSubtaskTest(task_id='poipd', subtask_id='jj435')
+
+    def test_that_ensure_retry_of_locked_calls_should_handle_multiprocessing_calls(self):
+        """
+        This test checks, if when a function handling with database is calling multiple times in the same time (by
+        different processes), decorator prevents of crash caused by race condition. Instead of crash processed
+        should call function one by one. Database lock is created only for particular element, not the whole database
+        """
+
+        for i in range(5):  # pylint: disable=unused-variable
+            t = Thread(target=self.hepler1.run_store_or_update_subtask, args=())
+            t.start()
+
+        for i in range(3):  # pylint: disable=unused-variable
+            t = Thread(target=self.hepler1.run_store_or_update_subtask, args=())
+            t.start()
+            t = Thread(target=self.hepler2.run_store_or_update_subtask, args=())
+            t.start()
+            t = Thread(target=self.hepler3.run_store_or_update_subtask, args=())
+            t.start()
+
+        time.sleep(1)
+        self.assertEqual(threading.active_count(), 1)
+        database_handler = DatabaseHandler()
+        database_handler.deactivate_communication_with_database()
--- a/concent_api/core/tests/test_utils.py
+++ b/concent_api/core/tests/test_utils.py
@@ -2,18 +2,23 @@ import datetime
 import mock
 
 from django.conf import settings
+from django.core.management import BaseCommand
+from django.db import connection
 from django.test import override_settings
 from django.test import TestCase
 
 from golem_messages import constants
 from golem_messages import helpers
 
+from core.message_handlers import store_or_update_subtask
+from core.models import Subtask
 from core.tests.utils import ConcentIntegrationTestCase
 from core.utils import calculate_maximum_download_time
 from core.utils import calculate_subtask_verification_time
 from common.helpers import get_current_utc_timestamp
 from common.helpers import parse_timestamp_to_utc_datetime
 from common.testing_helpers import generate_ecc_key_pair
+from common.testing_helpers import generate_priv_and_pub_eth_account_key
 
 (CONCENT_PRIVATE_KEY, CONCENT_PUBLIC_KEY) = generate_ecc_key_pair()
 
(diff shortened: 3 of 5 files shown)


concent_api/verifier/decorators.py
```diff
@@ -24,3 +28,38 @@ def wrapper(*args, **kwargs):
...
+                repeat_handler.is_max_number_of_tries_exceeded()
+                time.sleep(wait_time_in_sec)
+        return result
+    return wrapper
+
+
+class RepeatHandler:
+
+    def __init__(self):
+        self.max_number_of_retries = 100
```
rwrzesien: This should be in settings.

concent_api/verifier/decorators.py
```diff
@@ -24,3 +28,38 @@ def wrapper(*args, **kwargs):
...
+        self.current_number_of_retries = 0
+
+    def is_max_number_of_tries_exceeded(self):
+        if self.current_number_of_retries < self.max_number_of_retries:
+            self.current_number_of_retries = self.current_number_of_retries + 1
```
rwrzesien: This can be rewritten to:
```
self.current_number_of_retries += 1
```

concent_api/verifier/decorators.py
```diff
@@ -24,3 +28,38 @@ def wrapper(*args, **kwargs):
...
+        else:
+            raise Http500(
```
rwrzesien: Do we want to return 500 here ? @cameel @dybi  ?

concent_api/core/tests/test_utils.py
```diff
@@ -120,3 +125,40 @@ def test_that_both_subtask_verification_time_implementation_should_return_same_r
...
+            requestor_public_key=self.REQUESTOR_PUBLIC_KEY,
+            state=Subtask.SubtaskState.ADDITIONAL_VERIFICATION,
+            next_deadline=get_current_utc_timestamp() + settings.CONCENT_MESSAGING_TIME,
+            task_to_compute=self.report_computed_task.task_to_compute,
+            report_computed_task=self.report_computed_task,
+        )
+
+
+class DatabaseHandler(BaseCommand):   # pylint: disable=abstract-method
+
```
rwrzesien: Could you please add a comment here what this command does ?

concent_api/core/tests/test_unit_message_handlers.py
```diff
@@ -11,3 +18,36 @@ def test_that_function_returns_true_when_ids_are_diffrent(self):
     def test_that_function_returns_false_when_ids_are_the_same(self):
         response = are_items_unique([1, 2, 3, 2, 5])
         self.assertFalse(response)
+
+
+class TestEnsureRetryOfLockedCalls(TransactionTestCase):
+
+    def setUp(self):
+        super().setUp()
+        self.hepler1 = StoreOrUpdateSubtaskTest(task_id='g112d', subtask_id='hrff4')
```
rwrzesien: Type `hepler` -> `helper`.

concent_api/core/tests/test_utils.py
```diff
@@ -120,3 +125,40 @@ def test_that_both_subtask_verification_time_implementation_should_return_same_r
                 calculate_subtask_verification_time(report_computed_task),
                 int(helpers.subtask_verification_time(report_computed_task).total_seconds())
             )
+
+
+class StoreOrUpdateSubtaskTest(ConcentIntegrationTestCase):
```
rwrzesien: Why doing this using subclass of `ConcentIntegrationTestCase` ? Couldn't this be done using method defined in `TestEnsureRetryOfLockedCalls` ?

concent_api/verifier/decorators.py
```diff
@@ -24,3 +28,38 @@ def wrapper(*args, **kwargs):
...
+        should_repeat = True
+        result = None
+
+        while should_repeat:
+            try:
+                result = func(*args, **kwargs)
+                should_repeat = False
+            except DatabaseError:
+                repeat_handler.is_max_number_of_tries_exceeded()
+                time.sleep(wait_time_in_sec)
```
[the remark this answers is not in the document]
pawelkisielewicz: I tested it and instant retry requires 20-30 retries, when i test it in 2-3 threads. The sleep time (should be set in settings as well) set as 0.05 second is like 'instant', but requires only 3-5 retries.

concent_api/verifier/decorators.py
```diff
@@ -24,3 +28,38 @@ def wrapper(*args, **kwargs):
...
+    return wrapper
+
+
+class RepeatHandler:
+
+    def __init__(self):
+        self.max_number_of_retries = 100
+        self.current_number_of_retries = 0
+
+    def is_max_number_of_tries_exceeded(self):
```
cameel: This function starts with `is_...` which suggests that it returns a boolean value but it does not return anything.

concent_api/verifier/decorators.py
cameel: Also, it would be better if this class had more control over the retry.

E.g. instead of just checking the number of retries it could have a function that returns either the next delay or `None`. `None` would mean no more retrying, just fail. Next delay would be determined based on the current retry count or other factors. This way we could have different implementations providing different retry strategies. E.g. constant delay, delay by powers of 2, etc.

concent_api/verifier/decorators.py
cameel: Raising an exception is fine (as long as the name of the function is changed not to suggest that it returns a boolean), but definitely not something like `Http500`. Returning HTTP 500 is the result not the cause and deciding what the result should be is the job of the handler that catches the exception. The exception itself should indicate the cause instead. E.g. `RetryCountExceeded`.

concent_api/verifier/decorators.py
```diff
@@ -24,3 +28,38 @@ def wrapper(*args, **kwargs):
...
+        if self.current_number_of_retries < self.max_number_of_retries:
+            self.current_number_of_retries = self.current_number_of_retries + 1
+        else:
+            raise Http500(
+                "Maximum number of retries of function updating Subtask extended",
```
cameel: `extended` -> `exceeded`

concent_api/verifier/decorators.py
```diff
@@ -24,3 +28,38 @@ def wrapper(*args, **kwargs):
...
+                result = func(*args, **kwargs)
+                should_repeat = False
+            except DatabaseError:
+                repeat_handler.is_max_number_of_tries_exceeded()
+                time.sleep(wait_time_in_sec)
+        return result
+    return wrapper
+
+
+class RepeatHandler:
```
cameel: I think that something like `RetryStrategy` would be a better name.

concent_api/verifier/decorators.py
```diff
@@ -24,3 +28,38 @@ def wrapper(*args, **kwargs):
...
     return wrapper
+
+
+def ensure_retry_of_locked_calls(func):
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        repeat_handler = RepeatHandler()
+        wait_time_in_sec = 0.05
+        should_repeat = True
+        result = None
```
cameel: Let's call it `retry_...`, not `repeat_...`

concent_api/verifier/decorators.py
```diff
@@ -24,3 +28,38 @@ def wrapper(*args, **kwargs):
             # Remove any files left in VERIFIER_STORAGE_PATH.
             clean_directory(settings.VERIFIER_STORAGE_PATH)
     return wrapper
+
+
+def ensure_retry_of_locked_calls(func):
```
cameel: I'd make the name more explicit. How about `retry_on_database_error()`?

concent_api/verifier/decorators.py
```diff
@@ -24,3 +28,38 @@ def wrapper(*args, **kwargs):
...
+        repeat_handler = RepeatHandler()
+        wait_time_in_sec = 0.05
+        should_repeat = True
+        result = None
+
+        while should_repeat:
+            try:
+                result = func(*args, **kwargs)
+                should_repeat = False
+            except DatabaseError:
```
cameel: Are you absolutely sure that `DatabaseError` always indicates a failure due to a timed out lock? Because I think that there are a lot more things that can cause this error and blindly retrying when they happen might actually be harmful.

We want to retry only if it's a problem with the lock. If there's a different problem we want to raise an exception and exit immediately.

concent_api/verifier/decorators.py
```diff
@@ -24,3 +28,38 @@ def wrapper(*args, **kwargs):
             # Remove any files left in VERIFIER_STORAGE_PATH.
             clean_directory(settings.VERIFIER_STORAGE_PATH)
     return wrapper
+
+
+def ensure_retry_of_locked_calls(func):
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        repeat_handler = RepeatHandler()
```
cameel: I think this is not a good use case for a decorator. A decorator covers the whole function and that scope is too broad. And if it fails, you retry the whole function, not just the piece of code that obtains a lock.

It will be a source of bugs because doing this requires you to be extra careful when writing the function. You have to ensure that it can be retried without permanent side-effects. That's not an easy thing to do in a generic way.

There's also a possibility of easy mistakes with decorator order. If you use it **after** you start a transaction (using `@atomic`) database changes made before acquiring the lock will not be rolled back.

If you really want to have a reusable wrapper that manages a lock, you should make it a context manager. See https://jeffknupp.com/blog/2016/03/07/python-with-context-managers/. You could make a context manager that gets a query set as a parameter and performs a select query with a lock.

concent_api/core/exceptions.py
```diff
@@ -9,6 +9,10 @@ class Http400(ConcentBaseException):
     pass
 
 
+class Http500(ConcentBaseException):
```
kbeker: Shouldn't we name exceptions more specific? This name is too general. If there exist need to add more exceptions just do it :)

concent_api/verifier/decorators.py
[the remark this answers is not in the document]
cameel: It will depend on what is inside the critical section.

A good strategy for waiting is to start with something small and double it after each try. For example Kubernetes does it like this.

concent_api/core/tests/test_utils.py
```diff
@@ -120,3 +125,40 @@ def test_that_both_subtask_verification_time_implementation_should_return_same_r
...
+
+class DatabaseHandler(BaseCommand):   # pylint: disable=abstract-method
+
+    def deactivate_communication_with_database(self):  # pylint: disable=abstract-method, no-self-use
+        cursor = connection.cursor()
+        database_name = f"{settings.DATABASES['control']['NAME']}"
+
+        cursor.execute(
+            "SELECT pg_terminate_backend(pg_stat_activity.pid) FROM pg_stat_activity "
+            "WHERE pg_stat_activity.datname = %s AND pid <> pg_backend_pid();", [database_name])
```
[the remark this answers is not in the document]
cameel: This is very hacky and non-portable. You're making many assumptions here and reaching into the internals you should not have to touch.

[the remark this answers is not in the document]
kbeker: Please use f-string

concent_api/verifier/decorators.py
```diff
@@ -24,3 +28,38 @@ def wrapper(*args, **kwargs):
...
+        wait_time_in_sec = 0.05
```
kbeker: This should be setting

concent_api/verifier/decorators.py
```diff
@@ -24,3 +28,38 @@ def wrapper(*args, **kwargs):
...
+        self.max_number_of_retries = 100
+        self.current_number_of_retries = 0
+
+    def is_max_number_of_tries_exceeded(self):
+        if self.current_number_of_retries < self.max_number_of_retries:
+            self.current_number_of_retries = self.current_number_of_retries + 1
+        else:
+            raise Http500(
+                "Maximum number of retries of function updating Subtask extended",
+                error_code=ErrorCode.CONCENT_APPLICATION_CRASH,
```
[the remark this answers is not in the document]
rwrzesien: I think yes.

pawelkisielewicz:
closed, because solving https://github.com/golemfactory/concent/issues/412 is more general and should solve this issue as well. Part of work on this issue and branch should be useful to solve https://github.com/golemfactory/concent/issues/412

Issue #553, opened 2018-06-25
Show creation and modification time in the admin panel


cameel:
Currently the creation and modification time of a subtask is not displayed on the subtask list in the admin panel. This might be useful for other models too, like `PendingResponse` or `StoredMessage`.

Add the missing columns.

cameel:
Well, it's easy enough to add them to the column list. It's simpler if you can see all the fields at a glance and there's nothing hidden. It also lets you easily control their position.

I have nothing against mixins but it might be an overkill in this case.

rwrzesien:
I have misread the description, I thought we want to add them to models.

rwrzesien:
Actually only one of the models has only `created_at` field. I think it is a good practice in Django to have most of the models both of those fields, so I have added `BaseModel` class containing those two fields and made other models inherit from it.

cameel:
It's better to have them by default than not to have them at all. But that's just an easy solution for when you don't want to think about them too much. They are not always needed and I'm adding them only to models that actually need them.

Sometimes there's another timestamp field that already stores creation or modification date. Sometimes we need different timestamps (e.g. the time when a transaction was submitted). And sometimes adding them makes the object unnecessarily heavy - for example why would we need creation and modification date for each row that stores a frame number for the verifier? It's enough to store it for the `VerificationRequest` itself.

So let's keep it as is. If you think there are models that should have the timestamps but don't have them now, tell me and I'll add them to the spec.

rwrzesien:
> for example why would we need creation and modification date for each row that stores a frame number for the verifier?

This is a good example where it is not necessary, however in most cases I think it is good for debugging. I think sooner or later this information will become helpful and it would be better to already have it. As I have said, it is only matter of good practice.

> Sometimes there's another timestamp field that already stores creation or modification date. 

That is one of the reasons of doing this in master class - they get unified and stay consistent. 

> Sometimes we need different timestamps (e.g. the time when a transaction was submitted).

There is no problem with having both.

> So let's keep it as is. If you think there are models that should have the timestamps but don't have them now, tell me and I'll add them to the spec.

Ok, so my proposal is to add them to:
- StoredMessage
- Client
- Subtask
- PendingResponse
- PaymentInfo
- VerificationRequest
- BlenderSubtaskDefinition
- UploadReport

and not add them to:
- Frame

If you disagree with any we can discuss each other separately.

cameel:
> >  Sometimes there's another timestamp field that already stores creation or modification date.
[...]

PR #602, opened 2018-07-05
Feature show creation and modification time in the admin panel

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/553

Please check issue description, because implementation changes from what is originally described there.

### DIFF
--- a/concent_api/core/admin.py
+++ b/concent_api/core/admin.py
@@ -89,6 +89,24 @@ class PendingResponseAdmin(ModelAdminReadOnlyMixin, admin.ModelAdmin):
     get_client_public_key.short_description = 'Client public key'  # type: ignore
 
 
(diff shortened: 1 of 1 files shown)


Jakub89:
Shouldn't there be `created_at` and `modified_at` for `Subtask` model as well?

rwrzesien:
@Jakub89 It is, by inheriting from `BaseModel`. All models which inherit from it will have those fields, and because `abstract = True` is used there, separate fields will be created in each table.

Jakub89:
@rwrzesien Okey, but I don't see it displayed in admin panel.

rwrzesien:
@Jakub89 Oh, you mean that `created_at` and `modified_at` should be visible on Subtasks in admin panel ? Ok, I will add it but after @cameel response in issue because it depends if we are going to do all of them anyway.

PR #603, opened 2018-07-05
Add type annotations to verifier utils

### LINKED ISSUE #583: Add annotations for `blender_verification_order` (verifier) helper functions
functions that (for sure) need annotations:
- `download_archives_from_storage`
- `validate_downloaded_archives`
- `unpack_archives`
- `render_image`
- `delete_source_files`
- `try_to_upload_blender_output_file`
-  `load_images`
- `compare_images`

kbeker:
Resolves #583


concent_api/verifier/utils.py
```diff
@@ -193,7 +196,7 @@ def compare_minimum_ssim_with_results(ssim_list, subtask_id):
         raise VerificationMismatch(subtask_id=subtask_id)
 
 
-def load_images(blender_output_file_name, result_file, subtask_id):
+def load_images(blender_output_file_name: str, result_file: str, subtask_id: str) -> Union[ndarray, ndarray]:
```
rwrzesien: I think `Union` means that you return either one or another value. I think it should be `tuple` in this case.

concent_api/verifier/utils.py
```diff
@@ -313,7 +315,7 @@ def render_image(frame_number, output_format, scene_file, subtask_id, verificati
         )
 
 
-def unpack_archives(file_paths, subtask_id):
+def unpack_archives(file_paths: Iterable[str], subtask_id: str) -> None:
```
rwrzesien: How about changing this into `unpack_archives(subtask_id: str, *args)` where `args` will be list of paths ? What do you guys think ?

concent_api/verifier/utils.py
```diff
@@ -334,7 +336,7 @@ def unpack_archives(file_paths, subtask_id):
             )
 
 
-def validate_downloaded_archives(subtask_id, archives_list):
+def validate_downloaded_archives(subtask_id: str, archives_list: Iterable[str]) -> None:
```
rwrzesien: Same here, how about changing `archives_list` to `*args` ?

concent_api/verifier/utils.py
kbeker: Hmm, I think that it is good to know exactly what parameters function gets, but it would be also good solution. I'm leaving this decision to you guys @rwrzesien @dybi :)

rwrzesien: It is just a proposal, I have done it in the same way in function that validates message signature. But your comment has also a point. So maybe we shouldn't do it yet.

dybi: If we know exactly the number and type of arguments, I would be for leaving as it is ;)

concent_api/verifier/utils.py
dybi: my pov is like above ;)

Issue #593, opened 2018-07-03
Refactor Concent core validators


rwrzesien:
Extension of https://github.com/golemfactory/concent/issues/394.

Validation in Concent `core` module can be improved. Here is the list of things to be done:
- Validators should not raise `http400` error, but either return `True/False` or raise `ValidationError` (or our custom subclass for it with `error_code`) - this one is for discussion first. I prefer raising `ValidationError` as the native Django validators do.
- `validate_int_value` is actually used only once, for validating `deadline` in `ComputeTaskDef`, so it might be either renamed nad refactored or used directly in validation of `ComputeTaskDef`.
- In both `validate_key_with_desired_parameters` and `validate_int_value` there is a type checking, it should be extracted to separate function and reused.
- `validate_report_computed_task_time_window`, `validate_golem_message_subtask_results_rejected` are used only once, and checks view logic, so they can be moved directly to view.
- `validate_golem_message_client_authorization`, `validate_list_task_to_compute_ids` are not used at all, they can be removed.
- `validate_subtask_price_task_to_compute` does the same checks as `validate_int_value`, they should be combined.
- `validate_file_transfer_token` is used in both `core` and `gatekeeper`, so should be in `common` instead of `core`, along with `validate_secure_hash_algorithm`.

dybi:
> Validators should not raise `http400` error, but either return `True/False` or raise `ValidationError` (or our custom subclass for it with `error_code`) - this one is for discussion first. I prefer raising `ValidationError` as the native Django validators do.

I am totaly for raising exceptioin (and totaly against returning `True/False`). Please implement it via raising exception

Jakub89:
>In both `validate_key_with_desired_parameters` and `validate_int_value` there is a type checking, it should be extracted to separate function and reused. 
`validate_subtask_price_task_to_compute` does the same checks as `validate_int_value`, they should be combined.

The problem with those refactors is that `validate_int_value` checks also if value can be converted to int (from `str` for example) where other functions raise exceptions immediately. So I'm not sure if they all can be combined in to one function without some weird case checking?

rwrzesien:
@Jakub89 Right, I have missed that, however it is used only once now - to check `deadline` in `ComputeTaskDef`. So how about combining those two without casting to int, and add only checking if casting to int is possible for `deadline` ?

PR #606, opened 2018-07-05
Chore refactor concent core validators

Jakub89:
Resolves: #593

### DIFF
--- a/concent_api/common/decorators.py
+++ b/concent_api/common/decorators.py
@@ -17,8 +17,10 @@ from golem_messages.exceptions      import MessageTooOldError
 from golem_messages.exceptions      import TimestampError
 
 from common.constants import ErrorCode
+from common.exceptions import ConcentBaseException
 from common.exceptions import ConcentFeatureIsNotAvailable
 from common.exceptions import ConcentInSoftShutdownMode
+from common.exceptions import ConcentValidationError
 from common.helpers import join_messages
 from common import logging
 from common.logging import get_json_from_message_without_redundant_fields_for_logging
@@ -27,10 +29,6 @@ from common.logging import log_json_message
 from common.logging import log_message_received_in_endpoint
 from common.logging import log_string_message
 from common.shortcuts import load_without_public_key
-from core.exceptions import FileTransferTokenError
-from core.exceptions import GolemMessageValidationError
-from core.exceptions import HashingAlgorithmError
-from core.exceptions import Http400
 from core.validation import get_validated_client_public_key_from_client_message
 from core.validation import is_golem_message_signed_with_key
 
@@ -127,7 +125,7 @@ def require_golem_message(view):
                     golem_message.task_id if 'task_id' in dir(golem_message) else None,
                     golem_message.subtask_id if 'subtask_id' in dir(golem_message) else None
                 )
-            except Http400 as exception:
+            except ConcentValidationError as exception:
                 log_string_message(logger, f"error_code: {exception.error_code.value} error: {exception.error_message} ")
                 return JsonResponse(
                     {
@@ -174,7 +172,7 @@ def handle_errors_and_responses(database_name):
                 response_from_view = view(request, client_message, client_public_key, *args, **kwargs)
                 if database_name is not None:
                     transaction.savepoint_commit(sid, using=database_name)
-            except (Http400, FileTransferTokenError, HashingAlgorithmError, GolemMessageValidationError) as exception:
+            except ConcentBaseException as exception:
                 log_400_error(
                     logger,
                     view.__name__,
--- a/concent_api/common/exceptions.py
+++ b/concent_api/common/exceptions.py
@@ -20,3 +20,7 @@ class ConcentBaseException(Exception):
         self.error_code = error_code
         self.error_message = '' if error_message is None else error_message
         super().__init__(error_message)
+
+
+class ConcentValidationError(ConcentBaseException):
+    pass
--- a/concent_api/common/validations.py
+++ b/concent_api/common/validations.py
@@ -1,0 +1,112 @@
+from django.conf import settings
+from django.core.exceptions import ValidationError
+from django.core.validators import URLValidator
+
+from golem_messages import message
+from golem_messages.message import FileTransferToken
+
+from common.constants import ErrorCode
+from core.constants import VALID_SHA1_HASH_REGEX
+from core.enums import HashingAlgorithm
+from core.exceptions import FileTransferTokenError
+from core.exceptions import HashingAlgorithmError
+
+
+def validate_file_transfer_token(file_transfer_token: message.concents.FileTransferToken):
+    """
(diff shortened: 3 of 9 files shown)


concent_api/common/decorators.py
```diff
@@ -127,7 +128,7 @@ def wrapper(request, *args, **kwargs):
                     golem_message.task_id if 'task_id' in dir(golem_message) else None,
                     golem_message.subtask_id if 'subtask_id' in dir(golem_message) else None
                 )
-            except Http400 as exception:
+            except (Http400, ConcentValidationError) as exception:
```
rwrzesien: Is it still possible to catch `Http400` here ?

concent_api/common/exceptions.py
```diff
@@ -20,3 +20,7 @@ def __init__(self, error_message: Optional[str], error_code: ErrorCode) -> None:
         self.error_code = error_code
         self.error_message = '' if error_message is None else error_message
         super().__init__(error_message)
+
+
+class ConcentValidationError(ConcentBaseException):
```
rwrzesien: Please unify the name and location with @dybi as he is doing something similar in https://github.com/golemfactory/concent/pull/608.

concent_api/common/validations.py
```diff
@@ -0,0 +1,112 @@
...
+from django.core.validators import URLValidator
+
+from golem_messages import message
+from golem_messages.message import FileTransferToken
+
+from common.constants import ErrorCode
+from core.constants import VALID_SHA1_HASH_REGEX
+from core.enums import HashingAlgorithm
+from core.exceptions import HashingAlgorithmError
+from core.exceptions import FileTransferTokenError
```
rwrzesien: `F` before `H` :)

concent_api/core/validation.py
```diff
@@ -39,38 +32,34 @@ def validate_int_value(value):
...
-            "Wrong type, expected non-negative integer but negative integer provided.",
-            error_code=ErrorCode.MESSAGE_VALUE_WRONG_TYPE,
-        )
+    validate_positive_integer_value(value)
 
 
 def validate_id_value(value, field_name):
     if not isinstance(value, str):
-        raise Http400(
+        raise ConcentValidationError(
```
rwrzesien: You can replace this with `validate_expected_value_type`.

concent_api/core/validation.py
```diff
@@ -116,43 +100,33 @@ def validate_task_to_compute(task_to_compute: message.TaskToCompute):
...
 
     validate_hex_public_key(task_to_compute.provider_public_key, 'provider_public_key')
     validate_hex_public_key(task_to_compute.requestor_public_key, 'requestor_public_key')
     validate_secure_hash_algorithm(task_to_compute.package_hash)
-    validate_subtask_price_task_to_compute(task_to_compute)
+    validate_positive_integer_value(task_to_compute.price)
     validate_frames(task_to_compute.compute_task_def['extra_data']['frames'])
 
 
 def validate_report_computed_task_time_window(report_computed_task):
```
rwrzesien: This whole check is part of logic view, and used only there, it should be moved there (the inside of the function), and raise directly `http400` in view.

concent_api/core/validation.py
```diff
@@ -208,59 +182,39 @@ def is_golem_message_signed_with_key(
...
-        raise Http400(
-            "Price cannot be a negative value",
-            error_code=ErrorCode.MESSAGE_VALUE_NEGATIVE,
-        )
-
-
 def validate_ethereum_addresses(requestor_ethereum_address, provider_ethereum_address):
     if not isinstance(requestor_ethereum_address, str):
-        raise Http400(
+        raise ConcentValidationError(
```
rwrzesien: Those 4 checks could be paired and checked in `validate_key_with_desired_parameters`.

concent_api/common/decorators.py
```diff
@@ -174,7 +175,9 @@ def wrapper(request, client_message, client_public_key, *args, **kwargs):
                 response_from_view = view(request, client_message, client_public_key, *args, **kwargs)
                 if database_name is not None:
                     transaction.savepoint_commit(sid, using=database_name)
-            except (Http400, FileTransferTokenError, HashingAlgorithmError, GolemMessageValidationError) as exception:
+            except (
+                Http400, FileTransferTokenError, HashingAlgorithmError, GolemMessageValidationError, ConcentValidationError
```
rwrzesien: We need to make decision if we allow `ConcentValidationError` to be caught here instead of view first. @dybi @cameel What do you guys think ? I have added same question but more detailed in https://github.com/golemfactory/concent/pull/608#pullrequestreview-134832811 .

concent_api/core/validation.py
```diff
@@ -1,35 +1,28 @@
...
+from common.exceptions import ConcentValidationError
 from common.logging import log_error_message
-from core.constants                 import ETHEREUM_ADDRESS_LENGTH
-from core.constants                 import GOLEM_PUBLIC_KEY_LENGTH
-from core.constants                 import GOLEM_PUBLIC_KEY_HEX_LENGTH
-from core.constants                 import MESSAGE_TASK_ID_MAX_LENGTH
+from common.validations import validate_secure_hash_algorithm
+from core.constants import ETHEREUM_ADDRESS_LENGTH
+from core.constants import GOLEM_PUBLIC_KEY_LENGTH
+from core.constants import GOLEM_PUBLIC_KEY_HEX_LENGTH
```
kbeker: Alphabetic order :)

concent_api/common/decorators.py
Jakub89: No it's not, You'r right.

concent_api/common/exceptions.py
Jakub89: Ok, we agree on leaving this one.

concent_api/common/validations.py
Jakub89: Done

concent_api/core/validation.py

concent_api/core/validation.py
dybi: Well, quite frankly it is. But on the other hand, it is good that it was extracted to a function. Maybe we need a separate module for business logic validation and have all such functions stored there?

concent_api/common/decorators.py
dybi: I think we can even catch `ConcentBaseException` - it will be more succinct

concent_api/common/exceptions.py
dybi: Please move to `core.exceptions` if you are using it only in `core` module

Jakub89: Already did it ;)

concent_api/common/decorators.py
cameel: Depends on what inherits from `ConcentBaseException`. If there are any errors that indicate failures, they are not supposed to get a HTTP 4xx response.

Issue #600, opened 2018-07-04
Validate ComputeTaskDef before starting validation


rwrzesien:
Currently we are getting a several sentry reports related to exceptions with missing fields in `ComputeTaskDef` or its inner containers. 

Related sentry reports:
https://talkback.golem.network/sentry/concent/issues/27092/
https://talkback.golem.network/sentry/concent/issues/27144/

This should be either validated in `validate_task_to_compute`, and in case follow up current convention.

dybi:
@rwrzesien, @cameel, I would be for addind validation of `ComputeTaskDef` within `validate_task_to_compute` - this way it would be checked for basically every message

PaweuB:
Pivotal link:
https://www.pivotaltracker.com/story/show/158832069

cameel:
Yeah. The whole `TaskToCompute` should be validated by Concent Core immediately after it receives a message that starts a use case. It's the same situation as in #382.

And we should be doing it at the beginning of every use case (not only in verification).

PR #608, opened 2018-07-05
Add validation of ComputeTaskDef in for all Golem message handlers

dybi:
- New function `validate_compute_task_def` has been added; it is called within `validate_task_to_compute`.
- Unit tests have been added
- Existing tests have been adapted.

resolves: https://github.com/golemfactory/concent/issues/600

### DIFF
--- a/concent_api/common/tests/test_api_view.py
+++ b/concent_api/common/tests/test_api_view.py
@@ -222,6 +222,8 @@ class ApiViewTransactionTestCase(TransactionTestCase):
         compute_task_def['deadline'] = message_timestamp
         compute_task_def['extra_data'] = {
             'frames': [1],
+            'output_format': 'PNG',
+            'scene_file': 'kitten.blend',
         }
         task_to_compute = tasks.TaskToComputeFactory(
             compute_task_def=compute_task_def,
--- a/concent_api/core/tests/test_core_views.py
+++ b/concent_api/core/tests/test_core_views.py
@@ -73,6 +73,8 @@ class CoreViewSendTest(ConcentIntegrationTestCase):
         compute_task_def['deadline'] = self.message_timestamp + 600
         compute_task_def['extra_data'] = {
             'frames': [1],
+            'output_format': 'PNG',
+            'scene_file': 'kitten.blend',
         }
         task_to_compute = self._get_deserialized_task_to_compute(
             compute_task_def = compute_task_def
@@ -469,6 +471,8 @@ class CoreViewSendTest(ConcentIntegrationTestCase):
             compute_task_def['deadline']    = deadline
             compute_task_def['extra_data'] = {
                 'frames': [1],
+                'output_format': 'PNG',
+                'scene_file': 'kitten.blend',
             }
 
             deserialized_task_to_compute = self._get_deserialized_task_to_compute(
--- a/concent_api/core/tests/test_unit_validation.py
+++ b/concent_api/core/tests/test_unit_validation.py
@@ -1,8 +1,11 @@
 from unittest import TestCase
-
+from assertpy import assert_that
 import mock
+import pytest
+
 from django.test import override_settings
 
+from golem_messages.factories.tasks import ComputeTaskDefFactory
 from golem_messages.factories.tasks import TaskToComputeFactory
 from golem_messages.message.tasks import ReportComputedTask
 from golem_messages.message.tasks import SubtaskResultsRejected
@@ -20,6 +23,7 @@ from core.tests.utils import ConcentIntegrationTestCase
 from core.exceptions import FrameNumberValidationError
 from core.exceptions import HashingAlgorithmError
 from core.validation import validate_all_messages_identical
+from core.validation import validate_compute_task_def
 from core.validation import validate_ethereum_addresses
 from core.validation import validate_golem_message_subtask_results_rejected
 from core.validation import validate_id_value
@@ -262,3 +266,56 @@ class TestFramesListValidaation(TestCase):
     def test_that_if_frames_are_not_integers_method_should_raise_exception(self):
         with self.assertRaises(FrameNumberValidationError):
             validate_frames(['1', '2'])
+
+
+class TestValidateComputeTaskDef(object):
+    compute_task_def = None
+
+    @pytest.fixture(autouse=True)
+    def setup(self):
+        self.compute_task_def = ComputeTaskDefFactory()
+        self.compute_task_def["extra_data"] = {
+            "output_format": "PNG",
+            "scene_file": "nice_photo.blend",
+            "frames": [1, 2, 3],
+        }
+
+    def test_that_valid_compute_task_def_doesnt_raise_any_exception(self):
+        try:
+            validate_compute_task_def(self.compute_task_def)
+        except Exception as exception:  # pylint: disable=broad-except
+            assert False, f"Unexpected exception has been raised: {str(exception)}"
+
+    def test_that_mising_extra_data_causes_message_validation_error(self):
+        del self.compute_task_def['extra_data']
+        with pytest.raises(ConcentValidationError) as exception_wrapper:
+            validate_compute_task_def(self.compute_task_def)
+        assert_that(exception_wrapper.value.error_message).contains(f"extra_data")
+        assert_that(exception_wrapper.value.error_code).is_equal_to(ErrorCode.MESSAGE_INVALID)
+
+    @pytest.mark.parametrize(
+        "missing_data", [
+            "output_format",
+            "scene_file",
+            "frames",
+        ]
+    )
+    def test_that_missing_entries_in_extra_data_causes_message_validation_error(self, missing_data):
+        del self.compute_task_def["extra_data"][missing_data]
+        with pytest.raises(ConcentValidationError) as exception_wrapper:
+            validate_compute_task_def(self.compute_task_def)
+        assert_that(exception_wrapper.value.error_message).contains(f"{missing_data}")
+        assert_that(exception_wrapper.value.error_code).is_equal_to(ErrorCode.MESSAGE_INVALID)
+
+    @pytest.mark.parametrize(
+        "value_with_wrong_type", [
+            "output_format",
+            "scene_file",
+        ]
+    )
+    def test_that_wrong_field_types_causes_message_validation_error(self, value_with_wrong_type):
+        self.compute_task_def["extra_data"][value_with_wrong_type] = mock.sentinel.wrongtype
+        with pytest.raises(ConcentValidationError) as exception_wrapper:
+            validate_compute_task_def(self.compute_task_def)
+        assert_that(exception_wrapper.value.error_message).contains(f"{value_with_wrong_type}")
+        assert_that(exception_wrapper.value.error_code).is_equal_to(ErrorCode.MESSAGE_VALUE_NOT_STRING)
--- a/concent_api/core/validation.py
+++ b/concent_api/core/validation.py
@@ -101,16 +101,12 @@ def validate_task_to_compute(task_to_compute: message.TaskToCompute):
             error_code=ErrorCode.MESSAGE_WRONG_FIELDS,
         )
 
-    validate_value_is_int_convertible_and_positive(task_to_compute.compute_task_def['deadline'])
-
-    validate_id_value(task_to_compute.compute_task_def['task_id'], 'task_id')
-    validate_id_value(task_to_compute.compute_task_def['subtask_id'], 'subtask_id')
+    validate_compute_task_def(task_to_compute.compute_task_def)
 
     validate_hex_public_key(task_to_compute.provider_public_key, 'provider_public_key')
     validate_hex_public_key(task_to_compute.requestor_public_key, 'requestor_public_key')
     validate_secure_hash_algorithm(task_to_compute.package_hash)
     validate_positive_integer_value(task_to_compute.price)
-    validate_frames(task_to_compute.compute_task_def['extra_data']['frames'])
 
 
 def validate_report_computed_task_time_window(report_computed_task):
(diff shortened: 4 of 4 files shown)


concent_api/common/tests/test_api_view.py
```diff
@@ -222,6 +222,8 @@ def setUp(self):
         compute_task_def['deadline'] = message_timestamp
         compute_task_def['extra_data'] = {
             'frames': [1],
+            'output_format': "PNG",
```
rwrzesien: Why mixing two different types of quotes ? Please choose one and keep it consistent. I think we use more of `''` in project than `""`.

concent_api/core/tests/test_unit_validation.py
```diff
@@ -1,9 +1,11 @@
 from unittest import TestCase
-
+from assertpy import assert_that
 import mock
+import pytest
+
 from django.test import override_settings
 
-from golem_messages.factories.tasks import TaskToComputeFactory
+from golem_messages.factories.tasks import TaskToComputeFactory, ComputeTaskDefFactory
```
rwrzesien: Imports should be split into separate lines.

concent_api/core/tests/test_unit_validation.py
```diff
@@ -18,7 +20,8 @@
 from core.exceptions import FrameNumberValidationError
 from core.exceptions import HashingAlgorithmError
 from core.exceptions import Http400
-from core.validation import validate_all_messages_identical
+from core.exceptions import MessageValidationError
+from core.validation import validate_all_messages_identical, validate_compute_task_def
```
rwrzesien: Same here, :+1: for making them alphabetical.

concent_api/core/tests/test_unit_validation.py
```diff
@@ -268,3 +271,50 @@ def test_that_if_frames_are_not_one_after_the_other_method_should_pass(self):
     def test_that_if_frames_are_not_integers_method_should_raise_exception(self):
         with self.assertRaises(FrameNumberValidationError):
             validate_frames(['1', '2'])
+
+
+class TestValidateComputeTaskDef(object):
+    compute_task_def = None
```
rwrzesien: I think there is no need to define it here.

concent_api/core/tests/test_unit_validation.py
```diff
@@ -268,3 +271,50 @@ def test_that_if_frames_are_not_one_after_the_other_method_should_pass(self):
     def test_that_if_frames_are_not_integers_method_should_raise_exception(self):
         with self.assertRaises(FrameNumberValidationError):
             validate_frames(['1', '2'])
+
+
+class TestValidateComputeTaskDef(object):
```
rwrzesien: In python3 inheriting from `object` is not needed.

concent_api/core/exceptions.py
```diff
@@ -23,3 +23,7 @@ class GolemMessageValidationError(ConcentBaseException):
 
 class FrameNumberValidationError(ConcentBaseException):
     pass
+
+
+class MessageValidationError(ConcentBaseException):
```
rwrzesien: How about reusing `ValidationError` from Django ?

concent_api/core/tests/test_unit_validation.py
```diff
@@ -268,3 +271,50 @@ def test_that_if_frames_are_not_one_after_the_other_method_should_pass(self):
...
+        assert_that(e.value.error_code).is_equal_to(ErrorCode.MESSAGE_INVALID)
+
+    @pytest.mark.parametrize(
+        "value_with_wrong_type", [
+            "output_format",
+            "scene_file",
+        ]
+    )
+    def test_that_wrong_field_types_causes_message_validation_error(self, value_with_wrong_type):
+        self.compute_task_def["extra_data"][value_with_wrong_type] = mock.sentinel.wrongtype
```
rwrzesien: Could you explain what `mock.sentinel.wrongtype` does ? I was trying to find some docs for it over the web but I couldn't. Looks like a cool feature.

concent_api/core/validation.py
```diff
@@ -18,7 +18,7 @@
 from core.constants import VALID_ID_REGEX
 from core.constants import VALID_SHA1_HASH_REGEX
 from core.enums import HashingAlgorithm
-from core.exceptions import FileTransferTokenError
+from core.exceptions import FileTransferTokenError, MessageValidationError
```
rwrzesien: Same here.

concent_api/core/validation.py
```diff
@@ -434,3 +430,33 @@ def validate_frames(frames_list: List[int]):
...
+    string_fields = ["output_format", "scene_file"]
+    integer_fields = ["frames"]
+
+    validate_int_value(compute_task_def['deadline'])
+
+    validate_id_value(compute_task_def['task_id'], 'task_id')
+    validate_id_value(compute_task_def['subtask_id'], 'subtask_id')
+
+    extra_data = compute_task_def.get("extra_data")
+    if extra_data is None:
```
rwrzesien: I think it would be more readable if you put empty line before each `if` and `for`.

concent_api/core/validation.py
```diff
@@ -434,3 +430,33 @@ def validate_frames(frames_list: List[int]):
...
+            "'extra_data' is missing in ComputeTaskDef",
+            ErrorCode.MESSAGE_INVALID
+        )
+    for mandatory_data in string_fields + integer_fields:
+        if mandatory_data not in extra_data:
+            raise MessageValidationError(
+                f"{mandatory_data} is missing in ComputeTaskDef",
+                ErrorCode.MESSAGE_INVALID
+            )
+    validate_frames(extra_data["frames"])
```
rwrzesien: Here it should have separate line before and after this call.

concent_api/core/validation.py
```diff
@@ -434,3 +430,33 @@ def validate_frames(frames_list: List[int]):
...
+    validate_id_value(compute_task_def['task_id'], 'task_id')
+    validate_id_value(compute_task_def['subtask_id'], 'subtask_id')
+
+    extra_data = compute_task_def.get("extra_data")
+    if extra_data is None:
+        raise MessageValidationError(
+            "'extra_data' is missing in ComputeTaskDef",
+            ErrorCode.MESSAGE_INVALID
+        )
+    for mandatory_data in string_fields + integer_fields:
```
rwrzesien: Please add empty line between those two for readability.

concent_api/core/validation.py
```diff
@@ -434,3 +430,33 @@ def validate_frames(frames_list: List[int]):
                 'Frame number must be grater than 0',
                 ErrorCode.MESSAGE_FRAME_VALUE_NOT_POSITIVE_INTEGER
             )
+
+
+def validate_compute_task_def(compute_task_def: message.tasks.ComputeTaskDef) -> None:
+    string_fields = ["output_format", "scene_file"]
+    integer_fields = ["frames"]
```
rwrzesien: I think this variable name is misleading in comparison to previous one. In `string_fields` you have list of fields which values should be strings, so it suggest that with `integer_fields` you have a list of fields which values should be integers. But it should be a list of integers. So maybe call it `list_of_integer_fields` or simply `other_mandatory_fields` because later you don't check a type of values from those fields.

concent_api/core/validation.py
```diff
@@ -434,3 +430,33 @@ def validate_frames(frames_list: List[int]):
...
+
+    validate_int_value(compute_task_def['deadline'])
+
+    validate_id_value(compute_task_def['task_id'], 'task_id')
+    validate_id_value(compute_task_def['subtask_id'], 'subtask_id')
+
+    extra_data = compute_task_def.get("extra_data")
+    if extra_data is None:
+        raise MessageValidationError(
```
rwrzesien: This should be also handled somehow in the views or decorators. With this version it will just fail with exception, so it would be the same effect as we have now. 

I am thinking if it would be enough to add catching it in `handle_errors_and_responses` along with other `ConcentBaseException` subclasses. This way it will be still possible to handle it differently in view when needed in particular case. We should make the decision here because we want to change our validation soon and keep it consistent. @cameel What do you think ? It would be ok to catch `ValidationError` (os similar) in decorator or should it be caught in view first ?

Btw. @Jakub89 is doing something similar in https://github.com/golemfactory/concent/pull/606, so you guys should unify at least the exception name.

concent_api/core/tests/test_unit_validation.py
dybi: I know, but it's not a mistake either

concent_api/core/tests/test_unit_validation.py
```diff
@@ -268,3 +271,50 @@ def test_that_if_frames_are_not_one_after_the_other_method_should_pass(self):
...
+        self.compute_task_def = ComputeTaskDefFactory()
+        self.compute_task_def["extra_data"] = {
+            "output_format": "PNG",
+            "scene_file": "nice_photo.blend",
+            "frames": [1, 2, 3],
+        }
+
+    def test_that_mising_extra_data_causes_message_validation_error(self):
+        del self.compute_task_def['extra_data']
+        with pytest.raises(MessageValidationError) as e:
```
kbeker: As in another files we don't use shortcuts, please use `exception` :)

concent_api/core/tests/test_unit_validation.py
```diff
@@ -268,3 +271,50 @@ def test_that_if_frames_are_not_one_after_the_other_method_should_pass(self):
...
+    @pytest.mark.parametrize(
+        "missing_data", [
+            "output_format",
+            "scene_file",
+            "frames",
+        ]
+    )
+    def test_that_missing_entries_in_extra_data_causes_message_validation_error(self, missing_data):
+        del self.compute_task_def["extra_data"][missing_data]
+        with pytest.raises(MessageValidationError) as e:
```
kbeker: Same here

concent_api/core/tests/test_unit_validation.py
```diff
@@ -268,3 +271,50 @@ def test_that_if_frames_are_not_one_after_the_other_method_should_pass(self):
...
+
+    @pytest.mark.parametrize(
+        "value_with_wrong_type", [
+            "output_format",
+            "scene_file",
+        ]
+    )
+    def test_that_wrong_field_types_causes_message_validation_error(self, value_with_wrong_type):
+        self.compute_task_def["extra_data"][value_with_wrong_type] = mock.sentinel.wrongtype
+        with pytest.raises(MessageValidationError) as e:
```
kbeker: And here

concent_api/common/tests/test_api_view.py
dybi: done

concent_api/core/tests/test_unit_validation.py
dybi: done

concent_api/core/tests/test_unit_validation.py
dybi: done

concent_api/core/validation.py
dybi: done

concent_api/core/validation.py
dybi: done

concent_api/core/validation.py
dybi: done

concent_api/core/validation.py
dybi: good catch; done

concent_api/core/validation.py
dybi: Good point, I will change the decorator

dybi: done

concent_api/core/tests/test_unit_validation.py
dybi: done

concent_api/core/tests/test_unit_validation.py
dybi: done

concent_api/core/tests/test_unit_validation.py
dybi: done

concent_api/core/tests/test_unit_validation.py
dybi: it is ;)
either this or:
`W:283, 8: Attribute 'compute_task_def' defined outside __init__ (attribute-defined-outside-init)`

concent_api/core/exceptions.py
dybi: Primarily, `ValidationError` from Django is for forms and models' fields validation. We might use it but we would need to:
1. handle it in `handle_errors_and_responses` decorator
2. make sure, that catching it in the above mentioned decorator would not have any side-effects (i.e. catching model validation errors and returning it as bad request)

All in all, I would be for separate exception, specific to client's message validation error.

concent_api/core/tests/test_unit_validation.py
dybi: https://docs.python.org/3/library/unittest.mock.html#unittest.mock.sentinel
Sentinels are dummy objects (~objects with no value).
Basically, I would say you can use them in two scenarios:
1. For testing that if some function has been feed with certain object, the exact object is returned. For example:
```python
d = { "a": 1, "b": 2}
def get_ value_from_alphabetically_lowest_key(d):
    ....
    return d[key]

assert id(d["a"]) ==  id(value_from_alphabetically_lowest_key(d))
```
2. For testing with non-existing value (my use case) - when we want to check how something behaves when wrong value is given

concent_api/core/tests/test_core_views.py
```diff
@@ -73,6 +73,8 @@ def setUp(self):
         compute_task_def['deadline'] = self.message_timestamp + 600
         compute_task_def['extra_data'] = {
             'frames': [1],
+            'output_format': "PNG",
```
rwrzesien: Same here.

concent_api/core/tests/test_core_views.py
```diff
@@ -469,6 +471,8 @@ def test_send_should_return_http_202_if_task_to_compute_deadline_is_correct(self
             compute_task_def['deadline']    = deadline
             compute_task_def['extra_data'] = {
                 'frames': [1],
+                'output_format': "PNG",
```
rwrzesien: And here.

concent_api/core/tests/test_unit_validation.py
rwrzesien: It is unnecessary code, so I would be for removing it.

concent_api/core/tests/test_unit_validation.py
rwrzesien: Right, because you don't inherit from `TestCase`. It is ok then.

Issue #582, opened 2018-06-29
Add unit test for `blender_verification_order` (verifier) helper functions 


dybi:
functions that (for sure) need unit tests:
- `download_archives_from_storage`
- `validate_downloaded_archives`
- `unpack_archives`
- `render_image`
- `delete_source_files`
- `try_to_upload_blender_output_file`
-  `load_images`
- `compare_images`

kbeker:
Issue updated with new functions to test

PaweuB:
1) Pivotal link: https://www.pivotaltracker.com/story/show/158823515
2) Added subtasks in Pivotal Story to match progress while interrupting this resolve. 
3) Issue stopped in resolving due to high priority stories before mainnet.

PR #612, opened 2018-07-06
Add unit tests to method in verifier.utils

kbeker:
Method tested: 
1. compare_images 
2. generate_verifier_storage_file_path 
3. generate_upload_file_path
4. generate_base_blender_output_file_name,
5. generate_full_blender_output_file_name
6. adjust_format_name

This is part of https://github.com/golemfactory/concent/issues/582 issue

### DIFF
--- a/concent_api/core/validation.py
+++ b/concent_api/core/validation.py
@@ -293,6 +293,14 @@ def validate_positive_integer_value(value):
         )
 
 
+def validate_scene_file(scene_file):
+    if not scene_file.endswith('.blend'):
+        raise ConcentValidationError(
+            f'{scene_file} must ends with ".blend" filename extension',
+            ErrorCode.MESSAGE_INVALID
+        )
+
+
 def validate_compute_task_def(compute_task_def: message.tasks.ComputeTaskDef) -> None:
     string_fields = ["output_format", "scene_file"]
     other_mandatory_fields = ["frames"]
@@ -324,3 +332,5 @@ def validate_compute_task_def(compute_task_def: message.tasks.ComputeTaskDef) ->
                 f"{string_field} should be string",
                 ErrorCode.MESSAGE_VALUE_NOT_STRING
             )
+
+    validate_scene_file(extra_data['scene_file'])
--- a/concent_api/verifier/tests/test_utils.py
+++ b/concent_api/verifier/tests/test_utils.py
@@ -1,22 +1,34 @@
 from unittest import TestCase
 
+from assertpy import assert_that
+import pytest
 from mock import mock
+from django.conf import settings
 from django.test import override_settings
+from numpy import ones
+from numpy import zeros
 from numpy.core.records import ndarray
 
 from core.constants import VerificationResult
+from verifier.exceptions import VerificationError
 from verifier.exceptions import VerificationMismatch
+from verifier.utils import adjust_format_name
 from verifier.utils import are_image_sizes_and_color_channels_equal
 from verifier.utils import compare_all_rendered_images_with_user_results_files
+from verifier.utils import compare_images
 from verifier.utils import compare_minimum_ssim_with_results
+from verifier.utils import ensure_enough_result_files_provided
+from verifier.utils import ensure_frames_have_related_files_to_compare
+from verifier.utils import generate_base_blender_output_file_name
+from verifier.utils import generate_full_blender_output_file_name
+from verifier.utils import generate_upload_file_path
+from verifier.utils import generate_verifier_storage_file_path
 from verifier.utils import render_images_by_frames
 from verifier.utils import parse_result_files_with_frames
 from verifier.utils import upload_blender_output_file
-from verifier.utils import ensure_enough_result_files_provided
-from verifier.utils import ensure_frames_have_related_files_to_compare
 
 
-class VerifierVerificationIntegrationTest(TestCase):
+class VerifierUtilsTest(TestCase):
 
     def setUp(self):
         super().setUp()
@@ -55,6 +67,9 @@ class VerifierVerificationIntegrationTest(TestCase):
             (self.image, self.image),
             (self.image, self.image),
         ]
+        self.image_ones = ones(shape=(1920, 1080, 3), dtype='uint8')
+        self.image_zeros = zeros(shape=(1920, 1080, 3), dtype='uint8')
+        self.image_diffrent_size = zeros(shape=(1080, 1920, 3), dtype='uint8')
 
     def test_that_are_image_sizes_and_color_channels_equal_should_return_false_if_sizes_in_pixels_are_not_equal(self):
         result = are_image_sizes_and_color_channels_equal(self.image, self.image_diffrent_size)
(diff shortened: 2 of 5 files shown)


concent_api/verifier/utils.py
```diff
@@ -145,10 +145,14 @@ def delete_file(file_path: str) -> None:
 
 def generate_full_blender_output_file_name(scene_file: str, frame_number: int, output_format: str) -> str:
     base_blender_output_file_name = generate_base_blender_output_file_name(scene_file)
-    return f'{base_blender_output_file_name}{frame_number:>04}.{output_format}'
+    return f'{base_blender_output_file_name}{frame_number:>04}.{output_format.lower()}'
```
rwrzesien: Currently `.lower()` is also called in the only place where this function is used (`render_images_by_frames` function`), so please remove it from there - it is now called twice, and off course it is better when it is called inside `generate_full_blender_output_file_name`.

concent_api/verifier/utils.py
```diff
@@ -145,10 +145,14 @@ def delete_file(file_path: str) -> None:
...
 
 
 def generate_base_blender_output_file_name(scene_file: str) -> str:
+    if scene_file.startswith(('\\', '/')):
```
rwrzesien: When can this possibly happen ?

Wouldn't it be better to validate if `compute_task_def['extra_data']['scene_file']` is correct file name in `validate_compute_task_def` function ?

Also, I have noticed now that we don't check if `scene_file` exists in `VERIFIER_STORAGE_PATH` after unpacking archives. Shouldn't we do this and if it is not instantly end verification with error ? @cameel @dybi what do you guys think ?

concent_api/verifier/utils.py
```diff
@@ -157,6 +161,8 @@ def generate_upload_file_path(subtask_id: str, extension: str, frame_number: int
 
 
 def generate_verifier_storage_file_path(file_name: str) -> str:
+    if file_name.startswith(('\\', '/')):
```
rwrzesien: Same here, when can this happen ?

concent_api/verifier/tests/test_utils.py
```diff
@@ -199,3 +214,94 @@ def test_that_method_should_accept_ssim_list_and_delay_verification_result_with_
                 self.subtask_id,
                 VerificationResult.MATCH.name,
             )
+
+    def test_that_method_returns_ssim_equal_one(self):
```
dybi: maybe `test_that_for_the_same_images_metod_produces_ssim_equal_one`?

concent_api/verifier/tests/test_utils.py
```diff
@@ -199,3 +214,94 @@ def test_that_method_should_accept_ssim_list_and_delay_verification_result_with_
...
+        ssim = compare_images(self.image_ones, self.image_ones, 'subtask_id')
+        self.assertEqual(ssim, 1.0)
+
+    @override_settings(
+        VERIFIER_MIN_SSIM=0.95
+    )
+    def test_that_method_returns_ssim_below_min_verifier_ssim_level(self):
```
dybi: `test_that_for_different_images_calculated_ssim_is_below_min`?

concent_api/verifier/tests/test_utils.py
```diff
@@ -199,3 +214,94 @@ def test_that_method_should_accept_ssim_list_and_delay_verification_result_with_
...
+        ('tmp', '/test_file.png', 'tmp/test_file.png'),
+    ])  # pylint: disable=no-self-use
+    def test_that_method_returns_correct_verifier_storage_file_path(self, storage_path, file_name, expected):
+        with override_settings(VERIFIER_STORAGE_PATH=storage_path):
+            verifier_storage_file_path = generate_verifier_storage_file_path(file_name=file_name)
+
+            assert_that(verifier_storage_file_path).is_equal_to(expected)
+
+    @pytest.mark.parametrize(('subtask_id', 'extension', 'frame_number', 'expected'), [
+        ('subtask_id', 'PNG', '22', 'blender/verifier-output/subtask_id/subtask_id_0022.png'),
```
dybi: actually, frame number shouldn't be string, should it?

concent_api/verifier/tests/test_utils.py
```diff
@@ -199,3 +214,94 @@ def test_that_method_should_accept_ssim_list_and_delay_verification_result_with_
...
+    def test_that_method_returns_ssim_below_min_verifier_ssim_level(self):
+        ssim = compare_images(self.image_ones, self.image_zeros, 'subtask_id')
+        self.assertTrue(ssim < settings.VERIFIER_MIN_SSIM)
+
+    def test_that_method_raise_verification_error_when_images_have_diffrent_sizes(self):
+        with self.assertRaises(VerificationError):
+            compare_images(self.image_ones, self.image_diffrent_size, 'subtask_id')
+
+
+class TestGenerateFilePathMethods():
```
dybi: big :+1: for parametric tests :)

concent_api/verifier/utils.py
dybi: 1. It might be a good idea, provided we know what the correct name is ;)

2. I agree, we shold stop execution as soon as possible,

dybi: One more thing - in case there is no `scene_file` I think we should return `VerificationResult.MISMATCH`, because it would mean provider has intentionally uploaded bad archive

concent_api/verifier/utils.py
kbeker: Done

concent_api/verifier/utils.py
kbeker: So I will do it in this PR as a part of tests of this functions.

kbeker: Done by @dybi here -> https://github.com/golemfactory/concent/pull/613#pullrequestreview-135335272

kbeker: 1. Done

concent_api/verifier/utils.py
kbeker: Moved to `compute_task_def` validation

concent_api/verifier/tests/test_utils.py
kbeker: Done

concent_api/verifier/tests/test_utils.py
kbeker: Done

concent_api/verifier/tests/test_utils.py
kbeker: It shouldn't and I can delete it but this test proves that even frame which some how will be string it will still be valid to this method :)

dybi: please remove it - tests are also a low level specs for source code. Having string passed to this function could be missleading about which argument this function normally expects...

concent_api/core/validation.py
```diff
@@ -293,6 +293,21 @@ def validate_positive_integer_value(value):
         )
 
 
+def validate_scene_file(scene_file):
+    if not scene_file.startswith(('\\', '/')):
```
dybi: I would not assume this...
@cameel ?

concent_api/verifier/tests/test_utils.py
dybi: please remove :)

concent_api/verifier/utils.py
rwrzesien: @kbeker Ok, I was just curious when it can possibly happen for the `scene_file` to have bad name ?

cameel: @rwrzesien 
> Ok, I was just curious when it can possibly happen for the `scene_file` to have bad name ?

We should use the same rules as Golem here. If Golem is able to find the scene file then we should too. If not, it's fine to treat it as a `MISMATCH` even if the provider just has a mistake in the name - he should fix his client.

concent_api/core/validation.py
cameel: I do not know if this is a valid assumption or not. We would have to find this in Golem's source to be sure. Maybe ask on Rocket?

concent_api/verifier/utils.py
kbeker: @rwrzesien after consulting with @dybi we decided to delete it because someone can just upload file from current directory and then we won't have `slash` or `backslash`

Issue #610, opened 2018-07-06
Incomplete validation of zip files in in verification


dybi:
There is a missing check - downloaded archives are not check if they are valid `ZipFile`. 
In case zip file is bad, `VerificationResult.MISMATCH` should be scheduled, as provider must have intentionally uploaded a bad zip file (checksum of uploaded file is valid at this point).

PaweuB:
Added to Pivotal:
https://www.pivotaltracker.com/story/show/158903050

PR #613, opened 2018-07-09
Add additional archives validation in verifier

dybi:
- if download package is not a valid .zip file, VerificationMismatch is raised
- if there is no scene file among archives' files, VerificationMismatch is raised
- unit tests have been added

resolves: https://github.com/golemfactory/concent/issues/610

### DIFF
--- a/concent_api/conductor/tests/test_tasks.py
+++ b/concent_api/conductor/tests/test_tasks.py
@@ -1,5 +1,5 @@
 import mock
-from mock import patch
+from freezegun import freeze_time
 from django.conf import settings
 from conductor.models import BlenderSubtaskDefinition
 from conductor.models import VerificationRequest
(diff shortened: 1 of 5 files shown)


concent_api/verifier/utils.py
```diff
@@ -336,19 +336,29 @@ def unpack_archives(file_paths: Iterable[str], subtask_id: str) -> None:
...
-                ErrorCode.VERIFIER_UNPACKING_ARCHIVE_FAILED,
-                subtask_id,
+def validate_downloaded_archives(subtask_id: str, archives_list: Iterable[str], scene_file: str) -> None:
+    # If archive is broken, it means that Provider must have intentionally uploaded damaged zip file.
+    # In such case verification end with MISMATCH result.
+    package_files_list = []  # type: List[str]
+    try:
+        # If any file which is supposed to be unpacked from archives already exists, finish with error and raise exception.
+        for package_file_path in archives_list:
+            package_files_list = get_files_list_from_archive(
```
kbeker: You are here rewriting variable `package_files_list` every loop. Was it done on purpose?

dybi: whoops, there is a mistake - I will append files to the list

concent_api/verifier/utils.py
```diff
@@ -336,19 +336,31 @@ def unpack_archives(file_paths: Iterable[str], subtask_id: str) -> None:
...
-def validate_downloaded_archives(subtask_id: str, archives_list: Iterable[str]) -> None:
-    # If any file which is supposed to be unpacked from archives already exists, finish with error and raise exception.
+def validate_downloaded_archives(subtask_id: str, archives_list: Iterable[str], scene_file: str) -> None:
+    # If archive is broken, it means that Provider must have intentionally uploaded damaged zip file.
+    # In such case verification end with MISMATCH result.
+    package_files_list = []  # type: List[str]
+    try:
+        # If any file which is supposed to be unpacked from archives already exists, finish with error and raise exception.
+        for package_file_path in archives_list:
+            package_files_list += get_files_list_from_archive(
```
kbeker: Wouldn't be better to use built-in `append()` function? :)

dybi: why? `+=` seems more concise ;)

concent_api/verifier/utils.py
```diff
@@ -336,19 +336,31 @@ def unpack_archives(file_paths: Iterable[str], subtask_id: str) -> None:
...
+                generate_verifier_storage_file_path(package_file_path)
+            )
+    except zipfile.BadZipFile:
+        raise VerificationMismatch(subtask_id)
 
-    for package_file_path in archives_list:
-        package_files_list = get_files_list_from_archive(
-            generate_verifier_storage_file_path(package_file_path)
+    already_existing_files = set(os.listdir(settings.VERIFIER_STORAGE_PATH)).intersection(package_files_list)
```
rwrzesien: Maybe please add comment that it should not happen normally because we are cleaning directory at the beginning but we do a sanity check here.

concent_api/conductor/tests/test_tasks.py
```diff
@@ -53,51 +53,52 @@ def setUp(self):
...
-            task_id=self.report_computed_task.task_to_compute.compute_task_def['task_id'],
-            subtask_id=self.report_computed_task.task_to_compute.compute_task_def['subtask_id'],
-            provider_public_key=self.PROVIDER_PUBLIC_KEY,
-            requestor_public_key=self.REQUESTOR_PUBLIC_KEY,
-            state=Subtask.SubtaskState.ADDITIONAL_VERIFICATION,
-            next_deadline=get_current_utc_timestamp() + settings.CONCENT_MESSAGING_TIME,
-            task_to_compute=self.report_computed_task.task_to_compute,
-            report_computed_task=self.report_computed_task,
-        )
+        with freeze_time():
```
rwrzesien: Why is this needed now ?

dybi: because this test fails from time to time ;) without `freeze_time` there is sometimes 1 sec difference between times

concent_api/verifier/utils.py
dybi: ok

Issue #584, opened 2018-06-29
DatabaseTransactionsStorage for smart-contracts-interface


cameel:
golem-smart-contracts-interface needs to keep some local state - a list of pending transactions and a nonce. This state is kept in an object that implements the [TransactionsStorage](https://github.com/golemfactory/golem-smart-contracts-interface/blob/v1.4.2/golem_sci/transactionsstorage.py#L16-L53) interface. To protect multiple instances of SCI using the same contract from race conditions the implementation of this interface should allow sharing the state between instances.

The default implementation ([JsonTransactionsStorage](https://github.com/golemfactory/golem-smart-contracts-interface/blob/v1.4.2/golem_sci/transactionsstorage.py#L56-L143)) keeps data in a file on disk (`transactions.json`). Since each instance of Concent runs in a separate container, this is not enough to provide shared state.

### `DatabaseTransactionsStorage`
We need to create a new implementation that stores its state in the database. The database structure is as follows:

#### `GlobalTransactionState` model
| Field                           | Type      | Size           |
|---------------------------------|-----------|----------------|
| `nonce`                         | int       | up to 32 bytes |

There should always be at most one object of this type with `id = 0`. If it does not exist, it should be created and `nonce` initialized with a value obtained from SCI (`Client.get_transaction_count()`).
- Use the typical get-or-create pattern - try to get it from the database and create a new one if it does not exist. Since the `id` is unique, it will fail if another process tries to create it at the same time. When that happens just retry your query and use the object created by the other process.

#### `PendingEthereumTransaction` model
| Field                           | Type      | Size           |
|---------------------------------|-----------|----------------|
| `nonce`                         | int       | up to 32 bytes |
| `gasprice`                      | int       | up to 32 bytes |
| `startgas`                      | int       | up to 32 bytes |
| `to`                            | string    | 20 bytes       |
| `value`                         | int       | up to 32 bytes |
| `data`                          | binary    | unlimited      |
| `v`                             | int       | 1 byte         |
| `r`                             | int       | 32 bytes       |
| `s`                             | int       | 32 bytes       |
| `created_at`                    | timestamp |                |

Names are the traditional ones for Ethereum transaction fields. See the definition of [Transaction class in pyethereum](https://github.com/ethereum/pyethereum/blob/develop/ethereum/transactions.py#L20-L50).

[...]

PR #617, opened 2018-07-09
Feature database transaction storage

### LINKED ISSUE #605: Updating geth to version v1.8.11 causes `replacement transaction underpriced` error in `api-integration-force-report-computed-task-test.py`
The problem occurs only when `api-integration-force-report-computed-task-test.py` test is running  after update geth to the new version. It occurs since  `v1.8.9` release.
```
 Traceback (most recent call last): File "/srv/http/virtualenv/lib/python3.6/site-packages/django/core/handlers/base.py",
line 185, in _get_response response = wrapped_callback(request, *callback_args, **callback_kwargs)
File "/usr/lib/python3.6/contextlib.py", line 52, in inner return func(*args, **kwds)
File "/srv/http/concent_api/common/decorators.py", line 275, in wrapper return _function(*args, **kwargs)
File "/srv/http/virtualenv/lib/python3.6/site-packages/django/views/decorators/csrf.py", line 58, in wrapped_view return view_func(*args, **kwargs)
File "/srv/http/virtualenv/lib/python3.6/site-packages/django/views/decorators/http.py", line 40, in inner return func(request, *args, **kwargs)
File "/srv/http/concent_api/common/decorators.py", line 158, in wrapper return view(request, golem_message, client_public_key, *args, *kwargs)
File "/srv/http/concent_api/common/decorators.py", line 174, in wrapper response_from_view = view(request, client_message, client_public_key, *args, **kwargs)
File "/srv/http/concent_api/common/decorators.py", line 286, in wrapper response_from_view = view(request, golem_message, client_public_key)
File "/srv/http/concent_api/core/views.py", line 41, in send return handle_message(client_message)
File "/srv/http/concent_api/core/message_handlers.py", line 1422, in handle_message return handle_send_force_subtask_results(client_message)
File "/srv/http/concent_api/core/message_handlers.py", line 495, in handle_send_force_subtask_results payment_ts=current_time,
File "/srv/http/concent_api/core/payments/base.py", line 15, in wrapper return func(backend, *args, **kwargs)
File "/srv/http/concent_api/core/payments/base.py", line 49, in make_force_payment_to_provider payment_ts = payment_ts,
File "/srv/http/concent_api/core/payments/sci_backend.py", line 70, in make_force_payment_to_provider closure_time = payment_ts,
File "/srv/http/virtualenv/src/golem-smart-contracts-interface/golem_sci/implementation.py", line 562, in force_payment self.GAS_FORCE_PAYMENT,
File "/srv/http/virtualenv/src/golem-smart-contracts-interface/golem_sci/implementation.py", line 322, in _create_and_send_transaction return self._sign_and_send_transaction(tx)
File "/srv/http/virtualenv/src/golem-smart-contracts-interface/golem_sci/implementation.py", line 298, in _sign_and_send_transaction return self._geth_client.send(tx)
File "/srv/http/virtualenv/src/golem-smart-contracts-interface/golem_sci/client.py", line 94, in send return self.web3.eth.sendRawTransaction(hex_data)
File "/srv/http/virtualenv/lib/python3.6/site-packages/web3/eth.py", line 227, in sendRawTransaction [raw_transaction],
File "/srv/http/virtualenv/lib/python3.6/site-packages/web3/manager.py",
line 96, in request_blocking raise ValueError(response["error"])
ValueError: {'code': -32000, 'message': 'replacement transaction underpriced'}

```

### LINKED ISSUE #607: SCI fails to connect to geth on the cluster after update to `1.4.1` version
We don't know exactly why SCI fails to connect to geth on the cluster. The problem may occur on the`concent` or `concent-deployment` side.

```
Traceback (most recent call last): File "/srv/http/virtualenv/lib/python3.6/site-packages/django/core/handlers/base.py",
line 185, in _get_response response = wrapped_callback(request, *callback_args, **callback_kwargs)
File "/usr/lib/python3.6/contextlib.py", line 52, in inner return func(*args, **kwds)
File "/srv/http/concent_api/common/decorators.py", line 275, in wrapper return _function(*args, **kwargs)
File "/srv/http/virtualenv/lib/python3.6/site-packages/django/views/decorators/csrf.py", line 58, in wrapped_view return view_func(*args, **kwargs)
File "/srv/http/virtualenv/lib/python3.6/site-packages/django/views/decorators/http.py", line 40, in inner return func(request, *args, **kwargs)
File "/srv/http/concent_api/common/decorators.py", line 158, in wrapper return view(request, golem_message, client_public_key, *args, *kwargs)
File "/srv/http/concent_api/common/decorators.py", line 174, in wrapper response_from_view = view(request, client_message, client_public_key, *args, **kwargs)
File "/srv/http/concent_api/common/decorators.py", line 286, in wrapper response_from_view = view(request, golem_message, client_public_key)
File "/srv/http/concent_api/core/views.py", line 41, in send return handle_message(client_message)
File "/srv/http/concent_api/core/message_handlers.py", line 1431, in handle_message return handle_send_force_payment(client_message)
File "/srv/http/concent_api/core/message_handlers.py", line 725, in handle_send_force_payment transaction_type = TransactionType.BATCH,
File "/srv/http/concent_api/core/payments/base.py", line 15, in wrapper return func(backend, *args, **kwargs)
File "/srv/http/concent_api/core/payments/base.py", line 33, in get_list_of_payments transaction_type = transaction_type,
File "/srv/http/concent_api/core/payments/sci_backend.py", line 25, in get_list_of_payments concent_rpc = ConcentRPC()
File "/srv/http/concent_api/common/singleton.py", line 13, in __new__ lambda tx: tx.sign(settings.CONCENT_ETHEREUM_PRIVATE_KEY)
File "/srv/http/virtualenv/src/golem-smart-contracts-interface/golem_sci/factory.py", line 48, in new_sci_rpc return new_sci(datadir, Web3(HTTPProvider(rpc)), address, tx_sign, chain)
File "/srv/http/virtualenv/src/golem-smart-contracts-interface/golem_sci/factory.py", line 62, in new_sci _ensure_connection(web3)
File "/srv/http/virtualenv/src/golem-smart-contracts-interface/golem_sci/factory.py", line 95, in _ensure_connection
raise Exception('Could not connect to geth: {}'.format(web3.providers))
Exception: Could not connect to geth: [<web3.providers.rpc.HTTPProvider object at 0x7f2454adbd68>]
```

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/584

Adds GlobalTransactionState and PendingEthereumTransaction models.
Adds DatabaseTransactionsStorage and its usage in ConcentRPC.
Adds get_transaction_count to payments.service.

Marked [WIP] because:
- is waiting for answer from related issue about soft delete.
- there is some code to uncomment after udpate to SCI 1.4.1 is resolved (link to issue),
- it need to be tested on cluster.

Also:
Resolves #605
Resolves #607


concent_api/core/models.py
```diff
@@ -518,3 +519,39 @@ def clean(self):
...
+    """
+
+    nonce = DecimalField(max_digits=32, decimal_places=0, unique=True)
+
+    gasprice = CharField(max_length=32)
+    startgas = CharField(max_length=32)
+    value = CharField(max_length=32)
+
+    to = BinaryField(max_length=20)
+    data = BinaryField()
```
kbeker: After testing new SCI I have discovered that data is not `BinaryField` it is really long string -> probably as a `data` we will need `TextField()`

concent_api/core/models.py
```diff
@@ -518,3 +519,39 @@ def clean(self):
...
+    Represents pending Ethereum transaction state.
+    """
+
+    nonce = DecimalField(max_digits=32, decimal_places=0, unique=True)
+
+    gasprice = CharField(max_length=32)
+    startgas = CharField(max_length=32)
+    value = CharField(max_length=32)
+
+    to = BinaryField(max_length=20)
```
kbeker: Field `to` in `transaction.json` is stored as a string.

concent_api/core/models.py
```diff
@@ -518,3 +519,39 @@ def clean(self):
...
+
+
+class PendingEthereumTransaction(Model):
+    """
+    Represents pending Ethereum transaction state.
+    """
+
+    nonce = DecimalField(max_digits=32, decimal_places=0, unique=True)
+
+    gasprice = CharField(max_length=32)
```
kbeker: `Gasprice` in new `SCI` is stored as integer

concent_api/core/models.py
```diff
@@ -518,3 +519,39 @@ def clean(self):
...
+    startgas = CharField(max_length=32)
```
kbeker: `startgas` in new `SCI` is stored as integer

concent_api/core/payments/storage.py
```diff
@@ -0,0 +1,145 @@
...
+                f'Trying to revert last PendingEthereumTransaction with nonce {global_transaction_state.nonce - 1} but it does not exist.'
+            )
+            raise
+
+    @staticmethod
+    def _get_locked_global_transaction_state() -> GlobalTransactionState:
+        try:
+            return GlobalTransactionState.objects.select_for_update().get(
+                pk=0,
+            )
```
cameel: You're never releasing the lock explicitly. From the docs I see that the lock lasts "until the end of the transaction block". The thing is that we will likely have more than one transaction block and they're nested: one explicitly started with `@transaction.atomic` above and one that's by default on the view (due to `ATOMIC_REQUESTS`). Could you verify that it's the former that releases the lock? We definitely don't want to keep the row locked until the end of the view.

concent_api/core/models.py
```diff
@@ -518,3 +519,39 @@ def clean(self):
...
+    value = CharField(max_length=32)
+
+    to = BinaryField(max_length=20)
+    data = BinaryField()
+
+    v = IntegerField()
+    r = CharField(max_length=32)
```
kbeker: Field `r` in `transaction.json` is stored as a integer. In my tests i had `r` field which equaled 105276041803796697890139158600495981346175539693000174052040367753737207356915

concent_api/core/models.py
```diff
@@ -518,3 +519,39 @@ def clean(self):
...
+    s = CharField(max_length=32)
```
kbeker: With `s` field same story as with `r` field

concent_api/core/payments/storage.py
```diff
@@ -0,0 +1,145 @@
...
+        Remove the last transaction that was added.
+        This shouldn't be ever called if everything is being used correctly,
+        i.e. we don't try to send invalid transactions.
+        """
+        global_transaction_state = self._get_locked_global_transaction_state()
+
+        try:
+            pending_ethereum_transaction = PendingEthereumTransaction.objects.get(
+                nonce=global_transaction_state.nonce - 1
+            )
```
cameel: Your're not decreasing the nonce. This means that you can't for example call `revert_last_tx()` twice to revert two last transactions.

Is this intentional (maybe it's how SCI works)? Or is it a bug?

concent_api/core/payments/storage.py
```diff
@@ -0,0 +1,145 @@
...
+
+    @staticmethod
+    def _get_locked_global_transaction_state() -> GlobalTransactionState:
+        try:
+            return GlobalTransactionState.objects.select_for_update().get(
+                pk=0,
+            )
+        except GlobalTransactionState.DoesNotExist:
+            logger.error(f'Trying to get GlobalTransactionState but it does not exist.')
+            raise
```
kbeker: I think we will need also method which will get oldest transaction which hasn't been proceed yet. @cameel what do you think?

concent_api/core/payments/storage.py
```diff
@@ -0,0 +1,145 @@
...
+            Transaction(
+                nonce    = int(ethereum_transaction.nonce),
+                gasprice = int(ethereum_transaction.gasprice),
+                startgas = int(ethereum_transaction.startgas),
+                value    = int(ethereum_transaction.value),
+                v        = int(ethereum_transaction.v),
+                r        = int(ethereum_transaction.r),
+                s        = int(ethereum_transaction.s),
+                to       = encode_hex(ethereum_transaction.to.tobytes()),
+                data     = ethereum_transaction.data.tobytes(),
```
kbeker: Style:)

concent_api/core/payments/storage.py
```diff
@@ -0,0 +1,145 @@
...
+        pending_ethereum_transaction = PendingEthereumTransaction(
+            nonce       = tx.nonce,
+            gasprice    = tx.gasprice,
+            startgas    = tx.startgas,
+            value       = tx.value,
+            v           = tx.v,
+            r           = tx.r,
+            s           = tx.s,
+            data        = tx.data,
+            to          = tx.to,
```
kbeker: Style:)

concent_api/core/payments/storage.py
```diff
@@ -0,0 +1,145 @@
...
+        ]
+
+    @transaction.atomic(using='control')
+    def put_tx_and_inc_nonce(self, tx: Transaction) -> None:
+        """
+        Save a valid transaction and increase the nonce.
+        """
+        global_transaction_state = self._get_locked_global_transaction_state()
+
+        if global_transaction_state.nonce != tx.nonce:
```
cameel: Shouldn't it be `tx.nonce - 1`?

concent_api/core/tests/test_payments_database_transaction_storage.py
```diff
@@ -0,0 +1,193 @@
...
+        with mock.patch('core.payments.service.get_transaction_count', return_value=5):
+            self.storage = DatabaseTransactionsStorage()
+
+        self.global_transaction_state = GlobalTransactionState.objects.get(pk=0)
+
+    def _create_pending_ethereum_transaction(self):
+        pending_ethereum_transaction = PendingEthereumTransaction(
+            nonce=self.global_transaction_state.nonce,
+            gasprice=10 ** 9,
+            startgas=21000,
```
cameel: Please make at least one test that checks the full range of integers. We want to be sure that it does not break with very large numbers.

concent_api/core/models.py
```diff
@@ -518,3 +519,39 @@ def clean(self):
...
+
```
kbeker: @cameel @rwrzesien 
Don't we need also a field which will tell SCI which transactions was already done and which are awaiting to proceed?

cameel:
> Refactors core.payments file structure.

I think it would be better to extract this into a separate pull request so that it does not have to wait until this pull request is merged. I did not see any problems with the refactoring itself so it should go through easily and it's a distraction from the important changes.

concent_api/core/models.py
cameel: We only need the stuff that's in the current JSON. If there's such a field, please make a comment in the issue and I'll update the description.

concent_api/core/payments/storage.py
cameel: The `nonce` does that. By checking its value you can figure out which transaction was added last.

concent_api/core/payments/storage.py
cameel: Well, it's technically against our rules but in this case it clearly improves readability. Maybe we should allow for some exceptions?

@dybi What do you think?

concent_api/core/models.py
kbeker: I have checked that even `BigIntegerField` can store numbers in range `from -9223372036854775808 to 9223372036854775807` so it is too small for our needs. I think the best would be to convert this number to string and then store as `TextField()`

concent_api/core/payments/storage.py
rwrzesien: Yes, exactly, my point here was not to follow old rules but to occasionally break new rules for better readability. But if anyone will insist, I will change it.

@cameel It would be even less readable that in the second part of your example, because there should be no spaces around `=`.

concent_api/core/payments/storage.py
rwrzesien: Same here, done for readability, but if you or anyone insist I will change it.

concent_api/core/models.py
rwrzesien: Yep, that what I was already pointing out on daily few days ago, `BigIntegerField` is not enough for this, but `Decimal` could do in this case, as I now know how sample this data will look exactly in `Transaction`.

@kbeker I see you have liked `TextField` very much but believe me it is not a cure for everything ;)

concent_api/core/models.py
rwrzesien: Right, I will change it to `Decimal` too.

concent_api/core/payments/storage.py
rwrzesien: @kbeker Also, we are not implementing our custom interface for this, but we follow https://github.com/golemfactory/golem-smart-contracts-interface/blob/v1.4.2/golem_sci/transactionsstorage.py#L16-L53

concent_api/core/models.py
rwrzesien: Lets make it `Decimal` then with `max_digits` set to 32.

concent_api/core/models.py
rwrzesien: Lets make it `Decimal` then with `max_digits` set to 32.

concent_api/core/models.py
rwrzesien: No, there is no such field there.

concent_api/core/tests/test_payments_database_transaction_storage.py
rwrzesien: Ok, but as you see in the discussion in other comments I am still not sure what are maximum values here.

concent_api/core/models.py
```diff
@@ -518,3 +519,39 @@ def clean(self):
...
+class PendingEthereumTransaction(Model):
+    """
+    Represents pending Ethereum transaction state.
+    """
+
+    nonce = DecimalField(max_digits=32, decimal_places=0, unique=True)
+
+    gasprice = CharField(max_length=32)
+    startgas = CharField(max_length=32)
+    value = CharField(max_length=32)
```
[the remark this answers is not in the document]
rwrzesien: It supports, but big integer is not enough for the data from sample provided by @kbeker . I will switch to `Decimal` whenever possible.

concent_api/core/payments/backends/mock.py
```diff
@@ -9,3 +9,7 @@ def make_force_payment_to_provider(requestor_eth_address = None, provider_eth_ad
 
 def is_account_status_positive(client_eth_address = None, pending_value = None):  # pylint: disable=unused-argument
     return pending_value > 0
+
+
+def get_transaction_count() -> int:
+    return 10
```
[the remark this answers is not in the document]
rwrzesien: :+1:

concent_api/core/payments/backends/sci_backend.py
```diff
@@ -83,6 +83,10 @@ def is_account_status_positive(
     return client_acc_balance > pending_value
 
 
+def get_transaction_count() -> int:
+    return ConcentRPC().get_transaction_count()
```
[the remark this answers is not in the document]
rwrzesien: I actually have no idea, but if we want to change it lets do this in separate issue.

@kbeker Maybe you can explain ?

@cameel Isn't it RemoTe Procedure Call ?

concent_api/core/payments/storage.py
```diff
@@ -0,0 +1,145 @@
...
+            raise Exception(
+                f'Transaction nonce does not match. '
+                f'Current={global_transaction_state.nonce}, '
+                f'Transaction={tx.nonce}'
```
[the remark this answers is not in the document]
rwrzesien: I have followed up the fact that `JsonTransactionsStorage` is raising pure `Exception`, and assumed that it might matter, but if you think it doesn't then off course, lets make it a custom exception class.

concent_api/core/models.py
rwrzesien: That's most probably because JSON doesn't support binary data, so it has to be converted to string. I think our point is to be able to store and recreate `Transaction` to be exactly the same, not storing the data in the same format as `JsonTransactionsStorage`, but if I am wrong then please correct me. @cameel ?

Btw. 'I think our point is to be able to store and recreate `Transaction`' is a good scenario for test, I will add it.

concent_api/core/models.py
rwrzesien: But how does it look in `Transaction` class ? I still can't find a sample of that.

rwrzesien: It is string most probably from the same reason as in https://github.com/golemfactory/concent/pull/617#discussion_r201184212

I have decoded the `data` from sample which you have provided, so I have retrieved the original value from transaction, used it in tests, and it worked fine.

concent_api/core/payments/storage.py
rwrzesien: Why ? As I understand, the transaction `nonce` should match 'current' `nonce`. It is done the same way in https://github.com/golemfactory/golem-smart-contracts-interface/blob/v1.4.2/golem_sci/transactionsstorage.py#L98

concent_api/core/payments/storage.py
rwrzesien: My mistake, I thought `nonce` can never ever be used twice, but I have now checked that `JsonTransactionsStorage` is doing it.

concent_api/core/payments/storage.py
kbeker: For me is okay I didn't want to change our style and for me it is more readable :) I have just pointed this because it's not matched with our current style

concent_api/core/payments/storage.py
```diff
@@ -0,0 +1,145 @@
...
+    database using Django models.
+    """
+
+    def __init__(self, *args: List, **kwargs: Dict) -> None:
+        super().__init__(*args, **kwargs)
+
+        if not GlobalTransactionState.objects.filter(pk=0).exists():
+            global_transaction_state = GlobalTransactionState(
+                pk=0,
+                nonce=payments_service.get_transaction_count(),  # pylint: disable=no-value-for-parameter
```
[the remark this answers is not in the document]
rwrzesien: It is possible for a class, however this is a module. I can add `backend = None` in all function signatures in 'core.payments.service`, and that will do the job. Is that ok ?

concent_api/core/payments/storage.py
rwrzesien: I have checked it positively however I don't really know how to test it negatively (to cause crash because of lock). Please consider the following code:

```
with transaction.atomic(using='control'):
    with transaction.atomic(using='control'):
        s = Subtask.objects.select_for_update().get(pk=7)
    s = Subtask.objects.get(pk=7)
```

It works just fine. Do you think its enough ?

concent_api/core/models.py
cameel: You're right. We want to store the original transaction, not a transaction mangled by being converted to JSON.

concent_api/core/models.py
cameel: `Decimal` is fine. But 32 digits is too little. A 32-byte integer has 256 binary digits and the number of decimal digits is about 1/3 of that:

``` python
>>> len(str(2 ** 256 - 1))
78
```

concent_api/core/models.py
cameel: See my comment above: https://github.com/golemfactory/concent/pull/617#discussion_r201396775

concent_api/core/models.py
cameel: Maximum value for a 32-byte unsigned int is just `2**(32*8)-1`. 78 decimal digits is enough.

concent_api/core/models.py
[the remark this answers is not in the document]
cameel: OK. `Decimal` is fine.

concent_api/core/payments/backends/sci_backend.py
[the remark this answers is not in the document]
cameel: > @cameel Isn't it RemoTe Procedure Call ?

Yeah. Fixed my comment.

> I actually have no idea, but if we want to change it lets do this in separate issue.

So let's create an issue. It's a simple thing but it's been bugging me for a long time :)

concent_api/core/payments/storage.py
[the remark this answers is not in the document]
cameel: No, we don't want unnecessary defaults if we can avoid it.

What about the status of this in pylint? Are there any chances that they'll fix it soon?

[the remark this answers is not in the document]
rwrzesien: I think not soon according to https://github.com/PyCQA/pylint/issues/259#issuecomment-386978676

concent_api/core/payments/storage.py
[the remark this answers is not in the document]
cameel: Let's use a custom class. Even if it was checked in any way, a custom exception would inherit from `Exception` anyway.

 We actually should add it to our coding style: https://github.com/Code-Poets/coding-style/issues/10

concent_api/core/payments/storage.py
cameel: Like this?

``` python
with transaction.atomic(using='control'):
    with transaction.atomic(using='control'):
        s = Subtask.objects.select_for_update().get(pk=7)
    s = Subtask.objects.select_for_update().get(pk=7)
```

This should deadlock if you run it and unless the first `atomic()` releases the lock.

concent_api/core/payments/storage.py
cameel: But isn't `tx` the new transaction that already has a nonce set to one higher?

I was assuming that `nonce` is increased only after the new transaction gets added.

concent_api/core/tests/test_payments_database_transaction_storage.py
cameel: The sizes imply maximum values. For example for a 32-byte unsigned integer the maximum value is `2**(32*8)-1`

concent_api/core/payments/storage.py
rwrzesien: This code works without errors, so the inner lock is released with the end of context manager.

concent_api/core/payments/storage.py
rwrzesien: > I was assuming that nonce is increased only after the new transaction gets added.

Exactly, so the `current` transaction should have `current` nonce, right ? At least it looks like this in `JsonTransactionsStorage`.

concent_api/common/singleton.py
```diff
@@ -1,15 +1,45 @@
 from django.conf        import settings
-from golem_sci.factory  import new_sci_rpc
+from golem_sci.client import Client
+from golem_sci.contracts.provider import ContractDataProvider
+
```
rwrzesien: Please remove blank line.

concent_api/common/singleton.py
```diff
@@ -1,15 +1,45 @@
...
+from web3 import Web3
+from web3.middleware import geth_poa_middleware
+
+from core.payments.storage import DatabaseTransactionsStorage
+
+
+def concent_sci(
+        storage,
+        web3,
+        address,
```
rwrzesien: Fix indentation.

concent_api/common/singleton.py
```diff
@@ -1,15 +1,45 @@
...
+        tx_sign,
+):
+    if geth_poa_middleware not in web3.middleware_stack:
```
rwrzesien: You can add assert that `storage` is subclass of `TransactionStorage`.

concent_api/core/payments/storage.py
```diff
@@ -0,0 +1,152 @@
...
+            global_transaction_state.full_clean()
+            global_transaction_state.save()
+
+    @transaction.atomic(using='control')
+    def get_nonce(self) -> int:
+        """
+        Returns the nonce for the next transaction.
+        """
+        global_transaction_state = self._get_locked_global_transaction_state()
+        return int(global_transaction_state.nonce)
```
rwrzesien: Add method to `GlobalTransactionState` called `get_nonce` which will change it to int and reuse it in all places.

concent_api/common/singleton.py
kbeker: Done

concent_api/core/payments/storage.py
kbeker: Done

concent_api/core/payments/storage.py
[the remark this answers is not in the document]
cameel: OK. So let's ignore it.

BTW, I think I have already responded to this. Where did my comment go?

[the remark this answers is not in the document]
rwrzesien: Here :) https://github.com/golemfactory/concent/pull/617#discussion_r201404104

concent_api/core/payments/backends/sci_backend.py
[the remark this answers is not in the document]
rwrzesien: Created https://github.com/golemfactory/concent/issues/640

concent_api/core/payments/storage.py
cameel: I'd prefer for it to be allowed but you have a point. @rwrzesien please raise this issue on the retro.

concent_api/core/payments/storage.py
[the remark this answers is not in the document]
cameel: Oh, right :) That's where it went. Sorry :)

concent_api/core/models.py
```diff
@@ -518,3 +519,39 @@ def clean(self):
...
+    gasprice = DecimalField(max_digits=32, decimal_places=0)
+    startgas = DecimalField(max_digits=32, decimal_places=0)
+    value = IntegerField()
+
+    to = BinaryField(max_length=20)
+    data = BinaryField()
+
+    v = IntegerField()
+    r = DecimalField(max_digits=128, decimal_places=0)
+    s = DecimalField(max_digits=128, decimal_places=0)
```
cameel: Too many digits.

concent_api/core/models.py
```diff
@@ -518,3 +519,39 @@ def clean(self):
...
+
+class PendingEthereumTransaction(Model):
+    """
+    Represents pending Ethereum transaction state.
+    """
+
+    nonce = DecimalField(max_digits=32, decimal_places=0)
+
+    gasprice = DecimalField(max_digits=32, decimal_places=0)
+    startgas = DecimalField(max_digits=32, decimal_places=0)
```
cameel: Not enough digits.

concent_api/core/payments/backends/sci_backend.py
```diff
@@ -1,6 +1,8 @@
 from enum import Enum
 
 from golem_sci.blockshelper import BlocksHelper
+from web3 import Web3
+
```
cameel: If we're doing this, we should add web3 to our dependencies. And remove it later when we stop using it.

concent_api/requirements.lock
```diff
@@ -83,4 +90,5 @@ toolz==0.9.0
 typed-ast==1.1.0
 urllib3==1.23
 vine==1.1.4
-web3==3.16.4
+web3==4.2.1
+websockets==5.0.1
```
cameel: Why does SCI depend on `websockets`?

concent_api/requirements.lock
```diff
@@ -20,29 +21,35 @@ cytoolz==0.9.0.1
 dask==0.18.1
 decorator==4.3.0
 Django==1.11.13
-django-constance[database]==2.2.0
-eth-abi==0.5.0
-eth-keyfile==0.4.1
-eth-keys==0.1.0b4
-eth-tester==0.1.0b15
-eth-utils==0.7.4
+django-constance==2.2.0
```
cameel: Why is the dependency on `django-constance` being changed here?

concent_api/common/constants.py
```diff
@@ -60,6 +60,7 @@ class ErrorCode(enum.Enum):
     QUEUE_WRONG_STATE                                                   = 'queue.wrong_state'
     REQUEST_BODY_NOT_EMPTY                                              = 'request_body.not_empty'
     SUBTASK_DUPLICATE_REQUEST                                           = 'subtask.duplicate_request'
+    TRANSACTION_NONCE_NOT_MATCH                                         = 'transaction.nonce.not_match'
```
cameel: `TRANSACTION_NONCE_DOES_NOT_MATCH` or `TRANSACTION_NONCE_MISMATCH` would be a better name.

concent_api/core/exceptions.py
```diff
@@ -23,3 +23,7 @@ class GolemMessageValidationError(ConcentBaseException):
 
 class FrameNumberValidationError(ConcentBaseException):
     pass
+
+
+class TransactionNonceNotMatchError(ConcentBaseException):
```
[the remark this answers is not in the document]
kbeker: Done

concent_api/core/models.py

concent_api/core/models.py

concent_api/core/payments/backends/sci_backend.py
kbeker: Added to `requirements.txt`

concent_api/requirements.lock
kbeker: As we talked recently on `slack` and you decided that if something is in `pip freeze` it should be also in `requirements.lock` This is how `django-constance[database]` is installed by `pip`: 
```
django-constance==2.2.0
django-picklefield==1.0.0
```

concent_api/requirements.lock
kbeker: `websockets` depends on `web3` and is installing during it's installation. Look: 
`Collecting websockets>=4.0.1 (from web3==4.2.1)`

concent_api/core/payments/storage.py
```diff
@@ -0,0 +1,152 @@
...
+
+    @transaction.atomic(using='control')  # pylint: disable=no-self-use
+    def get_all_tx(self) -> List[Transaction]:
+        """
+        Returns the list of all transactions.
+        """
+        return [
+            Transaction(
+                nonce    = int(ethereum_transaction.nonce),
```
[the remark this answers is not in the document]
kbeker: Done

concent_api/common/singleton.py
```diff
@@ -1,15 +1,50 @@
...
+
+
+#  Temporary SCI factory for Concent needs.
+#  Gets storage as a parameter
+def concent_sci(
+    storage: TransactionsStorage,
+    web3: Web3,
+    address: Web3,
+    tx_sign: Callable[[Transaction], None],
+):
```
rwrzesien: You can add `-> SCIImplementation`

concent_api/core/payments/storage.py
```diff
@@ -0,0 +1,152 @@
...
+        Save a valid transaction and increase the nonce.
+        """
+        global_transaction_state = self._get_locked_global_transaction_state()
+
+        if global_transaction_state.get_nonce != tx.nonce:
+            raise TransactionNonceNotMatchError(
+                f'Transaction nonce does not match. '
+                f'Current={global_transaction_state.nonce}, '
+                f'Transaction={tx.nonce}',
+                ErrorCode.TRANSACTION_NONCE_MISMATCH,
```
rwrzesien: It doesn't need error code.

kbeker: Yes, but won't it be better to have such a thing? :)

concent_api/common/singleton.py
kbeker: Done

concent_api/core/payments/storage.py
rwrzesien: But it won't be visible anywhere, it will be just passed to Exception `__init__` and disappear.

kbeker: OK, so I delete it :)

concent_api/core/models.py
```diff
@@ -518,3 +519,67 @@ def clean(self):
...
+
+class GlobalTransactionState(Model):
+    """
+    Represents state of whole transaction service by storing last `nonce` used.
+
+    There should always be at most one object of this type with id = 0. If it does not exist, it should be created and
+    nonce initialized with a value obtained from SCI (Client.get_transaction_count()).
+    """
+
+    nonce = DecimalField(max_digits=32, decimal_places=0, default=0, unique=True)
```
rwrzesien: Shouldn't nonce have `max_digits=78` too ?

concent_api/core/payments/storage.py
```diff
@@ -0,0 +1,150 @@
...
+from typing import List
+from hexbytes import HexBytes
+from ethereum.transactions import Transaction
+from golem_sci.transactionsstorage import TransactionsStorage
+
+from django.db import transaction
+
+from core.models import GlobalTransactionState
+from core.models import PendingEthereumTransaction
+
```
rwrzesien: Please remove blank line and order alphabetically.

concent_api/requirements.lock
cameel: But it's completely unrelated to this pull request. You should make this change separately.

concent_api/requirements.lock
cameel: Weird but OK.

concent_api/common/singleton.py
```diff
@@ -1,15 +1,50 @@
...
+from web3 import HTTPProvider
+from web3 import Web3
+from web3.middleware import geth_poa_middleware
+
+from core.payments.storage import DatabaseTransactionsStorage
+
+
+#  Temporary SCI factory for Concent needs.
+#  Gets storage as a parameter
+def concent_sci(
```
cameel: Might be a good idea to say in a comment where the original code came from.

concent_api/core/models.py
```diff
@@ -518,3 +519,67 @@ def clean(self):
...
+
+    def save(self, *args, **kwargs):  # pylint: disable=arguments-differ
```
cameel: Why does this warning need to be disabled here?

concent_api/core/models.py
```diff
@@ -518,3 +519,67 @@ def clean(self):
...
+
+    v = IntegerField()
+    r = DecimalField(max_digits=78, decimal_places=0)
+    s = DecimalField(max_digits=78, decimal_places=0)
+
+    created_at = DateTimeField(auto_now_add=True)
+
+    @property
+    def get_nonce(self):
+        return int(self.nonce)
```
cameel: This sounds like a function name. Properties should be named like variables (i.e. using nouns, not verbs). E.g. `nonce_as_int`.

cameel: I think that these properties take more code than they actually save. It would be better to just call `int()` whenever you access a decimal field.

concent_api/core/payments/backends/sci_backend.py
```diff
@@ -28,15 +30,15 @@ def get_list_of_payments(
 
     if transaction_type == TransactionType.FORCE:
         payments_list = concent_rpc.get_forced_payments(  # pylint: disable=no-member
-            requestor_address   = requestor_eth_address,
-            provider_address    = provider_eth_address,
+            requestor_address   = Web3.toChecksumAddress(requestor_eth_address),
+            provider_address    = Web3.toChecksumAddress(provider_eth_address),
```
cameel: Wouldn't it be better to let `get_forced_payments()` do the address conversion internally? It looks like an implementation detail to me.

concent_api/core/payments/storage.py
```diff
@@ -0,0 +1,150 @@
...
+    @transaction.atomic(using='control')
+    def put_tx_and_inc_nonce(self, tx: Transaction) -> None:
+        """
+        Save a valid transaction and increase the nonce.
+        """
+        global_transaction_state = self._get_locked_global_transaction_state()
+
+        if global_transaction_state.get_nonce != tx.nonce:
+            raise TransactionNonceNotMatchError(
```
cameel: Maybe `TransactionNonceMismatch` would be a better name?

concent_api/common/singleton.py
kbeker: Done

concent_api/core/models.py
kbeker: PILINT gives warning if there is no disable:
`Parameters differ from overridden 'save' method (arguments-differ)`

concent_api/core/models.py
kbeker: Deleted, just as you said in next comment

concent_api/core/payments/backends/sci_backend.py
kbeker: But, we need to do this conversion before we give addresses as parameters. If not some exception is raised.

concent_api/core/payments/storage.py
kbeker: Done

concent_api/core/payments/storage.py
kbeker: Done

concent_api/core/models.py

concent_api/requirements.lock
kbeker: Extracted from this PR

concent_api/core/payments/backends/sci_backend.py
cameel: But `get_force_payments()` is our function. If it raises an exception then maybe it shouldn't?

concent_api/core/models.py
cameel: So please define it with the same parameters.

Issue #599, opened 2018-07-04
Signing Service: Base application


cameel:
This is a part of the implementation of the "Concent Signing Service" blueprint (#597).

#### Overview
The goal of this task is to prepare the base on which we can build the service. For now it's just a test application that connects to Middleman, sends a test message, prints the response and immediately disconnects.

This is just a base for the full service. The actual protocol between the service and the Middleman and signing of Ethereum transactions will be implemented on top of this base in subsequent issues.

#### New project
Create a new project not based on Django or any other web framework.
- Put it in a directory called `signing_service/` at the top level of Concent repository.
- The project should have its own `requirements.txt`/`requirements.lock`.
- Configure the project for unit testing
    - Add necessary dependencies.
    - Plug it into `full_check.sh`, `lint.sh`, `run_tests.sh`, `mypy.ini`, etc.
- Configure logging.
- Configure it to be able to send error reports to Sentry.
- It does not need a configuration file.

#### Command-line parameters
The project should have a top-level Python script that starts the server. It should support at least the following command-line parameters:
- Address of a Concent cluster (no default)
- Port to use for the connection (use the same default as in Middleman)
- Sentry DSN (optional)

#### Communication with Concent
This is just a simple test program to ensure that the application works.

1. Initiate a TCP connection to the Concent cluster.
2. Send a test message
3. Wait for the response and print it on the screen.
4. Close the connection and exit.

rwrzesien:
@cameel 
> Wait for the response and print it on the screen.

The case is that Concent doesn't respond to messages sent this way. I think that if it is working as HTTP server it will respond only to correct HTTP request.

cameel:
Update: There should be separate parameters for port and address.

cameel:
Update: Added Sentry DSN to command-line parameters.

Issue #616, opened 2018-07-09
Middleman: Accepting connections from the Signing Service


cameel:
This is a part of the implementation of the "Concent Signing Service" blueprint (#597).

#### Overview
The goal of this task is to prepare a Django app that starts a TCP server and accepts a single connection from the Signing Service. Connection is kept open until the service closes it. New connections are rejected while one is active. At this point there's no protocol and authentication and all received data is simply echoed back to the client.

This issue can be done independently of #615 but if #615 is not done before, you'll need to do some initial work from it here (creating the app, implementing the main loop, etc.). The stuff that's already been mentioned in #615 is omitted here so please read that issue before starting work on this one.

#### Middleman app and its management command
1. Create an app called `middleman` (See #615).
2. Create a management command called `middleman` (See #615).
3. Support the following command-line parameters (in addition to the ones from #615):
    - `--external-port`: int - the port Signing Service will connect to, available both from inside and outside the cluster.

#### Main loop
When Middleman starts it should perform the following steps (in addition to the ones from #615):

1. Start a TCP server listening on the external port.
    - If it's not possible (e.g. something else listening on that port), exit with an error.
2. When a new connection is established on the external port run the Signing Service connection handler
3. Handle exceptions and signals (See #615).

#### Concent connection handler
1. If a connection is already established close the connection and return
2. Otherwise continue
3. Read data from the client
4. Echo data back to the client
5. Handle exceptions
    - When the client closes the connection, return without errors

PaweuB:
Link in Pivotal:
https://www.pivotaltracker.com/story/show/158908178

PR #619, opened 2018-07-09
Create Signing Service project

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/599


run_tests.sh
```diff
@@ -44,3 +44,12 @@ pytest                              \
     ${TEST_RUNNER_EXTRA_ARGUMENTS}
 rm .coverage
 cd ..
+
+cd signing_service/
+pytest -p no:django                 \
```
dybi: we will have two separate coverage reports... Is this what we want?

rwrzesien: Yes, one for `concent_api` and one for `signing_service`. I am not sure if that is what we really want but `.coverage` is created in the directory when tests are run, and I am not sure if they can be run from parent directory correctly. Do you want me to check if they can work this way ?

rwrzesien:
@Jakub89 It took some time to test it in all our runners, so I managed only to prepare this, but it should be enough base to start working on other things, and off course I will start adding more parts as soon as possible :)

run_tests.sh
cameel: We should have one, but if it's hard to achieve, we could deal with it in a separate issue.

signing_service/src/signing_service.py
```diff
@@ -0,0 +1,16 @@
...
+crash_logger = logging.getLogger('crash')
+
+
+def main():
+    while True:
+        pass
+
+
+if __name__ == '__main__':
+    logging.config.fileConfig(os.path.join('..', 'logging.ini'))
```
cameel: This will load `logging.ini` from a directory relative to the working directory (the one where you run the script from). I think it should be relative to the place where the script is located instead.

rwrzesien: :+1: done

rwrzesien:
> Well, I still think this should be done along with some initial communication code so that we can make sure it actually works. For example, have you checked that stuff actually get logged?

I will add it as next thing, just didn't had enough time. And yes, logging is working fine, but I had no chance to test logging to sentry without setting DSN.

cameel:
OK. I have updated the issue and there's not much stuff missing here (all the complicated communication is in separate issues now) so I think it won't be hard to still add it in this pull request.

signing_service/src/signing_service.py
```diff
@@ -0,0 +1,34 @@
...
+    data = socket_to_concent.recv(1024)
+    logger.info(data)
+    socket_to_concent.close()
+
+
+def _parse_arguments():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('concent_cluster_address', help='Address of a Concent cluster')
+
+    return parser.parse_args()
```
kbeker: I think good thing might be validation to address which is giving by this function

rwrzesien: I will add it in https://github.com/golemfactory/concent/pull/634, this is just a prove of concept. Btw. how would you like this validation to be done ? Validate if is string ? Or if it is a host or ip ?

signing_service/src/signing_service.py
```diff
@@ -0,0 +1,34 @@
...
+import socket
+
+
+logger = logging.getLogger()
+crash_logger = logging.getLogger('crash')
+
+
+def main(concent_cluster_address):
+    address, port = concent_cluster_address.split(':')
+    port = int(port)
```
cameel: I assumed we would not have to parse the address. I don't want to go into various special cases like missing port, etc.

Let's switch to separate parameters for address and port. I'll update the issue.

signing_service/src/signing_service.py
```diff
@@ -0,0 +1,34 @@
...
+
+    socket_to_concent = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    socket_to_concent.connect((address, port))
+
+    socket_to_concent.send(b'GET / HTTP/1.0\r\n\r\n')
```
cameel: Maybe let's not pretend that it's HTTP. Might be confusing during debugging. Please just send some random sentence.

signing_service/src/signing_service.py
cameel: Validation would be a good idea. But if you don't have validation please at least catch and handle the exception that is raised by the socket when you give it an invalid address

This issue was meant to be simple so you don't have to do it here. In #634 is fine.

signing_service/tests/tests.py
```diff
@@ -0,0 +1,4 @@
+
+
+def test_that_full_check_will_run_this_test():
+    assert True
```
cameel: Please include at least rudimentary tests.

signing_service/src/signing_service.py
rwrzesien: Yes, I prefer to do it in #634

signing_service/src/signing_service.py
rwrzesien: Can it also be done in #634 ?

signing_service/src/signing_service.py
rwrzesien: Concent, as a HTTP server, is not responding to a random sentence.

signing_service/tests/tests.py
rwrzesien: Can it also be done in #634 ? Actually #634 changes it completely so all tests written here will have to be removed in #634.

signing_service/src/signing_service.py
cameel: It's a simple change. Let's do it here.

I'd prefer not to add unnecessary noise to #634.

signing_service/src/signing_service.py
cameel: Oh, so you're abusing a HTTP sever to test it instead of writing a custom server :) Clever. It's fine then.

signing_service/tests/tests.py
cameel: But then how do you know this code works? We don't know when #634 will pass the review. I'd really prefer each task to be self-contained. Maybe at least one tiny test? :)

signing_service/src/signing_service.py
rwrzesien: I have written custom one for #634 ;) And issue actually mentions that it should communicate with the Concent.

cameel: With Concent cluster. And Middleman is running as a service on the Concent cluster.

cameel: From the perspective of the Signing Service it does not really matter what it talks to. Only that it's using the right protocol. And that protocol is not HTTP :)

rwrzesien: Right, but it is not available yet. Anyway, I have custom server for #634 . I also have an idea to talk about on retro to create a placeholder for reusable scripts, commands and tools like this that are not in our repository.

signing_service/tests/tests.py
rwrzesien: Ok, I will add some.

signing_service/src/signing_service.py
cameel: If it's really reusable then just create a directory called `tools/` and put it there. I don't think this would be a controversial thing go do unless you're adding something that is huge or is really a one-off.

rwrzesien: I actually mean something more like drafts, not necessary matching our requirements like coding style etc. Also, commands for testing (or set of commands) like longer `curl` requests etc. Generally everything that one developer has on his local machine and other developer might find useful. But lets discuss this on retro.

signing_service/src/signing_service.py
rwrzesien: :+1 done

signing_service/tests/tests.py
rwrzesien: :+1 done

signing_service/src/signing_service.py
rwrzesien: :+1: done

signing_service/src/tests/tests.py
```diff
@@ -0,0 +1,59 @@
...
+import sys
+
+from ..constants import SIGNING_SERVICE_DEFAULT_PORT
+from ..signing_service import _parse_arguments
+from ..signing_service import main
+
+
+class SigningServiceMainTestCase(TestCase):
+
+    def test_that_main_should_called_with_correct_address(self):
```
dybi: I would suggest sth like `test_that_wrong_ip_address_causes_system_exit`

signing_service/src/tests/tests.py
```diff
@@ -0,0 +1,59 @@
...
+        with self.assertRaises(SystemExit):
+            main('not_really_ip_or_host', 8000)
+
+    def test_that_main_should_exit_when_called_with_wrong_address(self):
```
dybi: Hm... shouldn't it be the name for the test above? ;)

signing_service/src/tests/tests.py
```diff
@@ -0,0 +1,59 @@
...
+        mock_socket_send.assert_called_once()
+        mock_socket_recv.assert_called_once()
+        mock_socket_close.assert_called_once()
+
+
+class SigningServiceParseArgumentsTestCase(TestCase):
+
+    def setUp(self):
+        super().setUp()
+        sys.argv = sys.argv[:1]
```
dybi: I am not sure if understand what is it for ;)

rwrzesien: `ArgumentParser` takes values directly from `sys.argv`, but the test runner has its own arguments, so they have to be replaced.

signing_service/src/tests/tests.py
rwrzesien: Yes :)

signing_service/src/tests/tests.py
rwrzesien: It can be a host name also.

signing_service/src/tests/tests.py
cameel: This should be commented in the code.

signing_service/src/signing_service.py
```diff
@@ -0,0 +1,45 @@
+import argparse
```
cameel: One thing I noticed today when defining https://github.com/golemfactory/concent-deployment/issues/185. You don't have any top-level runner script. I think that having something simple directly inside the `signing_service/` directory would make it easier to discover how to run it.

signing_service/src/constants.py
```diff
@@ -0,0 +1 @@
+SIGNING_SERVICE_DEFAULT_PORT = 8888
```
cameel: Looks like I have left my comment in the wrong pull request. See https://github.com/golemfactory/concent/pull/634/files#r202398516

> Middleman uses 8888 for accepting connections from Concent (the internal port). You should use the external port there.

rwrzesien: But is is not defined yet neither in any pull request and it doesn't have default in https://github.com/golemfactory/concent/issues/616. Which one should I use then ?

signing_service/src/signing_service.py
rwrzesien: Do you mean shell script ? If yes, then I have added very simple one.

cameel: Can be a python script, can be a shell script. I just mean any obvious script at the top level because right now someone who wants to run the application has to go into `src/` and figure out which one should be used.

`run.sh` is OK but maybe `signing_service.sh` would be a better name?

signing_service/src/constants.py
```diff
@@ -0,0 +1 @@
+SIGNING_SERVICE_DEFAULT_PORT = 8889
```
cameel: I finally had to think about the ports to define the deployment (https://github.com/golemfactory/concent-deployment/issues/187). I settled upon 9055 for the external port and 9054 for the internal one. These ports are unassigned and do not seem to be in wide use. Please use them as defaults.

cameel:
> And yes, logging is working fine, but I had no chance to test logging to sentry without setting DSN.

I missed this earlier for some reason. Yeah, we need the service to accept the DSN on the command line. I have updated the issue.

signing_service/src/constants.py
cameel: Use 9055 (see https://github.com/golemfactory/concent/pull/619/files#r202727409).

signing_service/src/signing_service.py
rwrzesien: :+1:  done

signing_service/src/constants.py
rwrzesien: :+1:  done

rwrzesien:
> I missed this earlier for some reason. Yeah, we need the service to accept the DSN on the command line. I have updated the issue.
:+1:  done . However I have to manually create `raven.Client` with DSN passed from command line and replace with it the one created with `SentryHandler` when it is created from logging config file. I hope it will work but I have a feeling that I am doing something untypical. To verify this we can either test it on cluster or you can give me DSN for testing.

signing_service/signing_service.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
bartoszbetka: Consider add `-e` flag.

cameel:
@rwrzesien I think that's how you're supposed to use it if you want to add if after logging configuration has already been loaded.

The alternative is to add `dsn` to the logging config:

``` python
    'handlers': {
        'sentry': {
            'level': 'ERROR',
            'class': 'raven.handlers.logging.SentryHandler',
            'dsn': 'https://<key>:<secret>@sentry.io/<project>',
            },
        },
```

See [Sentry > SDKs > Python > Integrations > Logging](https://docs.sentry.io/clients/python/integrations/logging/).

rwrzesien:
@cameel It works with current implementation
https://talkback.golem.network/sentry/concent/issues/28910/activity/

I was trying to avoid using this in config, and it looks like we won't have to.

signing_service/src/tests/tests.py

signing_service/src/constants.py
rwrzesien: :+1: done

Issue #623, opened 2018-07-10
Signing Service: Error handling and reconnections


cameel:
This is a part of the implementation of the "Concent Signing Service" blueprint (#597). This task builds upon #599.

#### Overview
The purpose of this task is to turn the simple client application build in #599 into a service that initiates the connection but then listens for incoming data. It should also handle errors properly and reestablish the connection if it gets interrupted.

#### Command-line parameters
The following new parameters should be added:
- Retry parameters - delay, number of attempts, etc.
    - Depends on how you implement the retry. I recommend a delay that doubles after each unsuccessful attempt and is reset after success.

#### Communication with Concent
1. Initiate a TCP connection to the Concent cluster
    - If you can't connect or the connection drops, retry, possibly with some configurable delay between attempts.
2. When a connection is established run the connection handler
    - An exception that escapes from the handler should interrupt the current connection.
3. Handle exceptions and signals
    - If the connection is interrupted due to a failure (e.g. disconnected cable), don't report it. Just reconnect.
    - If there was an unrecognized exception, log it and report to Sentry. Then reraise it and crash.
    - If a shutdown signal (SIGTERM) or Ctrl+C (`KeyboardInterrupt`) is caught, terminate the connection and exit without errors.

#### Connection handler
1. Read data from the connection
2. Echo data back over the connection
3. Handle exceptions
    - When the other side closes the connection, return without errors

#### asyncio
The client is completely synchronous but it may still be a good idea to use [asyncio](https://docs.python.org/3/library/asyncio.html) for blocking operations, like listening for data coming over a socket.

See also #615 and #616 which implement the similar mechanism on the server side.

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/158969952

rwrzesien:
@cameel 

> An exception that escapes from the handler should interrupt the current connection.

What do you mean by that ? Should it end process gracefully or close current socket and reopen a new one ?

cameel:
Update: I made it clear that the service should crash when it encounters an unexpected error. Please see also the updated blueprint for information about how the service should handle errors.

cameel:
> What do you mean by that ? Should it end process gracefully or close current socket and reopen a new one ?

I mean that it should end gracefully but after cleaning up still reraise the exception and let the service crash. Don't bother reconnecting. The main loop should send a crash report to Sentry but after that let the exception escape.

[...]

PR #634, opened 2018-07-11
Feature signing service error handling and reconnections

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/623

### DIFF
--- a/signing_service/src/constants.py
+++ b/signing_service/src/constants.py
@@ -1,1 +1,32 @@
+import socket
+
 SIGNING_SERVICE_DEFAULT_PORT = 9055
+
+SIGNING_SERVICE_MAXIMUM_RECONNECT_TIME = 2**6
+
+SIGNING_SERVICE_RECOVERABLE_ERRORS = [
+    # 86 Streams pipe error
+    socket.errno.ESTRPIPE,  # type: ignore
+    # 90 Message too long
+    socket.errno.EMSGSIZE,  # type: ignore
+    # 100 Network is down
+    socket.errno.ENETDOWN,  # type: ignore
+    # 101 Network is unreachable
+    socket.errno.ENETUNREACH,  # type: ignore
+    # 102 Network dropped connection on reset
+    socket.errno.ENETRESET,  # type: ignore
+    # 103 Software caused connection abort
+    socket.errno.ECONNABORTED,  # type: ignore
+    # 104 Connection reset by peer
+    socket.errno.ECONNRESET,  # type: ignore
+    # 108 Cannot send after transport endpoint shutdown
+    socket.errno.ESHUTDOWN,  # type: ignore
+    # 110 Connection timed out
+    socket.errno.ETIMEDOUT,  # type: ignore
+    # 111 Connection refused
+    socket.errno.ECONNREFUSED,  # type: ignore
+    # 112 Host is down
+    socket.errno.EHOSTDOWN,  # type: ignore
+    # 113 No route to host
+    socket.errno.EHOSTUNREACH,  # type: ignore
+]
--- a/signing_service/src/signing_service.py
+++ b/signing_service/src/signing_service.py
@@ -1,50 +1,153 @@
 import argparse
 import logging.config
 import os
+import signal
 import socket
+from contextlib import closing
+from time import sleep
 
+from exceptions import SigningServiceValidationError  # pylint: disable=no-name-in-module
 from raven import Client
 
 from constants import SIGNING_SERVICE_DEFAULT_PORT  # pylint: disable=no-name-in-module
+from constants import SIGNING_SERVICE_MAXIMUM_RECONNECT_TIME  # pylint: disable=no-name-in-module
+from constants import SIGNING_SERVICE_RECOVERABLE_ERRORS  # pylint: disable=no-name-in-module
 
 
 logger = logging.getLogger()
 crash_logger = logging.getLogger('crash')
 
 
-def main(host, port):
-    socket_to_concent = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+class SigningService:
+    """
+    The Signing Service connects to Middleman as a client but then listens for requests coming from Concent via
+    Middleman. The underlying protocol is TCP and data sent over that is expected to conform to the Wire protocol.
+    """
 
-    try:
-        socket_to_concent.connect((host, port))
-    except socket.gaierror as exception:
-        logger.error(f'Exception occurred when connecting to {host}:{port}: {exception}')
-        exit(f'Exception occurred when connecting to {host}:{port}: {exception}')
+    __slots__ = (
+        'host',
+        'port',
+        'initial_reconnect_delay',
+        'current_reconnect_delay',
+        'socket',
+        'was_sigterm_caught',
+    )
 
-    socket_to_concent.send(b'GET / HTTP/1.0\r\n\r\n')
-    data = socket_to_concent.recv(1024)
-    logger.info(data)
-    socket_to_concent.close()
+    def __init__(self, host, port, initial_reconnect_delay) -> None:
+        assert isinstance(host, str)
+        assert isinstance(port, int)
+        assert isinstance(initial_reconnect_delay, int)
+        self.host = host
+        self.port = port
+        self.initial_reconnect_delay = initial_reconnect_delay
+        self.current_reconnect_delay = None
+        self.was_sigterm_caught = False
+
+        self._validate_arguments()
+
+    def run(self) -> None:
+        """
+        Handles main connection loop and its exceptions.
+
+        If connection is interrupted due to a failure, it should try to reconnect.
+        If connection is closed by other side, exit gracefully.
+        If a shutdown signal or KeyboardInterrupt is caught, exit gracefully.
+        If there was an unrecognized exception, it logs it and report to Sentry, then reraise and crash.
+        """
+
+        def _set_was_sigterm_caught_true(signum, frame):  # pylint: disable=unused-argument
+            logger.info('Closing connection and exiting on SIGTERM.')
+            self.was_sigterm_caught = True
+
+        # Handle shutdown signal.
+        signal.signal(signal.SIGTERM, _set_was_sigterm_caught_true)
+
+        while not self._was_sigterm_caught():
+            with closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as tcp_socket:
+                try:
+                    self._connect(tcp_socket)
+                except socket.error as exception:
+                    logger.error(f'Socket error occurred: {exception}')
+
+                    # Only predefined list of exceptions should cause reconnect, others should be reraised.
+                    if isinstance(exception.args, tuple) and exception.args[0] not in SIGNING_SERVICE_RECOVERABLE_ERRORS:  # type: ignore
+                        raise
+
+                    # Increase delay and reconnect if the connection is interrupted due to a failure.
+                    self._increase_delay()
+                except KeyboardInterrupt:
+                    # Handle keyboard interrupt.
+                    logger.info('Closing connection and exiting on KeyboardInterrupt.')
+                    break
+                except Exception as exception:
+                    # If there was an unrecognized exception, log it and report to Sentry.
+                    crash_logger.error(f'Unrecognized exception occurred: {exception}')
+                    raise
+
+    def _connect(self, tcp_socket: socket.socket) -> None:
+        """ Creates socket and connects to given HOST and PORT. """
+        if self.current_reconnect_delay is not None:
+            logger.info(f'Waiting {self.current_reconnect_delay} before connecting.')
+            sleep(self.current_reconnect_delay)
+
+        logger.info(f'Connecting to {self.host}:{self.port}.')
+        tcp_socket.connect((self.host, self.port))
+        logger.info(f'Connection established.')
+
+        # Reset delay on successful connection and set flag that connection is established.
+        self.current_reconnect_delay = None
+        self._handle_connection(tcp_socket)
+
+    def _handle_connection(self, tcp_socket: socket.socket) -> None:  # pylint: disable=no-self-use
+        """ Inner loop that handles data exchange over socket. """
+        while True:
+            data = tcp_socket.recv(1024)
+            tcp_socket.send(data)
+
+    def _increase_delay(self) -> None:
+        """ Increase current delay if connection cannot be established. """
+        assert self.current_reconnect_delay is None or self.current_reconnect_delay >= 0
+
+        if self.current_reconnect_delay is None:
+            self.current_reconnect_delay = self.initial_reconnect_delay  # type: ignore
+        else:
+            self.current_reconnect_delay = min(self.current_reconnect_delay * 2, SIGNING_SERVICE_MAXIMUM_RECONNECT_TIME)
+
+    def _was_sigterm_caught(self):
+        """ Helper function which checks if SIGTERM signal was caught. """
+        return self.was_sigterm_caught
+
+    def _validate_arguments(self):
+        if not 0 < self.port < 65535:
+            raise SigningServiceValidationError('Port must be 0-65535.')
+
+        if self.initial_reconnect_delay < 0:
+            raise SigningServiceValidationError('reconnect_delay must be non-negative integer.')
 
 
-def _parse_arguments():
+def _parse_arguments() -> argparse.Namespace:
     parser = argparse.ArgumentParser()
     parser.add_argument(
         'concent_cluster_host',
-        help='Host or IP address of a service on Concent cluster, to which SigningService connects over TCP.'
+        help='Host or IP address of a service on Concent cluster, to which SigningService connects over TCP.',
+    )
+    parser.add_argument(
+        'initial_reconnect_delay',
+        type=int,
+        help='Initial delay between reconnections, doubles after each unsuccessful attempt and is reset after success.',
     )
     parser.add_argument(
         '--concent-cluster-port',
         default=SIGNING_SERVICE_DEFAULT_PORT,
         dest='concent_cluster_port',
         type=int,
-        help=f'Port on which Concent cluster is listening (default: {SIGNING_SERVICE_DEFAULT_PORT}).'
+        help=f'Port on which Concent cluster is listening (default: {SIGNING_SERVICE_DEFAULT_PORT}).',
     )
     parser.add_argument(
         '--sentry-dsn',
         default='',
         dest='sentry_dsn',
-        help=f'Sentry DSN for error reporting.'
+        help=f'Sentry DSN for error reporting.',
     )
 
     return parser.parse_args()
@@ -52,9 +155,15 @@ def _parse_arguments():
 
 if __name__ == '__main__':
     logging.config.fileConfig(os.path.join(os.path.dirname(os.path.abspath(__file__)), '..', 'logging.ini'))
+
+    # Parse required arguments.
     args = _parse_arguments()
 
     raven_client = Client(dsn=args.sentry_dsn)
     crash_logger.handlers[0].client = raven_client  # type: ignore
 
-    main(args.concent_cluster_host, args.concent_cluster_port)
+    arg_host = args.concent_cluster_host
+    arg_port = args.concent_cluster_port
+    arg_initial_reconnect_delay = args.initial_reconnect_delay
+
+    SigningService(arg_host, arg_port, arg_initial_reconnect_delay).run()
--- a/signing_service/src/tests/tests.py
+++ b/signing_service/src/tests/tests.py
@@ -1,30 +1,107 @@
 from unittest import TestCase
+import socket
 import sys
 
 import mock
 
 from ..constants import SIGNING_SERVICE_DEFAULT_PORT
+from ..constants import SIGNING_SERVICE_RECOVERABLE_ERRORS
+from ..constants import SIGNING_SERVICE_MAXIMUM_RECONNECT_TIME
 from ..signing_service import _parse_arguments
-from ..signing_service import main
+from ..signing_service import SigningService
 
 
 class SigningServiceMainTestCase(TestCase):
 
-    def test_that_main_should_exit_when_called_with_wrong_host(self):
-        with self.assertRaises(SystemExit):
-            main('not_really_ip_or_host', 8000)
+    def setUp(self):
+        self.host = '127.0.0.1'
+        self.port = 8000
+        self.initial_reconnect_delay = 2
 
-    def test_that_main_should_work_correctly_when_called_with_correct_parameters(self):  # pylint: disable=no-self-use
+    def test_that_signing_service_should_be_instantiated_correctly_with_all_parameters(self):
+        signing_service = SigningService(self.host, self.port, self.initial_reconnect_delay)
+
+        self.assertIsInstance(signing_service, SigningService)
+        self.assertEqual(signing_service.host, self.host)
+        self.assertEqual(signing_service.port, self.port)
+        self.assertEqual(signing_service.initial_reconnect_delay, self.initial_reconnect_delay)
+
+    def test_that_signing_service_should_run_full_loop_when_instantiated_with_all_parameters(self):
         with mock.patch('socket.socket.connect') as mock_socket_connect:
-            with mock.patch('socket.socket.send') as mock_socket_send:
-                with mock.patch('socket.socket.recv') as mock_socket_recv:
-                    with mock.patch('socket.socket.close') as mock_socket_close:
-                        main('127.0.0.1', 8000)
+            with mock.patch('src.signing_service.SigningService._handle_connection') as mock__handle_connection:
+                with mock.patch('socket.socket.close') as mock_socket_close:
+                    with mock.patch('src.signing_service.SigningService._was_sigterm_caught', side_effect=[False, True]):
+                        signing_service = SigningService(self.host, self.port, self.initial_reconnect_delay)
+                        signing_service.run()
 
         mock_socket_connect.assert_called_once_with(('127.0.0.1', 8000))
-        mock_socket_send.assert_called_once()
-        mock_socket_recv.assert_called_once()
         mock_socket_close.assert_called_once()
+        mock__handle_connection.assert_called_once()
+
+    def test_that_signing_service_should_exit_gracefully_on_keyboard_interrupt(self):
+        with mock.patch('socket.socket.connect', side_effect=KeyboardInterrupt()) as mock_socket_connect:
+            with mock.patch('socket.socket.close') as mock_socket_close:
+                signing_service = SigningService(self.host, self.port, self.initial_reconnect_delay)
+                signing_service.run()
+
+        mock_socket_connect.assert_called_once_with(('127.0.0.1', 8000))
+        mock_socket_close.assert_called_once()
+
+    def test_that_signing_service_should_reraise_unrecognized_exception(self):
+        with mock.patch('socket.socket.connect', side_effect=Exception()) as mock_socket_connect:
+            with mock.patch('socket.socket.close') as mock_socket_close:
+                signing_service = SigningService(self.host, self.port, self.initial_reconnect_delay)
+                with self.assertRaises(Exception):
+                    signing_service.run()
+
+        mock_socket_connect.assert_called_once_with(('127.0.0.1', 8000))
+        mock_socket_close.assert_called_once()
+
+    def test_that_signing_service_should_reconnect_when_expected_socket_error_was_caught(self):
+        assert socket.errno.ECONNREFUSED in SIGNING_SERVICE_RECOVERABLE_ERRORS
+
+        with mock.patch('socket.socket.connect', side_effect=socket.error(socket.errno.ECONNREFUSED)) as mock_socket_connect:
+            with mock.patch('src.signing_service.SigningService._was_sigterm_caught', side_effect=[False, False, True]):
+                signing_service = SigningService(self.host, self.port, self.initial_reconnect_delay)
+                signing_service.run()
+
+        self.assertEqual(mock_socket_connect.call_count, 2)
+
+    def test_that_signing_service_should_reraise_different_socket_erros(self):
+        assert socket.errno.EBUSY not in SIGNING_SERVICE_RECOVERABLE_ERRORS
+
+        with mock.patch('socket.socket.connect', side_effect=socket.error(socket.errno.EBUSY)) as mock_socket_connect:
+            with mock.patch('socket.socket.close') as mock_socket_close:
+                signing_service = SigningService(self.host, self.port, self.initial_reconnect_delay)
+                with self.assertRaises(socket.error):
+                    signing_service.run()
+
+        mock_socket_connect.assert_called_once_with(('127.0.0.1', 8000))
+        mock_socket_close.assert_called_once()
+
+
+class SigningServiceIncreaseDelayTestCase(TestCase):
+
+    def setUp(self):
+        self.signing_service = SigningService('127.0.0.1', 8000, 2)
+
+    def test_that_initial_reconnect_delay_should_be_set_to_passed_value(self):
+        self.assertEqual(self.signing_service.current_reconnect_delay, None)
+        self.assertEqual(self.signing_service.initial_reconnect_delay, 2)
+
+    def test_that_current_reconnect_delay_should_be_set_to_reconnect_delay_after_first_call_to__increase_delay(self):
+        self.signing_service._increase_delay()
+        self.assertEqual(self.signing_service.current_reconnect_delay, 2)
+
+    def test_that_current_reconnect_delay_should_be_doubled_after_next_call_to__increase_delay(self):
+        self.signing_service._increase_delay()
+        self.signing_service._increase_delay()
+        self.assertEqual(self.signing_service.current_reconnect_delay, 2 * 2)
+
+    def test_that_current_reconnect_delay_should_be_set_to_allowed_maximum_after_it_extends_it(self):
+        self.signing_service.current_reconnect_delay = SIGNING_SERVICE_MAXIMUM_RECONNECT_TIME - 1
+        self.signing_service._increase_delay()
+        self.assertEqual(self.signing_service.current_reconnect_delay, SIGNING_SERVICE_MAXIMUM_RECONNECT_TIME)
 
 
 class SigningServiceParseArgumentsTestCase(TestCase):
@@ -36,7 +113,7 @@ class SigningServiceParseArgumentsTestCase(TestCase):
         sys.argv = sys.argv[:1]
 
     def test_that_argument_parser_should_parse_correct_input(self):
-        sys.argv += ['127.0.0.1', '--concent-cluster-port', '8000']
+        sys.argv += ['127.0.0.1', '1', '--concent-cluster-port', '8000']
 
         args = _parse_arguments()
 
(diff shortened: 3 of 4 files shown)


signing_service/tests/tests.py
```diff
@@ -0,0 +1,4 @@
+
+
+def test_that_full_check_will_run_this_test():
+    assert True
```
dybi: Actually there could be some unit tests for the following cases:

- If connection is interrupted due to a failure, it should try to reconnect.
- If connection is closed by other side, exit gracefully.
- If a shutdown signal or KeyboardInterrupt is caught, exit gracefully.
- If there was an unrecognized exception, it logs it and report to Sentry, then reraise and crash.

signing_service/src/signing_service.py
```diff
@@ -0,0 +1,139 @@
...
+        arg_host, arg_port = args.concent_cluster_address.split(':')
+    except ValueError:
+        exit('`concent_cluster_address` must be in format HOST:PORT.')
+
+    try:
+        arg_port = int(arg_port)
+    except ValueError:
+        exit('Port must be integer.')
+
+    arg_reconnect_delay = args.reconnect_delay
```
dybi: this could be checked if it is convertible to `int` as well

signing_service/src/signing_service.py
```diff
@@ -0,0 +1,139 @@
...
+
+        # Reset delay on successful connection and set flag that connection is established.
+        self.current_reconnect_delay = 0
+        self.is_connection_established = True
+        self._handle_connection()
+
+    def _handle_connection(self) -> None:
+        """ Inner loop that handles data exchange over socket. """
+        while True:
+            data = self.socket.recv(1024)
```
dybi: you could use `asyncio` coroutines here

signing_service/src/signing_service.py
```diff
@@ -0,0 +1,139 @@
...
+                crash_logger.error(str(exception))
+                raise
+
+    def _connect(self) -> None:
+        """ Creates socket and connects to given HOST and PORT. """
+        self.socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+
+        if self.current_reconnect_delay > 0:
+            logger.warning(f'Waiting {self.current_reconnect_delay} before connecting.')
+            sleep(self.current_reconnect_delay)
```
dybi: you can use `asyncio.sleep()` here

signing_service/src/signing_service.py
```diff
@@ -0,0 +1,139 @@
...
+        self.port = port
+        self.reconnect_delay = reconnect_delay
+        self.current_reconnect_delay = 0
+        self.socket: socket.socket = None  # type: ignore
+        self.is_connection_established = False
+
+        # Handle shutdown signal.
+        signal.signal(signal.SIGTERM, self._terminate_connection_and_exit)  # type: ignore
+
+        self._run()
```
dybi: personaly, I would remove it from `__init__` and call it explicitly (`SigningService(arg_host, arg_port, arg_reconnect_delay).run()`)

signing_service/src/signing_service.py
```diff
@@ -0,0 +1,139 @@
...
+        'concent_cluster_address',
+        help='Address of a Concent cluster.',
+    )
+    parser.add_argument(
+        'reconnect_delay',
+        type=int,
+        help='Delay between reconnections, doubles after each unsuccessful attempt and is reset after success.',
+    )
+
+    return parser.parse_args()
```
kbeker: I would suggest to do some validation, because `-80` is still int, but negative and we can't have such port. Same thing with `reconnect_delay`

signing_service/src/signing_service.py
rwrzesien: Yes, will be added.

signing_service/src/signing_service.py
rwrzesien: Yes, will be added.

signing_service/src/signing_service.py
rwrzesien: I will add it. Please note that without it will still crash instantly when trying to connect with `OverflowError: getsockaddrarg: port must be 0-65535.`. But I agree its better to handle it.

signing_service/src/signing_service.py
rwrzesien: It is validated in 
```
parser.add_argument(
        'reconnect_delay',
        type=int,
        help='Delay between reconnections, doubles after each unsuccessful attempt and is reset after success.',
    )
```

signing_service/src/signing_service.py
rwrzesien: I can do this but could you provide some argumentation for this ?

signing_service/tests/tests.py
rwrzesien: I thought no one will notice ;) Yes, most of this can be tested but a good integration tests are not possible in this case I think. But I will add those which are possible.

dybi: at least some unit tests with mocks ;)

signing_service/src/signing_service.py
dybi: :+1:

signing_service/src/signing_service.py
dybi: It is more visible, that you actually start the service. If start is automatic (you cannot create object without it), it is less flexible and might be misleading sometimes.

signing_service/src/signing_service.py
```diff
@@ -0,0 +1,139 @@
...
+        self.is_connection_established = True
+        self._handle_connection()
+
+    def _handle_connection(self) -> None:
+        """ Inner loop that handles data exchange over socket. """
+        while True:
+            data = self.socket.recv(1024)
+            self.socket.send(data)
+
+    def _terminate_connection_and_exit(self) -> None:
```
dybi: https://stackoverflow.com/questions/18499497/how-to-process-sigterm-signal-gracefully - it will be called with 3 args: `self`, `signum` and `frame`, please change the sinature ;)

rwrzesien: Right, that was the part I didn't tested, good catch.

signing_service/src/signing_service.py
rwrzesien: Ok, sounds good, will do.

signing_service/src/signing_service.py
```diff
@@ -0,0 +1,139 @@
...
+
+        If connection is interrupted due to a failure, it should try to reconnect.
+        If connection is closed by other side, exit gracefully.
+        If a shutdown signal or KeyboardInterrupt is caught, exit gracefully.
+        If there was an unrecognized exception, it logs it and report to Sentry, then reraise and crash.
+        """
+        while True:
+            try:
+                if self.is_connection_established is False:
+                    self._connect()
```
cameel: Do we really need a flag for this? Can't this be checked on the socket object?

signing_service/src/signing_service.py
```diff
@@ -0,0 +1,139 @@
...
+        self.socket.close()
+        logger.info('Closing connection and exiting.')
+        exit()
+
+    def _increase_delay(self) -> None:
+        """ Increase current delay if connection cannot be established. """
+        if self.current_reconnect_delay == 0:
+            self.current_reconnect_delay = self.reconnect_delay
+        else:
+            self.current_reconnect_delay *= 2
```
cameel: Maybe let's add a maximum too?

signing_service/src/signing_service.py
```diff
@@ -0,0 +1,139 @@
...
+        If connection is closed by other side, exit gracefully.
+        If a shutdown signal or KeyboardInterrupt is caught, exit gracefully.
+        If there was an unrecognized exception, it logs it and report to Sentry, then reraise and crash.
+        """
+        while True:
+            try:
+                if self.is_connection_established is False:
+                    self._connect()
+            except socket.error as exception:
+                logger.error(f'Socket error occurred: {exception}')
```
cameel: Do all possible `socket.error` exceptions represent connection errors? I think we want to handle only a few specific `socket.errno` values and reraise all the others (handle them like the `Exception` handler below does).

signing_service/src/signing_service.py
```diff
@@ -0,0 +1,139 @@
...
+            sleep(self.current_reconnect_delay)
+
+        logger.warning(f'Connecting to {self.host}:{self.port}.')
+        self.socket.connect((self.host, self.port))
+        logger.warning(f'Connection established.')
+
+        # Reset delay on successful connection and set flag that connection is established.
+        self.current_reconnect_delay = 0
+        self.is_connection_established = True
+        self._handle_connection()
```
cameel: I think it would be better to pass the socket via a parameter instead of storing it in the instance. The code is easier to debug and understand if things are passed directly via parameters rather than as instance fields.

signing_service/tests/tests.py
rwrzesien: Do you mean tests that will check communication between SigningService, Middleman and Concent ? Something like our `cluster` tests ? I will add the issue but please clarify this is what should be in it.

signing_service/src/signing_service.py
rwrzesien: It is already done in https://github.com/golemfactory/concent/pull/619, I will rebase.

signing_service/src/signing_service.py
```diff
@@ -0,0 +1,139 @@
...
+def _parse_arguments() -> argparse.Namespace:
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        'concent_cluster_address',
+        help='Address of a Concent cluster.',
+    )
+    parser.add_argument(
+        'reconnect_delay',
+        type=int,
+        help='Delay between reconnections, doubles after each unsuccessful attempt and is reset after success.',
```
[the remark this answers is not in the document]
rwrzesien: :+1: done

signing_service/src/signing_service.py
```diff
@@ -0,0 +1,139 @@
...
+            self.current_reconnect_delay = self.reconnect_delay
+        else:
+            self.current_reconnect_delay *= 2
+
+
+def _parse_arguments() -> argparse.Namespace:
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        'concent_cluster_address',
+        help='Address of a Concent cluster.',
```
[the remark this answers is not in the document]
rwrzesien: :+1: done

signing_service/src/signing_service.py
```diff
@@ -0,0 +1,139 @@
...
+        """ Inner loop that handles data exchange over socket. """
+        while True:
+            data = self.socket.recv(1024)
+            self.socket.send(data)
+
+    def _terminate_connection_and_exit(self) -> None:
+        """ Close socket and exit. """
+        self.socket.close()
+        logger.info('Closing connection and exiting.')
+        exit()
```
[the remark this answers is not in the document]
rwrzesien: :+1: done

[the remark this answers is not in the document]
rwrzesien: :+1: done

signing_service/src/signing_service.py
```diff
@@ -0,0 +1,139 @@
...
+                self.is_connection_established = False
+            except KeyboardInterrupt:
+                # Handle keyboard interrupt.
+                self._terminate_connection_and_exit()
+            except Exception as exception:
+                # Try to close opened socket.
+                self.socket.close()
+                # If there was an unrecognized exception, log it and report to Sentry.
+                logger.error(f'Unrecognized exception occurred: {exception}')
+                crash_logger.error(str(exception))
```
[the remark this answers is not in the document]
rwrzesien: No, it doesn't. Why it should ? It is a job of root logger.

signing_service/src/signing_service.py
rwrzesien: :+1: done

signing_service/src/signing_service.py
rwrzesien: :+1: done

signing_service/src/signing_service.py
rwrzesien: :+1: done

signing_service/src/signing_service.py
rwrzesien: :+1: done

signing_service/src/signing_service.py
rwrzesien: Will this list be ok ?
```
74 Bad message
86 Streams pipe error
87 Too many users
90 Message too long
100 Network is down
101 Network is unreachable
102 Network dropped connection on reset
103 Software caused connection abort
104 Connection reset by peer
108 Cannot send after transport endpoint shutdown
110 Connection timed out
111 Connection refused
112 Host is down
113 No route to host
```

rwrzesien: Will this list be ok ?
```
74 Bad message
86 Streams pipe error
87 Too many users
90 Message too long
100 Network is down
101 Network is unreachable
102 Network dropped connection on reset
103 Software caused connection abort
104 Connection reset by peer
108 Cannot send after transport endpoint shutdown
110 Connection timed out
111 Connection refused
112 Host is down
113 No route to host
```

signing_service/src/signing_service.py
rwrzesien: Really ? I think there is nothing better then accessing stuff via `self.` :) You make functions signature smaller, you have better control on modification of that object (if you pass you have to think if it was mutable or not, and if you should return it or not to keep changes, then return it from function that called that function etc.).

And in scenario like this, when you have one global element in a standalone module and you don't reuse methods outside and you access that element from different functions, its even more usable.

cameel: I think it's easier to analyze and understand a function that is self-contained because you can limit yourself to the code of that function. Putting state on the instance means that it's no longer self-contained. Now you need to keep the whole class in your head as context.

For that reason I think it's better to limit the scope of each variable as much as possible. For example if you need a variable only in one function, it's better to declare it there rather than use a field on the instance. That's similar to the situation with globals and why we avoid them - the scope of a global variable is the whole program and when anything can access it, keeping track of it becomes very hard.

In this particular case having `socket` on the instance is simply unnecessary. It only exists while `run()` is running and outside it should be `None`. It would be different if you needed to preserve its value between the calls but that's not the case here. In a situation like this I think that keeping the data in the function and passing it around via parameters is a better design. You don't need stationary data - you can keep it all on the stack.

Overall I find classes that keep lots of state hard to debug. Often you stop inside in the debugger and have no way to know where the state came from. You can't go up the stack and check. It's simply there. You don't know how it got there and why. Stateless stuff that has only input and output and no side channels is simpler.

But anyway, that's my opinion. In this case the difference in scope is not that big so it's more a discussion in principle rather than a real issue. Let's see what @dybi has to say but I won't insist too much on this change because I don't want to block this pull request and also because I want to leave the internal design decisions like this up to you guys.

signing_service/src/signing_service.py
cameel: Seems fine, but I'm not sure about these two:
```
74 Bad message
87 Too many users
```

In general I'd err on the side of catching too little rather than too much. If you catch something you shouldn't, we may miss a bug. If you don't catch something you should, the only difference will be that the service will crash, send a report and get restarted by Kubernetes instead of retrying on its own.

signing_service/src/signing_service.py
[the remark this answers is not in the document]
cameel: I think in Django the `django.request` logger does.

I think it should. No need to log it twice. We want to log it once and see it both on the console and in Sentry.

But it's not a big issue. Just a possible improvement.

signing_service/src/constants.py
```diff
@@ -0,0 +1,3 @@
+SIGNING_SERVICE_DEFAULT_PORT = 8888
```
cameel: Middleman uses 8888 for accepting connections from Concent (the internal port). You should use the external port there.

signing_service/src/signing_service.py
```diff
@@ -0,0 +1,151 @@
...
+
+        def _set_is_sigterm_sent_true(signum, frame):  # pylint: disable=unused-argument
+            self.is_sigterm_sent = True
+
+        # Handle shutdown signal.
+        signal.signal(signal.SIGTERM, _set_is_sigterm_sent_true)
+
+        while True:
+            if self.is_sigterm_sent:
+                self.socket.close()
```
cameel: This is no longer an infinite loop. Maybe replace it with `while not self.is_sigterm_sent:`?

Then you could put the cleanup after the loop rather than at the beginning with a `break`.

signing_service/src/signing_service.py
```diff
@@ -0,0 +1,151 @@
...
+    """
+
+    def __init__(self, host, port, reconnect_delay) -> None:
+        assert isinstance(host, str)
+        assert isinstance(port, int)
+        assert isinstance(reconnect_delay, int)
+        self.host = host
+        self.port = port
+        self.reconnect_delay = reconnect_delay
+        self.current_reconnect_delay = 0
```
cameel: I think it might be a good idea to use `__slots__`. It lets you declare fields ahead of time and also has smaller memory footprint.

signing_service/src/signing_service.py
```diff
@@ -0,0 +1,151 @@
...
+        Handles main connection loop and its exceptions.
+
+        If connection is interrupted due to a failure, it should try to reconnect.
+        If connection is closed by other side, exit gracefully.
+        If a shutdown signal or KeyboardInterrupt is caught, exit gracefully.
+        If there was an unrecognized exception, it logs it and report to Sentry, then reraise and crash.
+        """
+
+        def _set_is_sigterm_sent_true(signum, frame):  # pylint: disable=unused-argument
+            self.is_sigterm_sent = True
```
cameel: Maybe `sigterm_caught` (or `was_sigterm_caught`, though it sounds weird with the prefix) would be a better name?

signing_service/src/signing_service.py
```diff
@@ -0,0 +1,151 @@
...
+        while True:
+            if self.is_sigterm_sent:
+                self.socket.close()
+                logger.info('Closing connection and exiting on SIGTERM.')
+                break
+            try:
+                if self.socket is None:
+                    self._connect()
+                else:
+                    self.socket.send(b'')
```
cameel: Why are you sending an empty string? I think there should be a comment here.

signing_service/src/signing_service.py
cameel: What if SIGTERM was received immediately, before the first cycle of the loop and the socket has not been opened yet? It may be `None`.

signing_service/src/signing_service.py
```diff
@@ -0,0 +1,151 @@
...
+                self._increase_delay()
+                self.socket = None
+            except KeyboardInterrupt:
+                # Handle keyboard interrupt.
+                self.socket.close()
+                logger.info('Closing connection and exiting on KeyboardInterrupt.')
+                break
+            except Exception as exception:
+                # Try to close opened socket.
+                self.socket.close()
```
cameel: You're doing `socket.close()` in all the cases. I think you should have a `try ... finally` block that closes the connection instead. It would be less repetitive and the cleanup would be guaranteed. I think that for example now `SystemExit` (which does not inherit from `Exception` but from `BaseException`) will bypass your handlers and exit without closing the socket.

signing_service/src/signing_service.py
```diff
@@ -0,0 +1,151 @@
...
+            self.socket.send(data)
+
+    def _increase_delay(self) -> None:
+        """ Increase current delay if connection cannot be established. """
+        if self.current_reconnect_delay == 0:
+            self.current_reconnect_delay = self.reconnect_delay
+        elif self.current_reconnect_delay < SIGNING_SERVICE_MAXIMUM_RECONNECT_TIME:
+            self.current_reconnect_delay *= 2
+        elif self.current_reconnect_delay >= SIGNING_SERVICE_MAXIMUM_RECONNECT_TIME:
+            self.current_reconnect_delay = SIGNING_SERVICE_MAXIMUM_RECONNECT_TIME
```
cameel: This could be simplified:
``` python
self.current_reconnect_delay = max(self.reconnect_delay * 2, SIGNING_SERVICE_MAXIMUM_RECONNECT_TIME)
```

signing_service/src/signing_service.py
```diff
@@ -0,0 +1,151 @@
...
+    def _handle_connection(self) -> None:
+        """ Inner loop that handles data exchange over socket. """
+        while True:
+            data = self.socket.recv(1024)
+            self.socket.send(data)
+
+    def _increase_delay(self) -> None:
+        """ Increase current delay if connection cannot be established. """
+        if self.current_reconnect_delay == 0:
+            self.current_reconnect_delay = self.reconnect_delay
```
cameel: I think that `None` would be a better initial value for `current_reconnect_delay` because trying to multiply it by 2 by mistake results in an error.

And I'd assert that it's never zero.

signing_service/src/signing_service.py
cameel: I'd rename `reconnect_delay` to `initial_reconnect_delay`. Otherwise, without context, these two variables sound like they're storing the same thing.

signing_service/src/signing_service.py
```diff
@@ -0,0 +1,151 @@
...
+
+    arg_host = args.concent_cluster_address
+    arg_port = args.concent_cluster_port
+    arg_reconnect_delay = args.reconnect_delay
+
+    if not 0 < arg_port < 65535:
+        exit('Port must be 0-65535.')
+
+    if arg_reconnect_delay < 0:
+        exit('reconnect_delay must be non-negative integer.')
```
cameel: Maybe we should have `_validate_arguments()`?

signing_service/tests/tests.py
rwrzesien: Created https://github.com/golemfactory/concent/issues/647

> Though if you can think of any good way to prepare integration tests like the ones we had initially for Concent (running on Django's test case) that would be nice too.

I think it might be done but when all the components get more mature.

signing_service/src/signing_service.py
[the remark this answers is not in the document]
rwrzesien: Actually it was logging, but I was creating it with wrong name. Fixed :+1:

signing_service/src/signing_service.py
rwrzesien: I agree, I will create a list without those two.

signing_service/src/constants.py
rwrzesien: I have answered in https://github.com/golemfactory/concent/pull/619#discussion_r202466505

signing_service/src/signing_service.py
rwrzesien: I am thinking if there is already enough code there to move it to separate function. Or maybe it should be a method of `SigningService` called in __init__ ?

signing_service/src/signing_service.py
rwrzesien: It can be, but actually in this way:
`min(self.current_reconnect_delay * 2, SIGNING_SERVICE_MAXIMUM_RECONNECT_TIME)`

rwrzesien: :+1: done

signing_service/src/signing_service.py
rwrzesien: :+1: done

signing_service/src/signing_service.py
rwrzesien: :+1: done

signing_service/src/signing_service.py
rwrzesien: `was_sigterm_caught` sounds better because it instantly gives you info that it is bool (or at least should be).
:+1: done

signing_service/src/signing_service.py
rwrzesien: Sending something is the only way to check if connection is still alive directly on the socket object. I will add a comment for this.

signing_service/src/signing_service.py
rwrzesien: :+1: done

signing_service/src/signing_service.py
rwrzesien: :+1: done

signing_service/src/signing_service.py
rwrzesien: :+1: done

signing_service/src/signing_service.py
rwrzesien: :+1: done

signing_service/src/signing_service.py
rwrzesien: :+1: done

signing_service/src/signing_service.py
rwrzesien: @dybi @cameel 
I have added it in separate commit https://github.com/golemfactory/concent/pull/634/commits/09b3193ecccff5257c1e4fd04b4e6274489f3b7d so you can either review it as a whole or only those changes. And I can drop it if we decide it to be not needed after all.

rwrzesien: @dybi @cameel Please remind me what we finally agreed on ? Because if it is up to me I would like not to add it as a part of this pull request, because
- it will not have big impact on how this client actually works looking from outside of it,
- it will be difficult to test it,
- it will require rewriting most of the current code.

So if we really want to do this, I suggest to do this in separate issue. If it is optional, I would skip it for now.

signing_service/src/signing_service.py
cameel: Maybe. But then it should raise exceptions and have its own, custom exception class. `exit()` should be only at the top level.

signing_service/src/signing_service.py
cameel: Yeah, sorry. I was thinking about defining a maximum but to do this you have to use `min()`, not `max()`. It often trips me up :)

signing_service/src/constants.py
```diff
@@ -0,0 +1,21 @@
+import socket
+
+
+SIGNING_SERVICE_DEFAULT_PORT = 8888
+
+SIGNING_SERVICE_MAXIMUM_RECONNECT_TIME = 2**6
+
+SIGNING_SERVICE_EXCEPTIONS_CAUGHT_TO_RECONNECT = [
```
cameel: Maybe `SIGNING_SERVICE_RECOVERABLE_ERRORS`?

signing_service/src/signing_service.py
```diff
@@ -0,0 +1,167 @@
...
+                logger.info('Closing connection and exiting on KeyboardInterrupt.')
+                break
+            except Exception as exception:
+                # If there was an unrecognized exception, log it and report to Sentry.
+                crash_logger.error(f'Unrecognized exception occurred: {exception}')
+                raise
+            finally:
+                if self.socket is not None:
+                    # Try to close opened socket.
+                    self.socket.close()
```
cameel: This comment is a bit too obvious :)

signing_service/src/signing_service.py
```diff
@@ -0,0 +1,167 @@
...
+
+        if self.was_sigterm_caught:
+            if self.socket is not None:
+                self.socket.close()
```
cameel: You would not need this if you put the whole loop in the `finally` block.

signing_service/src/signing_service.py
cameel: Separate issue is fine. It is optional so if you don't want it, you can skip it altogether, though I would recommend doing it just to practice using asyncio a bit.

signing_service/src/constants.py
rwrzesien: :+1: done

signing_service/src/signing_service.py
rwrzesien: :+1: done

signing_service/src/signing_service.py
rwrzesien: I am not sure if I follow you on this one. This is actually part of the loop.

signing_service/src/signing_service.py
rwrzesien: :+1: done

signing_service/src/tests/tests.py
```diff
@@ -0,0 +1,160 @@
...
+
+    def test_that_signing_service_should_be_instantiated_correctly_with_all_parameters(self):
+        signing_service = SigningService(self.host, self.port, self.initial_reconnect_delay)
+
+        self.assertIsInstance(signing_service, SigningService)
+        self.assertEqual(signing_service.host, self.host)
+        self.assertEqual(signing_service.port, self.port)
+        self.assertEqual(signing_service.initial_reconnect_delay, self.initial_reconnect_delay)
+
+    def test_that_signing_service_should_run_full_loop_when_be_instantiated_with_all_parameters(self):
```
dybi: please rename: remove `"_be_"`

signing_service/src/tests/tests.py
```diff
@@ -0,0 +1,160 @@
...
+        with mock.patch('socket.socket.connect') as mock_socket_connect:
+            with mock.patch('src.signing_service.SigningService._handle_connection') as mock__handle_connection:
```
dybi: double underscore in: `mock__handle_connection`

signing_service/src/tests/tests.py
```diff
@@ -0,0 +1,160 @@
...
+                with mock.patch('socket.socket.close') as mock_socket_close:
+                    with mock.patch('src.signing_service.SigningService._was_sigterm_caught', side_effect=[False, True]):
+                        signing_service = SigningService(self.host, self.port, self.initial_reconnect_delay)
+                        signing_service.run()
+
+        mock_socket_connect.assert_called_once_with(('127.0.0.1', 8000))
+        mock_socket_close.assert_called_once()
+        mock__handle_connection.assert_called_once()
+
+    def test_that_signing_service_should_run_send_check_if_connection_is_still_alive(self):
```
dybi: Could you explain what you mean by `run_send_check`?

signing_service/src/tests/tests.py
```diff
@@ -0,0 +1,160 @@
...
+                with mock.patch('socket.socket.send') as mock_socket_send:
+                    with mock.patch('socket.socket.close') as mock_socket_close:
+                        with mock.patch('src.signing_service.SigningService._was_sigterm_caught', side_effect=[False, False, True]):
+                            signing_service = SigningService(self.host, self.port, self.initial_reconnect_delay)
+                            signing_service.run()
+
+        mock_socket_connect.assert_called_once_with(('127.0.0.1', 8000))
+        mock_socket_send.assert_called_once_with(b'')
+        mock__handle_connection.assert_called_once()
+        self.assertEqual(mock_socket_close.call_count, 2)
```
dybi: why is socket closed 2 times?

signing_service/src/signing_service.py
cameel: I mean that the the whole loop should be inside a `try...finally` block. You should start such a block immediately after acquiring a resource. If you did this, this `close()` would not be necessary. The block would handle that for you.

And now that I think of it - this is really a place where you should use a context manager. Then the finally block would not be even necessary. Just like you do `with open(...):`. With socket it would look like this:

``` python
from contextlib import closing

with closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as tcp_socket:
```

See [How to use socket in Python as a context manager?](https://stackoverflow.com/questions/16772465/how-to-use-socket-in-python-as-a-context-manager/16772520).

This also shows what I meant in my comment about passing `socket` in parameters, not on `self`: the `with` block should be the scope for the socket and it should be destroyed automatically when the scope ends. Putting it on `self` may potentially extend its life past it. The `with` block guarantees that it won't happen, even by mistake.

signing_service/src/tests/tests.py
rwrzesien: That is the name of the method :)

signing_service/src/tests/tests.py

signing_service/src/tests/tests.py
rwrzesien: Should be `should_send_empty_bytes_to_check` :+1: done

signing_service/src/tests/tests.py
rwrzesien: `close` is run in `finally` block of the main loop. Normally it would not be be called until inner loop in `_handle_connection` is interrupted, but the whole function is mocked here.

signing_service/src/signing_service.py
rwrzesien: The context manager looks nice but I still don't understand how you want this code to reconnect on several errors caught in `try..except` block if loop will be inside `try`. The only option I can think is adding another `try..except` inside `try`. Is that what you have meant ?

signing_service/src/constants.py
rwrzesien: :+1: done

signing_service/src/signing_service.py
rwrzesien: :+1: done https://github.com/golemfactory/concent/issues/647

signing_service/src/signing_service.py
cameel: OK, sorry. `with` should be inside the loop.

rwrzesien: Ok, I have rebuild it and simplify a little bit :+1: done

signing_service/src/signing_service.py
```diff
@@ -1,60 +1,169 @@
...
+
+    def _connect(self, tcp_socket: socket.socket) -> None:
+        """ Creates socket and connects to given HOST and PORT. """
+        if self.current_reconnect_delay is not None:
+            logger.warning(f'Waiting {self.current_reconnect_delay} before connecting.')
+            sleep(self.current_reconnect_delay)
+
+        logger.warning(f'Connecting to {self.host}:{self.port}.')
+        tcp_socket.connect((self.host, self.port))
+        logger.warning(f'Connection established.')
```
cameel: Why are you logging at the `WARNING` level? It's definitely not a warning.

signing_service/src/signing_service.py
```diff
@@ -1,60 +1,169 @@
...
+    def _handle_connection(self, tcp_socket: socket.socket) -> None:  # pylint: disable=no-self-use
+        """ Inner loop that handles data exchange over socket. """
+        while True:
+            data = tcp_socket.recv(1024)
+            tcp_socket.send(data)
 
-def _parse_arguments():
+    def _increase_delay(self) -> None:
+        """ Increase current delay if connection cannot be established. """
+        assert self.current_reconnect_delay != 0
```
cameel: Why not `> 0`?

signing_service/src/signing_service.py
rwrzesien: My mistake, fixed :+1:

signing_service/src/signing_service.py
rwrzesien: Because it can be also None but it can also be 0 so I have must not updated this line for some time. :+1: done

cameel: It should not. But I mean that your assert was allowing negative values. They aren't valid, are they?

signing_service/src/signing_service.py
```diff
@@ -1,60 +1,169 @@
...
+                    break
+                except Exception as exception:
+                    # If there was an unrecognized exception, log it and report to Sentry.
+                    crash_logger.error(f'Unrecognized exception occurred: {exception}')
+                    raise
+
+    def _connect(self, tcp_socket: socket.socket) -> None:
+        """ Creates socket and connects to given HOST and PORT. """
+        if self.current_reconnect_delay is not None:
+            logger.warning(f'Waiting {self.current_reconnect_delay} before connecting.')
```
cameel: You're still logging at the warning level. Please review your loggers.

signing_service/src/signing_service.py
rwrzesien: As discussed, we allow 0.
Yes, it should not be negative but it can be None. 
Fixed already.

Issue #628, opened 2018-07-11
calculate_subtask_verification_time should use calculate_maximum_download_time


MichalW-CP:
In `calculate_subtask_verification_time` we use `maximum_download_time` function from `golem_messages` which uses constants for `DEFAULT_UPLOAD_RATE` and `DOWNLOAD_LEADIN_TIME`. Instead of `constants` we should use `settings`. To calculate mdt use `calculate_maximum_download_time` function from` core/utils.py`
```python
def calculate_subtask_verification_time(report_computed_task: message.ReportComputedTask) -> int:
    """
    This function calls `subtask_verification_time` helper function from golem messages or Concent custom implementation
    of it, depending on CUSTOM_PROTOCOL_TIMES setting value.
    The reason for using custom implementation is because it has hard-coded values and we cannot make it use values
    from our settings.
    """
    assert isinstance(report_computed_task, message.ReportComputedTask)

    if settings.CUSTOM_PROTOCOL_TIMES:
        mdt = maximum_download_time(
            size=report_computed_task.size,
        )
        ttc_dt = datetime.datetime.utcfromtimestamp(
            report_computed_task.task_to_compute.timestamp,
        )
        subtask_dt = datetime.datetime.utcfromtimestamp(
            report_computed_task.task_to_compute.compute_task_def['deadline'],
        )
        subtask_timeout = subtask_dt - ttc_dt

        return int(
            (4 * settings.CONCENT_MESSAGING_TIME) +
            (3 * mdt.total_seconds()) +
            (0.5 * subtask_timeout.total_seconds())
        )
    else:
        return int(subtask_verification_time(report_computed_task).total_seconds())
```

PR #637, opened 2018-07-12
Fix calculate_subtask_verification_time to use only concent settings

pawelkisielewicz:
Resolves https://github.com/golemfactory/concent/issues/628


concent_api/concent_api/settings/testing.py
```diff
@@ -60,14 +60,14 @@
...
 
-CONCENT_MESSAGING_TIME    = 2
-
 STORAGE_SERVER_INTERNAL_ADDRESS = 'http://localhost/'
 
 VERIFIER_STORAGE_PATH = '/tmp/'
 
 CONCENT_MESSAGING_TIME    = 0
 
+CONCENT_UPLOAD_RATE       = int(384 / 8)  # KB/s = kbps / 8
```
pawelkisielewicz: Default value the same as in golem constants

concent_api/core/tests/utils.py
```diff
@@ -167,7 +168,7 @@ def _get_deserialized_report_computed_task(
 
     def _get_deserialized_task_to_compute(
         self,
-        timestamp: Optional[str] = None,
+        timestamp: Optional[Union[str, datetime.date]] = None,
```
dybi: isn't it `datetime.datetime`?, besides, I think `Union[str, datetime.datetime, None]` would be more readable ;)

concent_api/concent_api/settings/testing.py
dybi: what about `development.py`?

concent_api/concent_api/settings/testing.py
```diff
@@ -60,14 +60,14 @@
 
 STORAGE_CLUSTER_ADDRESS = 'http://localhost/'
 
-CONCENT_MESSAGING_TIME    = 2
```
rwrzesien: Why are you removing this ?

kbeker: This should be still in our testing settings. It doesn't matter that we overriding settings in our tests (at least we should) but in `testing.py` we must have all settings what we have in `base.py`

pawelkisielewicz: It was duplicated. look at line 69 - there is ` CONCENT_MESSAGING_TIME    = 0`

concent_api/concent_api/settings/testing.py
pawelkisielewicz: done

concent_api/core/tests/utils.py
pawelkisielewicz: done

concent_api/concent_api/settings/testing.py
kbeker: So this should stay and delete this with value 0.

PR #638, opened 2018-07-12
Chore Celery queues should not be created 'ad hoc'

### LINKED ISSUE #445: Celery queues should not be created 'ad hoc'
There should be only fixed/manually defined celery queues, missing ones should not be created:
`app.conf.task_create_missing_queues = False` in `celery.py`.

This task, most probably, requires switching to [manual routing](http://docs.celeryproject.org/en/latest/userguide/routing.html#manual-routing) in `celery.py`

kbeker:
Resolves #445

### DIFF
--- a/concent_api/concent_api/celery.py
+++ b/concent_api/concent_api/celery.py
@@ -1,5 +1,6 @@
 import os
 from celery import Celery
+from kombu import Queue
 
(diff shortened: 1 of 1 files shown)


kbeker:
This branch has been merged.

PR #643, opened 2018-07-13
Add freeze_time to verifier test which gives nondeterministic results

kbeker:
Add freeze_time because this test gives sometimes different time at the end of it. `freeze_time()` resolves it

### DIFF
--- a/concent_api/verifier/tests/test_integration_verification.py
+++ b/concent_api/verifier/tests/test_integration_verification.py
@@ -1,6 +1,7 @@
 import mock
 
 from django.test import override_settings
+from freezegun import freeze_time
 from numpy.core.multiarray import ndarray
 
 from common.constants import ErrorCode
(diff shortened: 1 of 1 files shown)


concent_api/verifier/tests/test_integration_verification.py
```diff
@@ -370,44 +371,45 @@ def test_blender_verification_order_should_call_verification_result_with_result_
         )
 
     def test_that_blender_verification_order_should_work_properly_when_there_is_more_than_one_frame_to_render(self):
-        with mock.patch('verifier.tasks.download_archives_from_storage', autospec=True) as mock_download_archives_from_storage,\
```
cameel: When doing something non obvious like this you should always remember to add a comment explaining why you're doing it.

Also, say which part is non-deterministic and fails.

kbeker: Should I comment this inside test?

kbeker: Updated.

rwrzesien: We should put more comments in code in general :)

cameel: > Should I comment this inside test?

Yes.

> We should put more comments in code in general :)

Agreed. But please avoid comments that just say what you're doing. The comments should explain why and give context that's not available in the code they describe.

PR #654, opened 2018-07-16
Chore make usage of creation and modification timestamps

### LINKED ISSUE #604: Make usage of creation and modification timestamps in our database models more consistent
This is the result of discussion about timestamps from an unrelated issue: https://github.com/golemfactory/concent/issues/553#issuecomment-402573866

The following changes are needed:
- Add `created_at` and `modified_at` timestamps to:
    - `Subtask`
    - `PendingResponse`
- Add `created_at` timestamp to:
    - `Client`
    - `StoredMessage`
- Move `created_at` timestamp from `BlenderSubtaskDefinition` to `VerificationRequest`.
    - Add a migration that moves the timestamp from the old object to the new one.

When you're done, update the corresponding docs/blueprints (or drop me a comment so that I remember that I have to do it).

kbeker:
Resolves #604

Tested locally. Both migrations works.


concent_api/conductor/models.py
```diff
@@ -34,6 +34,9 @@ class VerificationRequest(Model):
     # Deadline on additional verification for related Subtask.
     verification_deadline = DateTimeField()
 
+    # Indicates when Conductor has received the request.
+    created_at = DateTimeField(default=timezone.now)
```
rwrzesien: We should use `auto_now_add=True`. You can change this on existing fields which use `timezone.now`.

concent_api/core/models.py
```diff
@@ -456,7 +472,19 @@ def response_type_enum(self):
...
+
+    modified_at = DateTimeField(default=timezone.now)
+
+    # Overwritten method to save object with modified timestamp
+    # created_at field will be fulfilled only when object will be created
+    # modified_at field will be overwritten every time when object will be modified
+    def save(self, *args, **kwargs):  # pylint: disable=arguments-differ
+        if not self.id:
+            self.created_at = timezone.now()
+        self.modified_at = timezone.now()
```
rwrzesien: For `modified_at` fields please use `auto_add=True`.

concent_api/core/models.py
```diff
@@ -456,7 +472,19 @@ def response_type_enum(self):
...
+
+    created_at = DateTimeField(default=timezone.now)
+
+    modified_at = DateTimeField(default=timezone.now)
+
+    # Overwritten method to save object with modified timestamp
+    # created_at field will be fulfilled only when object will be created
+    # modified_at field will be overwritten every time when object will be modified
+    def save(self, *args, **kwargs):  # pylint: disable=arguments-differ
+        if not self.id:
```
rwrzesien: This is not necessary when you have `default` or `auto_add_now` and `auto_add`.

concent_api/core/models.py
```diff
@@ -287,24 +290,37 @@ class Meta:
...
 
-    state                       = CharField(max_length = 32, choices = SubtaskState.choices())
+    state = CharField(max_length=32, choices=SubtaskState.choices())
 
     # If in an active state, it's the time at which Concent automatically transitions to a passive state if it does
     # not get the information it expects from the client (a timeout). Must be NULL in a passive state.
-    next_deadline               = DateTimeField(blank = True, null = True)
+    next_deadline = DateTimeField(blank=True, null=True)
+
+    created_at = DateTimeField(editable=False, default=timezone.now)
```
rwrzesien: Why is `editable=False` set only here ?

concent_api/core/models.py
```diff
@@ -287,24 +290,37 @@ class Meta:
...
+
```
rwrzesien: I would not separate those two with an extra blank line, those are a kind of logical group.

concent_api/core/migrations/0005_auto_20180716_1733.py
```diff
@@ -0,0 +1,41 @@
...
+        ),
+        migrations.AddField(
+            model_name='storedmessage',
+            name='created_at',
+            field=models.DateTimeField(default=django.utils.timezone.now),
+        ),
+        migrations.AddField(
+            model_name='subtask',
+            name='created_at',
+            field=models.DateTimeField(default=django.utils.timezone.now, editable=False),
```
dybi: Why only this one has `editable=False`?

concent_api/conductor/models.py
kbeker: I read yesterday that if we use `auto_now_add` option then this won't be visible in administration panel - it was some bug in django. Maybe now this bug is already fixed because this was information from before year or two.

Done

kbeker: I have discovered something strange. If you want to add `default` to field definition you can't do this when you have `auto_now` or `auto_now_add` option.

```(fields.E160) The options auto_now, auto_now_add, and default are mutually exclusive. Only one of these options may be present.```

but it is impossible to add field to existing model without adding default value and it is necessary to add value during `makemigrations` command - and there, as a default is `timezone.now`

File: concent_api/core/models.py (hunk 2 above)

File: concent_api/core/models.py (hunk 3 above)

File: concent_api/core/models.py (hunk 4 above)
Comment (kbeker) [reply concent_api/core/models.py]: Fixed, Done

File: concent_api/core/models.py (hunk 5 above)
Comment (kbeker) [reply concent_api/core/models.py]: Merged, Done

File: concent_api/core/migrations/0005_auto_20180716_1733.py (hunk 6 above)
Comment (kbeker) [reply concent_api/core/migrations/0005_auto_20180716_1733.py]: Fixed, Done

File: concent_api/conductor/migrations/0008_auto_20180717_0908.py [hunk 7]
```diff
@@ -23,11 +23,13 @@ class Migration(migrations.Migration):
     dependencies = [
         ('conductor', '0007_auto_20180709_1526'),
     ]
+
     operations = [
         migrations.AddField(
             model_name='verificationrequest',
             name='created_at',
-            field=models.DateTimeField(default=django.utils.timezone.now),
+            field=models.DateTimeField(auto_now_add=True, default=django.utils.timezone.now),
```
dybi: > The options `auto_now_add`, `auto_now`, and `default` are **mutually exclusive**. Any combination of these options will result in an error.

Choose one ;) (https://docs.djangoproject.com/en/2.0/ref/models/fields/#django.db.models.DateField)

kbeker: See:
```
I have discovered something strange. If you want to add default to field definition you can't do this when you have auto_now or auto_now_add option.

(fields.E160) The options auto_now, auto_now_add, and default are mutually exclusive. Only one of these options may be present.

but it is impossible to add field to existing model without adding default value and it is necessary to add value during makemigrations command - and there, as a default is timezone.now
``` 
I have wrote this above because I wanted to do this without default and Django didn't let me to do this and during `makemigrations` command Django made suggestion to add `default` with `timezone.now`

kbeker: I have tested locally with old database and records in it and then I applied migrations and everything worked.

concent_api/conductor/models.py
rwrzesien: Yes, exactly, you have to remove the defaults. `auto_now_add` will handle it on its own.

kbeker: OK, done. Really strange :D

concent_api/conductor/migrations/0008_auto_20180717_0908.py
cameel: Just a hint for the future: we never want fake defaults in migrations. If you think you need one, this needs to be discussed first. You may end up making irreversible damage to the data in production otherwise.

PR #655, opened 2018-07-17
Add pytest-asyncio to requirements

dybi:
pytest-asyncio allows better unit testing asyncio based code


kbeker:
@pawelkisielewicz but this problem is not related with this PR. If this problem occurs it should be another PR. :)

concent_api/requirements.lock
```diff
@@ -1,27 +1,28 @@
...
 cloudpickle==0.5.3
-coincurve==7.1.0
+coincurve==8.0.0
 coverage==4.5.1
 cycler==0.10.0
 cytoolz==0.9.0.1
 dask==0.18.1
 decorator==4.3.0
-Django==1.11.13
+Django==1.11.14
```
cameel: Looks like this pull request does much more than just add pytest-asyncio.

When adding a new dependency please don't update unrelated stuff. Here's how `requirements.lock` should be updated to avoid this:
``` bash
virtualenv /tmp/clean-env
source /tmp/clean-env/bin/activate
pip install -r requirements.lock
pip install pytest-asyncio
pip freeze >> requirements.lock
```

This way you can focus on the task at hand and not fight with unrelated problems caused by changes in our dependencies. If you really want to update other dependencies, you can still do it in a separate pull request (or at least a separate commit).

Issue #562, opened 2018-06-26
Add assertions in tests that would check if message is signed by the right party


dybi:
Issues comes form comment: https://github.com/golemfactory/concent/pull/513#discussion_r198089467

**TL;DR**
Whenever we create message in tests, we should make sure it is signed by the right party, i.e. 
``` python
sign_message(ack_report_computed_task.report_computed_task, PROVIDER_PRIVATE_KEY)
assert TaskMessage.OWNER_CHOICES.requestor in ack_report_computed_task.EXPECTED_OWNERS
```

The mechanism could be added as a part of `self._get_deserialized_XXX` or can be added manually whenerver message is created.
The mechanizm mentioned above could have a flag to disable the check (i. e. for tests where message is **purposely** signed by the **wrong** party),  sth like:
```python
def _get_deserialized_report_computed_task(
        self,
        ...
        should_check_signature=True

    ):
    if should_check_signature:
        assert TaskMessage.OWNER_CHOICES.provider in report_computed_task.EXPECTED_OWNERS
```

```

rwrzesien:
> assert TaskMessage.OWNER_CHOICES.requestor in ack_report_computed_task.EXPECTED_OWNERS

I think this will not give us any extra information apart from the fact that there is a bug in golem-messages if someone accidentally removes one of the `EXPECTED_OWNERS`. We might have a single test that will go through golem-messages with `EXPECTED_OWNERS` and do sanity check if they are what Concent expects.

And instead of doing that in test methods like `_get_deserialized_report_computed_task` I suggest that, after message has been signed, verify its ownership, using method described in https://github.com/golemfactory/concent/issues/563

cameel:
@rwrzesien @dybi Could you summarize the current status of this task?

pawelkisielewicz:
It was interrupted and never restored. Since this issue we fixed typing in other tasks so this issue can be closed.

PR #656, opened 2018-07-17
[WIP] Add assertions and Type declarations to methods

pawelkisielewicz:
Resolves https://github.com/golemfactory/concent/issues/562

### DIFF
--- a/concent_api/core/tests/utils.py
+++ b/concent_api/core/tests/utils.py
@@ -1,6 +1,5 @@
 from base64 import b64encode
 from typing import List
-from typing import Optional
 from typing import Union
 import datetime
 import functools
@@ -22,6 +21,22 @@ from golem_messages.factories.tasks import ComputeTaskDefFactory
 from golem_messages.factories.tasks import ReportComputedTaskFactory
 from golem_messages.factories.tasks import TaskToComputeFactory
 from golem_messages.message.base import Message
+from golem_messages.message.concents import ForceGetTaskResult
+from golem_messages.message.concents import ForcePayment
+from golem_messages.message.concents import ForceReportComputedTask
+from golem_messages.message.concents import ForceSubtaskResults
+from golem_messages.message.concents import ForceSubtaskResultsResponse
+from golem_messages.message.concents import SubtaskResultsVerify
+from golem_messages.message.tasks import AckReportComputedTask
+from golem_messages.message.tasks import CannotComputeTask
+from golem_messages.message.tasks import ComputeTaskDef
+from golem_messages.message.tasks import RejectReportComputedTask
+from golem_messages.message.tasks import ReportComputedTask
+from golem_messages.message.tasks import SubtaskResultsAccepted
+from golem_messages.message.tasks import SubtaskResultsRejected
+from golem_messages.message.tasks import TaskFailure
+from golem_messages.message.tasks import TaskMessage
+from golem_messages.message.tasks import TaskToCompute
 from golem_messages.utils import encode_hex
 
 from core.models            import Client
@@ -121,10 +136,11 @@ class ConcentIntegrationTestCase(TestCase):
 
     def _get_serialized_force_get_task_result(
         self,
-        report_computed_task,
-        timestamp,
-        requestor_private_key = None
-    ):
+        report_computed_task: ReportComputedTask,
+        timestamp: Union[str, datetime.datetime],
+        requestor_private_key: Union[bytes, None] = None,
+    ) -> ForceGetTaskResult:
+
         """ Returns MessageForceGetTaskResult serialized. """
         with freeze_time(timestamp):
             force_get_task_result = message.concents.ForceGetTaskResult(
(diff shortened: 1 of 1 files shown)


concent_api/core/tests/utils.py
```diff
@@ -22,6 +21,22 @@
...
+from golem_messages.message.tasks import AckReportComputedTask
+from golem_messages.message.tasks import CannotComputeTask
+from golem_messages.message.tasks import ComputeTaskDef
+from golem_messages.message.tasks import RejectReportComputedTask
+from golem_messages.message.tasks import ReportComputedTask
+from golem_messages.message.tasks import SubtaskResultsAccepted
+from golem_messages.message.tasks import SubtaskResultsRejected
+from golem_messages.message.tasks import TaskFailure
+from golem_messages.message.tasks import TaskMessage
+from golem_messages.message.tasks import TaskToCompute
```
kbeker: I think better way to do imports is to import just `from golem_messages import message`

concent_api/core/tests/utils.py
```diff
@@ -169,22 +189,23 @@ def _get_deserialized_report_computed_task(
...
         subtask_id: str = '2',
-        compute_task_def = None,
-        requestor_id: bytes = None,
-        requestor_public_key: bytes = None,
-        requestor_ethereum_public_key: bytes = None,
-        provider_id: bytes = None,
-        provider_public_key: bytes = None,
-        provider_ethereum_public_key: bytes = None,
-        price = 0,
+        compute_task_def: Union[ComputeTaskDef, None] = None,
```
kbeker: If you want to get [something, None] better way would be `Optional[something]` it means that you expect something or None

concent_api/core/tests/utils.py
```diff
@@ -169,22 +189,23 @@ def _get_deserialized_report_computed_task(
...
+        provider_id: Union[bytes, None] = None,
+        provider_public_key: Union[bytes, None] = None,
+        provider_ethereum_public_key: Union[bytes, None] = None,
+        price: int = 0,
         package_hash: str = 'sha1:230fb0cad8c7ed29810a2183f0ec1d39c9df3f4a',
-        sign_with_private_key = None,
-        size=1,
-        frames: Optional[List[int]] = None
-    ):
+        sign_with_private_key: Union[bytes, None] = None,
```
kbeker: `Optional[bytes]`

concent_api/core/tests/utils.py
```diff
@@ -169,22 +189,23 @@ def _get_deserialized_report_computed_task(
...
+        size: int = 1,
+        frames: Union[List[int], None] = None,
```
kbeker: `Optional[List[int]]`

concent_api/core/tests/utils.py
```diff
@@ -233,14 +254,18 @@ def _get_deserialized_task_to_compute(
...
     def _get_deserialized_ack_report_computed_task(
         self,
-        timestamp       = None,
-        deadline        = None,
-        subtask_id      = '1',
-        report_computed_task = None,
-        task_to_compute = None,
-        sign_with_private_key = None,
-    ):
+        timestamp: Union[str, datetime.datetime, None] = None,
```
kbeker: `Optional[Union[str, datetime.datetime]]`

concent_api/core/tests/utils.py
```diff
@@ -265,21 +290,21 @@ def _get_deserialized_ack_report_computed_task(
...
                 requestor_private_key if requestor_private_key is not None else self.REQUESTOR_PRIVATE_KEY,
                 settings.CONCENT_PUBLIC_KEY
             )
 
-    def _parse_iso_date_to_timestamp(self, date_string):    # pylint: disable=no-self-use
+    def _parse_iso_date_to_timestamp(self, date_string) -> int:    # pylint: disable=no-self-use
         return int(dateutil.parser.parse(date_string).timestamp())
 
-    def _get_timestamp_string(self):                        # pylint: disable=no-self-use
+    def _get_timestamp_string(self) -> str:                        # pylint: disable=no-self-use
```
kbeker: Please fix style with pylint warnings. Also check if disables still necessary

concent_api/core/tests/utils.py
```diff
@@ -426,13 +451,16 @@ def _test_undelivered_pending_responses(
 
     def _get_deserialized_force_subtask_results(
         self,
-        timestamp                   = None,
-        ack_report_computed_task    = None,
-    ):
+        timestamp: Union[str, datetime.datetime, None] = None,
+        ack_report_computed_task: Union[AckReportComputedTask, None] = None,
```
kbeker: `Optional[ARCT]`

concent_api/core/tests/utils.py
pawelkisielewicz: I did it like this, because this messages are multiple times used in code. I think it is better to import all of them like this and use in code like simple i.e. `SubtaskResultsRejected` instead of `message.tasks.SubtaskResultsRejected`. 
What do you think?

concent_api/core/tests/utils.py
pawelkisielewicz: will be done :)

pawelkisielewicz:
Close reason:
Keys will be checked other way 
Types add will be in separate pull request

concent_api/core/tests/utils.py
kbeker: So then better to import these two modules: 
`from golem_messages.messages import concents`
`from golem_messages.messages import tasks` 
It will be more readable I think :)

cameel: I personally prefer the style used by Pawel here. The prefixes only add noise, especially if you use them often.

It's the same thing we do in `models` - we import each thing separately instead of just importing `django.db.models` because it makes model definitions more readable.

Issue #660, opened 2018-07-18
Add computation_deadline and result_package_size fields to Subtask model


cameel:
To implement #329 we need to be able to access pieces of data from `ReportComputedTask` and `TaskToCompute` in a database query. To do this we need to maintain a copy of relevant fields on the `Subtask` table.

Add the following fields to `Subtask` model:
- `computation_deadline`: should be the same as `TaskToCompute.deadline`.
- `result_package_size`: should be the same as `ReportComputedTask.size`.

Add model validations that ensure that these fields always match the corresponding golem-messages.

Create a migration that fills these fields on existing model instances.

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/159162635

PR #663, opened 2018-07-19
Feature add additional fields to subtask model

MichalW-CP:
Resolves https://github.com/golemfactory/concent/issues/660
Add migrations for result_package_size and computation_deadline
Migrations need to be tested on Dev


concent_api/core/migrations/0006_add_and_update_deadline_and_size.py
```diff
@@ -0,0 +1,55 @@
...
+class Migration(migrations.Migration):
+    dependencies = [
+        ('core', '0005_auto_20180717_0908'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='subtask',
+            name='computation_deadline',
+            field=models.DateTimeField(blank=True, null=True),
```
rwrzesien: Same as in https://github.com/golemfactory/concent/pull/570#discussion_r203621222, there is mismatch between field definition and migration. What is the reason for it ?

kbeker: It was wrong. Already changed and divided to 2 migrations.

Issue #666, opened 2018-07-19
ForceSubtaskResultsResponse with both subtask_results_accepted and subtask_results_rejected


pawelkisielewicz:
Handling ForceSubtaskResultsResponse we are checking if:
`client_message.subtask_results_accepted is not None or client_message.subtask_results_rejected is not None`
This message should contain `subtask_results_accepted` OR `subtask_results_rejected`, but we are not validating case, in which message contains `subtask_results_accepted` AND `subtask_results_rejected`.
Please add missing validation and response to client (currently in this case it will be not handled at all)

kbeker:
Am I wrong or this would be weird if we would have `subtask_results_accepted` and `subtask_results_rejected` in one message?

pawelkisielewicz:
I agree with you- it would be weird and incorrect. This is the reason to validate, if `message` contains only and exactly one of them (`subtask_results_accepted` or `subtask_results_rejected`)

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/159228783

PR #676, opened 2018-07-20
Add missing validation in handle_send_force_subtask_results_response for both subtask_results_rejected and rejected or both None

pawelkisielewicz:
Resolves https://github.com/golemfactory/concent/issues/666

### DIFF
--- a/concent_api/core/tests/test_unit_message_handlers.py
+++ b/concent_api/core/tests/test_unit_message_handlers.py
@@ -2,7 +2,6 @@ from unittest import TestCase
 
 from django.conf import settings
 from django.test import override_settings
-
 from core.message_handlers import are_items_unique
 from core.message_handlers import store_subtask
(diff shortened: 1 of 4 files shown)


concent_api/core/tests/test_unit_message_handlers.py
```diff
@@ -64,3 +67,22 @@ def test_that_messages_are_stored_with_correct_timestamps(self):
             self._parse_iso_date_to_timestamp(subtask.report_computed_task.timestamp.isoformat()),
             self._parse_iso_date_to_timestamp(self.report_computed_task_timestamp)
         )
+
+
+class TestForceSubtaskResultsResponse(ConcentIntegrationTestCase):
```
rwrzesien: I think this should be added to `AcceptOrRejectIntegrationTest` test case. Then you can send full request and test if response contains proper error code.

Issue #661, opened 2018-07-18
Custom model manager for Subtask that adds MDT, SVT and download deadline columns


cameel:
This task depends on #660.

To implement #329 we need to know the download deadline for each task. This information must be available within a query that's performed in the admin panel. We can compute and expose this information as separate columns using Django's `QuerySet.annotate()`.

1. Create a custom manager for the `Subtask` model and make it available as `Subtask.objects_with_timing_columns`. Keep the original manager available as `Subtask.objects`.
2. Implement a set of functions on the manager, each one adds one column:
    - `with_maximum_download_time()` (based on [calculate_maximum_download_time()](https://github.com/golemfactory/concent/blob/master/concent_api/core/utils.py#L17-L40))
    - `with_subtask_verification_time()` (based on [calculate_subtask_verification_time()](https://github.com/golemfactory/concent/blob/master/concent_api/core/utils.py#L43-L71))
    - `with_download_deadline()`
    - Each one gets a `QuerySet` as a parameter and returns a new queryset that has a new column added.
3. Implement `with_timing_columns()` function on the manager. It simply returns
    ``` python
    (query_set.
        with_maximum_download_time().
        with_subtask_verification_time().
        with_download_deadline()
    )
    ```
4. `get_queryset()` on the manager should return the result of running `with_timing_columns()` on the initial `QuerySet`.
5. Add unit tests that ensure that the implementation of each method matches the one it's based on. See https://github.com/golemfactory/concent/pull/419#discussion_r190557331 for details on what to test.

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/159162677

PR #644, opened 2018-07-13
Add CONCENT_UPLOAD_RATE setting to base.py

kbeker:
Add missing setting to `base.py`

PR #677, opened 2018-07-20
Delete repeated CONCENT_UPLOAD_RATE setting

kbeker:
During adding test to this issue: https://github.com/golemfactory/concent/issues/661 I have found setting which is repeated and need to be delete.


cameel:
For reference, setting was added in #644.

Issue #662, opened 2018-07-19
Add validate if reason in RejectReportComputedTask is valid enum


pawelkisielewicz:
Generally `reason` in `RejectReportComputedTask` should be one of three `enums`.
Currently handling `RejectReportComputedTask` in `handle_send_reject_report_computed_task` we only check:
if reason is one of them, 
elif is second, 
if is third. 

There is missing validation, if reason is valid enum in general. The task is to add this validation and return message to client if reason is invalid.

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/159228829

Issue #394, opened 2018-05-17
Some message validators don't have enough information to return meaningful errors


rwrzesien:
Currently `validate_all_messages_identical` and `validate_golem_message_signed_with_key` are raising HTTP400 error when the validation fails. This disallows flexibility in using those functions. Instead, they should:
* validate_all_messages_identical - return which fields doesn't match, then raise http400 in the place where function is called.
* validate_golem_message_signed_with_key - return True/False, then raise http400 in the place where function is called (done in https://github.com/golemfactory/concent/pull/513).

pawelkisielewicz:
I don't understand the point of task. Now we have validation functions that raises exception if verification fails (BTW i think it should be something like `ValidationError` or `ValidationMissmatch`, not `http400`). 

What is the purpose of this?

Now it works like 'normal' validation function and it's call needs one line in our code. If i change it will require:
1. call function
2. check if there are not matching fields
3. create and raise exception

As it is like now points 2 and 3 are written once in validation function. Are you sure we want to move point 2 and 3 to all places where function is called? Now all information (including which fields are not matching) are saved to exception inside validation function and raised.  

I think the best idea is to change `http400` to other `error` or `exception` and reformat or add something to message if needed. What do you think?

rwrzesien:
Reopening because github closed automatically, but it is only partially done.

rwrzesien:
Updating description about `validate_golem_message_signed_with_key` being resolved in https://github.com/golemfactory/concent/pull/513.

rwrzesien:
@pawelkisielewicz To move the exceptions which should be raised in the views, back to the views. But raising one of the exceptions you have proposed is a good idea too. However we should be consistent all over. Currently most of validators raise `http404` so I guess we want to refactor them to new approach of our choice. I will add issue of both to discuss for the team in our coding style repository.

rwrzesien:
This issue is extended by https://github.com/golemfactory/concent/issues/593, so I am closing this one.

cameel:
> I don't understand the point of task.

I'm not sure if the work on this task went in the right direction. The point was that we had a validating function that received a set of values as input and raised an exception saying that the message containing them was invalid. But how could it even know if the values actually come from a message? As a result the error message did not make sense in some places where that function was being used.

@rwrzesien Has this root issue been fixed? I still see a similar pattern being used in new code, e.g. in: https://github.com/golemfactory/concent/pull/678#discussion_r204715088

[...]

PR #678, opened 2018-07-23
Bugfix validate if reason in reject report computed task is valid enum

pawelkisielewicz:
Resolves https://github.com/golemfactory/concent/issues/662

### DIFF
--- a/concent_api/core/message_handlers.py
+++ b/concent_api/core/message_handlers.py
@@ -48,6 +48,7 @@ from core.utils import calculate_additional_verification_call_time
 from core.utils import calculate_maximum_download_time
 from core.utils import calculate_subtask_verification_time
 from core.validation import is_golem_message_signed_with_key
+from core.validation import validate_that_golem_messages_are_signed_with_key
 from core.validation import validate_reject_report_computed_task
 from core.validation import validate_all_messages_identical
 from core.validation import validate_ethereum_addresses
@@ -1456,16 +1457,3 @@ def is_subtask_in_wrong_state(subtask_id, forbidden_states):
 
 def are_items_unique(items: list):
     return len(items) == len(set(items))
-
-
-def validate_that_golem_messages_are_signed_with_key(
-    public_key: bytes,
-    *golem_messages: message.base.Message,
-) -> None:
-    for golem_message in golem_messages:
-        if not is_golem_message_signed_with_key(public_key, golem_message):
-            raise Http400(
-                f'There was an exception when validating if golem_message {golem_message.__class__.__name__} is signed with '
-                f'public key {public_key}.',
-                error_code=ErrorCode.MESSAGE_SIGNATURE_WRONG,
-            )
--- a/concent_api/core/message_handlers.py
+++ b/concent_api/core/message_handlers.py
@@ -31,7 +31,6 @@ from common.helpers import parse_timestamp_to_utc_datetime
 from common.helpers import sign_message
 from common.validations import validate_secure_hash_algorithm
 from common import logging
-
 from core.exceptions import Http400
 from core.models import Client
 from core.models import PaymentInfo
@@ -49,6 +48,7 @@ from core.utils import calculate_additional_verification_call_time
 from core.utils import calculate_maximum_download_time
 from core.utils import calculate_subtask_verification_time
 from core.validation import is_golem_message_signed_with_key
+from core.validation import validate_reject_report_computed_task
 from core.validation import validate_all_messages_identical
 from core.validation import validate_ethereum_addresses
 from core.validation import validate_golem_message_subtask_results_rejected
@@ -220,27 +220,12 @@ def handle_send_ack_report_computed_task(client_message):
 
 
 def handle_send_reject_report_computed_task(client_message):
-    if (
-        isinstance(client_message.cannot_compute_task, message.CannotComputeTask) and
-        isinstance(client_message.task_failure, message.TaskFailure)
-    ):
-        raise Http400(
-            "RejectReportComputedTask cannot contain CannotComputeTask and TaskFailure at the same time.",
-            error_code=ErrorCode.MESSAGE_INVALID,
-        )
 
-    # Validate if task_to_compute is valid instance of TaskToCompute.
+    validate_reject_report_computed_task(client_message)
+
     task_to_compute = client_message.task_to_compute
-    validate_task_to_compute(task_to_compute)
-
     provider_public_key = hex_to_bytes_convert(task_to_compute.provider_public_key)
     requestor_public_key = hex_to_bytes_convert(task_to_compute.requestor_public_key)
-
-    # Validate if TaskToCompute signed by the requestor.
-    validate_that_golem_messages_are_signed_with_key(
-        requestor_public_key,
-        task_to_compute,
-    )
 
     # If reason is GotMessageCannotComputeTask,
     # cannot_compute_task is instance of CannotComputeTask signed by the provider.
@@ -272,6 +257,7 @@ def handle_send_reject_report_computed_task(client_message):
 
     # RejectReportComputedTask should contain empty cannot_compute_task and task_failure
     else:
+        assert client_message.reason == message.RejectReportComputedTask.REASON.SubtaskTimeLimitExceeded
         if client_message.cannot_compute_task is not None or client_message.task_failure is not None:
             raise Http400(
                 "RejectReportComputedTask requires empty 'cannot_compute_task' and 'task_failure' with {} reason.".format(
--- a/concent_api/core/tests/test_unit_message_handlers.py
+++ b/concent_api/core/tests/test_unit_message_handlers.py
@@ -2,12 +2,16 @@ from unittest import TestCase
 
 from django.conf import settings
 from django.test import override_settings
+from golem_messages import message
+
+from core.exceptions import GolemMessageValidationError
 from core.message_handlers import are_items_unique
 from core.message_handlers import store_subtask
 from core.models import Subtask
 from core.tests.utils import ConcentIntegrationTestCase
 from core.tests.utils import parse_iso_date_to_timestamp
 from core.utils import hex_to_bytes_convert
+from core.validation import validate_reject_report_computed_task
 
 
 class TestMessageHandlers(TestCase):
@@ -64,3 +68,29 @@ class TestMessagesStored(ConcentIntegrationTestCase):
             parse_iso_date_to_timestamp(subtask.report_computed_task.timestamp.isoformat()),
             parse_iso_date_to_timestamp(self.report_computed_task_timestamp)
         )
+
+
+class TestValidateRejectReportComputedTask(ConcentIntegrationTestCase):
+
+    def test_that_validation_passes_if_correct_message_given(self):
+        correct_reject_reported_computed_task = self._get_deserialized_reject_report_computed_task(
+            reason=message.RejectReportComputedTask.REASON.SubtaskTimeLimitExceeded,
+            task_to_compute=self._get_deserialized_task_to_compute()
+        )
+        validate_reject_report_computed_task(correct_reject_reported_computed_task)
+
+    def test_that_validation_raises_error_if_invalid_reason_instance_given(self):
+        incorrect_reject_reported_computed_task = self._get_deserialized_reject_report_computed_task(
+            reason=message.tasks.CannotAssignTask.REASON.NotMyTask,
+            task_to_compute=self._get_deserialized_task_to_compute()
+        )
+        with self.assertRaises(GolemMessageValidationError):
+            validate_reject_report_computed_task(incorrect_reject_reported_computed_task)
+
+    def test_that_validation_raises_error_if_reason_is_none(self):
+        incorrect_reject_reported_computed_task = self._get_deserialized_reject_report_computed_task(
+            reason=None,
+            task_to_compute=self._get_deserialized_task_to_compute()
+        )
+        with self.assertRaises(GolemMessageValidationError):
+            validate_reject_report_computed_task(incorrect_reject_reported_computed_task)
--- a/concent_api/core/validation.py
+++ b/concent_api/core/validation.py
@@ -16,8 +16,8 @@ from core.constants import GOLEM_PUBLIC_KEY_LENGTH
 from core.constants import MESSAGE_TASK_ID_MAX_LENGTH
 from core.constants import VALID_ID_REGEX
 from core.exceptions import FrameNumberValidationError
+from core.exceptions import Http400
 from core.exceptions import GolemMessageValidationError
-from core.message_handlers import validate_that_golem_messages_are_signed_with_key
 from core.utils import hex_to_bytes_convert
 
 
@@ -338,6 +338,19 @@ def validate_compute_task_def(compute_task_def: message.tasks.ComputeTaskDef) ->
     validate_scene_file(extra_data['scene_file'])
 
 
+def validate_that_golem_messages_are_signed_with_key(
+    public_key: bytes,
+    *golem_messages: message.base.Message,
+) -> None:
+    for golem_message in golem_messages:
+        if not is_golem_message_signed_with_key(public_key, golem_message):
+            raise Http400(
+                f'There was an exception when validating if golem_message {golem_message.__class__.__name__} is signed with '
+                f'public key {public_key}.',
+                error_code=ErrorCode.MESSAGE_SIGNATURE_WRONG,
+            )
+
+
 def validate_reject_report_computed_task(client_message: RejectReportComputedTask):
 
     if (
@@ -349,7 +362,7 @@ def validate_reject_report_computed_task(client_message: RejectReportComputedTas
             error_code=ErrorCode.MESSAGE_INVALID,
         )
 
-    if client_message.REASON is None:
+    if client_message.reason is None:
         raise GolemMessageValidationError(
             f'Error during handling RejectReportComputedTask. REASON is None, it should be message.RejectReportComputedTask.REASON instance',
             error_code=ErrorCode.MESSAGE_VALUE_WRONG_TYPE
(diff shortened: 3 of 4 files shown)


concent_api/core/message_handlers.py
```diff
@@ -1483,3 +1487,25 @@ def validate_that_golem_messages_are_signed_with_key(
                 f'public key {public_key}.',
                 error_code=ErrorCode.MESSAGE_SIGNATURE_WRONG,
             )
+
+
+def validate_if_given_object_is_given_enum_instance(
```
Jakub89: 1. Move this function to validation.py please.
2. Name of the function sound weird. Can't it be just `validate_if_given_object_is_enum_instance` or `validate_if_given_object_is_correct_enum_instance` or `validate_if_given_object_is_expected_enum_instance`?

concent_api/core/tests/test_unit_message_handlers.py
```diff
@@ -2,8 +2,13 @@
 
 from django.conf import settings
 from django.test import override_settings
+
+from common import constants
+from core.exceptions import Http400
```
Jakub89: @dybi would point to change it to some custom exception :P There are some validation exceptions that inherit from `ConcentBaseException`. Use one of those or create new that also inherit from it.

concent_api/core/message_handlers.py
```diff
@@ -272,6 +275,7 @@ def handle_send_reject_report_computed_task(client_message):
 
     # RejectReportComputedTask should contain empty cannot_compute_task and task_failure
     else:
+        assert client_message.reason == message.RejectReportComputedTask.REASON.SubtaskTimeLimitExceeded
```
Jakub89: I'm not sure about that. What's the point of adding validation above just to have assert below that limit our code to those three reasons that `RejectReportComputedTask` currently has. If golem add some new reasons in future, our code will crush on that assert.
@rwrzesien , @cameel ?

concent_api/core/tests/test_unit_message_handlers.py
cameel: Yeah, we were supposed to start using more specific exceptions.

concent_api/core/message_handlers.py
cameel: > If golem add some new reasons in future, our code will crush on that assert.

That's exactly why it's here. If Golem adds a new reason and we don't update our code to handle it, we have a bug. The current code only works for the three reasons we currently know. This assert can detect such a bug.

The problem is that someone can silently break our code by doing a change in some far away place in another library. We don't want that. It's better to crash than to ignore such an error. Also, this crash is meant to happen in development. When you update golem-messages you're supposed to see it and fix the problem.

We need to be doing it more. Basically any final `else:` in an `if` that's defined for all values of an enum should have such an assert.

concent_api/core/tests/test_unit_message_handlers.py
pawelkisielewicz: I also think so, the reason I used http400 was, we used it in almost all places and similar situations. For this `GolemMessageValidationError` will be perfect

concent_api/core/tests/test_unit_message_handlers.py
```diff
@@ -2,8 +2,13 @@
 
 from django.conf import settings
 from django.test import override_settings
+
+from common import constants
```
rwrzesien: You are actually using only `ErrorCode` from `constants` so why not importing it directly ?

concent_api/core/validation.py
```diff
@@ -334,3 +335,25 @@ def validate_compute_task_def(compute_task_def: message.tasks.ComputeTaskDef) ->
...
 
     validate_scene_file(extra_data['scene_file'])
+
+
+def validate_if_given_object_is_expected_enum_instance(
+    custom_object: enum.Enum,
+    base_enum: type,
+    allow_none_as_instance: bool = False,
+) -> None:
+
```
rwrzesien: A comment would be nice.

concent_api/core/tests/test_unit_message_handlers.py
```diff
@@ -2,8 +2,13 @@
...
+from core.exceptions import GolemMessageValidationError
 from core.message_handlers import are_items_unique
+from core.validation import validate_if_given_object_is_expected_enum_instance
```
rwrzesien: Please reorder alphabetically.

concent_api/core/validation.py
```diff
@@ -334,3 +335,25 @@ def validate_compute_task_def(compute_task_def: message.tasks.ComputeTaskDef) ->
...
+    if allow_none_as_instance is True and custom_object is None:
+        return
+
+    if allow_none_as_instance is False and custom_object is None:
+        raise GolemMessageValidationError(
+            f'{custom_object} is None, it should be {base_enum} instance',
+            error_code=ErrorCode.MESSAGE_WRONG_FIELDS
```
rwrzesien: Maybe, but just maybe, this would be worth a different error code ? Because the problem in this case is rather with value, not field. We have code `MESSAGE_VALUE_NOT_ALLOWED`. But please wait for @cameel opinion on this. Choosing proper error code is tricky.

cameel: I think this is not the right place to choose an error code. This function does not know in which context it's being called. How does it even know that the enum comes from inside a golem message?

This the same problem as in #394. The function should just say if it's correct or not and the caller should be responsible for raising the error.

concent_api/core/validation.py
```diff
@@ -334,3 +335,25 @@ def validate_compute_task_def(compute_task_def: message.tasks.ComputeTaskDef) ->
...
+def validate_if_given_object_is_expected_enum_instance(
+    custom_object: enum.Enum,
+    base_enum: type,
+    allow_none_as_instance: bool = False,
+) -> None:
+
+    if allow_none_as_instance is True and custom_object is None:
+        return
+
+    if allow_none_as_instance is False and custom_object is None:
```
cameel: Please replace `allow_none_as_instance is False` with simply `not allow_none_as_instance`.

We only do `if x is Y:` (e.g. `if custom_object is None:`) when `x` is not a boolean expression. Here it is so you don't have to add `is True` or `is False` to convert it.

concent_api/core/validation.py
rwrzesien: Good point, then it has to be different exception because `GolemMessageValidationError` requires error code.

concent_api/core/tests/test_unit_message_handlers.py
```diff
@@ -64,3 +69,24 @@ def test_that_messages_are_stored_with_correct_timestamps(self):
...
+
+    def test_that_validation_raises_http400_if_string_given(self):
+        with self.assertRaises(GolemMessageValidationError):
+            validate_if_given_object_is_expected_enum_instance('any_string', constants.ErrorCode)
+
+    def test_that_validation_raises_http400_if_other_enum_given(self):
+        with self.assertRaises(GolemMessageValidationError):
+            validate_if_given_object_is_expected_enum_instance(constants.ErrorCode.MESSAGE_INVALID, PendingResponse.ResponseType)
+
+    def test_that_validation_raises_http400_if_none_given_and_none_not_allowed(self):
```
Jakub89: Those are not http400 any more, do they? :D Just change to `test_that_validation_raises_exception_...`

concent_api/core/message_handlers.py
```diff
@@ -49,6 +49,7 @@
 from core.utils import calculate_maximum_download_time
 from core.utils import calculate_subtask_verification_time
 from core.validation import is_golem_message_signed_with_key
+from core.validation import validate_if_given_object_is_expected_enum_instance
```
kbeker: Alphabetic order

concent_api/core/tests/test_unit_message_handlers.py
```diff
@@ -64,3 +69,24 @@ def test_that_messages_are_stored_with_correct_timestamps(self):
             parse_iso_date_to_timestamp(subtask.report_computed_task.timestamp.isoformat()),
             parse_iso_date_to_timestamp(self.report_computed_task_timestamp)
         )
+
+
+class TestIfObjectIsEnumInstance(TestCase):
```
kbeker: Please rename to `TestObjectEnumInstance`

concent_api/core/tests/test_unit_message_handlers.py
pawelkisielewicz: good point, done

concent_api/core/validation.py
pawelkisielewicz: I personally don't like `not` in this context. For me `is False` is more readable as `not`, but it is only my personal preference.. It is changed :)

concent_api/core/tests/test_unit_message_handlers.py
pawelkisielewicz: As always right :D

concent_api/core/message_handlers.py

concent_api/core/tests/test_unit_message_handlers.py
pawelkisielewicz: done

concent_api/core/validation.py
```diff
@@ -334,3 +336,23 @@ def validate_compute_task_def(compute_task_def: message.tasks.ComputeTaskDef) ->
...
+) -> None:
+    """
+    Function validates, if given object (custom_object) is given Enum instance (base_enum).
+    If it is not, raises TypeError. If given object is None NoneAsValueNotAllowedError will be raised,
+    unless allow_none_as_instance will be True.
+    """
+    if allow_none_as_instance is True and custom_object is None:
+        return
+
+    if not allow_none_as_instance and custom_object is None:
```
cameel: I think this part would be more readable as:

``` python
if custom_object is None:
    if not allow_none_as_instance:
        raise NoneAsValueNotAllowedError(f'{custom_object} is None, it should be {base_enum} instance')
    else:
        return
```

This immediately shows that these two conditions validate the case where `custom_object` is `None`. In your version you have to pause for a moment to figure that out.

concent_api/core/message_handlers.py
```diff
@@ -228,6 +230,18 @@ def handle_send_reject_report_computed_task(client_message):
             "RejectReportComputedTask cannot contain CannotComputeTask and TaskFailure at the same time.",
             error_code=ErrorCode.MESSAGE_INVALID,
         )
+    try:
+        validate_if_given_object_is_expected_enum_instance(client_message.reason, message.RejectReportComputedTask.REASON)
+    except NoneAsValueNotAllowedError:
+        raise GolemMessageValidationError(
```
cameel: I think this way of defining validation functions is not the best way to do it. You have created a validator to move some code out of the handler but then you had to add a big `try ... except` block. Overall it's more code than if you did it the simplest way and it is not even reusable. To validate it in another handler you'd have to copy the whole `try ... except` block.

I'd replace it with something simpler. A function called `validate_reject_report_computed_task(message)` that simply does
``` python
if isinstance(message.reason, message.RejectReportComputedTask.REASON):
    raise GolemMessageValidationError(...)
```

That's simple, does have enough context to construct an error message (it knows where the enum values comes from) and is easily extensible (we can add more checks to this function later; I'd actually put the `if` that's a few lines above there too). That's basically how our validations looked like at the beginning of the project. I think it was good enough for a simple validation system and the only problem with it was that it used `Http400` rather than an exception that represents a validation error.

If you really want to make it generic and reusable, I'd create something similar to Django's validator system. Something where you can define validators for each field and then have some function run the validators, catch exceptions and prepare error messages based on the name of the field and the message from the exception. Here's how it could work in our case:
1. Create a class called `MessageValidator`.
    - Has a validator map: field name -> list of validation functions, e.g.
    - Has a class method called `validate(message)`.
        - This method goes over all fields listed in the map and for each runs all the corresponding validators.
        - If a validator function raises `GolemMessageValidationError` it takes the error message, combines it with the name of the field and name of the message class and then raises the error again with the combined message.
2. Derive a specific class for each message type. For example `RejectReportComputedTaskValidator`.
    - Defines validation functions for all fields that need them
    - Each function is added to the validator map using `@register_validator(field_name)` decorator.
        - A validation function takes field value as argument and either returns or raises `GolemMessageValidationError`.
3. Create a decorator called `@register_validator(field_name)`.
    - Adds the class method to the validator map under the specified field name.
    - If the name is not specified, the validator is for the whole message and the function will get the message instead of just a value of a single field.
4. It should also be possible to register another validator class for a specific field (when it's a nested message).

But I'd suggest doing something like this in a separate issue. To finish this pull request quickly I'd go with the simple one I suggested above.

@rwrzesien What do you think about a system of validators like this?

concent_api/core/validation.py
cameel: 1. I think reusing built-in errors like `TypeError` is not a good idea because almost always they're caused by bugs. We want to be able to discern them from other errors that may indicate failures. For example if you make a mistake in your validation and Python raises `TypeError`. your function will interpret it as a failed validation, which is not true (it's a bug in our code, not a failed validation caused by invalid input).
2. I think that it's not necessary to discern between `None` and other types. They both indicate something that not's a value from the enum.

concent_api/core/message_handlers.py
rwrzesien: @cameel Shouldn't it be ?
```
if not isinstance(message.reason, message.RejectReportComputedTask.REASON):
    raise GolemMessageValidationError(...)
```

My idea was to propose of using Django Forms (with probably several tweaks) for validation as soon as we will have each Golem message stored in separate Model. We can then gather all errors from message, not just first. But it doesn't look like it is going to be soon. However, I think your idea describes a similar mechanism. 

How I see it is:
- Define atomic generic validators.
- Create some base validation class handling validation flow.
- Define a specific class for each message type, as you have proposed. This function will have a list of fields defined as assigned a list of validators to them.

I would still check if any part of Django Forms can be used for doing that. This way if someday we will have separate Models for Golem messages we can easily switch/add Django Forms, which is more natural for Django Model validation.

cameel: > Shouldn't it be ?
> ```
> if not isinstance(message.reason, message.RejectReportComputedTask.REASON):
>     raise GolemMessageValidationError(...)
> ```

Yes. My mistake.

> However, I think your idea describes a similar mechanism.
> 
> How I see it is:

Yes. It was just an example of such a mechanism. I did not want to suggest Django Forms because they're fine overall but I think we're going to have lots of small problems with details like differences between `""` and `None`. Django Forms is pretty lax about that because it assumes that the data comes from a HTML form anyway. I'd look for a different library if I were to implement this myself.

But anyway, my point was that the simple mechanism we're using from the beginning where we have:
- a validation function for each message type which checks a series of conditions and is responsible for creating error messages and raising errors
- helper functions that check more complex conditions and only return `True`/`False`

is still good enough. It's better than creating reusable, nested validators ad-hoc because such validators don't have enough context without a bigger mechanism that puts it all together. So I'd stick with it until we get something better.

And as for something better - I'd leave it up to you. Use whatever you like, even Django Forms if you think you can make something good out of it. My only requirement would be that I'd like to avoid it growing to the scale of a big task that is constantly being pushed back like it happened with Django checks. I'd like you to design something on a smaller scale that can be implemented immediately and then over time extend it to make it better.

concent_api/core/validation.py
```diff
@@ -334,3 +336,23 @@ def validate_compute_task_def(compute_task_def: message.tasks.ComputeTaskDef) ->
...
+        raise NoneAsValueNotAllowedError(f'{custom_object} is None, it should be {base_enum} instance')
+
+    if not isinstance(custom_object, base_enum):
+        raise TypeError(f'{custom_object} should be {base_enum} instance')
```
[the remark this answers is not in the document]
pawelkisielewicz: done

concent_api/core/validation.py
pawelkisielewicz: You are right. But changing like below this part of code will not be used any more

concent_api/core/message_handlers.py
rwrzesien: >  I did not want to suggest Django Forms because they're fine overall but I think we're going to have lots of small problems with details like differences between "" and None.

I agree, but those are the parts that I think we can override and prepare for golem-messages. I am not saying that it must be Django Forms, I just think that they might have some parts of the wheel that we can use.

Lets discuss it with @dybi too when he gets back.

concent_api/core/validation.py
```diff
@@ -334,3 +336,47 @@ def validate_compute_task_def(compute_task_def: message.tasks.ComputeTaskDef) ->
...
+    if client_message.REASON is None:
+        raise GolemMessageValidationError(
+            f'Error during handling RejectReportComputedTask. REASON is None, it should be message.RejectReportComputedTask.REASON instance',
+            error_code=ErrorCode.MESSAGE_VALUE_WRONG_TYPE
+        )
+
+    if not isinstance(client_message.reason, RejectReportComputedTask.REASON):
+        raise GolemMessageValidationError(
+            f'Error during handling RejectReportComputedTask. REASON should be message.RejectReportComputedTask.REASON instance'
+            f'Currently it is {type(client_message.reason)} instance',
```
cameel: These two lines will be combined without any space between them. You should add a space (and maybe a dot to end the sentence) at the end of the first one.

If this has passed your tests then the tests are too weak. Please add more.

concent_api/core/validation.py
```diff
@@ -334,3 +336,47 @@ def validate_compute_task_def(compute_task_def: message.tasks.ComputeTaskDef) ->
...
+    if (
+        isinstance(client_message.cannot_compute_task, message.CannotComputeTask) and
+        isinstance(client_message.task_failure, message.TaskFailure)
+    ):
+        raise GolemMessageValidationError(
+            "RejectReportComputedTask cannot contain CannotComputeTask and TaskFailure at the same time.",
+            error_code=ErrorCode.MESSAGE_INVALID,
+        )
+
+    if client_message.REASON is None:
```
cameel: I think that `REASON` is an `Enum` **class**. It's not the content of the message. You should be checking `client_message.reason`.

pawelkisielewicz: done

concent_api/core/validation.py
pawelkisielewicz: done

Issue #625, opened 2018-07-10
Signing Service: Message-oriented protocol


cameel:
This is a part of the implementation of the "Concent Signing Service" blueprint (#597). This task depends on #623.

#### Overview
Currently the Signing Service is implemented as a simple echo server working with a raw byte stream. It's time to make it possible to exchange full messages.

#### Wire protocol
The wire protocol is described in the blueprint (#597). Use it to implement message-based exchange.

The same protocol will be used by the Middleman and by Concent so make the implementation reusable.

#### Implementation
See the description of #618. Implement the same checks and way of communication but in the Signing Service.

#### Command-line parameters
- Concent's public key.
- Signing service private key: the private key used for signing protocol frames and Golem messages.
    - Should be passed as in #671.
    - Note that the public key can be easily computed from the private key.

PaweuB:
Pivotal: 
https://www.pivotaltracker.com/story/show/158969994

rwrzesien:
@cameel 

> See the description of #618. Implement the same checks and way of communication but in the Signing Service.

Should be "Middleman".

cameel:
No. The checks and communication should be implemented in the Signing Service not in the Middleman :P

rwrzesien:
@cameel If the received message was not possible to deserialize, we should response with `ServiceRefused` wrapped in `GolemMessageMiddlemanMessage`. What value should be set as `request_id` then ?

rwrzesien:
@cameel How to get private key for signing `ServiceRefused` ? Adding it to SigningService is in next issue, and I can't use `django.conf.settings` because SigningService is not being run in Django context. Can I just hardcode it for now ?

cameel:
> Which private key should be used to sign Middleman protocol messages when sending them from SigningService to Middleman ?

Signing Service should have its own key for Golem messages. I have added it to the descrption now.

> Also, adding Concent public key is in following issues. Can I hardcode it for now ?

But why? Adding a parameter is easy and without it you will only be able to test against one specific Concent cluster.

cameel:
@rwrzesien Is this task done?

rwrzesien:
@cameel Yes, resolved in https://github.com/golemfactory/concent/pull/688, closing.

PR #680, opened 2018-07-24
Feature signing service prepare for middleman protocol

rwrzesien:
Resolves problem with paths and imports in Signing Service.
Adds more tests for Signing Service.
Adds `Concent's public key` command-line parameter described in https://github.com/golemfactory/concent/issues/625.


cameel:
@rwrzesien When you're submitting a new pull request, there's not point in including unsquashed fixups. It only makes it harder to review.

cameel:
Oh, I see that the fixups are from a different pull request.
In that case please set that other branch as your base branch. Otherwise github thinks that those changes are a part of this pull request.

signing_service/requirements.txt
```diff
@@ -1,7 +0,0 @@
-coverage
-mypy
-pytest
-pytest-cov
```
cameel: Why are you removing requirements? The stuff that's in `setup.py` is not really equivalent to this.

rwrzesien:
@cameel :+1: done

signing_service/setup.py
```diff
@@ -0,0 +1,39 @@
...
+
+setup(
+    name='Signing-Service',
+    version=get_version(),
+    url='',
+    maintainer='',
+    maintainer_email='',
+    packages=[
+        'signing_service',
+    ],
```
cameel: What about a script to run the service? How do you run it when it's installed from a package?

signing_service/setup.py
```diff
@@ -0,0 +1,39 @@
...
+    ) as version_file:
+        return version_file.read()
+
+
+setup(
+    name='Signing-Service',
+    version=get_version(),
+    url='',
+    maintainer='',
+    maintainer_email='',
```
cameel: These fields need to be filled too.

signing_service/setup.py
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env python3
+
+import os
+from setuptools import setup
+
+
+def get_version():
+    with open(
+        os.path.join(
+            os.path.dirname(os.path.dirname(os.path.abspath(__file__))),
```
cameel: It would be more readable as

``` python
def get_version():
    path = os.path.join(
        os.path.dirname(os.path.dirname(os.path.abspath(__file__))),
        'RELEASE-VERSION',
    )
    with open(path, 'r') as version_file:
        return version_file.read()
```

signing_service/setup.py
```diff
@@ -0,0 +1,39 @@
...
+            'RELEASE-VERSION',
+        ),
+        'r'
+    ) as version_file:
+        return version_file.read()
+
+
+setup(
+    name='Signing-Service',
+    version=get_version(),
```
cameel: Anything related to `setup.py` and requirements should really be in a separate commit. It's not related in any way to other changes.

signing_service/signing_service/signing_service.py
```diff
@@ -136,6 +144,11 @@ def _parse_arguments() -> argparse.Namespace:
         type=int,
         help='Initial delay between reconnections, doubles after each unsuccessful attempt and is reset after success.',
     )
+    parser.add_argument(
+        'concent_public_key',
+        type=str,
+        help="Concent's public key.",
```
cameel: Adding this parameter should be a separate commit.

signing_service/signing_service/signing_service.py
```diff
@@ -164,6 +177,8 @@ def _parse_arguments() -> argparse.Namespace:
 
     arg_host = args.concent_cluster_host
     arg_port = args.concent_cluster_port
+    arg_concent_public_key, _ = escape_decode(args.concent_public_key)
```
cameel: Please add a proper name for the second argument (can still start with `_` but it should have a name) and put parentheses around it (it's a tuple).

signing_service/setup.py
rwrzesien: It still works after fixing, :+1: done

signing_service/setup.py
rwrzesien: Right, but with what values ?

signing_service/setup.py
rwrzesien: Personal matter, but ok :) :+1: done

Done in MiddleMan protocol too.

signing_service/signing_service/signing_service.py
rwrzesien: :+1: done

signing_service/setup.py
rwrzesien: :+1: done

signing_service/signing_service/signing_service.py
rwrzesien: :+1: done

signing_service/requirements.txt
cameel: Then the removal should be a separate commit. As it is now both this and conversion to `setup.py` are mixed in a single commit and it's not at all clear that this is what happened here so it's impossible to review this change. Please split the commit so that each change as a proper description.

cameel: Also `setup.py` has hard-coded versions while `requirements` did not. I don't think we should hard-code versions of these dependencies. We should keep the `txt`/`lock` split like in Concent because signing service is not a library and should be running with specific versions.

signing_service/setup.py
cameel: - URL of the repository
- Code Poets as maintainer

signing_service/signing_service.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/bash -e
 
-python3 src/signing_service.py "$@"
+python3 -m signing_service.signing_service "$@"
```
cameel: It looks like this change is in a wrong commit.

signing_service/setup.py
rwrzesien: :+1: done

signing_service/signing_service.sh
rwrzesien: Why ? The commit name is `Add setup.py to SigningService **and update of running script**`

signing_service/requirements.txt
rwrzesien: > Then the removal should be a separate commit.

I think it should be the same, if you consider this as a change of the way in which this application is being set up. You can set it up both before and after this commit.

I have added back `txt/lock` and removed requirements from `setup.py`.

signing_service/signing_service/signing_service.py
```diff
@@ -31,6 +31,7 @@ class SigningService:
         'current_reconnect_delay',
         'socket',
         'was_sigterm_caught',
+        'concent_public_key',
```
kbeker: Shouldn't this be in different commit? In this commit you only changing layout. This should be moved to: `Add concent_public_key argument, its validation and tests`

signing_service/signing_service/tests/tests.py
```diff
@@ -4,11 +4,12 @@
...
 
-from ..constants import SIGNING_SERVICE_DEFAULT_PORT
-from ..constants import SIGNING_SERVICE_RECOVERABLE_ERRORS
-from ..constants import SIGNING_SERVICE_MAXIMUM_RECONNECT_TIME
-from ..signing_service import _parse_arguments
-from ..signing_service import SigningService
+from signing_service.constants import SIGNING_SERVICE_DEFAULT_PORT
+from signing_service.constants import SIGNING_SERVICE_RECOVERABLE_ERRORS
+from signing_service.constants import SIGNING_SERVICE_MAXIMUM_RECONNECT_TIME
+from signing_service.exceptions import SigningServiceValidationError
```
kbeker: This import also should not be in this commit. You are adding validation to tests.

signing_service/signing_service/tests/tests.py
```diff
@@ -108,6 +114,7 @@ def setUp(self):
             '127.0.0.1',
             8000,
             2,
+            b'\xf3\x97\x19\xcdX\xda\x86tiP\x1c&\xd39M\x9e\xa4\xddb\x89\xb5,]O\xd5cR\x84\xb85\xed\xc9\xa17e,\xb2s\xeb\n1\xcaN.l\xba\xc3\xb7\xc2\xba\xff\xabN\xde\xb3\x0b\xa6l\xbf6o\x81\xe0;',
```
kbeker: I always recommend to add assignment to parameters in function when you are calling some function because then it is always obvious what parameters are you calling in specific function, but this is only my opinion :)

signing_service/signing_service/utils.py
```diff
@@ -0,0 +1,14 @@
+from golem_messages.cryptography import verify_pubkey
+from golem_messages.exceptions import InvalidKeys
+
+
+def is_valid_public_key(key):
```
kbeker: Please do some unit test only to this function.

signing_service/signing_service/tests/tests.py
rwrzesien: It was previously on line 9.

`from exceptions import SigningServiceValidationError`

signing_service/signing_service/tests/tests.py
rwrzesien: This is good idea for adding as proposal to our coding style. Please note that if you use ie. `Pycharm` you can display function signature using `ctrl + move over`.

signing_service/signing_service/signing_service.py
rwrzesien: :+1: done

signing_service/signing_service/utils.py
rwrzesien: :+1: done

signing_service/signing_service/tests/tests.py
kbeker: Sure, but it is simpler to see it without going inside function :) I'm going to propose in ours coding style wiki.

rwrzesien: I agree, now we are sometimes adding it, sometimes not. We should be consistent about it.

signing_service/signing_service.sh
cameel: Because it looks like a part of the file reorganization that's in the next commit. I thought that you mistakenly included it in the wrong commit. Does this work even without the reorganization?

If it's not meant to be a part of that commit then I think it should be a separate commit. The fact that you need to use _and_ in the description indicates that you're doing two different things in the same commit.

rwrzesien: Right, it won't work before file reorganization, so I will move it to that commit. :+1: done

signing_service/signing_service/signing_service.py
```diff
@@ -124,6 +129,9 @@ def _validate_arguments(self):
         if self.initial_reconnect_delay < 0:
             raise SigningServiceValidationError('reconnect_delay must be non-negative integer.')
 
+        if not is_valid_public_key(self.concent_public_key):
+            raise SigningServiceValidationError('concent_public_key is not valid public key.')
```
cameel: I'd call this exception differently because this is parameter validation and soon you'll also have validation for data that comes over the connection. These errors should be processed very differently - data validation should crash the application if it for any reason reaches the outer handler because it's never supposed to reach that point and it's a bug.

PR #681, opened 2018-07-25
Add active downloads admin filter

MichalW-CP:
Resolves https://github.com/golemfactory/concent/issues/329

### DIFF
--- a/concent_api/core/admin.py
+++ b/concent_api/core/admin.py
@@ -1,34 +1,44 @@
 from django.contrib import admin
+from django.db.models import Q
+from django.db.models import QuerySet
+from common.admin import ModelAdminReadOnlyMixin
+from common.helpers import get_current_utc_timestamp
+from common.helpers import parse_timestamp_to_utc_datetime
+from .models import PendingResponse
+from .models import StoredMessage
+from .models import Subtask
 
-from common.admin    import ModelAdminReadOnlyMixin
-from .models        import PendingResponse
-from .models        import StoredMessage
-from .models        import Subtask
+ACTIVE_STATE_NAMES = [x.name for x in Subtask.ACTIVE_STATES]
+PASSIVE_STATE_NAMES = [x.name for x in Subtask.PASSIVE_STATES]
 
 
-class ActivePassiveStateFilter(admin.SimpleListFilter):
-
-    title          = 'Active/passive state'
+class ActivePassiveDownloadsStateFilter(admin.SimpleListFilter):
+    title = 'Active/passive/downloads state'
     parameter_name = 'active_passive_state'
 
     def lookups(self, request, model_admin):
         return (
-            ('active',  'Active'),
+            ('active', 'Active'),
             ('passive', 'Passive'),
+            ('active_or_downloads', 'Active or with active downloads')
         )
 
-    def queryset(self, request, queryset):
+    def queryset(self, request, queryset: QuerySet) -> QuerySet:
+        assert self.value() in {'active', 'passive', 'active_or_downloads', None}
         if self.value() == 'active':
-            active_state_names = [x.name for x in Subtask.ACTIVE_STATES]
-            return queryset.filter(state__in=active_state_names)
+            return queryset.filter(state__in=ACTIVE_STATE_NAMES)
         elif self.value() == 'passive':
-            passive_state_names = [x.name for x in Subtask.PASSIVE_STATES]
-            return queryset.filter(state__in=passive_state_names)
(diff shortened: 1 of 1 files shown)


concent_api/core/admin.py
```diff
@@ -1,44 +1,55 @@
...
         return (
-            ('active',  'Active'),
+            ('active', 'Active'),
             ('passive', 'Passive'),
+            ('active_or_downloads', 'Active or with active downloads')
         )
 
     def queryset(self, request, queryset):
+        active_state_names = [x.name for x in Subtask.ACTIVE_STATES]
+        passive_state_names = [x.name for x in Subtask.PASSIVE_STATES]
```
rwrzesien: Those two can be placed in constants.

concent_api/core/admin.py
```diff
@@ -1,44 +1,56 @@
...
+from .models import Subtask
 
+ACTIVE_STATE_NAMES = [x.name for x in Subtask.ACTIVE_STATES]
+PASSIVE_STATE_NAMES = [x.name for x in Subtask.PASSIVE_STATES]
 
-class ActivePassiveStateFilter(admin.SimpleListFilter):
 
-    title          = 'Active/passive state'
+class ActivePassiveDownloadsStateFilter(admin.SimpleListFilter):
+    title = 'Active/passive/downloads state'
```
cameel: There are a lot of style changes that only distract from real changes. Please put them all in a separate commit.

concent_api/core/admin.py
```diff
@@ -1,44 +1,56 @@
...
         if self.value() == 'active':
-            active_state_names = [x.name for x in Subtask.ACTIVE_STATES]
-            return queryset.filter(state__in=active_state_names)
+            return queryset.filter(state__in=ACTIVE_STATE_NAMES)
         elif self.value() == 'passive':
-            passive_state_names = [x.name for x in Subtask.PASSIVE_STATES]
-            return queryset.filter(state__in=passive_state_names)
+            return queryset.filter(state__in=PASSIVE_STATE_NAMES)
+        elif self.value() == 'active_or_downloads':
+            current_timestamp = get_current_utc_timestamp()
```
cameel: What if the value is different than those three?

You should have an `else` and put `assert False` in it so that the code fails if someone adds a new filter and forgets to include it here.

concent_api/core/admin.py
```diff
@@ -48,17 +60,27 @@ class SubtaskAdmin(ModelAdminReadOnlyMixin, admin.ModelAdmin):
         'task_id',
     ]
 
-    def get_provider_public_key(self, obj):  # pylint: disable=no-self-use
+    @classmethod
+    def get_provider_public_key(cls, obj):
```
cameel: This change is not related to filters. It's fixing a pylint warning. It should be in a separate commit.

concent_api/core/admin.py
MichalW-CP: Hmm I haven't seen any in this commit... I've made git diff with master and only one is this above but this change of styling is connected with change in variable value so I dont't think I need to do it in separate commit.

concent_api/core/admin.py
MichalW-CP: Using ```else``` pycharm warned me that code is unreachable so I put just ```assert False```

concent_api/core/admin.py
MichalW-CP: Done.

concent_api/core/admin.py
cameel: There are more:
- Indentation change in imports above
- Removing a space before `'Active'`.

But OK. That's not that much. Let's leave it as is.

concent_api/core/admin.py
cameel: You could also do:
``` python
assert self.value() in {'active', 'passive', 'active_or_downloads'}
```
at the beginning of the function. And that would actually be more readable, though more repetitive too.

cameel:
Your commit that adds the filter still has some stuff related to pylint warnings:
``` diff
-    def get_subtask_subtask_id(self, obj):  # pylint: disable=no-self-use
+    @classmethod
+    def get_subtask_subtask_id(cls, obj):
```

Also, the description of the commit that fixes warnings should say what you're fixing and why. Right now it's too vague.

concent_api/core/admin.py
cameel: The commit description is still misleading. I see now that you're not fixing warnings. You're just removing the unnecessary `disable=no-self-use`. So why not describe it as
```
Replace 'pylint: disable=no-self-use' with the use of @classmethod in SubtaskAdmin
```
?

Issue #690, opened 2018-07-31
Create data migration for setting correct timestamp on StoredMessage objects.


rwrzesien:
The already merged pull request https://github.com/golemfactory/concent/pull/641 is missing data migration which will set correct timestamp on StoredMessage objects. Timestamp can be retrieved from `StoredMessage.data` which is serialized Golem message.

kbeker:
It might be ready solution:
```
StoredMessage = apps.get_model('core', 'StoredMessage')
for stored_message in StoredMessage.objects.all():
    stored_message.timestamp = parse_timestamp_to_utc_datetime(
        deserialize_message(stored_message).timestamp
    )
    stored_message.full_clean()
    stored_message.save()
```

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/159431910

Jakub89:
Closed via: #691

Issue #627, opened 2018-07-11
Each message should be stored with its timestamp not with store timestamp


MichalW-CP:
For timestamp in store_message we use datetime.datetime.now(timezone.utc). It should be parse_timestamp_to_utc_datetime(golem_message.timestamp).
```python
message_handlers.py

def store_message(
    golem_message:          message.base.Message,
    task_id:                str,
    subtask_id:             str,
):
    assert golem_message.TYPE in message.registered_message_types

    message_timestamp = datetime.datetime.now(timezone.utc)
    stored_message = StoredMessage(
        type        = golem_message.TYPE,
        timestamp   = message_timestamp,
        data        = copy.copy(golem_message).serialize(),
        task_id     = task_id,
        subtask_id  = subtask_id,
    )
    stored_message.full_clean()
    stored_message.save()

    return stored_message
```
In tests timestamp for stored_message is the same for each message that is stored. It should get timestamps individually for each message. 
In tests we can check if stored_message timestamp equals to message timestamp.
```python
tests/utils.py

    def _test_last_stored_messages(self, expected_messages, task_id, subtask_id, timestamp):
        assert isinstance(expected_messages, list)
        assert isinstance(task_id,           str)
        assert isinstance(subtask_id,        str)

        expected_message_types = [expected_message.TYPE for expected_message in expected_messages]

        for stored_message in StoredMessage.objects.order_by('-id')[:len(expected_message_types)]:
            self.assertIn(stored_message.type,                      expected_message_types)
            self.assertEqual(stored_message.task_id,                task_id)
            self.assertEqual(stored_message.subtask_id,             subtask_id)
            self.assertEqual(stored_message.timestamp.timestamp(),  self._parse_iso_date_to_timestamp(timestamp))

            expected_message_types.remove(stored_message.type)

        assert expected_message_types == []
```

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/158971816

Jakub89:
@dybi suggested solution for tests:
- Remove `self.assertEqual(stored_message.timestamp.timestamp(), self._parse_iso_date_to_timestamp(timestamp))` from `_test_last_stored_messages()`
- Add unit test for `store_message` that checks timestamp mechanism

I personally like this solution, because checking `expected_messages` with something like `expected_timestamp` to those messages is a little problematic.

rwrzesien:
[...]

PR #691, opened 2018-07-31
Add missing data migration for StoredMessage timestamp

Jakub89:
Resolves: #690 
data migration for issue #627 

please double check if that migration works as intended.

### DIFF
--- a/concent_api/core/migrations/0008_store_message_timestamp_fix.py
+++ b/concent_api/core/migrations/0008_store_message_timestamp_fix.py
@@ -1,0 +1,32 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+from common.helpers import deserialize_message
+from common.helpers import parse_timestamp_to_utc_datetime
+
+
+def switch_stored_message_timestamp_to_message_creation_time(apps, _schema_editor):
+    StoredMessage = apps.get_model('core', 'StoredMessage')
+
+    for stored_message in StoredMessage.objects.all():
+        stored_message.timestamp = parse_timestamp_to_utc_datetime(
+            deserialize_message(stored_message.data.tobytes()).timestamp
+        )
+        stored_message.full_clean()
+        stored_message.save()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('core', '0007_auto_20180719_1241'),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            switch_stored_message_timestamp_to_message_creation_time,
+            reverse_code=migrations.RunPython.noop
+        ),
+    ]


concent_api/core/migrations/0008_store_message_timestamp_fix.py
```diff
@@ -0,0 +1,32 @@
...
+from common.helpers import deserialize_message
+from common.helpers import parse_timestamp_to_utc_datetime
+
+
+def switch_stored_message_timestamp_to_message_creation_time(apps, _schema_editor):
+    StoredMessage = apps.get_model('core', 'StoredMessage')
+
+    for stored_message in StoredMessage.objects.all():
+        stored_message.timestamp = parse_timestamp_to_utc_datetime(
+            deserialize_message(stored_message.data).timestamp
```
rwrzesien: Shouldn't it be `.tobytes()` ?

Jakub89: I've change it, but it looks like it didn't make any different. 
P.S.
One other think about this migration is that all existing messages `created_at` field is going to get change to "current" timestamp. I hope it's not a problem?

kbeker: `created_at` should not get current time it uses `auto_now_add` which adds new time only when actually row is created now modified. Are you sure that this is changing this time?

kbeker:
Also please add that this PR resolves https://github.com/golemfactory/concent/issues/690

concent_api/core/migrations/0008_store_message_timestamp_fix.py
dybi: https://docs.djangoproject.com/en/2.0/ref/models/fields/#django.db.models.DateField.auto_now_add this is how `auto_now_add` works... if we dno't want this, we would need to get back to `default=datetime.datetime.now`
@cameel ? @rwrzesien

cameel: It's fine. We don't want to change the `created_at` timestamp. The object is not being created after all.

We would want to change `modified_at` if there was one. But there isn't because the object is never supposed to be modified in normal circumstances.

Actually, I said in #604 that we don't need modification timestamps in models that are considered read-only like this one but I did not consider migrations. I assumed that anyone changing the state outside of normal code (e.g. with a command-line SQL client) would not update the timestamps anyway. But migrations are kinda in between those situations. Maybe they would have a use after all...

cameel:
By the way, `debugging on the cluster` label was meant for features whose main purpose is to help with debugging problems that we have not reproduced locally yet. E.g. logging, debug data dumps, administrative tokens for getting files from storage, etc.

I'm removing it and I've added that description to the label.

concent_api/core/migrations/0008_store_message_timestamp_fix.py
Jakub89: Okey, my bad. `created_at` change for me, when I tested for the first time, because I make initial messages before that field was in the model :/ that's why when they were created for the first time it took `datetime.now`.

Issue #692, opened 2018-07-31
signing-service: we need default value for `initial_reconnect_delay` setting


bartoszbetka:
Now, to start the signing service, we always have to provide the value of `initial_reconnect_delay` setting. We want to use it only for non-standard use.

rwrzesien:
@bartoszbetka The value is always required, without it SigningService won't start. But if you want, I can add a default for it. What value would you expect ? Is `1` ok ?

cameel:
Seems fine to me.

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/159459270

PR #696, opened 2018-08-01
SigningService: Add default value for `initial_reconnect_delay` setting

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/692


signing_service/src/tests/tests.py
```diff
@@ -113,23 +114,24 @@ def setUp(self):
...
+        sys.argv += ['127.0.0.1', '--initial_reconnect_delay', '1', '--concent-cluster-port', '8000']
 
         args = _parse_arguments()
 
         self.assertEqual(args.concent_cluster_host, '127.0.0.1')
         self.assertEqual(args.concent_cluster_port, 8000)
 
-    def test_that_argument_parser_should_parse_correct_input_and_use_default_port(self):
-        sys.argv += ['127.0.0.1', '1']
+    def test_that_argument_parser_should_parse_correct_input_and_use_default_values(self):
```
dybi: something's wrong here 
1. test is called: `test_that_argument_parser_should_parse_correct_input_and_use_default_values`
2. some value of `initial_reconnect_delay` is being passed (`1`)
3. there's an assert that checks if `initial_reconnect_delay` is equal to `SIGNING_SERVICE_DEFAULT_INITIAL_RECONNECT_DELAY`
So basically, in this tests we check if user has entered default: `SIGNING_SERVICE_DEFAULT_INITIAL_RECONNECT_DELAY` ;)

signing_service/src/tests/tests.py
```diff
@@ -113,23 +114,24 @@ def setUp(self):
         sys.argv = sys.argv[:1]
 
     def test_that_argument_parser_should_parse_correct_input(self):
-        sys.argv += ['127.0.0.1', '1', '--concent-cluster-port', '8000']
+        sys.argv += ['127.0.0.1', '--initial_reconnect_delay', '1', '--concent-cluster-port', '8000']
```
dybi: there should be check for `initial_reconnect_delay` as well

signing_service/src/tests/tests.py

signing_service/src/tests/tests.py

PR #697, opened 2018-08-02
feature database lock for get or create subtask

pawelkisielewicz:
Resolves part https://github.com/golemfactory/concent/issues/412

### DIFF
--- a/concent_api/api-integration-force-get-task-result-test.py
+++ b/concent_api/api-integration-force-get-task-result-test.py
@@ -3,56 +3,26 @@
 import os
 import sys
 import hashlib
-from freezegun import freeze_time
 
 from golem_messages import message
-from golem_messages.factories.tasks import ReportComputedTaskFactory
 
 from common.helpers import get_current_utc_timestamp
 from common.helpers import get_storage_result_file_path
 from common.helpers import upload_file_to_storage_cluster
-from common.helpers import sign_message
-from api_testing_common import api_request
+from api_testing_common import api_request, get_force_get_task_result
 from api_testing_common import assert_condition
 from api_testing_common import count_fails
 from api_testing_common import create_client_auth_message
-from api_testing_common import create_signed_task_to_compute
 from api_testing_common import get_task_id_and_subtask_id
 from api_testing_common import PROVIDER_PRIVATE_KEY
 from api_testing_common import PROVIDER_PUBLIC_KEY
 from api_testing_common import REQUESTOR_PRIVATE_KEY
 from api_testing_common import REQUESTOR_PUBLIC_KEY
 from api_testing_common import run_tests
-from api_testing_common import timestamp_to_isoformat
 
 import requests
 
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "concent_api.settings")
-
-
-def get_force_get_task_result(task_id, subtask_id, current_time, cluster_consts, size, package_hash):
-    task_to_compute = create_signed_task_to_compute(
-        task_id=task_id,
-        subtask_id=subtask_id,
-        deadline=current_time,
-        price=0,
-    )
-
-    with freeze_time(timestamp_to_isoformat(current_time - 1)):
-        report_computed_task = ReportComputedTaskFactory(
-            task_to_compute=task_to_compute,
-            size=size,
-            package_hash=package_hash,
-            subtask_id=subtask_id,
-        )
-        sign_message(report_computed_task, PROVIDER_PRIVATE_KEY)
-
-    with freeze_time(timestamp_to_isoformat(current_time)):
-        force_get_task_result = message.concents.ForceGetTaskResult(
-            report_computed_task=report_computed_task,
-        )
-
-    return force_get_task_result
 
 
 @count_fails
--- a/concent_api/api-integration-multiprocess.py
+++ b/concent_api/api-integration-multiprocess.py
@@ -11,6 +11,7 @@ import requests
 from golem_messages.shortcuts import dump
 
 from api_testing_common import api_request
+from api_testing_common import get_force_get_task_result
 from api_testing_common import REQUESTOR_PRIVATE_KEY
 from api_testing_common import run_tests
 from common.helpers import get_current_utc_timestamp
--- a/concent_api/api-integration-multiprocess.py
+++ b/concent_api/api-integration-multiprocess.py
@@ -1,0 +1,91 @@
+#!/usr/bin/env python3
+
+import hashlib
+import os
+import sys
+import time
+from datetime import datetime
+from threading import Thread
+
+import requests
+from golem_messages.shortcuts import dump
+
+from api_testing_common import api_request
+from api_testing_common import REQUESTOR_PRIVATE_KEY
+from api_testing_common import run_tests
+from common.helpers import get_current_utc_timestamp
+
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "concent_api.settings")
+
+now = datetime.now()
+testing_subtask_id = str(now.year) + str(now.month) + str(now.day) + str(now.hour) + str(now.minute) + str(
+    now.second) + str(now.microsecond)
+
+responses = []
+
+
+def multiprocess(func):
+    def wrapper(*args, **kwargs):
+        for i in range(3):
+            t = Thread(target=func, args=(*args,), kwargs={**kwargs})
+            t.start()
+
+    return wrapper
+
+
+def test_case_that_multiple_requests_by_one_subtask_will_not_be_cause_of_server_error(cluster_consts, cluster_url, test_id):
+
+    send_correct_request(cluster_consts=cluster_consts, cluster_url=cluster_url)
+
+    time.sleep(4)
+    print('Responses = ' + str(responses))
+    assert 'AckForceGetTaskResult' in responses
+    responses.remove('AckForceGetTaskResult')
+    assert all(x == 'ServiceRefused' for x in responses)
+    print('Test passed successfully')
+
+
+@multiprocess
+def send_correct_request(cluster_consts, cluster_url):
+    current_time = get_current_utc_timestamp()
+    (subtask_id, task_id) = (testing_subtask_id, '130')
+
+    file_content = task_id
+    file_size = len(file_content)
+    file_check_sum = 'sha1:' + hashlib.sha1(file_content.encode()).hexdigest()
+
+    data = dump(
+        get_force_get_task_result(
+            task_id,
+            subtask_id,
+            current_time,
+            cluster_consts,
+            size=file_size,
+            package_hash=file_check_sum,
+        ),
+        REQUESTOR_PRIVATE_KEY,
+        CONCENT_PUBLIC_KEY,
+    )
+
+    response = api_request(
+        cluster_url,
+        'send',
+        REQUESTOR_PRIVATE_KEY,
+        CONCENT_PUBLIC_KEY,
+        data,
+        headers={
+            'Content-Type': 'application/octet-stream',
+        },
+        expected_status=200,
+        expected_content_type='application/octet-stream',
+    )
+    responses.append(response.__class__.__name__)
+
+
+if __name__ == '__main__':
+    try:
+        from concent_api.settings import CONCENT_PUBLIC_KEY
+        run_tests(globals())
+    except requests.exceptions.ConnectionError as exception:
+        print("\nERROR: Failed connect to the server.\n", file=sys.stderr)
+        sys.exit(str(exception))
--- a/concent_api/common/decorators.py
+++ b/concent_api/common/decorators.py
@@ -1,8 +1,10 @@
+import time
 from functools                      import wraps
 from logging                        import getLogger
 import traceback
 
-from django.db                      import transaction
+from django.db import DatabaseError
+from django.db import transaction
 from django.http                    import JsonResponse
 from django.http                    import HttpResponse
 from django.http                    import HttpResponseNotAllowed
@@ -169,9 +171,45 @@ def handle_errors_and_responses(database_name):
             try:
                 if database_name is not None:
                     sid = transaction.savepoint(using=database_name)
-                response_from_view = view(request, client_message, client_public_key, *args, **kwargs)
+                try:
+                    response_from_view = view(request, client_message, client_public_key, *args, **kwargs)
+                except DatabaseError as e:
+                    # This section is to catch Database errors caused by multiple requests about one Subtask.
+                    # This requests are waiting and retrying.
+                    log_string_message(
+                        logger,
+                        f'DatabaseError during process client request. Client public key: {client_public_key}.'
+                        f'Error: {e}'
+                    )
+                    sleep_time = settings.INITIAL_VIEW_RETRY_DELAY
+                    subtask_got_successfully = False
+                    response_from_view = None
+
+                    for counter in range(settings.MAX_VIEW_RETRIES):
+                        try:
+                            time.sleep(sleep_time)
+                            sleep_time *= 2
+                            response_from_view = view(request, client_message, client_public_key, *args, **kwargs)
+                            subtask_got_successfully = True
+                            break
+                        except DatabaseError as e:
+                            log_string_message(
+                                logger,
+                                f'DatabaseError during process client request. Client public key: {client_public_key}.'
+                                f'Number of retries {counter+1}/{settings.MAX_VIEW_RETRIES}'
+                                f'Error: {e}'
+                            )
+                    if not subtask_got_successfully:
+                        log_string_message(
+                            logger,
+                            f'DatabaseError during process client request. Client public key: {client_public_key}.'
+                            f'Maximum number of retries exceeded.'
+                        )
+                        raise DatabaseError
+
                 if database_name is not None:
                     transaction.savepoint_commit(sid, using=database_name)
+
             except ConcentBaseException as exception:
                 log_400_error(
                     logger,
--- a/concent_api/core/message_handlers.py
+++ b/concent_api/core/message_handlers.py
@@ -9,6 +9,7 @@ from typing import Union
 
 from django.conf import settings
 from django.core.mail import mail_admins
+from django.db import transaction
 from django.http import HttpResponse
 
 from constance import config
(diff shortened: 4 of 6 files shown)


concent_api/api-integration-multiprocess.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def test_case_that_multiple_requests_by_one_subtask_will_not_be_cause_of_server_error(cluster_consts, cluster_url, test_id):
+
+    send_correct_request(cluster_consts=cluster_consts, cluster_url=cluster_url)
+
+    time.sleep(4)
+    print('Responses = ' + str(responses))
+    assert 'AckForceGetTaskResult' in responses
+    responses.remove('AckForceGetTaskResult')
```
dybi: why would you remove anything from `responses`?

concent_api/common/decorators.py
```diff
@@ -169,9 +171,45 @@ def wrapper(request, client_message, client_public_key, *args, **kwargs):
             try:
                 if database_name is not None:
                     sid = transaction.savepoint(using=database_name)
-                response_from_view = view(request, client_message, client_public_key, *args, **kwargs)
+                try:
+                    response_from_view = view(request, client_message, client_public_key, *args, **kwargs)
+                except DatabaseError as e:
+                    # This section is to catch Database errors caused by multiple requests about one Subtask.
+                    # This requests are waiting and retrying.
```
dybi: This -> These

concent_api/common/decorators.py
```diff
@@ -169,9 +171,45 @@ def wrapper(request, client_message, client_public_key, *args, **kwargs):
             try:
                 if database_name is not None:
                     sid = transaction.savepoint(using=database_name)
-                response_from_view = view(request, client_message, client_public_key, *args, **kwargs)
+                try:
+                    response_from_view = view(request, client_message, client_public_key, *args, **kwargs)
+                except DatabaseError as e:
```
dybi: whole handling should be extracted to a separate function

concent_api/common/decorators.py
```diff
@@ -169,9 +171,45 @@ def wrapper(request, client_message, client_public_key, *args, **kwargs):
...
+                    # This section is to catch Database errors caused by multiple requests about one Subtask.
+                    # This requests are waiting and retrying.
+                    log_string_message(
```
dybi: You are logging basically the same message twice -here and few lines below...

concent_api/common/decorators.py
```diff
@@ -169,9 +171,45 @@ def wrapper(request, client_message, client_public_key, *args, **kwargs):
...
+                            subtask_got_successfully = True
+                            break
+                        except DatabaseError as e:
+                            log_string_message(
+                                logger,
+                                f'DatabaseError during process client request. Client public key: {client_public_key}.'
+                                f'Number of retries {counter+1}/{settings.MAX_VIEW_RETRIES}'
+                                f'Error: {e}'
+                            )
+                    if not subtask_got_successfully:
```
dybi: this is not necessary - please use `for ... else` instead of `subtask_got_successfully` flag

concent_api/common/decorators.py
```diff
@@ -169,9 +171,45 @@ def wrapper(request, client_message, client_public_key, *args, **kwargs):
...
+                        log_string_message(
```
dybi: too much logging of the same information

concent_api/common/decorators.py
dybi: http://book.pythontips.com/en/latest/for_-_else.html

concent_api/api-integration-force-get-task-result-test.py
```diff
@@ -3,58 +3,28 @@
...
 
 from golem_messages import message
-from golem_messages.factories.tasks import ReportComputedTaskFactory
 
 from common.helpers import get_current_utc_timestamp
 from common.helpers import get_storage_result_file_path
 from common.helpers import upload_file_to_storage_cluster
-from common.helpers import sign_message
-from api_testing_common import api_request
+from api_testing_common import api_request, get_force_get_task_result
```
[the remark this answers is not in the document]
pawelkisielewicz: checked it 100 times for such mistakes... but something left. Done :)

concent_api/api-integration-multiprocess.py
```diff
@@ -0,0 +1,92 @@
...
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "concent_api.settings")
+
+now = datetime.now()
+testing_subtask_id = str(now.year) + str(now.month) + str(now.day) + str(now.hour) + str(now.minute) + str(
+    now.second) + str(now.microsecond)
+
+responses = []
+
+
+def multiprocess(func):
```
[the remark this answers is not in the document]
pawelkisielewicz: done

concent_api/api-integration-multiprocess.py
```diff
@@ -0,0 +1,92 @@
...
+    return wrapper
+
+
+def test_case_that_multiple_requests_by_one_subtask_will_not_be_cause_of_server_error(cluster_consts, cluster_url, test_id):
+
+    send_correct_request(cluster_consts=cluster_consts, cluster_url=cluster_url)
+
+    time.sleep(4)
+    print('Responses = ' + str(responses))
+    assert 'AckForceGetTaskResult' in responses
```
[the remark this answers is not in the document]
pawelkisielewicz: done

concent_api/api-integration-multiprocess.py
```diff
@@ -0,0 +1,92 @@
...
+def multiprocess(func):
+    def wrapper(*args, **kwargs):
+        for i in range(3):
+            t = Thread(target=func, args=(*args,), kwargs={**kwargs})
+            t.start()
+
+    return wrapper
+
+
+def test_case_that_multiple_requests_by_one_subtask_will_not_be_cause_of_server_error(cluster_consts, cluster_url, test_id):
```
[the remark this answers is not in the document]
pawelkisielewicz: done

concent_api/api-integration-multiprocess.py
pawelkisielewicz: change to compare the whole list. `AckForceGetTaskResult` should be always first, so i can simplify check the responses

concent_api/common/decorators.py
pawelkisielewicz: done :)

concent_api/common/decorators.py
pawelkisielewicz: done

concent_api/common/decorators.py
pawelkisielewicz: They are very similar, but this 'first' is after first handling error. Each other is after another catch and retry so i think it is ok. They are not duplicated  and each one is important.

concent_api/common/decorators.py
pawelkisielewicz: Right... java doesn't have such construction. Will remember, but in this case, after move to separate function not necessary any more

concent_api/common/decorators.py
pawelkisielewicz: I think that in this case this should be logged, but maybe you are right... deleted

concent_api/api-integration-multi-threads-test.py
```diff
@@ -0,0 +1,98 @@
...
+from api_testing_common import assert_condition
+from api_testing_common import count_fails
+from api_testing_common import get_force_get_task_result
+from api_testing_common import REQUESTOR_PRIVATE_KEY
+from api_testing_common import run_tests
+from common.helpers import get_current_utc_timestamp
+
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "concent_api.settings")
+
+now = datetime.now()
```
rwrzesien: We usually name it `current_time`.

concent_api/api-integration-multi-threads-test.py
```diff
@@ -0,0 +1,98 @@
...
+testing_subtask_id = str(now.year) + str(now.month) + str(now.day) + str(now.hour) + str(now.minute) + str(
```
rwrzesien: It should either be in single line or something like
```
testing_subtask_id = (
    str(now.year) + 
    str(now.month) + 
    ...
)
```

concent_api/api-integration-multi-threads-test.py
```diff
@@ -0,0 +1,98 @@
...
+    now.second) + str(now.microsecond)
+
+responses = []  # type: str
+
+
+def multi_threads(func):
+    def wrapper(*args, **kwargs):
+        for i in range(3):
```
rwrzesien: Please make it configurable.

concent_api/api-integration-multi-threads-test.py
```diff
@@ -0,0 +1,98 @@
...
+    return wrapper
+
+
+@count_fails
+def test_case_that_multiple_requests_by_one_subtask_will_not_be_cause_of_server_error(cluster_consts, cluster_url, test_id):
+
+    send_correct_request(cluster_consts=cluster_consts, cluster_url=cluster_url)
+
+    time.sleep(4)
+    print('Responses = ' + str(responses))
```
rwrzesien: I think you forget to assign to responses in the call above ?

concent_api/api-integration-multi-threads-test.py
```diff
@@ -0,0 +1,98 @@
...
+
+    return wrapper
+
+
+@count_fails
+def test_case_that_multiple_requests_by_one_subtask_will_not_be_cause_of_server_error(cluster_consts, cluster_url, test_id):
+
+    send_correct_request(cluster_consts=cluster_consts, cluster_url=cluster_url)
+
+    time.sleep(4)
```
rwrzesien: I would put it in a `while` loop until `len(responses) = expected`.

concent_api/common/decorators.py
```diff
@@ -169,9 +175,13 @@ def wrapper(request, client_message, client_public_key, *args, **kwargs):
             try:
                 if database_name is not None:
                     sid = transaction.savepoint(using=database_name)
-                response_from_view = view(request, client_message, client_public_key, *args, **kwargs)
+                try:
+                    response_from_view = view(request, client_message, client_public_key, *args, **kwargs)
+                except DatabaseError as e:
+                    response_from_view = get_response_from_view_with_db_retry(view, e, request, client_message, client_public_key, *args, **kwargs)
```
rwrzesien: Please split into multiple lines, this one is too long.

concent_api/common/decorators.py
```diff
@@ -298,3 +308,39 @@ def wrapper(*args, **kwargs):
...
+        logger,
+        f'DatabaseError during process client request. Client public key: {client_public_key}.'
+        f'Error: {error}'
+    )
+    sleep_time = settings.INITIAL_VIEW_RETRY_DELAY
+
+    for counter in range(settings.MAX_VIEW_RETRIES):
+        try:
+            time.sleep(sleep_time)
+            sleep_time *= 2
```
rwrzesien: Do we really want to wait over 30 minutes for response ? I think the maximum is way to much and I would suggest shorter `INITIAL_VIEW_RETRY_DELAY`.

concent_api/api-integration-multi-threads-test.py
pawelkisielewicz: done

concent_api/api-integration-multi-threads-test.py
pawelkisielewicz: done

concent_api/api-integration-multi-threads-test.py
pawelkisielewicz: done in for loop with `break` if length expected. If there will be some error instead of response test would be never end. But generally great idea with such loop, will be save and much faster :)

concent_api/api-integration-multi-threads-test.py
pawelkisielewicz: I'm not sure I understand the question.. I need only names of messages, whole messages are not necessary. In this line I print responses only for someone, who will use this test to see and remember, which messages are expected

concent_api/common/decorators.py
```diff
@@ -336,3 +308,39 @@ def wrapper(*args, **kwargs):
             )
             raise
     return wrapper
+
+
+def get_response_from_view_with_db_retry(
```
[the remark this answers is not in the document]
pawelkisielewicz: done, hope you will like it :)

concent_api/common/decorators.py
pawelkisielewicz: done

concent_api/common/decorators.py
pawelkisielewicz: Current settings are:
`MAX_VIEW_RETRIES = 10`
`INITIAL_VIEW_RETRY_DELAY = 1  # Value in seconds`
I think it was about 8 minutes.. But you are right. Is it ok??:
`MAX_VIEW_RETRIES = 10`
`INITIAL_VIEW_RETRY_DELAY = 0.5`

concent_api/common/decorators.py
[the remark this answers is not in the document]
pawelkisielewicz: I tried to use `else`, but pylint says:
`Else clause on loop without a break statement (useless-else-on-loop)`. I think it is better too leave it like this instead of another one disable pylint. What do you think?

concent_api/api-integration-multi-threads-test.py
```diff
@@ -0,0 +1,105 @@
...
+        return wrapper
+    return call_function_in_threads
+
+
+@count_fails
+def test_case_that_multiple_requests_by_one_subtask_will_not_be_cause_of_server_error(cluster_consts, cluster_url, test_id):
+
+    send_correct_request(cluster_consts=cluster_consts, cluster_url=cluster_url)
+
+    for i in range(3 * number_of_testing_threads):
```
rwrzesien: Wait, why it is `3 *` ? Wasn't it just `3` previously ?

concent_api/api-integration-multi-threads-test.py
```diff
@@ -0,0 +1,105 @@
...
+from common.helpers import get_current_utc_timestamp
+
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "concent_api.settings")
+
+now = datetime.now()
+testing_subtask_id = str(now.year) + str(now.month) + str(now.day) + str(now.hour) + str(now.minute) + str(now.second) + str(now.microsecond)
+
+responses = []  # type: list
+
+number_of_testing_threads = 3
```
rwrzesien: Please make it uppercase as it is constant now.

concent_api/api-integration-multi-threads-test.py
pawelkisielewicz: yes, but in this case I use it in only one line and it is important to be short. If you want to I will change it, but it was the reason.

concent_api/api-integration-multi-threads-test.py
rwrzesien: But I don't think that `for` loop guarantees that. It still asserts that response will come in 0,5 second. What if after all 3 or 9 runs there will not be all responses ? It will just break anyway. My idea was to run loop until all responses are there for sure. Off course you can count error responses as responses too. Or if you expect errors that are not returning responses, you can count exceptions or set a global time limit for it.

concent_api/common/decorators.py
rwrzesien: The maximum delay time you can reach is `2**9=512` but you have already waited `256`, `128` etc. I think you should choose values that would overall not be longer that 1 minute. But please verify with @cameel too. I am thinking about it more like a response for web page, maybe in Golem network this can take much longer.

concent_api/api-integration-multi-threads-test.py
pawelkisielewicz: It were hardcoded 3 threads. In this case it is really number of repeats --> waiting time. I decided to do it like this, because if someone would change `number_of_testing_threads` possible waiting time should be also different. And it should depend on number of testing threads

concent_api/api-integration-multi-threads-test.py
pawelkisielewicz: done :)

concent_api/api-integration-multi-threads-test.py
rwrzesien: Yes, I mean the whole flow of this. You create a list outside of this function, then add items inside the function, and then check list outside of the function again. What I am asking is if the function can't return the value that will be added to the list directly ?

concent_api/common/decorators.py
```diff
@@ -298,3 +312,30 @@ def wrapper(*args, **kwargs):
             )
             raise
     return wrapper
+
+
+def get_response_from_view_with_secure_retry_if_database_locked(
```
pawelkisielewicz: @cameel look at my last comment in https://github.com/golemfactory/concent/issues/412

concent_api/api-integration-multi-threads-test.py
```diff
@@ -0,0 +1,105 @@
...
+from api_testing_common import count_fails
+from api_testing_common import get_force_get_task_result
+from api_testing_common import REQUESTOR_PRIVATE_KEY
+from api_testing_common import run_tests
+from common.helpers import get_current_utc_timestamp
+
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "concent_api.settings")
+
+now = datetime.now()
+testing_subtask_id = str(now.year) + str(now.month) + str(now.day) + str(now.hour) + str(now.minute) + str(now.second) + str(now.microsecond)
```
kbeker: Maybe it is good time to start using UUID as ours id's?

rwrzesien: I have actually skipped the context of this line, so yes, generating `subtask_id` should look much simpler and what @kbeker proposed is good idea.

kbeker: @rwrzesien maybe we should close this issue as fast as it possible and open new issue with changing all places where `subtask_id` and `task_id` is created to UUID?

kbeker: I've just add issue with adding to our code UUID id's: https://github.com/golemfactory/concent/issues/705

rwrzesien: Works for me.

concent_api/api-integration-multi-threads-test.py
pawelkisielewicz: It is because of multi-threading. I couldn't get response directly from other threads, but it is possible to create list with access from all threads. Each thread saves it's response to this list and it is checked in main thread.

concent_api/api-integration-multi-threads-test.py
pawelkisielewicz: done as discussed. I only left time.sleep(0.1), without this in local tests everything was longer

concent_api/core/message_handlers.py
```diff
@@ -1242,49 +1243,50 @@ def store_or_update_subtask(
...
-            ack_report_computed_task=ack_report_computed_task,
-            reject_report_computed_task=reject_report_computed_task,
-            subtask_results_accepted=subtask_results_accepted,
-            subtask_results_rejected=subtask_results_rejected,
-            force_get_task_result=force_get_task_result,
-        )
-    return subtask
+    with transaction.atomic(using='control'):
+        try:
+            subtask = Subtask.objects.select_for_update().get(
```
cameel: Only create was supposed to be immediately committed. Update should be rolled back if the view fails.

concent_api/common/decorators.py
```diff
@@ -298,3 +312,30 @@ def wrapper(*args, **kwargs):
...
+    client_public_key: bytes,
+    *args: Optional[Any],
+    **kwargs: Optional[Any],
+) -> HttpResponse:
+    sleep_time = settings.INITIAL_VIEW_RETRY_DELAY
+    for counter in range(settings.MAX_VIEW_RETRIES):
+        try:
+            response_from_view = view(request, client_message, client_public_key, *args, **kwargs)
+            return response_from_view
+        except DatabaseError as e:
```
cameel: Please use full names (`exception` rather than `e`).

concent_api/core/message_handlers.py
```diff
@@ -1242,49 +1243,50 @@ def store_or_update_subtask(
...
-            task_to_compute=task_to_compute,
-            report_computed_task=report_computed_task,
-            ack_report_computed_task=ack_report_computed_task,
-            reject_report_computed_task=reject_report_computed_task,
-            subtask_results_accepted=subtask_results_accepted,
-            subtask_results_rejected=subtask_results_rejected,
-            force_get_task_result=force_get_task_result,
-        )
-    return subtask
+    with transaction.atomic(using='control'):
```
cameel: There's a big problem here. You have recently turned my attention to the fact that a savepoint gets rolled back when the outer transaction is rolled back. So using nested `@atomic` does no guarantee that the changes will remain in the database.

Unfortunately I don't see any workaround other than doing two separate transactions. You'll have to use [`@non_atomic_requests`](https://docs.djangoproject.com/en/1.11/topics/db/transactions/#django.db.transaction.non_atomic_requests) on the views and create two non-nested `@atomic` blocks. The first one should create the `Subtask`. The other should do all the other operations.

concent_api/common/decorators.py
cameel: Is this retry for the lock or for the `get_or_create()`?
- If it's for the lock, then we don't need it after all (as we have determined when working on the code for workers).
- If it's for the creation, then we should only retry once. And only the creation operation should be retried.

cameel:
@pawelkisielewicz When closing pull request please always refer to related pull requests or issues if they're available.

In this case the problem this pull request was meant to solve will be covered by #718.

PR #698, opened 2018-08-02
Add passing secrets from file, enviroment and console to Singing Service

### LINKED ISSUE #671: Signing Service: Passing secrets in files or environment variables
This is a part of the implementation of the "Concent Signing Service" blueprint (#597). This task depends on #633.

Currently Signing Service needs two secrets: Ethereum private key and Sentry DSN. The private key needs to be provided in a file and DSN directly on the command line. Let's make it more consistent. It should be possible to provide each secret in three ways:
- On the command line (`--<secret>` option)
- In an environment variable  (`--<secret>-from-env` option)
- In a file  (`--<secret>-path` option)

Only one option can be specified. Variable should be the default.

### Examples
``` bash
export ETHEREUM_PRIVATE_KEY=123-key-456
./signing-service.sh --ethereum-private-key-from-env
```

``` bash
echo "123-dsn-456" > /tmp/dsn.txt
./signing-service.sh --sentry-dsn-path=/tmp/dsn.txt
```

``` bash
./signing-service.sh --ethereum-private 123-key-456
```

kbeker:
Resolves #671


signing_service/src/signing_service.py
```diff
@@ -126,6 +127,40 @@ def _validate_arguments(self):
...
+                try:
+                    with open(values) as file:
+                        self.const = file.read()
+                except FileNotFoundError as file_not_found_error:
+                    crash_logger.error(f'{file_not_found_error}')
+                    raise
+            elif values:
+                self.const = values
+            elif self.env_variable_name:
+                self.const = os.environ.get(self.env_variable_name)
```
cameel: There's no `else`. What happens if none of these conditions is true?

If it's not possible then you should have `assert False` in that `else`.

signing_service/src/signing_service.py
```diff
@@ -126,6 +127,40 @@ def _validate_arguments(self):
 
 
 def _parse_arguments() -> argparse.Namespace:
+
+    def get_secret_variable_wrapper(env_variable_name=None, file=False, base64_convert=False):
+        def wrapper(**kwargs):
+            return GetSecretVariable(env_variable_name, file, base64_convert, **kwargs)
```
cameel: Why do you need this wrapper? It does not do anything special. Why not just call the constructor directly?

kbeker: Because this wrapper passing all arguments which I don't have to pass again in constructor like `option_strings`, `dest` or `nargs`. If I removed this wrapper then place where should I call this class would be very big piece of code - using wrapper I can do same thing in one line.

signing_service/src/signing_service.py
```diff
@@ -126,6 +127,40 @@ def _validate_arguments(self):
...
+
+        return wrapper
+
+    class GetSecretVariable(argparse.Action):
+        def __init__(self, env_variable_name, file, base64_convert, option_strings, dest, nargs=None):
+            self.env_variable_name = env_variable_name
+            self.file = file
+            self.base64_convert = base64_convert
+            if self.env_variable_name:
+                nargs = 0
```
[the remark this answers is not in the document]
kbeker: But it is necessary in this case because if we call `--ethereum-private-key-from-env` we don't need any parameter for this argument and exactly `nargs` is responsible for this

signing_service/src/signing_service.py
```diff
@@ -126,6 +127,40 @@ def _validate_arguments(self):
 
 
 def _parse_arguments() -> argparse.Namespace:
+
+    def get_secret_variable_wrapper(env_variable_name=None, file=False, base64_convert=False):
```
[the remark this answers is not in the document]
dybi: `from_file` might be an option here

signing_service/src/signing_service.py
```diff
@@ -126,6 +127,40 @@ def _validate_arguments(self):
...
+        def __init__(self, env_variable_name, file, base64_convert, option_strings, dest, nargs=None):
+            self.env_variable_name = env_variable_name
+            self.file = file
+            self.base64_convert = base64_convert
+            if self.env_variable_name:
+                nargs = 0
+            super(GetSecretVariable, self).__init__(option_strings=option_strings, dest=dest, nargs=nargs)
+
+        def __call__(self, parser, namespace, values, option_string=None):
+            if values and self.file:
```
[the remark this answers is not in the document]
kbeker: Done

signing_service/src/signing_service.py
```diff
@@ -143,11 +178,39 @@ def _parse_arguments() -> argparse.Namespace:
...
+    )
+    ethereum_private_key_parser_group.add_argument(
+        '--ethereum-private-key-path',
+        dest='ethereum_private_key',
+        action=get_secret_variable_wrapper(file=True, base64_convert=True),
+    )
+    ethereum_private_key_parser_group.add_argument(
+        '--ethereum-private-key-from-env',
+        dest='ethereum_private_key',
+        action=get_secret_variable_wrapper(env_variable_name='ETHEREUM_PRIVATE_KEY', base64_convert=True),
```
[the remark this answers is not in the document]
kbeker: Separated to new commit

signing_service/signing_service/tests/tests.py
```diff
@@ -174,13 +188,50 @@ def test_that_argument_parser_should_parse_correct_input_and_use_default_values(
...
+
+        self.assertEqual(args.sentry_dsn, self.sentry_dsn)
+        self.assertEqual(args.ethereum_private_key, self.ethereum_private_key)
+
+    def test_that_argument_parses_should_fail_if_file_with_secrets_is_missing(self):
+        sys.argv += ['127.0.0.1', self.concent_public_key_encoded, '--sentry-dsn-path', '/not_existing_path/file.txt']
+        with self.assertRaises(FileNotFoundError):
             _parse_arguments()
 
+    def test_that_argument_parser_should_parse_correct_sentry_dsn_and_ethereum_private_key_if_files_exist(self):
```
dybi: files are not deleted after test. Please fix it

kbeker: Oh, I completely forgot about it. Done

kbeker:
@cameel All requested changes are supplied

signing_service/signing_service/signing_service.py
```diff
@@ -135,6 +135,37 @@ def _validate_arguments(self):
...
+        return wrapper
+
+    class SecretProvider(argparse.Action):
+        def __init__(self, read_command_line, env_variable_name, use_file, base64_convert, option_strings, dest):
+            self.read_command_line = read_command_line
+            self.env_variable_name = env_variable_name
+            self.use_file = use_file
+            self.base64_convert = base64_convert
+            if self.env_variable_name is not None:
+                command_line_arguments = 0
```
cameel: This should be someting like `num_command_line_arguments`, `command_line_argument_count` or `argument_count`. It's a number, not a list.

signing_service/signing_service/signing_service.py
```diff
@@ -135,6 +135,37 @@ def _validate_arguments(self):
...
+                with open(values) as file:
+                    self.const = file.read()
+            elif values is not None and self.read_command_line:
+                self.const = values
+            elif self.env_variable_name is not None:
+                self.const = os.environ.get(self.env_variable_name)
+            else:
+                assert False
+            if self.base64_convert and isinstance(self.const, str):
+                self.const = b64decode(self.const)
```
cameel: What if `self.base64_convert` but not `isinstance(self.const, str)`? Should anything other than `str` be even allowed in such a case?

signing_service/signing_service/signing_service.py
```diff
@@ -135,16 +135,16 @@ def _validate_arguments(self):
...
-            return GetSecretVariable(env_variable_name, use_file, **kwargs)
+            return GetSecretVariable(env_variable_name, use_file, base64_convert, **kwargs)
         return wrapper
 
     class GetSecretVariable(argparse.Action):
-        def __init__(self, env_variable_name, use_file, option_strings, dest, nargs=None):
+        def __init__(self, env_variable_name, use_file, base64_convert, option_strings, dest, nargs=None):
             self.env_variable_name = env_variable_name
             self.use_file = use_file
+            self.base64_convert = base64_convert
```
cameel: Adding support for `base64_convert` looks like something that should be added in a separate commit. It's needed for the Ethereum key but it's an independent change.

signing_service/signing_service/signing_service.py
```diff
@@ -159,13 +190,42 @@ def _parse_arguments() -> argparse.Namespace:
...
+        action=secret_provider_factory(env_variable_name='ETHEREUM_PRIVATE_KEY', base64_convert=True),
+    )
+    parser.set_defaults(ethereum_private_key=b64decode(os.environ.get('ETHEREUM_PRIVATE_KEY')) if os.environ.get('ETHEREUM_PRIVATE_KEY') is not None else None)
+
+    sentry_dsn_parser_group = parser.add_mutually_exclusive_group()
+    sentry_dsn_parser_group.add_argument(
         '--sentry-dsn',
-        default='',
         dest='sentry_dsn',
-        help=f'Sentry DSN for error reporting.',
```
cameel: Why are you removing the description? It sounds useful. You should have it for every option.

signing_service/signing_service/signing_service.py
```diff
@@ -135,6 +135,37 @@ def _validate_arguments(self):
 
 
 def _parse_arguments() -> argparse.Namespace:
+    def secret_provider_factory(read_command_line=False, env_variable_name=None, use_file=False, base64_convert=False):
```
cameel: As it turned out when we talked last time - it's not really a factory. More like factory factory :)

Maybe call it `make_secret_provider_factory()`?

signing_service/signing_service/signing_service.py
kbeker: Done

signing_service/signing_service/signing_service.py
kbeker: It shouldn't. We expect that only string that is convertible from base64 to bytes will be passed here as `self.const`.

signing_service/signing_service/signing_service.py
kbeker: Restored

signing_service/signing_service/signing_service.py
kbeker: I remember, but I didn't know which name will be the best :P

signing_service/signing_service/signing_service.py
kbeker: OK, I will add this as separate commit.

signing_service/signing_service/signing_service.py
cameel: There's really no good name for it :) But `make_...` at least hints at what's going on inside.

signing_service/signing_service/signing_service.py
cameel: And can user specify anything that will not be converted to string?
- If not, add an `assert`.
- If yes, this should be treated as any other error in parameters.

cameel: I'd use the first one. But you could make it more concise:

``` python
if self.base64_convert:
    assert isinstance(self.constr, str)
    self.const = b64decode(self.const)
```

PR #700, opened 2018-08-03
Add database locks for concent workers

pawelkisielewicz:
Resolves part https://github.com/golemfactory/concent/issues/412
Second part resolved in https://github.com/golemfactory/concent/pull/697/

### DIFF
--- a/concent_api/conductor/tasks.py
+++ b/concent_api/conductor/tasks.py
@@ -3,9 +3,13 @@ from typing import List
 
 from celery import shared_task
 from mypy.types import Optional
+from django.db import DatabaseError
 from django.db import transaction
 
 from core import tasks
+from core.constants import CELERY_LOCKED_SUBTASK_DELAY
+from core.constants import MAXIMUM_VERIFICATION_RESULT_TASK_RETRIES
+from core.models import Subtask
 from core.validation import validate_frames
 from common.constants import ErrorCode
 from common.decorators import log_task_errors
@@ -24,11 +28,13 @@ from .models import VerificationRequest
 logger = logging.getLogger(__name__)
 
 
-@shared_task
+@shared_task(bind=True)
 @provides_concent_feature('conductor-worker')
 @log_task_errors
 @transaction.atomic(using='storage')
+@transaction.atomic(using='control')
 def blender_verification_request(
+    self,
     subtask_id: str,
     source_package_path: str,
     result_package_path: str,
@@ -49,6 +55,24 @@ def blender_verification_request(
         f'Verification_deadline: {verification_deadline}',
         f'With blender_crop_script: {bool(blender_crop_script)}',
     )
+
+    # Worker locks database row corresponding to the subtask in the subtask table.
+    if Subtask.objects.filter(subtask_id=subtask_id).exists():  # pylint: disable=no-member
+        try:
+            Subtask.objects.select_for_update(nowait=True).get(subtask_id=subtask_id)
+        except DatabaseError:
+            logging.warning(
+                f'Subtask object with ID {subtask_id} database row is locked, '
+                f'retrying task {self.request.retries}/{self.max_retries}'
+            )
+            # If the row is already locked, task fails so that Celery can retry later.
+            self.retry(
+                countdown=CELERY_LOCKED_SUBTASK_DELAY,
+                max_retries=MAXIMUM_VERIFICATION_RESULT_TASK_RETRIES,
+                throw=False,
+            )
+            return
+
     validate_frames(frames)
     assert isinstance(output_format, str)
     assert isinstance(verification_deadline, int)
@@ -103,10 +127,12 @@ def blender_verification_request(
         verification_request.save()
 
 
-@shared_task
+@shared_task(bind=True)
 @log_task_errors
 @transaction.atomic(using='storage')
+@transaction.atomic(using='control')
 def upload_acknowledged(
+    self,
     subtask_id: str,
     source_file_size: str,
     source_package_hash: str,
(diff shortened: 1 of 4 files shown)


concent_api/conductor/tasks.py
```diff
@@ -123,6 +149,22 @@ def upload_acknowledged(
     )
     assert isinstance(subtask_id, str)
 
+    # Worker locks database row corresponding to the subtask in the subtask table.
+    if Subtask.objects.filter(subtask_id=subtask_id).exists():  # pylint: disable=no-member
```
pawelkisielewicz: I think if we are starting tasks, Subtask corresponding with subtask_id should exist in our database. Now we have test `test_that_upload_acknowledged_task_should_log_error_when_verification_request_with_given_subtask_id_does_not_exist`, but `upload_acknowledged` is called in `upload_finished`. `upload_finished` is called in `blender_verification_request` or `report_upload`. I think we should be sure (by validation at the beginning of first celery task or just before call), that subtask exist in database.  It should be checked at the beginning, not in some function in the middle of flow. This problem does not concern this issue, so I decided to cover it like "easiest way", but I think it should be changed in separate issue

concent_api/conductor/tasks.py
```diff
@@ -49,6 +55,24 @@ def blender_verification_request(
...
+                f'retrying task {self.request.retries}/{self.max_retries}'
+            )
+            # If the row is already locked, task fails so that Celery can retry later.
+            self.retry(
+                countdown=CELERY_LOCKED_SUBTASK_DELAY,
+                max_retries=MAXIMUM_VERIFICATION_RESULT_TASK_RETRIES,
+                throw=False,
+            )
+            return
+
```
kbeker: Conductor doesn't have access to `Control` database. It won't work on cluster.

concent_api/conductor/tasks.py
```diff
@@ -123,6 +149,22 @@ def upload_acknowledged(
...
+                f'Subtask object with ID {subtask_id} database row is locked, '
+                f'retrying task {self.request.retries}/{self.max_retries}'
+            )
+            # If the row is already locked, task fails so that Celery can retry later.
+            self.retry(
+                countdown=CELERY_LOCKED_SUBTASK_DELAY,
+                max_retries=MAXIMUM_VERIFICATION_RESULT_TASK_RETRIES,
+                throw=False,
+            )
+            return
```
kbeker: Also here. It won't work on Conductor.

concent_api/core/tasks.py
```diff
@@ -29,12 +29,25 @@
...
-@shared_task
+@shared_task(bind=True)
 @log_task_errors
 @transaction.atomic(using='control')
-def upload_finished(subtask_id: str):
+def upload_finished(self, subtask_id: str):
+    # Worker locks database row corresponding to the subtask in the subtask table.
     try:
-        subtask = Subtask.objects.get(subtask_id=subtask_id)
+        subtask = Subtask.objects.select_for_update(nowait=True).get(subtask_id=subtask_id)
```
cameel: Why still `nowait`? Last time we talked about it, it turned out that `select_for_update()` can wait on its own and it's better to rely on it. So neither retry nor catching `DatabaseError` is necessary.

concent_api/core/tasks.py
```diff
@@ -29,12 +29,25 @@
...
+    except DatabaseError:
+        logging.warning(
+            f'Subtask object with ID {subtask_id} database row is locked, '
+            f'retrying task {self.request.retries}/{self.max_retries}'
```
cameel: The message is misleading. It can fail also for other reasons.

We still want to retry then but the error message should be different. You should definitely log the original message with `logger.exception()`.

concent_api/core/tasks.py
```diff
@@ -29,12 +29,25 @@
...
 @log_task_errors
 @transaction.atomic(using='control')
-def upload_finished(subtask_id: str):
+def upload_finished(self, subtask_id: str):
+    # Worker locks database row corresponding to the subtask in the subtask table.
     try:
-        subtask = Subtask.objects.get(subtask_id=subtask_id)
+        subtask = Subtask.objects.select_for_update(nowait=True).get(subtask_id=subtask_id)
+    except DatabaseError:
+        logging.warning(
```
cameel: Shouldn't it be `logger.` rather than `logging.`?

concent_api/core/tasks.py
pawelkisielewicz: right.. done

concent_api/core/tasks.py
pawelkisielewicz: done

concent_api/core/tasks.py
pawelkisielewicz: it should be. Done

concent_api/core/tasks.py
```diff
@@ -29,14 +29,27 @@
...
+        # If the row is already locked, task fails so that Celery can retry later.
+        self.retry(
+            countdown=CELERY_LOCKED_SUBTASK_DELAY,
+            max_retries=MAXIMUM_VERIFICATION_RESULT_TASK_RETRIES,
+            throw=False,
+        )
+        return
     except Subtask.DoesNotExist:
-        logging.error(f'Task `upload_finished` tried to get Subtask object with ID {subtask_id} but it does not exist.')
+        logger.error(f'Task `upload_finished` tried to get Subtask object with ID {subtask_id} but it does not exist.')
```
kbeker: This should be in different commit. This change not related with database locks

concent_api/conductor/tasks.py
```diff
@@ -55,23 +56,31 @@ def blender_verification_request(
...
+                scene_file=scene_file,
+                blender_crop_script=blender_crop_script,
+            )
+
+            store_frames(
+                blender_subtask_definition=blender_subtask_definition,
+                frame_list=frames,
+            )
+    except DatabaseError as exception:
+        log_string_message(
```
kbeker: Better to use here `logger.warning`. It should be unified everywhere. If you use in `core.tasks.upload_finished` `logger.warning` then here must be the same.

cameel:
The problem this pull request was meant to solve will be covered by #718.

PR #687, opened 2018-07-30
Add golem messages for signing transactions

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/632

Moved from golem-messages, original pull request https://github.com/golemfactory/golem-messages/pull/240.

PR #712, opened 2018-08-06
Add golem messages for signing transactions

### LINKED ISSUE #632: Signing Service: Protocol messages
This is a part of the implementation of the "Concent Signing Service" blueprint (#597).

#### Overview
The Signing Service exchanges data with Concent using Golem messages. We need to several new types for that purpose.

Ideally these messages would be added only in Concent. They are not a part of the protocol used by Golem and Concent. But last time we talked about it, golem-messages required refactoring to allow defining arbitrary messages without registering them from within the library (https://github.com/golemfactory/golem-messages/issues/103). If that refactoring has not been done yet, submit our new messages in a pull request for inclusion in golem-messages. This will be a temporary workaround and we'll move them to Concent later, when it's possible.

#### `TransactionSigningRequest`
| Field                           | Type                   | Size           |
|---------------------------------|------------------------|----------------|
| `nonce`                         | int                    | up to 32 bytes |
| `gasprice`                      | int                    | up to 32 bytes |
| `startgas`                      | int                    | up to 32 bytes |
| `to`                            | string                 | 20 bytes       |
| `value`                         | int                    | up to 32 bytes |
| `data`                          | binary                 | unlimited      |
| `from`                          | string                 | 20 bytes       |

#### `SignedTransaction`
| Field                           | Type                   | Size           |
|---------------------------------|------------------------|----------------|
| `nonce`                         | int                    | up to 32 bytes |
| `gasprice`                      | int                    | up to 32 bytes |
| `startgas`                      | int                    | up to 32 bytes |
| `to`                            | string                 | 20 bytes       |
| `value`                         | int                    | up to 32 bytes |
| `data`                          | binary                 | unlimited      |
| `v`                             | int                    | 1 byte         |
| `r`                             | int                    | 32 bytes       |
| `s`                             | int                    | 32 bytes       |

#### `TransactionRejected`
| Field                           | Type                         | Size           |
|---------------------------------|------------------------------|----------------|
| `nonce`                         | int                          | up to 32 bytes |
| `reason`                        | `TransactionRejectionReason` |

#### `TransactionRejectionReason` enum
- `InvalidTransaction` 
- `UnauthorizedAccount`

rwrzesien:
Resolves #632

Moved from golem-messages, original pull request golemfactory/golem-messages#240.

Reopened from https://github.com/golemfactory/concent/pull/687.

### DIFF
--- a/middleman_protocol/middleman_protocol/concent_golem_messages/message.py
+++ b/middleman_protocol/middleman_protocol/concent_golem_messages/message.py
@@ -1,0 +1,162 @@
+import enum
+
+from golem_messages.message.base import Message
+from golem_messages.register import library
+from golem_messages.validators import validate_integer
+from golem_messages.validators import validate_varchar
+
+from middleman_protocol.concent_golem_messages.validators import validate_bytes
+from middleman_protocol.concent_golem_messages.validators import validate_maximum_int_length
+
+
+CONCENT_CUSTOM_MSG_BASE = 9000
+
+
+class NonceAbstractMessage(Message):
+    """
+    Abstract message containing `nonce` field and its validation.
+    """
+
+    __slots__ = [
+        'nonce',
+    ] + Message.__slots__
+
+    def deserialize_slot(self, key, value):
+        value = super().deserialize_slot(key, value)
+        if key == 'nonce':
+            validate_integer(
+                field_name=key,
+                value=value,
+            )
+            validate_maximum_int_length(
+                field_name=key,
+                value=value,
+                maximum_allowed_length=78,
+            )
+        return value
+
+
+class TransactionAbstractMessage(NonceAbstractMessage):
+    """
+    Abstract message containing transaction data and its validation.
+    """
+
+    __slots__ = [
+        'gasprice',
+        'startgas',
+        'to',
+        'value',
+        'data',
+    ] + NonceAbstractMessage.__slots__
+
+    def deserialize_slot(self, key, value):
+        value = super().deserialize_slot(key, value)
+        if key == 'to':
+            validate_varchar(
+                field_name=key,
+                value=value,
+                max_length=20,
+            )
+        if key == 'data':
+            validate_bytes(
+                field_name=key,
+                value=value,
+            )
+        if key in ('gasprice', 'startgas', 'value'):
+            validate_integer(
+                field_name=key,
+                value=value
+            )
+            validate_maximum_int_length(
+                field_name=key,
+                value=value,
+                maximum_allowed_length=78,
+            )
+        return value
+
+
+@library.register(CONCENT_CUSTOM_MSG_BASE + 1)
+class TransactionSigningRequest(TransactionAbstractMessage):
+    """
+    Message sent from SCI transaction signing callback to a Concent,
+    containing data about transaction which client wants to sign using SigningService.
+    """
+
+    __slots__ = [
+        'from',
+    ] + TransactionAbstractMessage.__slots__
+
+    def deserialize_slot(self, key, value):
+        value = super().deserialize_slot(key, value)
+        if key == 'from':
+            validate_varchar(
+                field_name=key,
+                value=value,
+                max_length=20,
+            )
+        return value
+
+
+@library.register(CONCENT_CUSTOM_MSG_BASE + 2)
+class SignedTransaction(TransactionAbstractMessage):
+    """
+    Message sent from SigningService to the Concent,
+    if transaction was successfully signed,
+    containing data about transaction and its signature.
+
+    Concent should copy the signature data to the transaction object passed
+    to the callback by SCI.
+    """
+
+    __slots__ = [
+        'v',
+        'r',
+        's',
+    ] + TransactionAbstractMessage.__slots__
+
+    def deserialize_slot(self, key, value):
+        value = super().deserialize_slot(key, value)
+        if key in ('v', 'r', 's'):
+            validate_integer(
+                field_name=key,
+                value=value,
+            )
+        if key == 'v':
+            validate_maximum_int_length(
+                field_name=key,
+                value=value,
+                maximum_allowed_length=3,
+            )
+        if key in ('r', 's'):
+            validate_maximum_int_length(
+                field_name=key,
+                value=value,
+                maximum_allowed_length=78,
+            )
+        return value
+
+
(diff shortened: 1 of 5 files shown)


concent_api/concent_golem_messages/message.py
```diff
@@ -0,0 +1,133 @@
...
+    """
+    Abstract message containing `nonce` field and its validation.
+    """
+
+    __slots__ = [
+        'nonce',
+    ] + Message.__slots__
+
+    def deserialize_slot(self, key, value):
+        value = super().deserialize_slot(key, value)
```
rwrzesien: @kbeker https://github.com/golemfactory/concent/pull/687#discussion_r207895227
> As this messages are in Concent we can make more restrict validation. There can be added validation which would assume that nonce has 32 bytes ~ 78 numbers.

:+1: done

concent_api/concent_golem_messages/message.py
```diff
@@ -0,0 +1,133 @@
...
+            )
+        if key == 'data':
+            validate_bytes(
+                field_name=key,
+                value=value,
+            )
+        if key in ('gasprice', 'startgas', 'value'):
+            validate_integer(
+                field_name=key,
+                value=value
```
rwrzesien: @kbeker https://github.com/golemfactory/concent/pull/687#discussion_r207895899
> The same situation as in NonceAbstractMessage we can add here validation for gasprice, startgas and value that it could have maximum 78 numbers

:+1: done

concent_api/concent_golem_messages/message.py
```diff
@@ -0,0 +1,133 @@
...
+    __slots__ = [
+        'v',
+        'r',
+        's',
+    ] + TransactionAbstractMessage.__slots__
+
+    def deserialize_slot(self, key, value):
+        value = super().deserialize_slot(key, value)
+        if key in ('v', 'r', 's'):
+            validate_integer(field_name=key, value=value)
```
rwrzesien: @kbeker https://github.com/golemfactory/concent/pull/687#discussion_r207897446
> Same here as above
:+1:  done

concent_api/concent_golem_messages/tests/factories.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+class TransactionSigningRequestFactory(TransactionAbstractMessage):
+    class Meta:
+        model = TransactionSigningRequest
+
+    # pylint: disable=no-self-argument
+
+    @factory.post_generation
+    def arct_report_computed_task(msg, _create, _extracted, **kwargs):  # pylint: disable=unused-argument
```
rwrzesien: @kbeker https://github.com/golemfactory/concent/pull/687#discussion_r207899210
> Please write full name, or comment what this function does :)

The name was wrong, so thanks for catching this, I have changed it to more proper one and added a comment why it has to be done through method instead of setting this directly on class attributes. You can check what this decorator does directly here http://factoryboy.readthedocs.io/en/latest/reference.html#factory.post_generation. In short, it allows you to run some code on the newly created object.
:+1: done

concent_api/concent_golem_messages/tests/factories.py
```diff
@@ -0,0 +1,69 @@
...
+from concent_golem_messages.message import TransactionSigningRequest
+
+
+class NonceAbstractMessageFactory(MessageFactory):
+    class Meta:
+        model = NonceAbstractMessage
+
+    nonce = factory.LazyFunction(
+        lambda: random.randint(0, denoms.turing)
+    )
```
rwrzesien: @kbeker https://github.com/golemfactory/concent/pull/687#discussion_r207898049
> Please add tests for messages which will have nonce bigger than it is possible and check if it will raise exception

:+1: done, but tests are in `test_message.py`.

concent_api/concent_golem_messages/tests/factories.py
```diff
@@ -0,0 +1,69 @@
...
+    startgas = factory.LazyFunction(
+        lambda: random.randint(0, denoms.turing)
+    )
+    to = factory.fuzzy.FuzzyText(length=20, chars='0123456789abcdef')
+    value = factory.LazyFunction(
+        lambda: random.randint(0, denoms.turing)
+    )
+    data = b''
```
rwrzesien: @kbeker https://github.com/golemfactory/concent/pull/687#discussion_r207898326
> Please add tests which will raise exceptions if gasprice, startgas, to and value will be bigger than our assumption

:+1: done, but tests are in `test_message.py`.

concent_api/concent_golem_messages/validators.py
```diff
@@ -0,0 +1,26 @@
...
+
+    if not isinstance(value, bytes):
+        raise FieldError(
+            'Should be a bytes field',
+            field=field_name,
+            value=value,
+        )
+
+
+def validate_maximum_int_length(field_name: str, value: int, maximum_allowed_length: int):
```
kbeker: Hmm, I just thought now, maybe better way to validate maximum int value (restricted by bytes, like in our example) could be:
```python
maximum_allowed_value = 2 ** (allowed_bytes * 8 ) - 1
if 2 ** (allowed_bytes * 8 ) - 1 < value:
   raise FieldError(
      f'Maximum allowed value is {maximum_allowed_value} but value is {value}',
      field=field_name,
      value=value,
   )
```
because for example if we assume that we can pass maximum 32 bytes = 256 bits it is possible to pass maximum 78 numbers, but also for 259 bits will have 78 numbers, look:
```bash
>>> len(str(2**(259)-1))
78
```
It is just proposal, but I think that might be better validator for this kind values to check :)

rwrzesien: I think that in the future we might want to validate other Concent golem-messages that will not necessary be based on bytes length, so I would skip that. But thanks to your comment I thought about something different; currently we reuse value `78` here and in `core`. How about placing it in `common/constants.py`, making it being calculated like in the code you have posted above, and describing what it describes ? 

What do you guys think ? This should be a separate issue.

cameel: I see that right now it does this:
``` python
if maximum_allowed_length < len(str(value)):
```

I agree with @kbeker. Relying on string representation for comparing numbers is rarely a good idea. Does it even handle negative integers?

BTW, `if len(str(value)) > maximum_allowed_length:` reads much better as a condition. With the current one I have to go over it a few times to parse it mentally because of the weird order.

Issue #714, opened 2018-08-08
Use larger chunk size when downloading files with `requests` to reduce the speed overhead


kbeker:
There should be added setting which will say how big chunk size is. Problem occurred when Concent got bigger file to save in verifier_storage and then render. Default value for `chunk_size` in `response.iter_content()` is `1`. 

From requests package -> ```def iter_content(self, chunk_size=1, decode_unicode=False)```

It should be a modifiable setting. I think that `CONCENT_CHUNK_SIZE` might be good name for this setting. What do you think @cameel ? Also how big `chunk_size` should be? 1024 bytes?

rwrzesien:
If it is a setting, please add checks that:
- it is set only if `verifier` is in `AVAILABLE_CONCENT_FEATURES`.
- if set, it is positive int.

I think the name should be more related to verifier, ie. `VERIFIER_DOWNLOAD_CHUNK_SIZE`.

cameel:
From what I see the size is a trade-off between memory usage and speed/responsiveness. A low value will work with less memory but you will need to process more smaller chunks which will have a bigger overhead. 1 byte at a time is quite extreme. I think we should try higher values. We're not very constrained with memory so high values are fine for us. Verifier needs a lot of memory because of Blender anyway. I think that 1 megabyte should be OK. But please test it with a large file (>= 1 GB) and chunk sizes of 1, 10, 100, 1000, ..., up to 1 gigabyte. Look at transfer speed and memory usage of the process. Post results in a table in a comment below.

Defining a setting for this is fine. But it should have a default. Also, whether it should be a general setting or one for verfier depends on your results. If it turns out that using a lot of memory is better, we'd probably still prefer to use less if we do a download in a different component. But if values over 1 MB don't help much, it would be fine to use this setting in the future or other components too if we ever need to make a download in them.

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/159649243

kbeker:
#### `RAM usage in chunk_size function for file 2,8GB`
chunk_size [B]  | RAM usage [GB] | time [s] |
|------------------------|------------------|-----------------|
| 1                          | ~3 | too long didn't test
| 10                        | ~3 | 96.05483078956604
| 100                      | ~3 | 11.57115912437439
| 1000                    |~3  | 3.6279618740081787
| 10000                  | ~3 | 2.1257567405700684
| 100000                | ~3 | 1.8617238998413086
[...]

PR #722, opened 2018-08-10
Chore add verifier downloading chunk size setting

kbeker:
Resolves https://github.com/golemfactory/concent/issues/714

### DIFF
--- a/concent_api/concent_api/settings/base.py
+++ b/concent_api/concent_api/settings/base.py
@@ -294,6 +294,9 @@ STORAGE_CLUSTER_SSL_CERTIFICATE_PATH = ''
 # rendering results and any intermediate files.
 # VERIFIER_STORAGE_PATH = ''
 
+# A global constant defining chunk_size used for downloading files from nginx storage to verifier storage.
+VERIFIER_DOWNLOAD_CHUNK_SIZE = 1000000
+
 CUSTOM_PROTOCOL_TIMES = False
 
 # The minimum acceptable value of the SSIM metric. Values below this threshold for an image pair will result in a
--- a/concent_api/concent_api/system_check.py
+++ b/concent_api/concent_api/system_check.py
@@ -256,6 +256,30 @@ def create_error_39_storage_server_internal_address_is_not_set():
     )
 
 
+def create_error_40_verifier_download_chunk_size_is_not_defined():
+    return Error(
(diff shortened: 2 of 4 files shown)


concent_api/concent_api/settings/base.py
```diff
@@ -294,6 +294,9 @@
 # rendering results and any intermediate files.
 # VERIFIER_STORAGE_PATH = ''
 
+# A Global constant defining chunk_size used for downloading files from nginx storage to verifier_storage.
+VERIFIER_DOWNLOAD_CHUNK_SIZE = 1000000
```
dybi: since it's defined here, is there any point of defining it in `development` and `testing` (especially when it has the same value)?

concent_api/concent_api/system_check.py
```diff
@@ -256,6 +256,30 @@ def create_error_39_storage_server_internal_address_is_not_set():
...
+        'VERIFIER_DOWNLOAD_CHUNK_SIZE has wrong type',
+        hint=f"VERIFIER_DOWNLOAD_CHUNK_SIZE must be integer instead of {verifier_download_chunk_size_type}.",
+        id='concent.E041',
+    )
+
+
+def create_error_42_verifier_download_chunk_size_has_wrong_value(verifier_download_chunk_size_value):
+    return Error(
+        'VERIFIER_DOWNLOAD_CHUNK_SIZE setting has wrong value',
+        hint=f"VERIFIER_DOWNLOAD_CHUNK_SIZE must be greater or equal 1. Currently it has {verifier_download_chunk_size_value}.",
```
dybi: "be greater or equal **than** 1" or just "positive"

kbeker: Done

concent_api/concent_api/settings/base.py
kbeker: After a quick talk with @cameel he said that if setting is in `base.py` it shouldn't be in `development.py` or `testing.py`

cameel: Yeah. @dybi is right. We add stuff to `development.py` and `testing.py` only when either there's no default in `base.py` but we want one in development/unit tests or when we want a different default.

In this case default is defined and should be usable both in production and in development so there's no need to put it in those other files.

Issue #635, opened 2018-07-12
SCI Callback: Submitting Ethereum Transactions to the Signing Service


cameel:
This is a part of the implementation of the "Concent Signing Service" blueprint (#597). This task assumes that the wire protocol described in #618 is already implemented.

#### Overview
Currently Concent is signing the Etherem transactions on its own by supplying a trivial callback that simply calls `Transaction.sign()`. We want to replace the callback with one that submits the transaction to the Signing Service and does not rely on having access to the private key.

#### Settings
Create the following new settings:
- `MIDDLEMAN_ADDRESS`: string - IP or domain name that can be used to connect to Middleman.
    - No default in production. `localhost` in development.
- `MIDDLEMAN_PORT`: int - The port on which Middleman is accepting requests. This should be the internal port, meant for connections with Concent.
    - Default should be the same as the port Middleman uses if you start it without giving it a port.
- `SIGNING_SERVICE_PUBLIC_KEY`: string - public part of the key used by Signing Service to sign Middleman protocol frames and Golem messages
    - No default in production, in development you can hard-code a key (use the same you used for the Signing Service)

Also, remove the setting that contains the private ethereum key. Update concent-deployment to remove this key from our configuration on the cluster as well.

#### Signing callback
The callback should do the following:
1. Create a `TransactionSigningRequest`.
2. Establish a TCP connection with Middleman.
    - If you can't connect, raise an exception immediately, without retrying.
3. Send the `TransactionSigningRequest`.
4. Read the response.
5. Close the connection.
6. If the response is a valid `SignedTransaction`, copy the signature to the transaction object passed to the callback by SCI and return.
7. Otherwise handle errors by raising an exception and interrupting the callback:
    - If the connection with Middleman times out before a complete response is received.
    - If the received data frame is invalid or the signature does not match its content.
    - If the request ID in the received data frame does not match the one in the data frame that was sent.
    - If the received Golem message is malformed or not signed by the Signing Service.
    - If the received Golem message is anything else than `SignedTransaction`, `TransactionRejected` or and `ERROR` frame.
    - If the received Golem message is `TransactionRejected` or an `ERROR` frame.
    - If the transaction data in `SignedTransaction` does not match the information we sent.
        - In particular if the contract address is different than the one in the request.
    - If the Ethereum signature in `SignedTransaction` is not a valid signature.
    - If the transaction in `SignedTransaction` is not signed with the right Ethereum private key.

[...]

PR #727, opened 2018-08-15
SCI callback

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/635.


concent_api/core/payments/sci_callback.py
```diff
@@ -0,0 +1,149 @@
...
+        raise SCICallbackPayloadError() from exception
+
+    # Request ID in the received data frame does not match the one in the data frame that was sent.
+    if deserialized_message.request_id != transaction.nonce:  # TODO: How to generate request_id ??
+        raise SCICallbackRequestIdError('MiddleMan response request_id does not match requested.')
+
+    # If the received Golem message is TransactionRejected or an ERROR frame.
+    if isinstance(deserialized_message, ErrorFrame) or isinstance(deserialized_message.payload, TransactionRejected):
+        raise SCICallbackPayloadError(
+            'Received frame is either ErrorFrame or contains Golem message TransactionRejected.'
```
dybi: I think we should be more specific here

concent_api/core/payments/sci_callback.py
```diff
@@ -0,0 +1,149 @@
...
+
+    # If the received Golem message is anything else than SignedTransaction, TransactionRejected or and ERROR frame.
+    if not isinstance(deserialized_message.payload, SignedTransaction):
+        raise SCICallbackPayloadError('Received frame payload is not Golem message SignedTransaction instance.')
+
+    signed_transaction = deserialized_message.payload
+
+    # The transaction data in SignedTransaction does not match the information we sent.
+    if not (
+        signed_transaction.nonce == transaction.nonce and
```
dybi: please extract this to a helper function: `transactions_match` or sth similar

concent_api/core/payments/sci_callback.py
```diff
@@ -0,0 +1,149 @@
...
+        )
+
+    # If the response is a valid SignedTransaction, copy the signature to the transaction object passed to the
+    # callback by SCI and return.
+
+    transaction.v = signed_transaction.v
+    transaction.r = signed_transaction.r
+    transaction.s = signed_transaction.s
+
+    return transaction
```
dybi: hm... we are doing an interesting thing here:
1. we modify the object that was given as parameter
2. we return it
Maybe it would be better to make a copy of the `Transaction` object, modify it and return it?

concent_api/core/tests/test_payments_sci_callback.py
```diff
@@ -0,0 +1,192 @@
...
+
+    def test_that_sci_callback_should_raise_exception_on_timeout(self):
+        transaction = self._create_unsigned_transaction()
+
+        with mock.patch('core.payments.sci_callback.socket.socket.connect'):
+            with mock.patch('core.payments.sci_callback.send_over_stream', side_effect=socket.timeout):
+                with self.assertRaises(SCICallbackTimeoutError):
+                    sci_callback(transaction)
+
+    def test_that_sci_callback_should_raise_exception_on_receiving_incorrectly_signed_middleman_protocol_message(self):
```
dybi: Does this exception `MiddlemanProtocolError` always mean that MiddleMan protocol message is incorrectly signed?

concent_api/core/tests/test_payments_sci_callback.py
```diff
@@ -0,0 +1,192 @@
...
+        transaction = self._create_unsigned_transaction()
+
+        with mock.patch('core.payments.sci_callback.socket.socket.connect'):
+            with mock.patch('core.payments.sci_callback.send_over_stream'):
+                with mock.patch('core.payments.sci_callback.unescape_stream'):
+                    with mock.patch('middleman_protocol.message.AbstractFrame.deserialize', side_effect=MiddlemanProtocolError):
+                        with self.assertRaises(SCICallbackFrameError):
+                            sci_callback(transaction)
+
+    def test_that_sci_callback_should_raise_exception_on_receiving_incorrectly_signed_golem_message(self):
```
dybi: Does this exception `MessageError` always mean that Golem message is incorrectly signed?

concent_api/core/tests/test_payments_sci_callback.py
```diff
@@ -0,0 +1,192 @@
...
+            with mock.patch('core.payments.sci_callback.send_over_stream'):
+                with mock.patch('core.payments.sci_callback.unescape_stream'):
+                    with mock.patch(
+                        'middleman_protocol.message.AbstractFrame.deserialize',
+                        return_value=GolemMessageFrame(payload=self._create_signed_transaction(), request_id=100),
+                    ):
+                        with self.assertRaises(SCICallbackRequestIdError):
+                            sci_callback(transaction)
+
+    def test_that_sci_callback_should_raise_exception_when_response_is_transaction_rejected_or_error_frame(self):
```
dybi: In this test `ErrorFrame` case is being checked (please change the name)

concent_api/core/tests/test_payments_sci_callback.py
```diff
@@ -0,0 +1,192 @@
...
+            with mock.patch('core.payments.sci_callback.send_over_stream'):
+                with mock.patch('core.payments.sci_callback.unescape_stream'):
+                    with mock.patch(
+                        'middleman_protocol.message.AbstractFrame.deserialize',
+                        return_value=ErrorFrame(payload=(ErrorCode.UnexpectedMessage, 'error'), request_id=99),
+                    ):
+                        with self.assertRaises(SCICallbackPayloadError):
+                            sci_callback(transaction)
+
+    def test_that_sci_callback_should_raise_exception_when_response_is_not_signed_transaction_golem_message(self):
```
dybi: hm... "is_message_of_wrong_type"?

concent_api/core/payments/sci_callback.py
```diff
@@ -0,0 +1,149 @@
...
+from middleman_protocol.concent_golem_messages.message import TransactionRejected
+from middleman_protocol.concent_golem_messages.message import TransactionSigningRequest
+from middleman_protocol.exceptions import MiddlemanProtocolError
+from middleman_protocol.message import ErrorFrame
+from middleman_protocol.message import GolemMessageFrame
+from middleman_protocol.stream import send_over_stream
+from middleman_protocol.stream import unescape_stream
+
+
+def sci_callback(transaction: Transaction) -> Transaction:
```
dybi: this function is pretty big... Maybe we could have it split into smaller self-describing minifunctions (like in `blender_verification_order` case)?

concent_api/concent_api/tests/test_check_middleman_port.py
```diff
@@ -0,0 +1,69 @@
...
+        del settings.MIDDLEMAN_PORT
+
+        errors = check_middleman_port()
+
+        self.assertEqual(len(errors), 1)
+        self.assertEqual(errors[0], create_error_45_middleman_port_is_not_set())
+
+    @override_settings(
+        CONCENT_FEATURES=[]
+    )
```
kbeker: I'm wondering which of this styles would be better:
```python
@override_settings(CONCENT_FEATURES=[])
```
or
```python
@override_settings(
   CONCENT_FEATURES=[]
)
```
Because I think that breaking one line statement to another line is not necessary.

concent_api/concent_api/system_check.py
```diff
@@ -539,3 +579,29 @@ def check_verifier_download_chunk_size(app_configs=None, **kwargs):  # pylint: d
...
+
+@register()
+def check_middleman_port(app_configs=None, **kwargs):  # pylint: disable=unused-argument
+    if not hasattr(settings, 'MIDDLEMAN_PORT') and 'middleman' in settings.CONCENT_FEATURES:
+        return [create_error_45_middleman_port_is_not_set()]
+
+    if hasattr(settings, 'MIDDLEMAN_PORT'):
+        if not isinstance(settings.MIDDLEMAN_PORT, int):
+            return [create_error_46_middleman_port_has_wrong_type(settings.MIDDLEMAN_PORT)]
+        if not 0 < settings.MIDDLEMAN_PORT < 65535:
```
kbeker: We do not allow to use port number 0 or 65535, right?

concent_api/concent_api/settings/base.py
```diff
@@ -331,3 +331,12 @@
 
 # Verifier setting defining number of threads used by Blender
 BLENDER_THREADS = 1
+
+# Defines IP or domain name that can be used to connect to Middleman.
+# MIDDLEMAN_ADDRESS = ''
+
+# The port on which Middleman is accepting requests.
+# MIDDLEMAN_PORT =
```
kbeker: These two settings should be added to `base.py` in previous commit

concent_api/concent_api/tests/test_check_signing_service_public_key.py
```diff
@@ -0,0 +1,60 @@
...
+        del settings.SIGNING_SERVICE_PUBLIC_KEY
+
+        errors = check_signing_service_public_key()
+
+        self.assertEqual(len(errors), 1)
+        self.assertEqual(errors[0], create_error_48_signing_service_public_key_is_not_set())
+
+    @override_settings(
+        CONCENT_FEATURES=[]
+    )
```
kbeker: Same situation as previous. Which style do we want to use, because breaking one line statement might be overkilling

concent_api/core/tests/test_payments_sci_callback.py
```diff
@@ -0,0 +1,192 @@
...
+        return SignedTransaction(
+            nonce=99,
+            gasprice=10 ** 6,
+            startgas=80000,
+            value=10,
+            to=b'7917bc33eea648809c28',
+            v=self.v,
+            r=self.r,
+            s=self.s,
+            data=b'3078666333323333396130303030303030303030303030303030303030303030303032393662363963383738613734393865663463343531363436353231336564663834666334623330303030303030303030303030303030303030303030303030333431336437363161356332633362656130373531373064333239363566636161386261303533633030303030303030303030303030303030303030303030303030303030303030303030303030303030303030303030303030303030303030303030303237313030303030303030303030303030303030303030303030303030303030303030303030303030303030303030303030303030303030303030303562343336643666',
```
kbeker: If this is not some specific byte string for this transaction I think it might be better to generate this from `os` lib by function `urandom`

concent_api/core/tests/test_payments_sci_callback.py
```diff
@@ -0,0 +1,192 @@
...
+        )
+
+    def _create_unsigned_transaction(self):  # pylint: disable=no-self-use
+        return Transaction(
+            nonce=99,
+            gasprice=10 ** 6,
+            startgas=80000,
+            value=10,
+            to=b'7917bc33eea648809c28',
+            data=b'3078666333323333396130303030303030303030303030303030303030303030303032393662363963383738613734393865663463343531363436353231336564663834666334623330303030303030303030303030303030303030303030303030333431336437363161356332633362656130373531373064333239363566636161386261303533633030303030303030303030303030303030303030303030303030303030303030303030303030303030303030303030303030303030303030303030303237313030303030303030303030303030303030303030303030303030303030303030303030303030303030303030303030303030303030303030303562343336643666',
```
kbeker: Same here

concent_api/concent_api/tests/test_check_middleman_port.py
```diff
@@ -0,0 +1,69 @@
...
+        self.assertEqual(len(errors), 1)
+        self.assertEqual(errors[0], create_error_47_middleman_port_has_wrong_value(settings.MIDDLEMAN_PORT))
+
+    @override_settings(
+        CONCENT_FEATURES=[
+            'middleman'
+        ],
+        MIDDLEMAN_PORT=111
+    )
+    def test_that_middleman_port_as_string_will_not_produce_error(self):
```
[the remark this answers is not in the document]
rwrzesien: :+1:

concent_api/core/payments/sci_callback.py
```diff
@@ -0,0 +1,149 @@
...
+    transaction_signing_request = TransactionSigningRequest(
+        nonce    = transaction.nonce,
+        gasprice = transaction.gasprice,
+        startgas = transaction.startgas,
+        to       = transaction.to,
+        value    = transaction.value,
+        data     = transaction.data,
+    )
+
+    setattr(transaction_signing_request, 'from', settings.CONCENT_ETHEREUM_ADDRESS)
```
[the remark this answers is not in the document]
rwrzesien: `from` can't be used as it is a reserved keyword, unless you know better way ? :)

concent_api/core/payments/sci_callback.py
```diff
@@ -0,0 +1,149 @@
...
+        # Send the TransactionSigningRequest.
+        try:
+            send_over_stream(
+                connection=client_socket,
+                raw_message=middleman_message,
+                private_key=settings.CONCENT_PRIVATE_KEY,
+            )
+
+            # Read the response and close the connection.
+            raw_response = unescape_stream(connection=client_socket).__next__()
```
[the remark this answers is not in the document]
rwrzesien: :+1:

concent_api/core/payments/sci_callback.py
rwrzesien: I have split this into two conditions and added separate tests. :+1:

concent_api/core/payments/sci_callback.py
rwrzesien: :+1:

concent_api/core/payments/sci_callback.py
rwrzesien: What would be benefit of it ?

concent_api/core/tests/test_payments_sci_callback.py
rwrzesien: No, I have updated test name. :+1:

concent_api/core/tests/test_payments_sci_callback.py
rwrzesien: Same here, I have updated test name. :+1:

concent_api/core/tests/test_payments_sci_callback.py
rwrzesien: Updated. :+1:

concent_api/core/tests/test_payments_sci_callback.py
rwrzesien: Actually similar, but mine points out what message type is exactly expected.

concent_api/concent_api/system_check.py
rwrzesien: Right.

concent_api/concent_api/tests/test_check_middleman_port.py
rwrzesien: I always do this for readability. I like when code is not extended to much to the right, and I can read it from top to bottom, but I understand someone can have different feelings about it. Actually this specific breaking for `@override_settings` is already used in other places do I though it would be ok. But I think if we want to do make it clear it should be explicitly described in our coding style.

concent_api/concent_api/tests/test_check_signing_service_public_key.py
rwrzesien: As I have said, I think it is good for readability. It doesn't have any impact on performance or size of result code in `pyc` files.

concent_api/concent_api/tests/test_check_middleman_port.py
kbeker: I started to think about this when @cameel said that we sometimes breaks lines too much. That's why I started this discuss :)

concent_api/concent_api/settings/base.py
rwrzesien: :+1: done

concent_api/concent_api/tests/test_check_middleman_port.py
cameel: Both versions are fine and allowed in our style.

@rwrzesien's preferred version has the advantage that you don't have to rewrap it when the line gets longer. But it makes the code less compact.

I personally prefer `@override_settings(CONCENT_FEATURES=[])` when the line fits below the suggested length limit (120 characters) but it's not a rule, just my preference.

rwrzesien: > But it makes the code less compact.

I think this actually increases readability.

concent_api/concent_api/tests/test_check_middleman_port.py
```diff
@@ -0,0 +1,69 @@
...
+        self.assertEqual(len(errors), 1)
+        self.assertEqual(errors[0], create_error_46_middleman_port_has_wrong_type(settings.MIDDLEMAN_PORT))
+
+    @override_settings(
+        CONCENT_FEATURES=[
+            'middleman'
+        ],
+        MIDDLEMAN_PORT=100000
+    )
+    def test_that_out_of_range_middleman_port_setting_type_should_produce_error(self):
```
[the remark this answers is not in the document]
rwrzesien: But I think it would still require `django.test.TestCase` for `override_settings` to work correctly. So I think the only way to do this is to write more regular tests. Would that be ok ?

concent_api/core/payments/sci_callback.py
rwrzesien: It has exactly 100 lines when you cut of beginning docstring, and there are no parts of it that are repeated. In `blender_verification_order` there was some part of repeated code, and mixed some low-level code with high-level code. I think this is not the case here, but if you insist I will do it. Maybe lets discuss tomorrow.

rwrzesien: @dybi Lets discuss tomorrow.

concent_api/concent_api/tests/test_check_middleman_port.py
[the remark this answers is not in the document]
dybi: you don't have to use `django.test.TestCase` to be able to use `override_settings`. `override_settings` can be used with `pytest` as well, but only in form of context manager.

concent_api/core/payments/sci_callback.py
dybi: for me 100 lines is a lot ;) if I cannot convince you, maybe Uncle Bob will? ;) 
https://vimeo.com/12643301

concent_api/concent_api/tests/test_check_middleman_port.py
cameel: Not always. When the code is complex, it does. But when it's simple it may actually make it look more complex than it really is.

This is pretty compact:
``` python
@override_settings(CONCENT_FEATURES=[])
@override_settings(BLENDER_THREADS=5)
@override_settings(VERIFIER_DOWNLOAD_CHUNK_SIZE=1000)
@override_settings(VERIFIER_STORAGE_PATH='/var/verifier/')
@override_settings(STATICFILES_DIRS=['/a/b/c'])
@override_settings(AUTH_PASSWORD_VALIDATORS=[{'NAME': 'django.contrib.auth.password_validation.UserAttributeSimilarityValidator'}])
@override_settings(CUSTOM_PROTOCOL_TIMES=True)
def test(self):
    pass
```

and for me adding more whitespace does not make it more readable:
``` python
@override_settings(
    CONCENT_FEATURES=[]
)
@override_settings(
    BLENDER_THREADS=5
)
@override_settings(
    VERIFIER_DOWNLOAD_CHUNK_SIZE=1000
)
@override_settings(
    VERIFIER_STORAGE_PATH='/var/verifier/'
)
@override_settings(
    STATICFILES_DIRS=[
        '/a/b/c'
    ]
)
@override_settings(
    AUTH_PASSWORD_VALIDATORS=[
        {
            'NAME': 'django.contrib.auth.password_validation.UserAttributeSimilarityValidator'
        }
    ]
)
@override_settings(
    CUSTOM_PROTOCOL_TIMES=True
)
def test(self):
    pass
```

concent_api/concent_api/tests/test_check_middleman_port.py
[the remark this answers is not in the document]
rwrzesien: :+1: done

concent_api/core/payments/sci_callback.py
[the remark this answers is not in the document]
rwrzesien: It was done this way already in Golem messages in Middleman protocol and in some tests in SigningService and accepted, so if we want to change it now I suggest creating separate issue.

concent_api/core/payments/sci_callback.py
rwrzesien: But we don't know if that is not expected either ;) Ok, done :+1:

concent_api/core/payments/sci_callback.py
rwrzesien: :+1: done

concent_api/concent_api/tests/test_check_middleman_port.py
```diff
@@ -0,0 +1,81 @@
...
+
+    @pytest.mark.parametrize(('middleman_port_value', 'expect_error'), [
+        (-1, True),
+        (0, True),
+        (1, False),
+        (65534, False),
+        (65535, True),
+        (65536, True),
+    ])
+    def test_that_checking_different_middleman_port_setting_should_produce_error_or_not_depending_on_the_value(
```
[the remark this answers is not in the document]
rwrzesien: :+1: done

kbeker:
Please wait with merge until I will test it locally

concent_api/common/singleton.py
```diff
@@ -26,6 +26,9 @@ def __new__(cls, *args: Any, **kwargs: Any) -> SmartContractsInterface:  # pylin
                 ),
                 chain=chains.RINKEBY,
                 storage=DatabaseTransactionsStorage(),
-                tx_sign=sci_callback,
+                tx_sign=(
+                    sci_callback if settings.USE_SIGNING_SERVICE else
+                    lambda tx: tx.sign(settings.CONCENT_ETHEREUM_PRIVATE_KEY),
```
kbeker: With this comma it doesn't work. Please delete it

concent_api/core/payments/sci_callback.py
```diff
@@ -0,0 +1,260 @@
...
+
+    _request_id = 0
+
+    @classmethod
+    def generate_request_id(cls) -> int:
+        cls._request_id += 1
+        return cls._request_id
+
+
+def sci_callback(transaction: Transaction) -> Transaction:
```
dybi: For me comments in this functions are a bit redundant:
1) functions used within are well-named and pretty self-describing
2) they have docstring which explain their behavior.

concent_api/core/payments/sci_callback.py
```diff
@@ -0,0 +1,260 @@
...
+    signed_transaction = deserialize_response_and_handle_errors(raw_response, request_id)
+
+    assert isinstance(signed_transaction, SignedTransaction)
+
+    # Verify received SignedTransaction signature and handle related errors.
+    verify_data_and_signature(signed_transaction, transaction)
+
+    # If the response is a valid SignedTransaction, copy Ethereum Transaction object into new one and
+    # add SignedTransaction data.
+    return copy_transaction_and_signature(signed_transaction, transaction)
```
dybi: Okay, this one is a bit misleading - I would suggest changing this function, so it doesn't return anything and would rename it to sth more like:
`sign_ethereum_transaction` or `copy_transaction_signature`, so it would look like:
```python
...
sign_transaction(signed_transaction, ethereum_transaction)
return ethereum_transaction
```

concent_api/core/payments/sci_callback.py
```diff
@@ -0,0 +1,260 @@
...
+
+
+def create_transaction_signing_request(transaction: Transaction) -> TransactionSigningRequest:
+    """ Create TransactionSigningRequest message from given Ethereum Transaction object. """
+
+    assert isinstance(transaction, Transaction)
+
+    transaction_signing_request = TransactionSigningRequest(
```
dybi: can we please change `from` to `from_`, so it could be used in constructor (via named params)? Right know we are having an `TransactionSigningRequest`object constructed with two steps: call to "constructor" and call to `setattr`.

concent_api/core/payments/sci_callback.py
```diff
@@ -0,0 +1,260 @@
...
+def send_request_to_middleman(middleman_message: GolemMessageFrame) -> bytes:
+    """
+    Opens socket connection to Middleman, sends Frame to MiddleMan through MiddleMan Protocol and receive response.
+
+    Returns raw Frame as bytes.
+    """
+
+    assert isinstance(middleman_message, GolemMessageFrame)
+
+    # Establish a TCP connection with Middleman.
```
dybi: this comment seems unnecessary - the code is clear about itself ;)

concent_api/core/payments/sci_callback.py
```diff
@@ -0,0 +1,260 @@
...
+    with closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as client_socket:
+        client_socket.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, True)
+        client_socket.connect((settings.MIDDLEMAN_ADDRESS, settings.MIDDLEMAN_PORT))
+        client_socket.settimeout(SCI_CALLBACK_MAXIMUM_TIMEOUT)
+
+        # Send the TransactionSigningRequest.
```
dybi: same here

concent_api/core/payments/sci_callback.py
```diff
@@ -0,0 +1,260 @@
...
+            public_key=settings.CONCENT_PUBLIC_KEY,
+        )
+    # Received data frame is invalid or the signature does not match its content.
+    except MiddlemanProtocolError as exception:
+        raise SCICallbackFrameError() from exception
+    # If the received Golem message is malformed.
+    except MessageError as exception:
+        raise SCICallbackPayloadError() from exception
+
+    # Request ID in the received data frame does not match the one in the data frame that was sent.
```
dybi: this comment seems unnecessary - the code is clear about itself ;)

concent_api/core/payments/sci_callback.py
```diff
@@ -0,0 +1,260 @@
...
+    if deserialized_message.request_id != request_id:
+        raise SCICallbackRequestIdError('MiddleMan response request_id does not match requested.')
+
+    # If the received frame is ErrorFrame.
```
dybi: this comment seems unnecessary - the code is clear about itself ;)

concent_api/core/payments/sci_callback.py
```diff
@@ -0,0 +1,260 @@
...
+    if isinstance(deserialized_message, ErrorFrame):
+        raise SCICallbackPayloadError(
+            'Received frame is ErrorFrame.'
+        )
+
+    # If the received Golem message is TransactionRejected.
```
dybi: this comment seems unnecessary - the code is clear about itself ;)

concent_api/core/payments/sci_callback.py
```diff
@@ -0,0 +1,260 @@
...
+    if isinstance(deserialized_message.payload, TransactionRejected):
+        raise SCICallbackPayloadError(
+            'Received frame is contains Golem message TransactionRejected.'
+        )
+
+    # If the received Golem message is anything else than SignedTransaction, TransactionRejected or and ERROR frame.
```
dybi: this comment seems unnecessary - the code is clear about itself ;)

concent_api/core/payments/sci_callback.py
```diff
@@ -0,0 +1,260 @@
...
+    if not isinstance(deserialized_message.payload, SignedTransaction):
+        raise SCICallbackPayloadError('Received frame payload is not Golem message SignedTransaction instance.')
+
+    # If the received Golem message is not signed by the Signing Service.
```
dybi: this comment seems unnecessary - the code is clear about itself ;)

concent_api/core/payments/sci_callback.py
```diff
@@ -0,0 +1,260 @@
...
+    Verify received SignedTransaction data and signature and handle related errors.
+
+    1. Check if received SignedTransaction data match original Ethereum Transacion.
+    2. Check if received signature is created for given message with given keys pair.
+    """
+
+    assert isinstance(signed_transaction, SignedTransaction)
+    assert isinstance(transaction, Transaction)
+
+    # The transaction data in SignedTransaction does not match the information we sent.
```
dybi: this comment seems unnecessary - the code is clear about itself ;)

concent_api/core/payments/sci_callback.py
```diff
@@ -0,0 +1,260 @@
...
+            signed_transaction.s,
+        ),
+        settings.CONCENT_ETHEREUM_PUBLIC_KEY,
+    ):
+        raise SCICallbackTransactionSignatureError(
+            'Received SignedTransaction signature data is not signed by right Ethereum private key.'
+        )
+
+
+def is_signed_transaction_data_equal_to_transaction_data(
```
dybi: :+1: for crystal clear intentions revealing ;)

concent_api/core/payments/sci_callback.py
```diff
@@ -0,0 +1,260 @@
...
+        signed_transaction.nonce == transaction.nonce and
+        signed_transaction.gasprice == transaction.gasprice and
+        signed_transaction.startgas == transaction.startgas and
+        signed_transaction.to == transaction.to and
+        signed_transaction.value == transaction.value and
+        signed_transaction.data == transaction.data
+    )
+
+
+def copy_transaction_and_signature(signed_transaction: SignedTransaction, transaction: Transaction) -> Transaction:
```
dybi: as stated above (where the function is called): please rename it

concent_api/common/helpers.py
```diff
@@ -177,8 +177,15 @@ def upload_file_to_storage_cluster(
     )
 
 
+def generate_ethereum_address_from_ethereum_public_key_bytes(ethereum_public_key: str) -> bytes:
```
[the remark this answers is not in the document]
rwrzesien: :+1:

concent_api/concent_api/system_check.py
```diff
@@ -613,3 +677,47 @@ def check_concent_ethereum_public_key(app_configs: None=None, **kwargs: Any) ->
...
 
     return []
+
+
+@register()
+def check_middleman_address(app_configs: None=None, **kwargs: Any) -> list:  # pylint: disable=unused-argument
+    if not hasattr(settings, 'MIDDLEMAN_ADDRESS') and 'middleman' in settings.CONCENT_FEATURES:
+        return [create_error_49_middleman_address_is_not_set()]
+
+    if hasattr(settings, 'MIDDLEMAN_ADDRESS'):
```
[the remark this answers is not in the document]
rwrzesien: I think I have discussed it recently with @cameel and we did this only for few settings for verification. Generally that should not cause any problems, and there is a lot of other settings which work this way. But if you insist I can change it - please just propose the error message.
@cameel Can you confirm if we should be strict about it?

concent_api/concent_api/system_check.py
```diff
@@ -326,6 +326,70 @@ def create_error_47_concent_ethereum_public_key_has_wrong_length(concent_ethereu
     )
 
 
+def create_error_48_middleman_address_has_wrong_type(value: Any) -> Error:
+    return Error(
+        f"Setting MIDDLEMAN_ADDRESS has incorrect type `{type(value)}` instead of `str`.",
+        hint=f"Set setting MIDDLEMAN_ADDRESS to be string type.",
```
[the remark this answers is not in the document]

concent_api/concent_api/system_check.py
```diff
@@ -326,6 +326,70 @@ def create_error_47_concent_ethereum_public_key_has_wrong_length(concent_ethereu
...
+        id="concent.E048",
+    )
+
+
+def create_error_49_middleman_address_is_not_set() -> Error:
+    return Error(
+        'MIDDLEMAN_ADDRESS setting is not defined.',
+        hint='Set MIDDLEMAN_ADDRESS to host on which MiddleMan instance is working.',
```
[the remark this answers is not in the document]

concent_api/concent_api/system_check.py
```diff
@@ -326,6 +326,70 @@ def create_error_47_concent_ethereum_public_key_has_wrong_length(concent_ethereu
...
+        'MIDDLEMAN_PORT setting is not defined',
+        hint='Set MIDDLEMAN_PORT to port on which MiddleMan instance is accepting connections.',
+        id='concent.E050',
+    )
+
+
+def create_error_51_middleman_port_has_wrong_type(value: Any) -> Error:
+    return Error(
+        f"Setting MIDDLEMAN_PORT has incorrect type `{type(value)}` instead of `int`.",
+        hint=f"Set setting MIDDLEMAN_PORT to be integer type.",
```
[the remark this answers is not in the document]

concent_api/concent_api/system_check.py
```diff
@@ -326,6 +326,70 @@ def create_error_47_concent_ethereum_public_key_has_wrong_length(concent_ethereu
...
+        id="concent.E051",
+    )
+
+
+def create_error_52_middleman_port_has_wrong_value(value: int) -> Error:
+    return Error(
+        f"Setting MIDDLEMAN_PORT has incorrect value `{value}`.",
+        hint=f"Set setting MIDDLEMAN_PORT to be integer value between 0-65535.",
```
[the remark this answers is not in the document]
rwrzesien: :+1: So `between` should make sense :)

concent_api/concent_api/system_check.py
```diff
@@ -326,6 +326,70 @@ def create_error_47_concent_ethereum_public_key_has_wrong_length(concent_ethereu
...
+        id="concent.E052",
+    )
+
+
+def create_error_53_use_signing_service_not_set() -> Error:
+    return Error(
+        'USE_SIGNING_SERVICE setting is not defined',
+        hint='Set USE_SIGNING_SERVICE to True if you want to use SigningService, otherwise False.',
```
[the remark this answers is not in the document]

concent_api/concent_api/system_check.py
```diff
@@ -326,6 +326,70 @@ def create_error_47_concent_ethereum_public_key_has_wrong_length(concent_ethereu
...
+        id='concent.E053',
+    )
+
+
+def create_error_54_use_signing_service_has_wrong_type(value: Any) -> Error:
+    return Error(
+        f"Setting USE_SIGNING_SERVICE has incorrect type `{type(value)}` instead of `bool`.",
+        hint=f"Set setting USE_SIGNING_SERVICE to be bool type.",
```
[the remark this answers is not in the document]

concent_api/concent_api/system_check.py
```diff
@@ -613,3 +677,47 @@ def check_concent_ethereum_public_key(app_configs: None=None, **kwargs: Any) ->
...
+
+    if hasattr(settings, 'MIDDLEMAN_ADDRESS'):
+        if not isinstance(settings.MIDDLEMAN_ADDRESS, str):
+            return [create_error_48_middleman_address_has_wrong_type(settings.MIDDLEMAN_ADDRESS)]
+
+    return []
+
+
+@register()
+def check_middleman_port(app_configs: None=None, **kwargs: Any) -> list:  # pylint: disable=unused-argument
```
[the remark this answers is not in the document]

concent_api/concent_api/tests/test_check_middleman_port.py
```diff
@@ -0,0 +1,87 @@
...
+    def test_that_non_int_middleman_port_rate_setting_type_should_produce_error(self):
+        errors = check_middleman_port()
+
+        self.assertEqual(len(errors), 1)
+        self.assertEqual(errors[0], create_error_51_middleman_port_has_wrong_type(settings.MIDDLEMAN_PORT))
+
+
+class TestMiddlemanPortCheck:
+
+    @pytest.mark.parametrize(('middleman_port_value',), [
```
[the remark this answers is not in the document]
rwrzesien: I thought it has to be done this way :) Thanks for clarification :+1:

concent_api/concent_api/tests/test_check_siginig_service_key_availability_for_middleman.py
```diff
@@ -17,6 +17,7 @@ def test_that_if_middleman_is_disabled_signing_service_public_key_does_not_have_
             errors = check_signing_service_key_availability_for_middleman()
             assert_that(errors).is_empty()
 
+    @override_settings()
     def test_that_if_middleman_is_enabled_and_key_signing_service_public_key_is_not_defined_error_is_returned(self):
```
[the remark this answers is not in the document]
rwrzesien: :+1:

concent_api/core/payments/sci_callback.py
rwrzesien: I will remove the `flow` part as it may get outdated easily.

concent_api/core/payments/sci_callback.py
rwrzesien: Hm but it actually doesn't sign, just copies the signature from one object to another. But I agree that it might not return the value, so I will change that and remove `_and` from name (as it copies only signature now).

concent_api/core/payments/sci_callback.py
rwrzesien: :+1:

concent_api/core/payments/sci_callback.py
rwrzesien: I would keep this one, as you have to scroll up a lot to figure out this is `TransactionSigningRequest`.

concent_api/core/payments/sci_callback.py
rwrzesien: :+1:

concent_api/core/payments/sci_callback.py
rwrzesien: :+1:

concent_api/core/payments/sci_callback.py
rwrzesien: :+1:

concent_api/core/payments/sci_callback.py
rwrzesien: :+1:

concent_api/core/payments/sci_callback.py
rwrzesien: I would keep this one to indicate that `.payload` is Golem message here.

concent_api/core/payments/sci_callback.py
rwrzesien: :+1:

concent_api/core/payments/sci_callback.py
rwrzesien: :+1:

concent_api/core/payments/sci_callback.py
rwrzesien: How about changing the name of it to `source` or `sender`? @cameel what do you think?

concent_api/concent_api/system_check.py
```diff
@@ -353,31 +353,31 @@ def create_error_50_middleman_port_is_not_set() -> Error:
...
+        hint=f"Set setting MIDDLEMAN_PORT to be an integer.",
         id="concent.E051",
     )
 
 
 def create_error_52_middleman_port_has_wrong_value(value: int) -> Error:
     return Error(
         f"Setting MIDDLEMAN_PORT has incorrect value `{value}`.",
-        hint=f"Set setting MIDDLEMAN_PORT to be integer value between 0-65535.",
+        hint=f"Set setting MIDDLEMAN_PORT to be integer value between 0 and 65535.",
```
dybi: Sorry, but it really troubles me: could you add `"0 and 65535 (exclusive)."`? https://www.quora.com/English-Grammar-Does-between-two-numbers-a-and-b-include-a-and-b-themselves

concent_api/concent_api/system_check.py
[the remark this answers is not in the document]
cameel: The thing we talked about was whether to require the setting to be undefined (or set to an empty value in some cases) when the feature that uses it is not enabled. We have many containers running with slighly different `local_settings.py` in deployment and this would allow us to spot a mistake such as putting the value in the wrong configuration file. Initially my opinion was that this should be enforced and I was including such a check when defining issues. But then we have discovered a case where it's really inconvenient - when you want to have a global default in `base.py`. Getting an error simply because you did not remove a setting from `base.py` with `del` in your `local_settings.py` seems like a very unfriendly and possibly confusing thing to do. You could get that error even on a clean config with no extra settings added.

So eventually I stopped recommending it. A setting from a feature can be empty when the feature is disabled but does not have to.

The case here is slightly different. You're asking if the setting should be validated if you do decide to define it. My opinion is that it should. There's no reason to have an invalid value in it. If you can't provide a valid value when the feature is disabled then just don't set it.

concent_api/core/payments/sci_callback.py
cameel: Sure. I would personally go with `from_address`.

Please drop me a comment about it in #632 so that I remember about it when I'll be updating the docs.

concent_api/concent_api/system_check.py
rwrzesien: Sure :+1: done

concent_api/core/payments/sci_callback.py
rwrzesien: :+1: done

PR #732, opened 2018-08-17
Add shell script to install all requirements

### LINKED ISSUE #731: Create shell script to install requirements from Signing Service, MiddleMan Protocol and Concent_api
Currently to run all tests we need to install all requirements separately in Singing Service, MiddleMan Protocol and in Concent_api. We should have one script which would do this for us. 
Script should be placed in directory where is `full_check.sh`.
It should install:
- requirements from concent_api
- requirements from singing_service
- requirements from middleman_protocol

kbeker:
Resolves #731

### DIFF
--- a/install-requirements.sh
+++ b/install-requirements.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+printf "=================== INSTALL CONCENT_API REQUIREMENTS ====================\n"
+pip install -r concent_api/requirements.lock
+
+printf "=================== INSTALL MIDDLEMAN_PROTOCOL REQUIREMENTS ====================\n"
+cd middleman_protocol
+python setup.py develop
+cd ..
+
+printf "=================== INSTALL SIGNING_SERVICE REQUIREMENTS ====================\n"
+cd signing_service
+python setup.py develop
+cd ..
+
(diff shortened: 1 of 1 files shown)


install-requirements.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+printf "=================== INSTALL CONCENT_API REQUIREMENTS ====================\n"
+pip install -r concent_api/requirements.lock
+
```
pawelkisielewicz: don't we need `requirements-development.txt` as well. Even if they are not updated so often, I'm afraid, in e.g. two months we will remember only that we have one script with **all** requirements and we will not update `requirements-development.txt`.. I think it should be added, even if `requirements-development.txt` will be unchanged nothing bad will happen.

kbeker: Added. Done

kbeker: Added. Done

install-requirements.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
kbeker: Done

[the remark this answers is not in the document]
kbeker: Done

Issue #716, opened 2018-08-08
 Validator for scene_file field in Compute Task Def must be more restrictive


kbeker:
For this moment we validate only that if scene_file ends with `.blend`. This is much too weak validation. 
At first we should agreed on how we want to pass `scene_file` in ComputeTaskDef. 
- Should there only be relative path to file which is in archive?
- Should it be somehow related with `working_directory` which is also in ComputeTaskDef?

@cameel @rwrzesien @dybi Please add your ideas to do stronger validation for this.

@shadeofblue @jiivan please tell us how Golem pass this field to Concent

rwrzesien:
Isn't it that placing correct `.blend` file is in the interest of provider ?

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/159649324

cameel:
Yeah, it is. If he places it in the wrong place and we don't find it, it's his fault. But this issue is about how the location is defined in the first place. Right now we assume that anything goes and this is too permissive. For example what happens if the path is `/etc/passwd`? In general absolute paths should probably be forbidden. But if Golem is allowing them then we may have to allow them in some cases, and convert them to relative paths internally. E.g. if the path starts with `/home/`.

In any case, `TaskToCompute` that does not meet the requirements should be rejected by validations. We should not even bother starting verification if we can immediately see that something's wrong with the message.

shadeofblue:
@cameel @kbeker @Jakub89 re: the path that the `scene_file` points to relates to a directory within the Blender's docker environment

shadeofblue:
though, there's also the `path_root` which is the directory in which the original task resides -> though I'm unsure if it's needed from Provider's POV

PaweuB:
@shadeofblue Please provide some examples with that kind of path.

shadeofblue:
```'scene_file': '/golem/resources/wlochaty3.blend'```

cameel:
Thanks for the info. So, to summarize the current situation and the solution we want:
- `TaskToCompute` contains an absolute path like `/golem/resources/wlochaty3.blend`.
- The archive contains a file called `wlochaty3.blend` in its root directory.
- Therefore `/golem/resources/` is a constant part that must be the same in every `TaskToCompute`.
    - We should validate that it's present.
    - We should strip if from the path and internally only pass around the relative part (`wlochaty3.blend`).
[...]

Issue #675, opened 2018-07-20
Use Case 4  Verification Timeout


PaweuB:
Issue provided by @shadeofblue as described below in `SubtaskResultVerify`.
Please specify more information about condition that caused this problem. Does it fail in particular scenarios or every time?

```
SubtaskResultsVerify(
    header=MessageHeader(
        type_=4006,
        timestamp=1531851528
        encrypted=False),
    sig=None,
    slots=[
        ['subtask_results_rejected',SubtaskResultsRejected(
            header=MessageHeader(
            	type_=2011,
                timestamp=1531851528,
                encrypted=False
            ),
            sig=b'\xfa\x15\xdbkc7,6`7\x87\x01\xb0\xccy(\x10\xf7\x8e\xfd\xa6\xeb\xe6V\x04\xd2\x10\x92\x1a\x03%tl3\x83\xb7\xd3\x1d\xaec\x08\xc3\xcc\ti\x1e\x919nV\x96\xb5/\xa1\xa1\xbcn\x1e$1\x13\n5\x94\x00',
            slots=[[
                'report_computed_task',
                ReportComputedTask(
                header=MessageHeader(
                    type_=2004,
                    timestamp=1531851528,
                    encrypted=False
                ),
                sig=b'\x99z\xcc\xd5ZA\xb7O\xd5n\x15\x06\xd6\xa1\rS\xe7\x0e\x8b\xd0\x14\r\xf2\x18\xccIq\x07h7\xec=gq\x92^\xc1\xaf]\x82H\xf39\xf1\xa5\x05\xc0\xd1U\x9c\xda\xf9\x10\xae\x93\xa6\xb4\xa5\x8b\x99\xf4\x82\x8f\x00\x01',
                slots=[
                    ['result_type', 0],
                    ['node_name', 'Sheri Harris'],
                    ['address', '30.70.240.123'],
                    ['node_info', None],
                    ['port', 3099],
                    ['key_id', b'\xdcaxg\x83\xf7R\x0e3\x0c \xbb\x7f-Ch_\xf4\xa8K\x84{\xb4F\xc0uG\x86\x17\xbf\xb3\x1a}\xb1\x86Z\xef\xe7<O\x83\x7f\xee9\xd4\xf0\xe9r\xcc\x17\xa5{Y\xb2\xe0\xe5\xe2\xd4\x11\xd6\xc0>\xc0t'],
                    ['extra_data', None],
[...]

PR #734, opened 2018-08-17
Bugfix add more restrictive validatior for scene file

kbeker:
Resolves https://github.com/golemfactory/concent/issues/716 
Resolves https://github.com/golemfactory/concent/issues/675

### DIFF
--- a/concent_api/core/constants.py
+++ b/concent_api/core/constants.py
@@ -1,6 +1,11 @@
 from enum import IntEnum
 import re
 
+# Defines available scene file paths for Concent validators
+VALID_SCENE_FILE_PREFIXES = ['/golem/resources/']
+
+# Defines available scene file extension for Blender render
+SCENE_FILE_EXTENSION = '.blend'
 
 # Defines max length of task_id passed in Golem Messages.
 MESSAGE_TASK_ID_MAX_LENGTH = 128
--- a/concent_api/core/tests/test_integration_subtask_results_verify.py
+++ b/concent_api/core/tests/test_integration_subtask_results_verify.py
@@ -16,13 +16,13 @@ from core.tests.utils import add_time_offset_to_date
 from core.tests.utils import ConcentIntegrationTestCase
 from core.tests.utils import parse_iso_date_to_timestamp
 from core.transfer_operations import create_file_transfer_token_for_golem_client
+from core.utils import extract_name_from_scene_file_path
 from common.constants import ErrorCode
 from common.helpers import get_current_utc_timestamp
 from common.helpers import get_storage_result_file_path
 from common.helpers import get_storage_source_file_path
 from common.helpers import parse_timestamp_to_utc_datetime
 from common.testing_helpers import generate_ecc_key_pair
-
 
 (CONCENT_PRIVATE_KEY, CONCENT_PUBLIC_KEY) = generate_ecc_key_pair()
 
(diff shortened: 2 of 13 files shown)


concent_api/core/tests/test_integration_subtask_results_verify.py
```diff
@@ -50,6 +50,8 @@ def setUp(self):
             task_id=self.task_id,
         )
         self.report_computed_task = self._create_report_computed_task()
+        self.scene_file_without_golem_path = self.report_computed_task.task_to_compute.compute_task_def['extra_data'][
```
rwrzesien: Why is this replaced here ? Shouldn't it be replaced in `send_blender_verification_request` ?

rwrzesien: I mean that it is already being done there now https://github.com/golemfactory/concent/pull/734/files#diff-6f9c043a7cc2560c4e15c0127a57dc25R29. Why do you have to replace it before sending request ?

rwrzesien: After discussion, please extract code for replacing beginning of the path to separate function and reuse in both places.

concent_api/core/constants.py
```diff
@@ -1,6 +1,11 @@
 from enum import IntEnum
 import re
 
+# Defines available scene file paths for Concent validators
+AVAILABLE_SCENE_FILE_PATHS = ['/golem/resources/']
```
cameel: This could have a clearer name. It's not even a full path, just a prefix.
`VALID_SCENE_FILE_PREFIXES`? `VALID_RENDERING_ROOT_DIRS`?

concent_api/core/utils.py
```diff
@@ -102,3 +104,16 @@ def hex_to_bytes_convert(client_public_key: str):
     key_bytes = decode_hex(client_public_key)
     assert len(key_bytes) == GOLEM_PUBLIC_KEY_LENGTH
     return key_bytes
+
+
+def extract_name_from_scene_file_path(scene_file_path):
+    for golem_resources_path in AVAILABLE_SCENE_FILE_PATHS:
+        if scene_file_path.startswith(golem_resources_path):
+            scene_file = scene_file_path[len(golem_resources_path):]
```
cameel: This is misleading. `scene_file` and `scene_file_path` are both paths to a scene file. The difference is that one is a relative path within the archive and the other an absolute path within the docker image with Blender.

concent_api/requirements.txt
```diff
@@ -2,7 +2,7 @@
 assertpy
 Django >= 1.11, < 1.12
 psycopg2
--e git://github.com/golemfactory/golem-messages.git@v2.11.0#egg=golem_messages
+-e git://github.com/golemfactory/golem-messages.git@f120ac4823cf1b0037ea4077e14e191102594a79#egg=golem_messages
```
cameel: Why is a commit hard-coded here?

concent_api/core/tests/test_utils.py
```diff
@@ -8,8 +10,9 @@
 from golem_messages import constants
 from golem_messages import helpers
 
+from core.exceptions import SceneFilePathError
 from core.tests.utils import ConcentIntegrationTestCase
-from core.utils import calculate_maximum_download_time
+from core.utils import calculate_maximum_download_time, extract_name_from_scene_file_path
```
dybi: you know what to do ;)

concent_api/requirements.txt
kbeker: As I wrote in PR comment in the top 
>About requirements:
Temporary solution before new version of Golem Messages will tagged.

I need to wait with merging it till golem_messages won't have new tag.

concent_api/core/tests/test_utils.py
kbeker: Oh, my bad. Done :)

concent_api/core/utils.py
kbeker: Done

Issue #735, opened 2018-08-17
Replace CONCENT_ETHEREUM_ADDRESS with CONCENT_ETHEREUM_PUBLIC_KEY 


rwrzesien:
Replace `CONCENT_ETHEREUM_ADDRESS` with `CONCENT_ETHEREUM_PUBLIC_KEY`. Generate `CONCENT_ETHEREUM_ADDRESS` from `CONCENT_ETHEREUM_PUBLIC_KEY`.

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/159848266

cameel:
Please remember that this needs to be synchonized with deployment. We need to change the settings in our configs there.

PR #738, opened 2018-08-17
Replace CONCENT_ETHEREUM_ADDRESS with CONCENT_ETHEREUM_PUBLIC_KEY

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/735

### DIFF
--- a/concent_api/common/helpers.py
+++ b/concent_api/common/helpers.py
@@ -6,6 +6,7 @@ import time
 
 from django.conf import settings
 from django.utils                   import timezone
+from ethereum.utils import sha3
 from mypy.types import Optional
 from mypy.types                     import Union
 import requests
@@ -14,9 +15,11 @@ from golem_messages                 import message
 from golem_messages.datastructures  import FrozenDict
 from golem_messages.exceptions      import MessageError
 from golem_messages.shortcuts import dump
+from golem_messages.utils import decode_hex
 
-from core.exceptions                import Http400
-from common.constants                import ErrorCode
+from core.constants import ETHEREUM_PUBLIC_KEY_LENGTH
+from core.exceptions import Http400
+from common.constants import ErrorCode
 
 
 def is_base64(data: str) -> bool:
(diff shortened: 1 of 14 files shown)


concent_api/common/helpers.py
```diff
@@ -17,6 +18,7 @@
 
 from core.exceptions                import Http400
 from common.constants                import ErrorCode
+from common.constants import ETHEREUM_PUBLIC_KEY_LENGTH
```
cameel: Please fix the two imports above too.

I'd prefer to have the whole "block" reformatted rather than just one line because otherwise the result looks as if we had not consistent style at all.

concent_api/concent_api/settings/testing.py
```diff
@@ -77,6 +77,8 @@
 
 CUSTOM_PROTOCOL_TIMES = True
 
+CONCENT_ETHEREUM_PUBLIC_KEY = '04efb99d9860f4dec4cb548a5722c27e9ef58e37fbab9719c5b33d55c216db49311221a01f638ce5f255875b194e0acaa58b19a89d2e56a864427298f826a7f887'
+
```
cameel: This should only be hard-coded in `development.py`. Tests should set it themselves, just like they do for the message keys.

And in `development.py` - only hard-code if if there's some universal key usable out of the box.

rwrzesien: @cameel It is generated from `PRIVATE_KEY` from https://theethereum.wiki/w/index.php/Accounts,_Addresses,_Public_And_Private_Keys,_And_Tokens

Is that ok to hardcode it in development ?

cameel: The idea is that you should be able to use settings from `development.py` out of the box in most cases, without having to modify them.

rwrzesien: I think it would be better to fail, and set this value in `local_settings.py` if needed.

cameel: In that case we don't want it in `development.py` either.

At most we might add a `manage.py check` for this setting with a hint about where to get a real key from.

concent_api/common/helpers.py
rwrzesien: :+1: done

Issue #745, opened 2018-08-21
Getting version by `setup.py` in Singing Service and MiddleMan Protocol


kbeker:
Currently script `setup.py` in SS and MM Protocol gets version only when `RELEASE-VERSION` file exists. This way of importing version prevent installation these apps as a requirements from e.g. `requirements.lock` file. 
There should be added importing version by script same way as it is done in golem_messages repo -> https://github.com/golemfactory/golem-messages/blob/master/setup.py#L7

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/159919495

PR #748, opened 2018-08-21
Add generating version by get_version function from golem_message repository

kbeker:
Resolves https://github.com/golemfactory/concent/issues/745

### DIFF
--- a/middleman_protocol/setup.py
+++ b/middleman_protocol/setup.py
@@ -1,6 +1,5 @@
 #!/usr/bin/env python3
 
-import os
 from setuptools import setup
 
 from version import get_version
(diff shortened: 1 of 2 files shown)


middleman_protocol/setup.py
```diff
@@ -3,19 +3,12 @@
 import os
```
Jakub89:
```
./middleman_protocol/setup.py:3:1: F401 'os' imported but unused
```

signing_service/setup.py
```diff
@@ -3,19 +3,12 @@
 import os
```
Jakub89:
```
./signing_service/setup.py:3:1: F401 'os' imported but unused
```

Issue #708, opened 2018-08-06
Verifier does not always handle VerificationMismatch caused by BadZipFile


kbeker:
When verification shows that result should be `MISMATCH` or `ERROR` then `handle_verification_results` decorator should catch `VerificationMismatch` exception and pass to `verification_result` function `MISMATCH` or `ERROR` result without raising exception and causing error on sentry. 

All needed informations on sentry: https://talkback.golem.network/sentry/concent/issues/28905/

Please add new api-integration-tests` to `MISMATCH` result

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/159559406

PR #749, opened 2018-08-22
Bug fix in assertion

MichalW-CP:
Resolves https://github.com/golemfactory/concent/issues/708

### DIFF
--- a/concent_api/core/tasks.py
+++ b/concent_api/core/tasks.py
@@ -125,7 +125,7 @@ def verification_result(
 
     result_enum = VerificationResult[result]
 
-    assert result_enum != VerificationResult.ERROR or all([error_message, error_code])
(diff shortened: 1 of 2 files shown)


concent_api/core/tasks.py
```diff
@@ -125,7 +125,7 @@ def verification_result(
 
     result_enum = VerificationResult[result]
 
-    assert result_enum != VerificationResult.ERROR or all([error_message, error_code])
+    assert result_enum != VerificationResult.ERROR or all([isinstance(error_message, str), isinstance(error_code, str)])
```
rwrzesien: Fix looks good, but I think we should also determine on what occasion there is an empty `error_message` or `error_code` and fix that too. From what I see we always set `error_message` or `error_code` when raising `VerificationError` so this should not happen.

Btw. linked sentry issue doesn't look related to this case.

rwrzesien: After discussion, looks like the cause is already fixed.

Issue #679, opened 2018-07-23
[WIP] MiddleMan raise IncompleteReadError when signing service close connection.


Jakub89:
When running middleman with default settings, which are: 
``` shell
INFO     | MiddleMan is serving on ('127.0.0.1', 9054)
```
and then running signing service with `./signing_service.sh 127.0.0.1 1 --concent-cluster-port=9054`:

1. Nothing happens, there's no response from either sides.

2. When closing signing service with Ctrl-C, we get this error:
```  shell
15:54:03 ERROR    | Exception occurred: 0 bytes read on a total of None expected bytes, Traceback: Traceback (most recent call last):
  File "/home/jakub/Documents/work/projects/concent/concent_api/middleman/middleman_server.py", line 89, in _handle_user_request
    data = await reader.readuntil()
  File "/usr/lib/python3.6/asyncio/streams.py", line 578, in readuntil
    raise IncompleteReadError(chunk, None)
asyncio.streams.IncompleteReadError: 0 bytes read on a total of None expected bytes
 
15:54:03 ERROR    | Task exception was never retrieved
future: <Task finished coro=<MiddleMan._handle_user_request() done, defined at /home/jakub/Documents/work/projects/concent/concent_api/middleman/middleman_server.py:87> exception=IncompleteReadError('0 bytes read on a total of None expected bytes',)> 
Traceback (most recent call last):
  File "/home/jakub/Documents/work/projects/concent/concent_api/middleman/middleman_server.py", line 89, in _handle_user_request
    data = await reader.readuntil()
  File "/usr/lib/python3.6/asyncio/streams.py", line 578, in readuntil
    raise IncompleteReadError(chunk, None)
asyncio.streams.IncompleteReadError: 0 bytes read on a total of None expected bytes
```

rwrzesien:
https://docs.python.org/3/library/asyncio-stream.html#asyncio.StreamReader.readuntil

> If an EOF occurs and the complete separator is still not found, an IncompleteReadError exception will be raised, and the internal buffer will be reset. The IncompleteReadError.partial attribute may contain the separator partially.

I think we could add handling of `IncompleteReadError` and close gracefully when it is caught.

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/159264674

cameel:
[...]

PR #751, opened 2018-08-23
Bugfix middle man raise incomplete read error when signing service close connection

rwrzesien:
Resovles https://github.com/golemfactory/concent/issues/679

### DIFF
--- a/concent_api/middleman/middleman_server.py
+++ b/concent_api/middleman/middleman_server.py
@@ -147,6 +147,8 @@ class MiddleMan:
                 await self._respond_to_user(frame, writer)
             except (BrokenEscapingInFrameMiddlemanProtocolError, asyncio.LimitOverrunError, MiddlemanProtocolError) as exception:
                 await self._send_immediate_error(writer, exception)
+            except asyncio.streams.IncompleteReadError:
+                pass
             except Exception as exception:  # pylint: disable=broad-except
                 crash_logger.error(
                     f"Exception occurred: {exception}, Traceback: {traceback.format_exc()}"
--- a/concent_api/middleman/middleman_server.py
+++ b/concent_api/middleman/middleman_server.py
@@ -127,6 +127,8 @@ class MiddleMan:
             await self._respond_to_user(frame, writer)
         except (BrokenEscapingInFrameMiddlemanProtocolError, asyncio.LimitOverrunError, MiddlemanProtocolError) as exception:
             await self._send_immediate_error(writer, exception)
+        except asyncio.streams.IncompleteReadError:
+            pass
         except Exception as exception:  # pylint: disable=broad-except
             crash_logger.error(
                 f"Exception occurred: {exception}, Traceback: {traceback.format_exc()}"


kbeker:
@rwrzesien You also added `concent_api/tests_resources/result.zip` and `concent_api/tests_resources/source.zip`. Please delete them :)

rwrzesien:
@kbeker done :+1:

concent_api/middleman/middleman_server.py
```diff
@@ -127,6 +127,8 @@ def _close_middleman(self) -> None:
             await self._respond_to_user(frame, writer)
         except (BrokenEscapingInFrameMiddlemanProtocolError, asyncio.LimitOverrunError, MiddlemanProtocolError) as exception:
             await self._send_immediate_error(writer, exception)
+        except asyncio.streams.IncompleteReadError:
+            pass
```
Jakub89: You've added that handling only in `_handle_concent_connection()` (default port:9054), but the exception happen when we close `signing_service` (default port:9055). Either copy or move this to `_handle_service_connection()`.

rwrzesien: :+1: done

pawelkisielewicz: Don't we want to add logging for this situation? I'm not sure we should, just asking

rwrzesien: The case is that as @dybi described it yesterday, the problem is 'by design', so it is not actually an real error. Lets see how the situation with this will develop, maybe after merging his latest changes there will be no need for this pull request.

cameel:
> Resovles #679

?

concent_api/middleman/middleman_server.py
cameel: This should be handled in the decorator, not here. It's not an error like the other exceptions being caught here.

Have you read what I wrote in the issue?

rwrzesien:
Closing as the issues is being solved with #766.

cameel:
@rwrzesien Could you extract a separate pull request with a fix for this? I really want to review this one but now I would have to review the whole middleman pull request.

Also, this issue is not really limited to Middleman. It's a problem in the synchronous version too (though maybe it raises a different exception).

rwrzesien:
@cameel 
I don't think it is possible to extract only this because most of the code in `MiddleMan` changed and taking that fix only would be completely out of context. But it is only two exceptions checking and handling, exactly those lines:
https://github.com/golemfactory/concent/pull/766/files#diff-2834a4ae48c4f51214b35d7783a8a89cR51
https://github.com/golemfactory/concent/pull/766/files#diff-2834a4ae48c4f51214b35d7783a8a89cR112

cameel:
And what about the Signing Service? EOF can happen in both.

In my opinion is the wrong place to handle it. EOF should be handled in the generator. Forcing the upper layer is like making a file object always raise an exception when you reach the end of a file.

rwrzesien:
Closing as fixed in https://github.com/golemfactory/concent/pull/766, other bugs will be described in separate issues.

PR #752, opened 2018-08-23
Bugfix verification deadline check in conductor test

Jakub89:
No issue for this one, just something that happens on my computer. That one test sometimes fails because values for comparison are created separately with `current_time`. And also I've add `SIGNING_SERVICE_PUBLIC_KEY` to testing.py settings so that `python manage.py test --settings concent_api.settings.testing` doesn't crush on system_check.


concent_api/concent_api/settings/testing.py
```diff
@@ -77,6 +77,8 @@
 
 CUSTOM_PROTOCOL_TIMES = True
 
+SIGNING_SERVICE_PUBLIC_KEY = b'\xc0$\xa3\xab\x8d\x04[q\x0b\x1b\x00#\x8b\x18\xd9\xcf\x95\xf1\\`\x08w@\xf0\x9d\xf6B7;\xe56\xc8\xb1pp\x97\xac\xb3\xd3\xc1\xb0\x96a6\xcat\x8b\xea\xbb\x98\x04\xcfD28\x97$\xa9\xb9b8\x17\xe1\xfc'
```
dybi: shouldn't this be overridden directly like `CONCENT_PRIVATE_KEY`?

concent_api/concent_api/tests/test_check_siginig_service_key_availability_for_middleman.py
```diff
@@ -17,9 +18,11 @@ def test_that_if_middleman_is_disabled_signing_service_public_key_does_not_have_
             assert_that(errors).is_empty()
 
     def test_that_if_middleman_is_enabled_and_key_signing_service_public_key_is_not_defined_error_is_returned(self):
-        errors = check_signing_service_key_availability_for_middleman()
-        assert_that(errors).is_length(1)
-        assert_that(errors[0]).is_equal_to(create_error_43_signing_service_public_key_is_missing())
+        with override_settings():
```
dybi: this change doesn't seem to belong to this PR ;)

concent_api/concent_api/settings/testing.py
Jakub89: It can't be, because of `system_check`. If `middleman` is specify in `CONCENT_FEATURES`, then `SIGNING_SERVICE_PUBLIC_KEY` is required. `CONCENT_PRIVATE_KEY` is not required similarly.

concent_api/concent_api/tests/test_check_siginig_service_key_availability_for_middleman.py
Jakub89: True, actually. I don't know why I've added that `override_settings()`. `del settings` works just fine without it.

Issue #707, opened 2018-08-06
Verifier fails to find the source package it has just downloaded


kbeker:
At this moment we do not validate situation when file does not exist in moment when verifier wants to unpack archive to get file list. There should be added validation before verifier wants to get list of files from archives. Also please investigate why and how this situation may happen. 

Exception raised: 
```python
FileNotFoundError: No such file or directory: 
'/srv/storage/source_e5b64e01-bf44-8598-5831-07f321291102.9af2a0b9-a556-f3dd-9b4d-7c97c97ebcaa.zip'
```
Function which produces error:
```python
def get_files_list_from_archive(file_path: str) -> List[str]:
    """ Returns list of files from given zip archive. """
    return zipfile.ZipFile(file_path).namelist()
```

Please add unit tests and be sure that all `api-integration-tests...` works.

Sentry link: https://talkback.golem.network/sentry/concent/issues/31103/events/189679/

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/159558351

cameel:
> There should be added validation before verifier wants to get list of files from archives.

Not a validation. `assert`. It's not client's error. It's our code that called this function even though the file did not exist.

PR #757, opened 2018-08-24
Add unit test for verifier get_files_list_from_archive() function

Jakub89:
Resolves: #707 (not really)
But also a part of #582


Jakub89:
Ad.1 We've spoke about that at least twice, on standup and retro. Because we can't reproduce #707 and it's probably non existence, we agree to close the issue by adding unit test for the method where it crushes.
Ad.2,3,4 I've rewrite that test, with no `mock()` this time.

Issue #693, opened 2018-08-01
DatabaseTransactionStorage can't handle transaction removal performed by two different processes in parallel


kbeker:
During multiple tests for transactions one problem with database locks occurred. One of Concent instances removed `PendingEthereumTransaction`:

```09:28:08 INFO | Successfully removed PendingEthereumTransaction with nonce 3202.```

and in the same moment different instance of Concent wants to do same thing but this row is already removed and Concent raised exception:

 `DoesNotExist: PendingEthereumTransaction matching query does not exist.` 

because of:

```09:28:08 ERROR    | Trying to remove PendingEthereumTransaction with nonce 3202 but it does not exist. ```

Two Concent instances tried to do the same thing in the same moment. This is not expected behavior. When one of instances removing row in `PendingEthereumTransaction` model second and any other instance should wait until first instance finish their work.

Needs to be investigated.

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/159461110

cameel:
@Krigpl 
> My suggestion would be to ignore this particular error (trying to remove non existing transaction). I think making sure that only one Concent instance is trying to remove the transaction would be awkward in the current design (one shared SCI module would solve it though). Ignoring this error is fine - the transaction is not in the storage after the call and that's the goal. And we can keep logs to see if we ever incorrectly tried to remove a transaction.

I agree. That seems to be the best solution. We should check if the transaction exists and if not - not try to remove it and not decrease the nonce. That solves the problem completely.

> While writing this I came up with another idea how to solve this. up to you to decide if it's feasible.
Each Concent instance has its own DB where it stores transactions only (no global nonce).
And another shared DB instance with just the nonce for all Concents.
That would work because nonce is still shared across all Concent instances, but each instance stores its own transactions only - so it will only try to remove its own.

It's possible even within a single database - just by marking each transaction with an additional "instance ID" so that we know which instance is responsible for managing it.

The problem with multiple transaction pools is that we would have to guarantee that there always exists an  instance responsible for that particular pool and the instance knows its ID. It's easy if the number of pools is static but when we enable autoscaling on the cluster that will no longer be the case. The number of instances will vary dynamically according to network load.

[...]

PR #760, opened 2018-08-27
DatabaseTransactionStorage: Do not raise exception when trying to rem…

rwrzesien:
…ove PendingEthereumTransaction which does not exist

Resolves https://github.com/golemfactory/concent/issues/693


rwrzesien:
@pawelkisielewicz Sorry I have missed your comment before merging. I don't think that much is necessary in this case, there is no problem here with concurrency (as described in related issue i one of the comments) but that requests come one after another, and in first item is removed, and is missing in second. The bug was actually about catching wrong `DoesNotExist` exception, which is kind of typical Django flow for retrieving single object using `.get()`.

Issue #718, opened 2018-08-09
Use SELECT FOR UPDATE to prevent lost updates


cameel:
This is another attempt at solving #412. The solution proposed here is simpler in many ways (no retries, not need for `finally`) but also broader in scope (applies to more than just `Subtask`) so I think it's betterto close that previous issue instead of rewriting the description and leaving a confusing thread of comments. The issue addresses the same underlying problem though.

Due to the broader scope, solving it will also solve #4.

### Overview
This article presents the problem and the solution really well: [A beginner’s guide to database locking and the lost update phenomena](https://vladmihalcea.com/a-beginners-guide-to-database-locking-and-the-lost-update-phenomena/).

The solution is simple: find every place in our code where we update an existing object (database row) and make sure that the data that is saved was fetched using `SELECT FOR UPDATE` and not a plain `SELECT`.

### Details
Until now our strategy was a little confused because I assumed that update operations do not lock the updated row until the end of transaction and that `select_for_update()` adds an overhead that we did not previously have and should be used sparingly. Turns out that we're already implicitly locking rows so there's not that much difference if we lock a little earlier.

Unlike in #412 we're not going to deal with timed out locks and retries here. Locks do not time out if we don't use `nowait=True` (they could but [`lock_timeout` in PostgreSQL](https://www.postgresql.org/docs/current/static/runtime-config-client.html#GUC-LOCK-TIMEOUT) is disabled by default).

For this reason automatic retries on locks are not necessary. We may need retries on creation operations to prevent problems like #486 but we'll deal with that in a separate issue. `INSERT`, unlike `UPDATE` is not a problem because two simultaneous creations will fail due to `UNIQUE` constraints on the primary key and the only reason for the retry is to complete the operation without returning HTTP 500 and having the client retry by itself. There's no risk of overwriting anything.

So in this issue we simply need to review our update statements. This applies not only to `Subtask`. There are several situations where we need `select_for_update()` (the list is not exhaustive):
- When marking pending responses as delivered in `receive/`.
- When updating `UploadReport`s in the `storage` database.
- When updating `upload_finished` flag in the `VerificationRequest`.

pawelkisielewicz:
All our requests in send concern particular subtask or more than one subtask (ForcePayment). In first pull request I created lock for core/send. 
[...]

PR #753, opened 2018-08-23
Feature use select for update to prevent lost updates

pawelkisielewicz:
Resolves https://github.com/golemfactory/concent/issues/718

Issue #759, opened 2018-08-27
Use one retry of message trying to store subtask


pawelkisielewicz:
There is another one problem similar to https://github.com/golemfactory/concent/issues/718, but in case when subtask does NOT exist in database. The case and solution should look like this:

1. Two or more messages are trying to create a new subtask (the same) in our database at one time. 
2. One of them will success, others will fail with IntegrityError. 
3. We should catch this IntegrityError in `store_subtask`, create a new custom Error only for this case and raise this new Error. 
4. New Error should be caught at the beginning of validations (e.g. beginning of `handle_message` or `view`. Important is to do it before database lock for existing subtasks). This custom Error is necessary to be 100%, we are catching error from `store_subtask` and not from other parts of application. I'm not sure if it would be possible, but this way we are 100% that 'new Error' comes from `store_subtask`
5. Message should be retried and validated again. It is possibility that concent will answer with e.g. `ServiceRefused.REASON.DuplicateRequest`, but message should try to pass all validations again with new database state (after one process successfully finished). There is no danger the same problem will be again, because subtask with this id will already exist in database and we use select_for_update lock to execute all processed one by one, if there will be more than one waiting.

In this case inside function store_subtask may be a good idea to create a new nested atomic block, but only inside this particular function to:
- gracefully rollback changes from this function if necessary
- catch IntegrityError and raise new custom Error to retry validations ONLY in this case

cameel:
1) This is really the same problem as in #486
2) This is not limited to `Subtask`. We should do this in all cases where we create a new object in the database.
3) You can check the PostgreSQL error code from the exception with `exception.__cause__.pgcode`.
    - [IntegrityError: distinguish between unique constraint and not null violations](https://stackoverflow.com/questions/12490172/integrityerror-distinguish-between-unique-constraint-and-not-null-violations)
    - [psycopg2.errorcodes](http://initd.org/psycopg/docs/errorcodes.html)
4) Unfortunately a nested atomic block is not enough. To get an `IntegrityError` you need to commit the whole transaction. To implement this correctly we really need `@atomic` blocks smaller than a view.
    - The transaction **must** be committed before you perform any action that may have irreversible side-effects. Like adding tasks to queue or making payments. Without this retry is not safe.

pawelkisielewicz:
OK. 
Ad. 1 - right :)
[...]

PR #673, opened 2018-07-20
Middleman protocol

rwrzesien:
Creates message class for Middleman protocol.
Creates helper generator for handling receiving Middleman protocol messages.
Add tests for messages and helper

PR #764, opened 2018-08-29
Feature select for update to prevent lost updates

pawelkisielewicz:
Resolves https://github.com/golemfactory/concent/issues/718
based on https://github.com/golemfactory/concent/pull/753 
This branch includes and fixes your comments from https://github.com/golemfactory/concent/pull/753

### DIFF
--- a/concent_api/api-integration-multi-threads-test.py
+++ b/concent_api/api-integration-multi-threads-test.py
@@ -14,7 +14,7 @@ from golem_messages.message.concents import ForceReportComputedTask
 from golem_messages.message.tasks import AckReportComputedTask
 from golem_messages.message.tasks import ReportComputedTask
 
-from api_testing_common import compare_lists_regardless_of_order
+from api_testing_common import assert_iterables_content_is_equal
 from api_testing_common import count_fails
 from api_testing_common import PROVIDER_PRIVATE_KEY
 from api_testing_common import REQUESTOR_PRIVATE_KEY
@@ -188,7 +188,7 @@ def test_case_multiple_requests_concerning_one_subtask_will_be_processed_one_by_
 
     print('Responses = ' + str(responses_global))
     error_message = f"Responses should be: {expected_responses}."
-    assert compare_lists_regardless_of_order(actual=responses_global, expected=expected_responses), error_message
+    assert_iterables_content_is_equal(actual=responses_global, expected=expected_responses), error_message
     print('Single test passed successfully')
 
 
--- a/concent_api/api-integration-multi-threads-test.py
+++ b/concent_api/api-integration-multi-threads-test.py
@@ -1,0 +1,214 @@
+
+# !/usr/bin/env python3
+import os
+import sys
+import time
+from threading import Thread
+from typing import Optional
+
+import requests
+from golem_messages import message
+from golem_messages.factories.tasks import ReportComputedTaskFactory
+from golem_messages.message import Message
+from golem_messages.message.concents import ForceGetTaskResult
+from golem_messages.message.concents import ForceReportComputedTask
+from golem_messages.message.tasks import AckReportComputedTask
+from golem_messages.message.tasks import ReportComputedTask
+
+from api_testing_common import count_fails
+from api_testing_common import PROVIDER_PRIVATE_KEY
+from api_testing_common import REQUESTOR_PRIVATE_KEY
+from api_testing_common import api_request
+from api_testing_common import assert_condition
+from api_testing_common import create_signed_task_to_compute
+from api_testing_common import run_tests
+from api_testing_common import timestamp_to_isoformat
+from common.helpers import get_current_utc_timestamp
+from common.helpers import sign_message
+
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "concent_api.settings")
+
+REPORT_COMPUTED_TASK_SIZE = 10
+
+NUMBER_OF_TESTING_THREADS = 3  # changing this variable remember to correct number of expected responses
+MAXIMUM_WAITING_TIME_FOR_ALL_RESPONSES = 3  # seconds
+
+responses_global = []  # type: list
+
+
+def clear_responses():
+    global responses_global
+    responses_global.clear()
+
+
+def call_function_in_treads(
+    func,
+    number_of_threads: int,
+    cluster_url: str,
+    golem_message: Message,
+) -> None:
+    for i in range(number_of_threads):
+        t = Thread(target=func, args=(cluster_url, golem_message,))
+        t.start()
+
+
+def get_ack_report_computed_task(report_computed_task: ReportComputedTask) -> AckReportComputedTask:
+    ack_report_computed_task = AckReportComputedTask(report_computed_task=report_computed_task)
+    return ack_report_computed_task
+
+
+def get_force_get_task_result(report_computed_task: ReportComputedTask) -> ForceGetTaskResult:
+    force_get_task_result = message.concents.ForceGetTaskResult(report_computed_task=report_computed_task)
+    return force_get_task_result
+
+
+def get_force_report_computed_task(report_computed_task: ReportComputedTask) -> ForceReportComputedTask:
+    force_report_computed_task = ForceReportComputedTask()
+    force_report_computed_task.report_computed_task = report_computed_task
+    return force_report_computed_task
+
+
+def create_report_computed_task(
+    task_id: str,
+    subtask_id: str
+) -> ReportComputedTask:
+    current_time = get_current_utc_timestamp()
+
+    task_to_compute = create_signed_task_to_compute(
+        timestamp=timestamp_to_isoformat(current_time),
+        task_id=task_id,
+        subtask_id=subtask_id,
+        deadline=(current_time + 100),
+        price=10000,
+    )
+
+    report_computed_task = ReportComputedTaskFactory(
+        task_to_compute=task_to_compute,
+    )
+    sign_message(report_computed_task, PROVIDER_PRIVATE_KEY)
+    return report_computed_task
+
+
+def send_correct_force_report_computed_task(
+    cluster_url: str,
+    report_computed_task: ReportComputedTask,
+    expected_status: Optional[int] = None
+) -> None:
+    response = api_request(
+        cluster_url,
+        'send',
+        PROVIDER_PRIVATE_KEY,
+        CONCENT_PUBLIC_KEY,
+        get_force_report_computed_task(
+            report_computed_task=report_computed_task,
+        ),
+        headers={
+            'Content-Type': 'application/octet-stream',
+        },
+        expected_status=expected_status,
+    )
+    responses_global.append(response['error_code'] if isinstance(response, dict) else response)
+
+
+def send_correct_ack_report_computed_task(
+    cluster_url: str,
+    report_computed_task: ReportComputedTask,
+    expected_status: Optional[int] = None
+) -> None:
+    api_request(
+        cluster_url,
+        'send',
+        REQUESTOR_PRIVATE_KEY,
+        CONCENT_PUBLIC_KEY,
+        get_ack_report_computed_task(
+            report_computed_task=report_computed_task
+        ),
+        headers={
+            'Content-Type': 'application/octet-stream',
+        },
+        expected_status=expected_status,
+    )
+
+
+def send_correct_force_get_task_result(
+    cluster_url: str,
+    report_computed_task: ReportComputedTask,
+    expected_status: Optional[int] = None,
+) -> None:
+    response = api_request(
+        cluster_url,
+        'send',
+        REQUESTOR_PRIVATE_KEY,
+        CONCENT_PUBLIC_KEY,
+        get_force_get_task_result(
+            report_computed_task=report_computed_task,
+        ),
+        headers={
+            'Content-Type': 'application/octet-stream',
+        },
+        expected_status=expected_status,
+    )
+    responses_global.append(response.__class__.__name__)
+
+
+@count_fails
+def test_case_multiple_requests_concerning_one_subtask_will_be_processed_one_by_one_if_subtask_exists_in_database(
+    cluster_consts: str,
+    cluster_url: str,
+    task_id: str,
+    subtask_id: str
+):
+    report_computed_task = create_report_computed_task(task_id=task_id, subtask_id=subtask_id)
+
+    """send messages to save subtask in database and change state to REPORTED. Next message ForceGetTaskResult
+       requires REPORTED state and this is the simplest way to get it """
+    send_correct_force_report_computed_task(
+        cluster_url=cluster_url,
+        report_computed_task=report_computed_task,
+        expected_status=202,
+    )  # Subtask state changed to: FORCING_REPORT
+    send_correct_ack_report_computed_task(
+        cluster_url=cluster_url,
+        report_computed_task=report_computed_task,
+        expected_status=202,
+    )  # Subtask state changed to: REPORTED
+
+    clear_responses()
+    """this is test- sending some messages in one time """
+    call_function_in_treads(
+        func=send_correct_force_get_task_result,  # Subtask state changed to: FORCING_RESULT_TRANSFER
+        number_of_threads=NUMBER_OF_TESTING_THREADS,
+        cluster_url=cluster_url,
+        golem_message=report_computed_task,
+    )
+    "waiting for responses from all threads"
+    end_time = time.time() + MAXIMUM_WAITING_TIME_FOR_ALL_RESPONSES
+    while len(responses_global) != NUMBER_OF_TESTING_THREADS:
+        time.sleep(0.1)
+        if time.time() >= end_time:
+            break
+
+    expected_responses = ['AckForceGetTaskResult', 'ServiceRefused', 'ServiceRefused']
+
+    print('Responses = ' + str(responses_global))
+    assert_condition(
+        actual=responses_global,
+        expected=expected_responses,
+        error_message=f"Responses should be: {expected_responses}. Probably reason is that something is wrong with code"
+        f" and test failed or maybe you changed number of testing threads and forgot to change expected_responses?",
+        compare_lists_regardless_of_order=True,
+    )
+    print('Single test passed successfully')
+
+
+if __name__ == '__main__':
+    try:
+        from concent_api.settings import CONCENT_PUBLIC_KEY
+
+        run_tests(globals())
+    except requests.exceptions.ConnectionError as exception:
+        print("\nERROR: Failed connect to the server.\n", file=sys.stderr)
+        sys.exit(str(exception))
+
+
+
--- a/concent_api/api-integration-multi-threads-test.py
+++ b/concent_api/api-integration-multi-threads-test.py
@@ -1,12 +1,11 @@
-
 # !/usr/bin/env python3
 import os
 import sys
 import time
 from threading import Thread
 from typing import Optional
-
 import requests
+
 from golem_messages import message
 from golem_messages.factories.tasks import ReportComputedTaskFactory
 from golem_messages.message import Message
@@ -15,11 +14,11 @@ from golem_messages.message.concents import ForceReportComputedTask
 from golem_messages.message.tasks import AckReportComputedTask
 from golem_messages.message.tasks import ReportComputedTask
 
+from api_testing_common import compare_lists_regardless_of_order
 from api_testing_common import count_fails
 from api_testing_common import PROVIDER_PRIVATE_KEY
 from api_testing_common import REQUESTOR_PRIVATE_KEY
 from api_testing_common import api_request
-from api_testing_common import assert_condition
 from api_testing_common import create_signed_task_to_compute
 from api_testing_common import run_tests
 from api_testing_common import timestamp_to_isoformat
@@ -48,24 +47,20 @@ def call_function_in_treads(
     golem_message: Message,
 ) -> None:
     for i in range(number_of_threads):
-        t = Thread(target=func, args=(cluster_url, golem_message,))
-        t.start()
+        thread = Thread(target=func, args=(cluster_url, golem_message,))
+        thread.start()
 
 
 def get_ack_report_computed_task(report_computed_task: ReportComputedTask) -> AckReportComputedTask:
-    ack_report_computed_task = AckReportComputedTask(report_computed_task=report_computed_task)
-    return ack_report_computed_task
+    return AckReportComputedTask(report_computed_task=report_computed_task)
 
 
 def get_force_get_task_result(report_computed_task: ReportComputedTask) -> ForceGetTaskResult:
-    force_get_task_result = message.concents.ForceGetTaskResult(report_computed_task=report_computed_task)
-    return force_get_task_result
+    return message.concents.ForceGetTaskResult(report_computed_task=report_computed_task)
 
 
 def get_force_report_computed_task(report_computed_task: ReportComputedTask) -> ForceReportComputedTask:
-    force_report_computed_task = ForceReportComputedTask()
-    force_report_computed_task.report_computed_task = report_computed_task
-    return force_report_computed_task
+    return ForceReportComputedTask(report_computed_task = report_computed_task)
 
 
 def create_report_computed_task(
@@ -157,11 +152,11 @@ def test_case_multiple_requests_concerning_one_subtask_will_be_processed_one_by_
     cluster_url: str,
     task_id: str,
     subtask_id: str
-):
+) -> None:
     report_computed_task = create_report_computed_task(task_id=task_id, subtask_id=subtask_id)
 
-    """send messages to save subtask in database and change state to REPORTED. Next message ForceGetTaskResult
-       requires REPORTED state and this is the simplest way to get it """
+    # send messages to save subtask in database and change state to REPORTED. Next message ForceGetTaskResult
+    # requires REPORTED state and this is the simplest way to get it
     send_correct_force_report_computed_task(
         cluster_url=cluster_url,
         report_computed_task=report_computed_task,
@@ -174,14 +169,14 @@ def test_case_multiple_requests_concerning_one_subtask_will_be_processed_one_by_
     )  # Subtask state changed to: REPORTED
 
     clear_responses()
-    """this is test- sending some messages in one time """
+    # this is test- sending some messages in one time
     call_function_in_treads(
         func=send_correct_force_get_task_result,  # Subtask state changed to: FORCING_RESULT_TRANSFER
         number_of_threads=NUMBER_OF_TESTING_THREADS,
         cluster_url=cluster_url,
         golem_message=report_computed_task,
     )
-    "waiting for responses from all threads"
+    # waiting for responses from all threads
     end_time = time.time() + MAXIMUM_WAITING_TIME_FOR_ALL_RESPONSES
     while len(responses_global) != NUMBER_OF_TESTING_THREADS:
         time.sleep(0.1)
@@ -189,26 +184,18 @@ def test_case_multiple_requests_concerning_one_subtask_will_be_processed_one_by_
             break
 
     expected_responses = ['AckForceGetTaskResult', 'ServiceRefused', 'ServiceRefused']
+    assert len(expected_responses) == NUMBER_OF_TESTING_THREADS,'Did you changed number of testing threads and forgot to change expected_responses?'
 
     print('Responses = ' + str(responses_global))
-    assert_condition(
-        actual=responses_global,
-        expected=expected_responses,
-        error_message=f"Responses should be: {expected_responses}. Probably reason is that something is wrong with code"
-        f" and test failed or maybe you changed number of testing threads and forgot to change expected_responses?",
-        compare_lists_regardless_of_order=True,
-    )
+    error_message = f"Responses should be: {expected_responses}."
+    assert compare_lists_regardless_of_order(actual=responses_global, expected=expected_responses), error_message
     print('Single test passed successfully')
 
 
 if __name__ == '__main__':
     try:
         from concent_api.settings import CONCENT_PUBLIC_KEY
-
         run_tests(globals())
     except requests.exceptions.ConnectionError as exception:
         print("\nERROR: Failed connect to the server.\n", file=sys.stderr)
-        sys.exit(str(exception))
-
-
-
+        sys.exit(exception)
--- a/concent_api/api-integration-multi-threads-test.py
+++ b/concent_api/api-integration-multi-threads-test.py
@@ -35,7 +35,7 @@ MAXIMUM_WAITING_TIME_FOR_ALL_RESPONSES = 3  # seconds
 responses_global = []  # type: list
 
 
-def clear_responses():
+def clear_responses() -> None:
     global responses_global
     responses_global.clear()
 
--- a/concent_api/api-integration-multi-threads-test.py
+++ b/concent_api/api-integration-multi-threads-test.py
@@ -51,15 +51,15 @@ def call_function_in_treads(
         thread.start()
 
 
-def get_ack_report_computed_task(report_computed_task: ReportComputedTask) -> AckReportComputedTask:
+def create_ack_report_computed_task(report_computed_task: ReportComputedTask) -> AckReportComputedTask:
     return AckReportComputedTask(report_computed_task=report_computed_task)
 
 
-def get_force_get_task_result(report_computed_task: ReportComputedTask) -> ForceGetTaskResult:
+def create_force_get_task_result(report_computed_task: ReportComputedTask) -> ForceGetTaskResult:
     return message.concents.ForceGetTaskResult(report_computed_task=report_computed_task)
 
 
-def get_force_report_computed_task(report_computed_task: ReportComputedTask) -> ForceReportComputedTask:
+def create_force_report_computed_task(report_computed_task: ReportComputedTask) -> ForceReportComputedTask:
     return ForceReportComputedTask(report_computed_task = report_computed_task)
 
 
@@ -73,7 +73,7 @@ def create_report_computed_task(
         timestamp=timestamp_to_isoformat(current_time),
         task_id=task_id,
         subtask_id=subtask_id,
-        deadline=(current_time + 100),
+        deadline=(current_time + CONCENT_MESSAGING_TIME),
         price=10000,
     )
 
@@ -94,7 +94,7 @@ def send_correct_force_report_computed_task(
         'send',
         PROVIDER_PRIVATE_KEY,
         CONCENT_PUBLIC_KEY,
-        get_force_report_computed_task(
+        create_force_report_computed_task(
             report_computed_task=report_computed_task,
         ),
         headers={
@@ -115,7 +115,7 @@ def send_correct_ack_report_computed_task(
         'send',
         REQUESTOR_PRIVATE_KEY,
         CONCENT_PUBLIC_KEY,
-        get_ack_report_computed_task(
+        create_ack_report_computed_task(
             report_computed_task=report_computed_task
         ),
         headers={
@@ -135,7 +135,7 @@ def send_correct_force_get_task_result(
         'send',
         REQUESTOR_PRIVATE_KEY,
         CONCENT_PUBLIC_KEY,
-        get_force_get_task_result(
+        create_force_get_task_result(
             report_computed_task=report_computed_task,
         ),
         headers={
@@ -194,7 +194,8 @@ def test_case_multiple_requests_concerning_one_subtask_will_be_processed_one_by_
 
 if __name__ == '__main__':
     try:
-        from concent_api.settings import CONCENT_PUBLIC_KEY
+        from concent_api.settings import CONCENT_PUBLIC_KEY, CONCENT_MESSAGING_TIME
+
         run_tests(globals())
     except requests.exceptions.ConnectionError as exception:
         print("\nERROR: Failed connect to the server.\n", file=sys.stderr)
--- a/concent_api/api-integration-multi-threads-test.py
+++ b/concent_api/api-integration-multi-threads-test.py
@@ -184,7 +184,7 @@ def test_case_multiple_requests_concerning_one_subtask_will_be_processed_one_by_
             break
 
     expected_responses = ['AckForceGetTaskResult', 'ServiceRefused', 'ServiceRefused']
-    assert len(expected_responses) == NUMBER_OF_TESTING_THREADS,'Did you changed number of testing threads and forgot to change expected_responses?'
+    assert len(expected_responses) == NUMBER_OF_TESTING_THREADS, 'Did you changed number of testing threads and forgot to change expected_responses?'
 
     print('Responses = ' + str(responses_global))
     error_message = f"Responses should be: {expected_responses}."
--- a/concent_api/api_testing_common.py
+++ b/concent_api/api_testing_common.py
@@ -76,7 +76,7 @@ def assert_condition(actual, expected, error_message = None):
 
 def assert_iterables_content_is_equal(actual: Iterable[str], expected: Iterable[str]) -> None:
     if not sorted(actual) == sorted(expected):
-        raise TestAssertionException(f'Content of iterables is not equal. Actual {actual}. Expected {expected}')
+        raise TestAssertionException(f'Content of iterables is not equal. "Actual: {actual}". "Expected: {expected}"')
 
 
 def print_golem_message(message, indent = 4):
--- a/concent_api/api_testing_common.py
+++ b/concent_api/api_testing_common.py
@@ -1,8 +1,9 @@
 import argparse
-import collections
 import sys
 
 import uuid
+from typing import List
+
 from freezegun import freeze_time
 
 from golem_messages.exceptions      import MessageError
@@ -67,14 +68,18 @@ class count_fails(object):
         print(f'Total failed tests : {cls.get_fails()} out of {cls.number_of_run_tests}')
 
 
-def assert_condition(actual, expected, error_message = None, compare_lists_regardless_of_order: bool = False):
+def assert_condition(actual, expected, error_message = None):
     message = error_message or f"Actual: {actual} != expected: {expected}"
-    if not compare_lists_regardless_of_order:
-        if actual != expected:
-            raise TestAssertionException(message)
-    else:
-        if collections.Counter(actual) != collections.Counter(expected):
-            raise TestAssertionException(message)
+    if actual != expected:
+        raise TestAssertionException(message)
+
+
+def compare_lists_regardless_of_order(actual: List[str], expected: List[str]) -> bool:
+    assert isinstance(actual, list)
+    assert all(isinstance(item, str) for item in actual)
+    assert isinstance(expected, list)
+    assert all(isinstance(item, str) for item in expected)
+    return sorted(actual) == sorted(expected)
 
 
 def print_golem_message(message, indent = 4):
--- a/concent_api/api_testing_common.py
+++ b/concent_api/api_testing_common.py
@@ -1,4 +1,5 @@
 import argparse
+import collections
 import sys
 
 import uuid
@@ -66,10 +67,14 @@ class count_fails(object):
         print(f'Total failed tests : {cls.get_fails()} out of {cls.number_of_run_tests}')
 
 
-def assert_condition(actual, expected, error_message = None):
+def assert_condition(actual, expected, error_message = None, compare_lists_regardless_of_order: bool = False):
     message = error_message or f"Actual: {actual} != expected: {expected}"
-    if actual != expected:
-        raise TestAssertionException(message)
+    if not compare_lists_regardless_of_order:
+        if actual != expected:
+            raise TestAssertionException(message)
+    else:
+        if collections.Counter(actual) != collections.Counter(expected):
+            raise TestAssertionException(message)
 
 
 def print_golem_message(message, indent = 4):
--- a/concent_api/api_testing_common.py
+++ b/concent_api/api_testing_common.py
@@ -2,7 +2,7 @@ import argparse
 import sys
 
 import uuid
-from typing import List
+from typing import Iterable
 
 from freezegun import freeze_time
 
@@ -74,12 +74,9 @@ def assert_condition(actual, expected, error_message = None):
         raise TestAssertionException(message)
 
 
-def compare_lists_regardless_of_order(actual: List[str], expected: List[str]) -> bool:
-    assert isinstance(actual, list)
-    assert all(isinstance(item, str) for item in actual)
-    assert isinstance(expected, list)
-    assert all(isinstance(item, str) for item in expected)
-    return sorted(actual) == sorted(expected)
+def assert_iterables_content_is_equal(actual: Iterable[str], expected: Iterable[str]) -> None:
+    if not sorted(actual) == sorted(expected):
+        raise TestAssertionException(f'actual and expected lists are not equal. Actual {actual}. Expected {expected}')
 
 
 def print_golem_message(message, indent = 4):
--- a/concent_api/api_testing_common.py
+++ b/concent_api/api_testing_common.py
@@ -76,7 +76,7 @@ def assert_condition(actual, expected, error_message = None):
 
 def assert_iterables_content_is_equal(actual: Iterable[str], expected: Iterable[str]) -> None:
     if not sorted(actual) == sorted(expected):
-        raise TestAssertionException(f'actual and expected lists are not equal. Actual {actual}. Expected {expected}')
+        raise TestAssertionException(f'Content of iterables is not equal. Actual {actual}. Expected {expected}')
 
 
 def print_golem_message(message, indent = 4):
--- a/concent_api/conductor/tasks.py
+++ b/concent_api/conductor/tasks.py
@@ -76,7 +76,7 @@ def blender_verification_request(
     # If there are already UploadReports corresponding to some files, the app links them with the VerificationRequest
     # by setting the value of the foreign key in UploadReport.
     for path in [source_package_path, result_package_path]:
-        UploadReport.objects.filter(
+        UploadReport.objects.select_for_update().filter(
             path=path,
             verification_request=None,
         ).update(
--- a/concent_api/conductor/tasks.py
+++ b/concent_api/conductor/tasks.py
@@ -124,7 +124,7 @@ def upload_acknowledged(
     assert isinstance(subtask_id, str)
 
     try:
-        verification_request = VerificationRequest.objects.get(subtask_id=subtask_id)
+        verification_request = VerificationRequest.objects.select_for_update().get(subtask_id=subtask_id)
     except VerificationRequest.DoesNotExist:
         log_error_message(
             logger,
--- a/concent_api/conductor/views.py
+++ b/concent_api/conductor/views.py
@@ -1,5 +1,6 @@
 from logging import getLogger
 
+from django.db import transaction
 from django.db.models import Q
 from django.http.response import HttpResponse
 from django.views.decorators.csrf import csrf_exempt
@@ -19,12 +20,13 @@ logger = getLogger(__name__)
 @provides_concent_feature('conductor-urls')
 @require_POST
 @csrf_exempt
+@transaction.atomic(using='storage')
 def report_upload(_request, file_path):
 
     log_request_received(logger,  file_path, FileTransferToken.Operation.upload)
     # If there's a corresponding VerificationRequest, the load it and link it to UploadReport.
     try:
-        verification_request = VerificationRequest.objects.get(
+        verification_request = VerificationRequest.objects.select_for_update().get(
             Q(source_package_path=file_path) | Q(result_package_path=file_path)
         )
     except VerificationRequest.DoesNotExist:
--- a/concent_api/core/message_handlers.py
+++ b/concent_api/core/message_handlers.py
@@ -79,7 +79,7 @@ def handle_send_force_report_computed_task(client_message):
         task_to_compute,
     )
 
-    if Subtask.objects.filter(  # pylint: disable=no-member
+    if Subtask.objects.select_for_update().filter(  # pylint: disable=no-member
         subtask_id=task_to_compute.compute_task_def['subtask_id'],
     ).exists():
         raise Http400(
@@ -140,7 +140,7 @@ def handle_send_ack_report_computed_task(client_message):
 
     if get_current_utc_timestamp() <= task_to_compute.compute_task_def['deadline'] + settings.CONCENT_MESSAGING_TIME:
         try:
-            subtask = Subtask.objects.get(
+            subtask = Subtask.objects.select_for_update().get(
                 subtask_id = task_to_compute.compute_task_def['subtask_id'],
             )
         except Subtask.DoesNotExist:
@@ -269,7 +269,7 @@ def handle_send_reject_report_computed_task(client_message):
             )
 
     try:
-        subtask = Subtask.objects.get(
+        subtask = Subtask.objects.select_for_update().get(
             subtask_id = task_to_compute.compute_task_def['subtask_id'],
         )
     except Subtask.DoesNotExist:
@@ -389,6 +389,9 @@ def handle_send_force_get_task_result(client_message: message.concents.ForceGetT
         task_to_compute,
     )
 
+    if Subtask.objects.filter(subtask_id=task_to_compute.compute_task_def['subtask_id']).exists():  # pylint: disable=no-member
+        Subtask.objects.select_for_update().get(subtask_id=task_to_compute.compute_task_def['subtask_id'])  # pylint: disable=no-member
+
     if Subtask.objects.filter(  # pylint: disable=no-member
         subtask_id = task_to_compute.compute_task_def['subtask_id'],
         state      = Subtask.SubtaskState.FORCING_RESULT_TRANSFER.name,  # pylint: disable=no-member
@@ -463,6 +466,9 @@ def handle_send_force_subtask_results(client_message: message.concents.ForceSubt
     )
 
     current_time = get_current_utc_timestamp()
+
+    if Subtask.objects.filter(subtask_id=task_to_compute.compute_task_def['subtask_id']).exists():  # pylint: disable=no-member
+        Subtask.objects.select_for_update().get(subtask_id=task_to_compute.compute_task_def['subtask_id'])  # pylint: disable=no-member
 
     if Subtask.objects.filter(  # pylint: disable=no-member
         subtask_id=task_to_compute.compute_task_def['subtask_id'],
@@ -588,7 +594,7 @@ def handle_send_force_subtask_results_response(client_message):
         )
 
     try:
-        subtask = Subtask.objects.get(
+        subtask = Subtask.objects.select_for_update().get(
             subtask_id = task_to_compute.compute_task_def['subtask_id'],
         )
     except Subtask.DoesNotExist:
@@ -720,6 +726,12 @@ def handle_send_force_payment(
     oldest_payments_ts = min(
         subtask_results_accepted.payment_ts for subtask_results_accepted in client_message.subtask_results_accepted_list
     )
+
+    for subtask_results_accepted in client_message.subtask_results_accepted_list:
+        try:
+            Subtask.objects.select_for_update().get(subtask_id=subtask_results_accepted.subtask_id)
+        except Subtask.DoesNotExist:
+            pass
 
     # Concent gets list of transactions from payment API where timestamp >= T0.
     list_of_transactions = payments_service.get_list_of_payments(  # pylint: disable=no-value-for-parameter
@@ -1246,7 +1258,7 @@ def store_or_update_subtask(
     force_get_task_result: Optional[message.concents.ForceGetTaskResult] = None,
 ) -> Subtask:
     try:
-        subtask = Subtask.objects.get(
+        subtask = Subtask.objects.select_for_update().get(
             subtask_id = subtask_id,
         )
     except Subtask.DoesNotExist:
@@ -1353,6 +1365,9 @@ def handle_send_subtask_results_verify(
         return message.concents.ServiceRefused(
             reason=message.concents.ServiceRefused.REASON.InvalidRequest,
         )
+
+    if Subtask.objects.filter(subtask_id=compute_task_def['subtask_id']).exists():  # pylint: disable=no-member
+        Subtask.objects.select_for_update().get(subtask_id=compute_task_def['subtask_id'])  # pylint: disable=no-member
 
     if Subtask.objects.filter(  # pylint: disable=no-member
         subtask_id=compute_task_def['subtask_id'],
--- a/concent_api/core/message_handlers.py
+++ b/concent_api/core/message_handlers.py
@@ -79,7 +79,7 @@ def handle_send_force_report_computed_task(client_message):
         task_to_compute,
     )
 
-    if Subtask.objects.select_for_update().filter(  # pylint: disable=no-member
+    if Subtask.objects.filter(  # pylint: disable=no-member
         subtask_id=task_to_compute.compute_task_def['subtask_id'],
     ).exists():
         raise Http400(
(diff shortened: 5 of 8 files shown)


concent_api/api-integration-multi-threads-test.py
```diff
@@ -0,0 +1,214 @@
...
+    number_of_threads: int,
+    cluster_url: str,
+    golem_message: Message,
+) -> None:
+    for i in range(number_of_threads):
+        t = Thread(target=func, args=(cluster_url, golem_message,))
+        t.start()
+
+
+def get_ack_report_computed_task(report_computed_task: ReportComputedTask) -> AckReportComputedTask:
```
rwrzesien: There are functions for creating `AckReportComputedTask`, `ForceGetTaskResult` and `ForceReportComputedTask` in other tests, please reuse them.

concent_api/api-integration-multi-threads-test.py
```diff
@@ -0,0 +1,214 @@
...
+    responses_global.append(response.__class__.__name__)
+
+
+@count_fails
+def test_case_multiple_requests_concerning_one_subtask_will_be_processed_one_by_one_if_subtask_exists_in_database(
+    cluster_consts: str,
+    cluster_url: str,
+    task_id: str,
+    subtask_id: str
+):
```
rwrzesien: Please add return type here.

concent_api/api-integration-multi-threads-test.py
```diff
@@ -0,0 +1,214 @@
...
+        f" and test failed or maybe you changed number of testing threads and forgot to change expected_responses?",
+        compare_lists_regardless_of_order=True,
+    )
+    print('Single test passed successfully')
+
+
+if __name__ == '__main__':
+    try:
+        from concent_api.settings import CONCENT_PUBLIC_KEY
+
```
rwrzesien: Please remove blank line.

concent_api/core/message_handlers.py
```diff
@@ -389,6 +389,9 @@ def handle_send_force_get_task_result(client_message: message.concents.ForceGetT
         task_to_compute,
     )
 
+    if Subtask.objects.filter(subtask_id=task_to_compute.compute_task_def['subtask_id']).exists():  # pylint: disable=no-member
```
rwrzesien: The `Subtask` should not exist in this moment. If it does, Concent returns `ServiceRefused` (in the lines below) and this Subtask is not updated. So locking it here is not necessary.

concent_api/core/message_handlers.py
```diff
@@ -721,6 +727,12 @@ def handle_send_force_payment(
         subtask_results_accepted.payment_ts for subtask_results_accepted in client_message.subtask_results_accepted_list
     )
 
+    for subtask_results_accepted in client_message.subtask_results_accepted_list:
```
rwrzesien: There is no changes being done on those `Subtasks`, so locking them in this particular use case is not necessary.

concent_api/core/subtask_helpers.py
```diff
@@ -33,6 +33,15 @@ def update_timed_out_subtasks(
         next_deadline__lte      = timezone.now()
     )
 
+    for subtask in clients_subtask_list:  # create locks in database for all subtasks of client
+        try:
+            Subtask.objects.select_for_update().get(pk=subtask.pk)
```
rwrzesien: Why not locking them in the query above ?

concent_api/core/subtask_helpers.py
```diff
@@ -80,7 +89,7 @@ def update_timed_out_subtasks(
                 subtask             = subtask,
             )
         elif subtask.state == Subtask.SubtaskState.VERIFICATION_FILE_TRANSFER.name:  # pylint: disable=no-member
-            locked_subtask = Subtask.objects.select_for_update().get(pk=subtask.pk)
+            locked_subtask = Subtask.objects.get(pk=subtask.pk)
```
rwrzesien: Locking in the query at the beginning of this function should cover this and below cases.

concent_api/core/message_handlers.py
pawelkisielewicz: This is exactly the sense of this task. This line is like a barrier making all-except of one processes to stop. If there were e.g. two requests exactly in the same time concent should response with one ACK and one Refuse(DuplicateRequest). Without this concent would response with two ACKs, because during this validation below there will be not this subtask in database. This requests would have to be exactly in the same moment, probability is very, very low, but it is possible. 
Second thing is that in line below it is also checked state. I want to lock subtasks regardless of subtasks state.

rwrzesien: Makes sense, thanks for clarification.

concent_api/api-integration-multi-threads-test.py
pawelkisielewicz: I tried to reuse them, but they are different depending on test, actually each api test has it's own functions creating messages. Trying to reuse them i had problem with nested messages (sometimes they were created, sometimes given) and with signatures. Creating my own functions looked for me easiest and simplest.

concent_api/api-integration-multi-threads-test.py
pawelkisielewicz: done

concent_api/api-integration-multi-threads-test.py
pawelkisielewicz: done, sorry for this..

concent_api/core/message_handlers.py
pawelkisielewicz: Right.. I checked it again and I see only one possible problem- without locking it might be danger that other process will change something in this subtask during this flow.. And we are not changing anything is subtask, but we are storing pending_response with values basing on values in subtask.. If it is real problem and this problem is even possible- I'm really not sure. On the other hand, generally even if it is not necessary it 'costed' us not much and might prevent data corruption.. 
What do you think? I'm not sure.. Should I leave it or remove?

concent_api/core/subtask_helpers.py
pawelkisielewicz: I though't about this, but this query is already pretty complicated and I decided to do it like this, locking is more distinct. But I also hesitated in this place, if you think it would be better to do it in query I will

concent_api/core/subtask_helpers.py
pawelkisielewicz: Right, done. :+1:  for this

concent_api/api_testing_common.py
```diff
@@ -165,6 +165,8 @@ def _print_data(data, url):
     print()
     if response.status_code in [202, 204]:
         return None
+    elif response.status_code == 400:
+        return json.loads(response._content)
```
cameel: You should use `Content-Type` header to decide how to decode the response, not the error code. Responses with the same code can have different content in different situations.

concent_api/api-integration-multi-threads-test.py
```diff
@@ -0,0 +1,214 @@
...
+def call_function_in_treads(
+    func,
+    number_of_threads: int,
+    cluster_url: str,
+    golem_message: Message,
+) -> None:
+    for i in range(number_of_threads):
+        t = Thread(target=func, args=(cluster_url, golem_message,))
+        t.start()
```
cameel: Please don't use single-letter variable names outside of one-liners.

concent_api/api_testing_common.py
```diff
@@ -165,6 +165,8 @@ def _print_data(data, url):
...
     else:
         return try_to_decode_golem_message(private_key, public_key, response.content)
```
cameel: This is wrong. We shouldn't even try to decode the response as a Golem message if `Content-Type` is not `application/octet-stream`.

If you see something like this in other places, please fix it too.

concent_api/api_testing_common.py
```diff
@@ -165,6 +165,8 @@ def _print_data(data, url):
     print()
     if response.status_code in [202, 204]:
         return None
```
cameel: What about 200? Or 401? The whole 200..299 range should be treated as a success. 400..499 as a client error (so the test either has an error or is deliberately trying to cause a 4xx). 500..599 as a server-side bug that should be reported as an issue.

concent_api/core/message_handlers.py
```diff
@@ -79,7 +79,7 @@ def handle_send_force_report_computed_task(client_message):
         task_to_compute,
     )
 
-    if Subtask.objects.filter(  # pylint: disable=no-member
+    if Subtask.objects.select_for_update().filter(  # pylint: disable=no-member
         subtask_id=task_to_compute.compute_task_def['subtask_id'],
     ).exists():
```
cameel: This is not wrong but not really necessary. If it exists, there's not need to lock it because we exit immediately. If it does not, you won't be able to lock it anyway.

concent_api/core/message_handlers.py
```diff
@@ -389,6 +389,9 @@ def handle_send_force_get_task_result(client_message: message.concents.ForceGetT
         task_to_compute,
     )
 
+    if Subtask.objects.filter(subtask_id=task_to_compute.compute_task_def['subtask_id']).exists():  # pylint: disable=no-member
+        Subtask.objects.select_for_update().get(subtask_id=task_to_compute.compute_task_def['subtask_id'])  # pylint: disable=no-member
```
cameel: I'd replace this with `try: subtask = Subtask.objects.select_for_update().get(...) except DoesNotExist: ...` construct.

Why make two queries when you can get away with making just one.

concent_api/core/message_handlers.py
```diff
@@ -389,6 +389,9 @@ def handle_send_force_get_task_result(client_message: message.concents.ForceGetT
...
+
     if Subtask.objects.filter(  # pylint: disable=no-member
         subtask_id = task_to_compute.compute_task_def['subtask_id'],
         state      = Subtask.SubtaskState.FORCING_RESULT_TRANSFER.name,  # pylint: disable=no-member
```
cameel: You have already performed a query above and locked the subtask. This code can now just use the task you fetched instead of fetching it again. You could replace it with 
``` python
if subtask.state == Subtask.SubtaskState.FORCING_RESULT_TRANSFER.name:
```

BTW, don't we have a custom field that lets you drop the `.name` here?

concent_api/core/message_handlers.py
cameel: There's one case you did not take into account:
- You check if subtask exists. It does not.
- Another process creates the subtask.
- You call `store_or_update_subtask()`. It overwrites the subtask instead of returning `DuplicateRequest`.

To avoid this we have stop using `store_or_update_subtask()`. We need separate create and update functions.

concent_api/api-integration-multi-threads-test.py
rwrzesien: Some are reused and placed in `api_testing_common.py`, but if you need a different ones than its ok.

concent_api/api-integration-multi-threads-test.py
rwrzesien: No problem, it happens :)

concent_api/core/message_handlers.py
cameel: This handler is not supposed to even look at `Subtask` objects in the database. We never store anything submitted in this use case. Note that there's no corresponding state in the state machine. The pending response table has a special case for responses created here so that they do not need to refer to a subtask.

So @rwrzesien is right. This is not necessary and should be removed.

If you find anything in this use case that makes a query to the `Subtask` table, please report it as a bug.

concent_api/core/subtask_helpers.py
rwrzesien: It would be better to do this in single query. Good formatting will make it less complicated ;)

concent_api/core/message_handlers.py
```diff
@@ -1354,6 +1366,9 @@ def handle_send_subtask_results_verify(
             reason=message.concents.ServiceRefused.REASON.InvalidRequest,
         )
 
+    if Subtask.objects.filter(subtask_id=compute_task_def['subtask_id']).exists():  # pylint: disable=no-member
+        Subtask.objects.select_for_update().get(subtask_id=compute_task_def['subtask_id'])  # pylint: disable=no-member
+
```
cameel: You can replace these three queries with one. Same as in one of my comments above.

concent_api/core/message_handlers.py
```diff
@@ -1354,6 +1366,9 @@ def handle_send_subtask_results_verify(
...
     if Subtask.objects.filter(  # pylint: disable=no-member
```
cameel: It would be better if this function received `subtask` in a parameter instead of fetching it on its own.

As it is now, the lock is hidden in the validator and not well visible. I think the caller (view/handler) should be responsible for locking it before using the validator.

concent_api/core/message_handlers.py
rwrzesien: The `store_pending_message` in this case do not use anything from `Subtask` table, only from client message. Please check the call of it - `subtask` parameter is skipped there on purpose, because there can be many. So I think there is no need for locking this one, unless there is a reason I can't see now :)

concent_api/core/subtask_helpers.py
cameel: We want to minimize the number of queries as much as possible. And especially we don't want to make n queries to fetch n objects. We want only one query to fetch them all unless there are reasons preventing us from doing this (e.g. the need to commit a transaction - which is not the case here).

I agree with @rwrzesien, the lock should be in the query above, no matter how complex it is.

concent_api/core/message_handlers.py
```diff
@@ -464,6 +467,9 @@ def handle_send_force_subtask_results(client_message: message.concents.ForceSubt
 
     current_time = get_current_utc_timestamp()
 
+    if Subtask.objects.filter(subtask_id=task_to_compute.compute_task_def['subtask_id']).exists():  # pylint: disable=no-member
+        Subtask.objects.select_for_update().get(subtask_id=task_to_compute.compute_task_def['subtask_id'])  # pylint: disable=no-member
+
     if Subtask.objects.filter(  # pylint: disable=no-member
```
cameel: These three queries can be replaced with one.

concent_api/api-integration-multi-threads-test.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+if __name__ == '__main__':
+    try:
+        from concent_api.settings import CONCENT_PUBLIC_KEY
+
+        run_tests(globals())
+    except requests.exceptions.ConnectionError as exception:
+        print("\nERROR: Failed connect to the server.\n", file=sys.stderr)
+        sys.exit(str(exception))
```
cameel: I think that `exit(exception)` would work too.

concent_api/api-integration-multi-threads-test.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+
```
cameel: Why the extra empty lines?

concent_api/api_testing_common.py
```diff
@@ -66,10 +67,14 @@ def print_fails(cls):
...
-def assert_condition(actual, expected, error_message = None):
+def assert_condition(actual, expected, error_message = None, compare_lists_regardless_of_order: bool = False):
     message = error_message or f"Actual: {actual} != expected: {expected}"
-    if actual != expected:
-        raise TestAssertionException(message)
+    if not compare_lists_regardless_of_order:
+        if actual != expected:
+            raise TestAssertionException(message)
+    else:
+        if collections.Counter(actual) != collections.Counter(expected):
```
cameel: You're not checking if `actual` and `expected` are really lists or collections.

concent_api/api_testing_common.py
```diff
@@ -66,10 +67,14 @@ def print_fails(cls):
         print(f'Total failed tests : {cls.get_fails()} out of {cls.number_of_run_tests}')
 
 
-def assert_condition(actual, expected, error_message = None):
+def assert_condition(actual, expected, error_message = None, compare_lists_regardless_of_order: bool = False):
```
cameel: I think that it would be better to create a separate function for comparing lists.

concent_api/core/message_handlers.py
rwrzesien: Maybe it would be better to change it to `try/except DoesNotExists` flow.

concent_api/core/message_handlers.py
```diff
@@ -464,6 +467,9 @@ def handle_send_force_subtask_results(client_message: message.concents.ForceSubt
 
     current_time = get_current_utc_timestamp()
 
+    if Subtask.objects.filter(subtask_id=task_to_compute.compute_task_def['subtask_id']).exists():  # pylint: disable=no-member
```
rwrzesien: Maybe it would be better to change this and similar places to `try get()/except DoesNotExists` flow, then it would be only single database query.

concent_api/core/message_handlers.py
cameel: I'd personally just get the object first with `filter()` and then check the count. The `try ... except` version is very verbose. I'd leave exceptions for cases where they're really needed.

rwrzesien: So it would be still two queries, which makes version with `exists` better. Doing this with `try/except DoesNotExists` is a typical Django flow.

concent_api/core/message_handlers.py
pawelkisielewicz: You are right, but in this PR I assume that Subtask exist. Case when subtask does not exist i described in https://github.com/golemfactory/concent/issues/759. I don't think split of `store_or_update_subtask()` is necessary, my idea to solve this problem is in my issue. What do you think?

concent_api/core/message_handlers.py
pawelkisielewicz: done, thanks for good explanation.

concent_api/core/subtask_helpers.py
pawelkisielewicz: Agree, done :)

concent_api/core/message_handlers.py
pawelkisielewicz: Yes, but as @cameel wrote in first comment- it is not necessary. If subtask exist flow ends with duplicate request. If not it will not be locked anyway. I simply deleted `select_for_update` in this place ane left it as at the beginning

concent_api/core/message_handlers.py

concent_api/core/message_handlers.py
pawelkisielewicz: right, after small refactor changed. And there is possibility to drop `.name` by comparing `enums` instead of `strings`

concent_api/api-integration-multi-threads-test.py
```diff
@@ -0,0 +1,214 @@
...
```
kbeker: Delete this blank line

concent_api/api-integration-multi-threads-test.py
```diff
@@ -0,0 +1,214 @@
...
+from api_testing_common import run_tests
+from api_testing_common import timestamp_to_isoformat
+from common.helpers import get_current_utc_timestamp
+from common.helpers import sign_message
+
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "concent_api.settings")
+
+REPORT_COMPUTED_TASK_SIZE = 10
+
+NUMBER_OF_TESTING_THREADS = 3  # changing this variable remember to correct number of expected responses
```
kbeker: Please add assert to this statement

concent_api/api-integration-multi-threads-test.py
```diff
@@ -0,0 +1,214 @@
...
+    golem_message: Message,
+) -> None:
+    for i in range(number_of_threads):
+        t = Thread(target=func, args=(cluster_url, golem_message,))
+        t.start()
+
+
+def get_ack_report_computed_task(report_computed_task: ReportComputedTask) -> AckReportComputedTask:
+    ack_report_computed_task = AckReportComputedTask(report_computed_task=report_computed_task)
+    return ack_report_computed_task
```
kbeker: If you are creating messages this way please do this in one line:
```python
return AckReportComputedTask(report_computed_task=report_computed_task)
```

concent_api/api-integration-multi-threads-test.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+def get_force_get_task_result(report_computed_task: ReportComputedTask) -> ForceGetTaskResult:
+    force_get_task_result = message.concents.ForceGetTaskResult(report_computed_task=report_computed_task)
```
kbeker: Same here

concent_api/api-integration-multi-threads-test.py
```diff
@@ -0,0 +1,214 @@
...
+    return force_get_task_result
+
+
+def get_force_report_computed_task(report_computed_task: ReportComputedTask) -> ForceReportComputedTask:
+    force_report_computed_task = ForceReportComputedTask()
+    force_report_computed_task.report_computed_task = report_computed_task
+    return force_report_computed_task
```
kbeker: Please change it to:
```python
return ForceReportComputedTask(report_computed_task=report_computed_task)
```

concent_api/api-integration-multi-threads-test.py
```diff
@@ -0,0 +1,214 @@
...
+    task_id: str,
+    subtask_id: str
+) -> ReportComputedTask:
+    current_time = get_current_utc_timestamp()
+
+    task_to_compute = create_signed_task_to_compute(
+        timestamp=timestamp_to_isoformat(current_time),
+        task_id=task_id,
+        subtask_id=subtask_id,
+        deadline=(current_time + 100),
```
kbeker: Maybe it would be good to use some concent_constans here, because if some of constans will change this test might won't pass.

concent_api/api-integration-multi-threads-test.py
```diff
@@ -0,0 +1,214 @@
...
+
+    clear_responses()
+    """this is test- sending some messages in one time """
+    call_function_in_treads(
+        func=send_correct_force_get_task_result,  # Subtask state changed to: FORCING_RESULT_TRANSFER
+        number_of_threads=NUMBER_OF_TESTING_THREADS,
+        cluster_url=cluster_url,
+        golem_message=report_computed_task,
+    )
+    "waiting for responses from all threads"
```
kbeker: Please change it to `#  waiting for responses from all threads`

concent_api/api-integration-multi-threads-test.py
```diff
@@ -0,0 +1,214 @@
...
+        expected_status=202,
+    )  # Subtask state changed to: FORCING_REPORT
+    send_correct_ack_report_computed_task(
+        cluster_url=cluster_url,
+        report_computed_task=report_computed_task,
+        expected_status=202,
+    )  # Subtask state changed to: REPORTED
+
+    clear_responses()
+    """this is test- sending some messages in one time """
```
kbeker: Please change it to: `#  this is test- sending some messages in one time`

concent_api/core/message_handlers.py

concent_api/core/message_handlers.py
pawelkisielewicz: I really don't understand.. :/ I think i did it like this at the beginning and you asked to change it to create as small atomic blocks as possible. The difference was exactly about this validation, I think you asked they should be before atomic and lock.. Or am I wrong?

concent_api/core/message_handlers.py

concent_api/core/message_handlers.py
pawelkisielewicz: right, done

concent_api/api-integration-multi-threads-test.py
pawelkisielewicz: done

concent_api/api-integration-multi-threads-test.py
pawelkisielewicz: mistake.. fixed

concent_api/api_testing_common.py
pawelkisielewicz: created separate function with this checks

concent_api/api_testing_common.py
pawelkisielewicz: as above - created separate function with this check

concent_api/api-integration-multi-threads-test.py
pawelkisielewicz: done

concent_api/core/message_handlers.py
cameel: Your issue describes a different problem.

#759 describes this case:
1. Client 1 submits message x with `subtask_id = A`.
2. Client 2 submits message y with `subtask_id = A`.
3. Process 1 checks if subtask with `subtask_id = A` exists. It does not.
4. Process 2 checks if subtask with `subtask_id = A` exists. It does not.
5. Process 1 prepares subtask X with `subtask_id = A` and data from message x.
6. Process 2 prepares subtask Y with `subtask_id = A` and data from message y,
7. Process 1 tries to save subtask X and succeeds.
8. Process 2 tries to save subtask Y and `IntegrityError` is raised.
9. Result
    - Client 1 gets HTTP 2xx, client 2 gets HTTP 5xx. Only one client succeeded.
    - Client 1 may have triggered irreversible actions (like adding tasks to queue or making payments). Client 2 has not triggered any irreversible actions.
    - Data from message x is in the database. Data from message y is not. One subtask has been created.
    - Client 2 can safely retry by sending message y again.

Here we have something different:
1. Client 1 submits message x with `subtask_id = A`.
2. Client 2 submits message y with `subtask_id = A`.
3. Process 1 checks if subtask with `subtask_id = A` exists. It does not.
4. Process 2 checks if subtask with `subtask_id = A` exists. It does not.
5. Process 1 prepares subtask X with `subtask_id = A` and data from message x.
6. Process 2 prepares subtask Y with `subtask_id = A` and data from message y,
7. Process 1 calls  `store_or_update_subtask()`
    1. `store_or_update_subtask()` checks if subtask with `subtask_id = A` exists. It does not.
    2. `store_or_update_subtask()` tries to save subtask X and succeeds.
8. Process 2 calls  `store_or_update_subtask()`
    1. `store_or_update_subtask()` checks if subtask with `subtask_id = A` exists. It does.
    2. `store_or_update_subtask()` updates subtask X with data from subtask Y and succeeds.
9. Result
    - Both clients get HTTP 2xx. Both succeeded.
    - Both clients may have triggered Irreversible actions for the same subtask.
    - Data from message x has been overwritten by data from message y. One subtask has been created but irreversible actions were executed twice.

concent_api/api-integration-multi-threads-test.py
pawelkisielewicz: It is not good idea. At the end of test I compare list of responses from views with `expected list`- `expected list` is hardcoded in test. E.g. if you are changing number of threads from 3 to 2, you should change `expected_responses = ['AckForceGetTaskResult', 'ServiceRefused', 'ServiceRefused']` to `expected_responses = ['AckForceGetTaskResult', 'ServiceRefused']`. 

Second thing is, if something gone wrong (test failed), number of responses may be incorrect. But it is not simple assert for numbers, it has to be validated. I added in error_message in this validation comment, that maybe test failed and maybe someone simple forgot to change number of threads    

This comment is to reme

concent_api/api-integration-multi-threads-test.py
pawelkisielewicz: done

concent_api/api-integration-multi-threads-test.py
pawelkisielewicz: done

concent_api/api-integration-multi-threads-test.py
pawelkisielewicz: done

concent_api/api-integration-multi-threads-test.py
pawelkisielewicz: done

concent_api/api-integration-multi-threads-test.py
pawelkisielewicz: done

concent_api/core/message_handlers.py
cameel: I have created a separate task for this so that here you can focus on `select_for_update()`: #770.

concent_api/api_testing_common.py
cameel: BTW, you can do it as a separate pull request. I'm sure we have this problem in many more places and it would be a good idea to fix them all.

concent_api/core/message_handlers.py
cameel: > Doing this with try/except `DoesNotExists` is a typical Django flow.

Honestly, I often see people arguing for either as more typical or better. They're not always equivalent but when they are I think it's mostly a personal preference. I prefer not using exceptions in this particular case but I don't want to force my preference here so I'll just explain why I prefer it: the `try ... except` really stands out to me as something "heavy" and more complex. A version with `exists()` seems plain and ordinary. When you start catching stuff it looks like you're handing an exceptional case and such cases often need closer review. If it's a simple existence check with an `if`, I know that there isn't much that can go wrong and I can simply move on without focusing on it.

It's not wrong to do it with an exception but it just seems like an overkill to me.

concent_api/core/message_handlers.py
cameel: Yeah. I thought we refactored it in other places already. After you're done with `select_for_update()` you could make a commit that removes those unnecessary `.name`s. We want to use enums in comparisons if possible.

concent_api/api-integration-multi-threads-test.py
cameel: "This is **a** test".

BTW, why is this docstring not at the beginning of a function? Is it just a normal comment?

concent_api/core/message_handlers.py
rwrzesien: Right, but it is still two queries to database against one, which in applications which use database a lot can have serious impact (if those double queries vs one is repeated several times). We can always switch to:
```
subtask = Subtask.objects.filter(subtask_id=value).first()
if subtask is not None:
    do_stuff()
```

Which is single query and doesn't involve exceptions BUT doesn't raise exception when more than 1 object is returned (here `subtask_id` is unique so it won't happen, but its easy to forget that)

Actually the `try/except` is suggested in official Django documentation.

cameel: That's why I never use `first()`. I'd write it as:

``` python
subtasks = Subtask.objects.filter(subtask_id=value)
assert len(subtasks) <= 1
if len(subtasks) == 1:
    do_stuff()
```

It's just one query and you also get the check correctly as an `assert` (because it's not necessary once the code has been debugged).

I usually create helper like this:
``` python
def get_one_or_none(query):
    assert len(query) <= 1
    if len(query) == 0:
        return None
    return query[0]
```

or:
``` python
def get_one_or_none(model, filter_condition):
    query = model.objects.filter(filter_condition)
    assert len(query) <= 1
    if len(query) == 0:
        return None
    return query[0]
```

which reduces the code to:
``` python
if get_one_or_none(Subtask, subtask_id=value) is not None:
    do_stuff()
```

and is nicer and more concise than:
``` python
try:
    subtask = Subtask.objects.get(subtask_id=value)
except Subtask.DoesNotExist:
    pass
else:
    do_stuff()
```

You could put the version with an exception into a function too - and maybe that would be the best compromise because it would still use exceptions but would hide the `try ... except` block.

concent_api/api-integration-multi-threads-test.py
cameel: @pawelkisielewicz 

> It is not good idea. At the end of test I compare list of responses from views with `expected list`- `expected list` is hardcoded in test. E.g. if you are changing number of threads from 3 to 2, you should change `expected_responses = ['AckForceGetTaskResult', 'ServiceRefused', 'ServiceRefused']` to `expected_responses = ['AckForceGetTaskResult', 'ServiceRefused']`. 

But that's exactly why you need to use `assert`. It does not need to be here. You can put it in the test. If you add
``` python
assert len(expected_responses) == NUMBER_OF_TESTING_THREADS
```
you won't have to worry about someone changing the constant and not updating the list - python will be able to detect it and also anyone reading the code will see that the list length should match the number of threads. It serves both as documentation and as a method to enforce it.

> Second thing is, if something gone wrong (test failed), number of responses may be incorrect. But it is not simple assert for numbers, it has to be validated. I added in error_message in this validation comment, that maybe test failed and maybe someone simple forgot to change number of threads

Looks like you have just discovered why we need `assert` in tests and why it's different from test assertions ;)

I have already explained this in the past. Please see this comment: https://github.com/golemfactory/concent/pull/673#discussion_r205289860

Simply put:
- Use `assert` to check if the test itself is correct and it it makes sense to even launch it (i.e. if the list contains expected value for every thread).
- Use test assertions (`assertEqual`, etc.) to check if the results returned by the code you're testing are correct (i.e. if the threads returned correct values).

concent_api/core/message_handlers.py
pawelkisielewicz: Ok, that is 'other problem' and be resolved in separate PR based on your issue

concent_api/api_testing_common.py
pawelkisielewicz: resolved in https://github.com/golemfactory/concent/pull/773

concent_api/api_testing_common.py
pawelkisielewicz: resolved in https://github.com/golemfactory/concent/pull/773

concent_api/api_testing_common.py
pawelkisielewicz: resolved in https://github.com/golemfactory/concent/pull/773

concent_api/api-integration-multi-threads-test.py
pawelkisielewicz: Thanks, done :)

concent_api/api-integration-multi-threads-test.py
rwrzesien: Actually this should stay, because it separates Python stdlib imports from 3rd party imports.

concent_api/api-integration-multi-threads-test.py
```diff
@@ -0,0 +1,201 @@
...
+    number_of_threads: int,
+    cluster_url: str,
+    golem_message: Message,
+) -> None:
+    for i in range(number_of_threads):
+        thread = Thread(target=func, args=(cluster_url, golem_message,))
+        thread.start()
+
+
+def get_ack_report_computed_task(report_computed_task: ReportComputedTask) -> AckReportComputedTask:
```
rwrzesien: Why don't you use factories for this ? Each of those is used only once and does nothing extra except from passing arguments.

concent_api/api_testing_common.py
```diff
@@ -72,6 +74,14 @@ def assert_condition(actual, expected, error_message = None):
         raise TestAssertionException(message)
 
 
+def compare_lists_regardless_of_order(actual: List[str], expected: List[str]) -> bool:
+    assert isinstance(actual, List)
```
rwrzesien: Should it be `list` ? Isn't `List` only a typing ?

concent_api/api_testing_common.py
```diff
@@ -72,6 +74,14 @@ def assert_condition(actual, expected, error_message = None):
...
+    assert all(isinstance(item, str) for item in actual)
+    assert isinstance(expected, List)
```
rwrzesien: Same here.

concent_api/core/message_handlers.py
```diff
@@ -389,13 +389,14 @@ def handle_send_force_get_task_result(client_message: message.concents.ForceGetT
...
 
-    if Subtask.objects.filter(  # pylint: disable=no-member
-        subtask_id = task_to_compute.compute_task_def['subtask_id'],
-        state      = Subtask.SubtaskState.FORCING_RESULT_TRANSFER.name,  # pylint: disable=no-member
-    ).exists():
-        return message.concents.ServiceRefused(
-            reason = message.concents.ServiceRefused.REASON.DuplicateRequest,
-        )
+    try:
+        subtask=Subtask.objects.select_for_update().get(subtask_id=task_to_compute.compute_task_def['subtask_id'])
```
rwrzesien: Please add spaces around `subtask=Subtask`.

concent_api/core/transfer_operations.py
```diff
@@ -39,15 +39,15 @@ def verify_file_status(
...
         state                  = Subtask.SubtaskState.FORCING_RESULT_TRANSFER.name,  # pylint: disable=no-member
     )
 
-    for get_task_result in force_get_task_result_list:
-        report_computed_task    = deserialize_message(get_task_result.report_computed_task.data.tobytes())
+    for subtask in subtasks_list:
+        report_computed_task    = deserialize_message(subtask.report_computed_task.data.tobytes())
         if request_upload_status(report_computed_task):
-            subtask               = get_task_result
+            subtask               = subtask
```
rwrzesien: Is this necessary ?

concent_api/api-integration-multi-threads-test.py
```diff
@@ -0,0 +1,201 @@
...
+@count_fails
+def test_case_multiple_requests_concerning_one_subtask_will_be_processed_one_by_one_if_subtask_exists_in_database(
+    cluster_consts: str,
+    cluster_url: str,
+    task_id: str,
+    subtask_id: str
+) -> None:
+    report_computed_task = create_report_computed_task(task_id=task_id, subtask_id=subtask_id)
+
+    """send messages to save subtask in database and change state to REPORTED. Next message ForceGetTaskResult
```
rwrzesien: Please change to `#` comment

concent_api/api-integration-multi-threads-test.py
```diff
@@ -0,0 +1,201 @@
...
+
+    clear_responses()
+    # this is test- sending some messages in one time
+    call_function_in_treads(
+        func=send_correct_force_get_task_result,  # Subtask state changed to: FORCING_RESULT_TRANSFER
+        number_of_threads=NUMBER_OF_TESTING_THREADS,
+        cluster_url=cluster_url,
+        golem_message=report_computed_task,
+    )
+    # waiting for responses from all threads"""
```
rwrzesien: Remove `"""`.

concent_api/api-integration-multi-threads-test.py
pawelkisielewicz: I spoke with @kbeker about this and he said, it will be problem with keys. I don't remember exactly details, but this is the reason. I use factory at the beginning to create `report_computed_task`, and all next messages are like box for previous

rwrzesien: Ok, I thought it is safe to use those factories in general.

@kbeker Can you describe the problem with keys ?

concent_api/api-integration-multi-threads-test.py
pawelkisielewicz: this line was at the beginning of file. But I found place with this separate and add line

concent_api/api_testing_common.py
pawelkisielewicz: done

concent_api/api_testing_common.py
pawelkisielewicz: done

concent_api/core/message_handlers.py
pawelkisielewicz: good eyes :) done

concent_api/core/transfer_operations.py
pawelkisielewicz: I think it is. It is not directly the task, but we have in our code variables called `get_task_result`, and they were really 'subtasks'. So I changed the names of variables to corresponding with objects they are representing

concent_api/api-integration-multi-threads-test.py
pawelkisielewicz: done

concent_api/api-integration-multi-threads-test.py
pawelkisielewicz: done

concent_api/core/transfer_operations.py
rwrzesien: But why you are assigning `subtask` to `subtask` ?

concent_api/core/message_handlers.py
rwrzesien: Rather good glasses ;)

concent_api/core/transfer_operations.py
pawelkisielewicz: right, it is not necessary. I cleaned it but didn't finished. Sorry :/ 
Done :)

concent_api/api-integration-multi-threads-test.py
kbeker: This should be secured from this. Deadline should not be responsible for test fails. If Concent change its constants then this test also fail, so please add here some CONCENT_MESSAGING_TIME :)

concent_api/api-integration-multi-threads-test.py
```diff
@@ -0,0 +1,201 @@
...
+    number_of_threads: int,
+    cluster_url: str,
+    golem_message: Message,
+) -> None:
+    for i in range(number_of_threads):
+        thread = Thread(target=func, args=(cluster_url, golem_message,))
+        thread.start()
+
+
+def get_ack_report_computed_task(report_computed_task: ReportComputedTask) -> AckReportComputedTask:
```
dybi: please rename it to `create_...`

concent_api/api-integration-multi-threads-test.py
```diff
@@ -0,0 +1,201 @@
...
+    return AckReportComputedTask(report_computed_task=report_computed_task)
+
+
+def get_force_get_task_result(report_computed_task: ReportComputedTask) -> ForceGetTaskResult:
```
dybi: please rename it to `create_...`

concent_api/api-integration-multi-threads-test.py
```diff
@@ -0,0 +1,201 @@
...
+    return message.concents.ForceGetTaskResult(report_computed_task=report_computed_task)
+
+
+def get_force_report_computed_task(report_computed_task: ReportComputedTask) -> ForceReportComputedTask:
```
dybi: please rename it to `create_...`

concent_api/api-integration-multi-threads-test.py
```diff
@@ -0,0 +1,201 @@
...
+        ),
+        headers={
+            'Content-Type': 'application/octet-stream',
+        },
+        expected_status=expected_status,
+    )
+    responses_global.append(response['error_code'] if isinstance(response, dict) else response)
+
+
+def send_correct_ack_report_computed_task(
```
dybi: why not adding response to responses list in this case?

concent_api/api-integration-multi-threads-test.py
```diff
@@ -0,0 +1,201 @@
...
+        time.sleep(0.1)
+        if time.time() >= end_time:
+            break
+
+    expected_responses = ['AckForceGetTaskResult', 'ServiceRefused', 'ServiceRefused']
+    assert len(expected_responses) == NUMBER_OF_TESTING_THREADS,'Did you changed number of testing threads and forgot to change expected_responses?'
+
+    print('Responses = ' + str(responses_global))
+    error_message = f"Responses should be: {expected_responses}."
+    assert compare_lists_regardless_of_order(actual=responses_global, expected=expected_responses), error_message
```
dybi: this is not OK. I think `assert_condition` should be used here (from `commons`). We cannot use `assert` statement to both
- check if assumptions are correct (sanity check)
- check the result of the test case.
For the latter you have `assertEqual` and similar methods in unit case. Here you don't inherit from `unittest.TestCase` (which is fine, because these are not unit test, but API tests), so a different mechanism has been introduced - `assert_condition` and it should be used in such cases.

concent_api/api_testing_common.py
```diff
@@ -72,6 +74,14 @@ def assert_condition(actual, expected, error_message = None):
         raise TestAssertionException(message)
 
 
+def compare_lists_regardless_of_order(actual: List[str], expected: List[str]) -> bool:
```
dybi: do you really need all those assertions?

concent_api/api-integration-multi-threads-test.py
kbeker: @rwrzesien 
Problem appears when we use `ReportComputedTask` factories. The point is that it gives to `provider_public_key` and `requestor_public_key` random keys and we don't have `private_key` for message serialization.
I checked and there is one work around for keys problems, when all keys are overriden in `ReportComputedTaskFactory` constructor.

Also, I think that if you use `AckReportComputedTask` factory here it might not be a problem because you have to override `report_computed_task` to have same parameters as in other messages.

concent_api/api-integration-multi-threads-test.py
pawelkisielewicz: done

concent_api/api-integration-multi-threads-test.py
pawelkisielewicz: done

concent_api/api-integration-multi-threads-test.py
pawelkisielewicz: this function is only to change the state before 'real' test. It would be not used in this place.

concent_api/api-integration-multi-threads-test.py
kbeker: That is strange. When I have refreshed page it showed me different line from this which I have marked.

concent_api/api_testing_common.py
pawelkisielewicz: I know you hate assertions :D. In this case... I think them useful, they are 'sanity check'. In this place I expect list of strings and I check if it is. I think this asserts will never fail, but in this case they can save lot of debugging time. I would leave them here, it is test-util, but if you think they should be deleted I will :)

concent_api/api-integration-multi-threads-test.py
pawelkisielewicz: thanks, done :)

concent_api/api-integration-multi-threads-test.py
pawelkisielewicz: done

concent_api/api-integration-multi-threads-test.py
pawelkisielewicz: Ok, this is correct way.

But in this case I think it is easier and simpler to create 2 lines function:
> `def create_ack_report_computed_task(report_computed_task: ReportComputedTask) -> AckReportComputedTask:`
> `   return AckReportComputedTask(report_computed_task=report_computed_task)`

then use factory and change some fields. Factories are useful, I used them to create `ReportComputedTask`. But creating other messages is easier without factories. I would leave it as it is now..

concent_api/api-integration-multi-threads-test.py
dybi: @pawelkisielewicz, even if so, your function should use the exact mechanism as `assert_condition`, meaning: it should raise `TestAssertionException`, because it is caught in `count_fails` wrapper.

concent_api/api_testing_common.py
dybi: I think this is an overkill - you have simple function, which's core is basically one line. Piling so many asserts makes it harder to get. I really see no point in adding theme here.

pawelkisielewicz: done, deleted

concent_api/api_testing_common.py
```diff
@@ -72,6 +74,11 @@ def assert_condition(actual, expected, error_message = None):
         raise TestAssertionException(message)
 
 
+def assert_iterables_content_is_equal(actual: Iterable[str], expected: Iterable[str]) -> None:
+    if not sorted(actual) == sorted(expected):
+        raise TestAssertionException(f'actual and expected lists are not equal. Actual {actual}. Expected {expected}')
```
dybi: Please reformat the message:
`f' Content of iterables is not equal: expected={expected}, actual={actual} '`

concent_api/core/message_handlers.py
```diff
@@ -389,13 +389,14 @@ def handle_send_force_get_task_result(client_message: message.concents.ForceGetT
...
     )
 
-    if Subtask.objects.filter(  # pylint: disable=no-member
-        subtask_id = task_to_compute.compute_task_def['subtask_id'],
-        state      = Subtask.SubtaskState.FORCING_RESULT_TRANSFER.name,  # pylint: disable=no-member
-    ).exists():
-        return message.concents.ServiceRefused(
-            reason = message.concents.ServiceRefused.REASON.DuplicateRequest,
-        )
+    try:
```
dybi: why `try ... except` instead of filtering?

concent_api/core/message_handlers.py
```diff
@@ -464,13 +465,14 @@ def handle_send_force_subtask_results(client_message: message.concents.ForceSubt
...
     current_time = get_current_utc_timestamp()
 
-    if Subtask.objects.filter(  # pylint: disable=no-member
-        subtask_id=task_to_compute.compute_task_def['subtask_id'],
-        state=Subtask.SubtaskState.FORCING_ACCEPTANCE.name,  # pylint: disable=no-member
-    ).exists():
-        return message.concents.ServiceRefused(
-            reason=message.concents.ServiceRefused.REASON.DuplicateRequest,
-        )
+    try:
```
dybi: same here - why?

concent_api/core/message_handlers.py
```diff
@@ -1354,28 +1356,28 @@ def handle_send_subtask_results_verify(
...
-        subtask_id=compute_task_def['subtask_id'],
-        state__in=[
-            Subtask.SubtaskState.VERIFICATION_FILE_TRANSFER.name,  # pylint: disable=no-member
-            Subtask.SubtaskState.ADDITIONAL_VERIFICATION.name,     # pylint: disable=no-member
-        ]
-    ).exists():
-        return message.concents.ServiceRefused(
-            reason=message.concents.ServiceRefused.REASON.DuplicateRequest,
-        )
+    try:
```
dybi: same here

concent_api/api_testing_common.py
pawelkisielewicz: done

concent_api/core/message_handlers.py
pawelkisielewicz: It is to reduce queries to database. I only one ask database for subtask, and then I check it's state- but I check local variable, not database. This is not perfect solution, it will be changed in https://github.com/golemfactory/concent/pull/778 anyway.

concent_api/core/message_handlers.py
pawelkisielewicz: It is to reduce queries to database. I only one ask database for subtask, and then I check it's state- but I check local variable, not database. This is not perfect solution, it will be changed in https://github.com/golemfactory/concent/pull/778 anyway.

concent_api/core/message_handlers.py
pawelkisielewicz: It is to reduce queries to database. I only one ask database for subtask, and then I check it's state- but I check local variable, not database. This is not perfect solution, it will be changed in https://github.com/golemfactory/concent/pull/778 anyway.

concent_api/core/subtask_helpers.py
```diff
@@ -111,20 +109,20 @@ def update_timed_out_subtasks(
...
-                subtask             = locked_subtask,
+                subtask             = subtask,
             )
             store_pending_message(
                 response_type       = PendingResponse.ResponseType.SubtaskResultsSettled,
-                client_public_key   = locked_subtask.requestor.public_key_bytes,
+                client_public_key   = subtask.requestor.public_key_bytes,
                 queue               = PendingResponse.Queue.ReceiveOutOfBand,
-                subtask             = locked_subtask,
+                subtask             = subtask,
```
kbeker: You can use here our new style also.

concent_api/api_testing_common.py
```diff
@@ -72,6 +74,11 @@ def assert_condition(actual, expected, error_message = None):
         raise TestAssertionException(message)
 
 
+def assert_iterables_content_is_equal(actual: Iterable[str], expected: Iterable[str]) -> None:
+    if not sorted(actual) == sorted(expected):
+        raise TestAssertionException(f'Content of iterables is not equal. Actual {actual}. Expected {expected}')
```
dybi: maybe add at least **:** - `"Actual:.."` and `"Expected :..."`

concent_api/core/message_handlers.py
dybi: actually here the number of requests to DB has not been changed

concent_api/core/message_handlers.py
dybi: here neither

concent_api/core/message_handlers.py
dybi: ok, here it has been reduced by one ;)

concent_api/api_testing_common.py
pawelkisielewicz: done

concent_api/api-integration-multi-threads-test.py
```diff
@@ -0,0 +1,202 @@
...
+
+REPORT_COMPUTED_TASK_SIZE = 10
+
+NUMBER_OF_TESTING_THREADS = 3  # changing this variable remember to correct number of expected responses
+MAXIMUM_WAITING_TIME_FOR_ALL_RESPONSES = 3  # seconds
+
+responses_global = []  # type: list
+
+
+def clear_responses():
```
rwrzesien: Please add typing.

Jakub89:
@dybi 
> I approve, but some small changes are still needed.
Generally, I have preferred the approach with filtering instead of `try ... except DoesNotExist: pass` - for me it was more readable - but I can accept it if other reviewers are OK with it.

Sure, if `try/except` block can be avoided, we should do it. But is it okey to write QuerySet to variable for later comparison and modifications?

rwrzesien:
> Sure, if try/except block can be avoided, we should do it. But is it okey to write QuerySet to variable for later comparison and modifications?

Please remember that this is common way of doing this in Django.

cameel:
> Generally, I have preferred the approach with filtering instead of try ... except DoesNotExist: pass - for me it was more readable - but I can accept it if other reviewers are OK with it.

@dybi @rwrzesien 
What do you think about `get_one_or_none()` as described here: https://github.com/golemfactory/concent/pull/764#discussion_r214070540? It can be implemented by catching the exception inside. Is it a good middle ground?

And I mean in geneal, not necessarily in this particular pull request. In the code that has already been written I'm okay with either approach as long as it works.

concent_api/api-integration-multi-threads-test.py
cameel: Just as @dybi says, both functions should use the same mechanism - i.e. raise the same exceptions on failure.

I did not mean that it was wrong to use `assert_condition()` here but simply that you started adding new, unrelated functionality to it. It's better to have a set of assert functions, each with clearly defined purpose (exact comparison, comparison that ignores order, etc) if the implementations are completely different rather than one, monster function that tries to support every corner case.

concent_api/api-integration-multi-threads-test.py
cameel: That's because of how Github stores code comments. It does not simply store file name and line number. It stores the whole patch you see above. That's also how it can detect that a comment is "outdated" - it sees that the diff it has is no longer present in the code. That way it can still find the right place in the code even if you move it without fixing the problem.

The downside of this approach is that when the diff is trivial, like above (the patch consists of a single empty line being added), the diff might match some other place in the code. So when @pawelkisielewicz has removed the empty line Github simply found another empty line and attached the comment there :)

rwrzesien:
> What do you think about get_one_or_none() as described here: #764 (comment)? It can be implemented by catching the exception inside. Is it a good middle ground?

Yes, that is what I have suggested @pawelkisielewicz to do in https://github.com/golemfactory/concent/pull/778#discussion_r215277045

I think that what Django developers have in minds when introducing this way of getting object is:
1. If you want only to check if object exist, use `.exists()`.
2. If you want an instance, then you most probably want some flow when it does exist and when it is not. So why not doing this in two steps, which are 'exist' part in `try` block, `does not exist` part in except block, instead of doing it in three steps, which are get, check if exits and do something, so something else if it doesn't.

concent_api/api-integration-multi-threads-test.py
pawelkisielewicz: done

pawelkisielewicz:
this function `get_one_or_none()` is not implemented in this pull request. It is implemented in https://github.com/golemfactory/concent/pull/778 and suggestion concerning it's implementations should be there. (BTW it is changed as you asked :) ). Do we have any more blockers to merge **this** pull request?

concent_api/core/message_handlers.py
pawelkisielewicz: maybe.. this will be resolved in https://github.com/golemfactory/concent/pull/778. In this place it is unified to be as it is in `handle_send_subtask_results_verify`

rwrzesien:
@pawelkisielewicz 
> Do we have any more blockers to merge this pull request?

Not from me, but this was just aside discussion I think.

concent_api/core/message_handlers.py
```diff
@@ -389,13 +389,14 @@ def handle_send_force_get_task_result(client_message: message.concents.ForceGetT
...
-    ).exists():
-        return message.concents.ServiceRefused(
-            reason = message.concents.ServiceRefused.REASON.DuplicateRequest,
-        )
+    try:
+        subtask = Subtask.objects.select_for_update().get(subtask_id=task_to_compute.compute_task_def['subtask_id'])
+        if subtask.state_enum == Subtask.SubtaskState.FORCING_RESULT_TRANSFER:
+            return message.concents.ServiceRefused(
+                reason=message.concents.ServiceRefused.REASON.DuplicateRequest,
+            )
```
cameel: You should try to keep the `try ... except` part as small as possible to avoid catching unrelated exceptions.

This `if` should be in the `else:` clause:
``` python
try:
    # Get object
except ...:
    # Do something if it does not exist
else:
    # Do something with it if it exists
```

concent_api/core/message_handlers.py
```diff
@@ -464,13 +465,14 @@ def handle_send_force_subtask_results(client_message: message.concents.ForceSubt
...
-    ).exists():
-        return message.concents.ServiceRefused(
-            reason=message.concents.ServiceRefused.REASON.DuplicateRequest,
-        )
+    try:
+        subtask = Subtask.objects.select_for_update().get(subtask_id=task_to_compute.compute_task_def['subtask_id'])
+        if subtask.state_enum == Subtask.SubtaskState.FORCING_ACCEPTANCE:
+            return message.concents.ServiceRefused(
+                reason=message.concents.ServiceRefused.REASON.DuplicateRequest,
+            )
```
cameel: This part should be in `else` clause.

concent_api/core/message_handlers.py
```diff
@@ -1354,28 +1356,28 @@ def handle_send_subtask_results_verify(
...
-            Subtask.SubtaskState.ADDITIONAL_VERIFICATION.name,     # pylint: disable=no-member
-        ]
-    ).exists():
-        return message.concents.ServiceRefused(
-            reason=message.concents.ServiceRefused.REASON.DuplicateRequest,
-        )
+    try:
+        subtask = Subtask.objects.select_for_update().get(subtask_id=compute_task_def['subtask_id'])
+
+        if subtask.state_enum in[
```
cameel: Fix whitespace here.

concent_api/core/message_handlers.py
```diff
@@ -1354,28 +1356,28 @@ def handle_send_subtask_results_verify(
...
-            Subtask.SubtaskState.FAILED.name,  # pylint: disable=no-member
-        ]
-    ):
-        raise Http400(
-            "SubtaskResultsVerify is not allowed in current state",
-            error_code=ErrorCode.QUEUE_SUBTASK_STATE_TRANSITION_NOT_ALLOWED,
-        )
+        if subtask.state_enum in [
+            Subtask.SubtaskState.ACCEPTED,  # pylint: disable=no-member
+            Subtask.SubtaskState.FAILED,  # pylint: disable=no-member
```
cameel: Are these disables still necessary?

concent_api/core/message_handlers.py
```diff
@@ -1354,28 +1356,28 @@ def handle_send_subtask_results_verify(
...
-        return message.concents.ServiceRefused(
-            reason=message.concents.ServiceRefused.REASON.DuplicateRequest,
-        )
+    try:
+        subtask = Subtask.objects.select_for_update().get(subtask_id=compute_task_def['subtask_id'])
+
+        if subtask.state_enum in[
+            Subtask.SubtaskState.VERIFICATION_FILE_TRANSFER,
+            Subtask.SubtaskState.ADDITIONAL_VERIFICATION,
+        ]:
```
cameel: This `if` and the one below should be in the `else` clause.

concent_api/core/transfer_operations.py
```diff
@@ -39,15 +39,14 @@ def verify_file_status(
     """
 
     encoded_client_public_key = b64encode(client_public_key)
-    force_get_task_result_list = Subtask.objects.filter(
+    subtasks_list = Subtask.objects.select_for_update().filter(
         requestor__public_key  =encoded_client_public_key,
         state                  = Subtask.SubtaskState.FORCING_RESULT_TRANSFER.name,  # pylint: disable=no-member
```
cameel: Could you fix whitespace here too?

concent_api/core/transfer_operations.py
```diff
@@ -39,15 +39,14 @@ def verify_file_status(
...
     )
 
-    for get_task_result in force_get_task_result_list:
-        report_computed_task    = deserialize_message(get_task_result.report_computed_task.data.tobytes())
+    for subtask in subtasks_list:
+        report_computed_task    = deserialize_message(subtask.report_computed_task.data.tobytes())
```
cameel: Please remove the whitespace here according to our new style.

concent_api/api-integration-multi-threads-test.py
```diff
@@ -0,0 +1,202 @@
...
+    )
+    # waiting for responses from all threads
+    end_time = time.time() + MAXIMUM_WAITING_TIME_FOR_ALL_RESPONSES
+    while len(responses_global) != NUMBER_OF_TESTING_THREADS:
+        time.sleep(0.1)
+        if time.time() >= end_time:
+            break
+
+    expected_responses = ['AckForceGetTaskResult', 'ServiceRefused', 'ServiceRefused']
+    assert len(expected_responses) == NUMBER_OF_TESTING_THREADS, 'Did you changed number of testing threads and forgot to change expected_responses?'
```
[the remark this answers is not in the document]
pawelkisielewicz: done

concent_api/api_testing_common.py
```diff
@@ -72,6 +74,11 @@ def assert_condition(actual, expected, error_message = None):
         raise TestAssertionException(message)
 
 
+def assert_iterables_content_is_equal(actual: Iterable[str], expected: Iterable[str]) -> None:
```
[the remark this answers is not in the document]
pawelkisielewicz: changed to `assert_content_equal`

concent_api/conductor/views.py
```diff
@@ -19,12 +20,13 @@
 @provides_concent_feature('conductor-urls')
 @require_POST
 @csrf_exempt
+@transaction.atomic(using='storage')
 def report_upload(_request, file_path):
```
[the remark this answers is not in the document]
pawelkisielewicz: right, shame on me... done

concent_api/conductor/tasks.py
```diff
@@ -76,7 +76,7 @@ def blender_verification_request(
     # If there are already UploadReports corresponding to some files, the app links them with the VerificationRequest
     # by setting the value of the foreign key in UploadReport.
     for path in [source_package_path, result_package_path]:
-        UploadReport.objects.filter(
+        UploadReport.objects.select_for_update().filter(
             path=path,
             verification_request=None,
         ).update(
```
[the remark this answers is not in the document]
pawelkisielewicz: right, done

concent_api/core/message_handlers.py
pawelkisielewicz: ok, done. In this case I don't think it is necessary, because i catch only `Subtask.DoesNotExist`.  For me the old way was more readable... but changed :)

concent_api/core/message_handlers.py

concent_api/core/message_handlers.py
pawelkisielewicz: not any more. deleted

concent_api/core/message_handlers.py

concent_api/core/message_handlers.py

concent_api/core/transfer_operations.py
pawelkisielewicz: done

concent_api/core/transfer_operations.py
pawelkisielewicz: done

concent_api/core/message_handlers.py
cameel: Code changes. Now you have only this `if`. Later someone could add a function call. Then someone else comes and adds a database query in this function, not even realizing that it's used like this somewhere.

As a general rule it's safer to keep any extra code out of the `try ... except` block.

pawelkisielewicz:
It is merged, should be closed automatically but I forgot to push modified version

Issue #762, opened 2018-08-27
Passing environment and component names to the Signing Service


cameel:
In deployment we always configure `RAVEN_CONFIG` with the name of the current environment (`concent-dev`, `concent-staging`, etc.) and component (`concent-api`, `gatekeeper`, `conductor`, etc.):

See [an example of `RAVEN_CONFIG` in concent-api/local_settings.py](https://github.com/golemfactory/concent-deployment/blob/v0.8.0/kubernetes/config-maps/concent-api/local_settings.py.j2#L42-L46).

This allows us to see in Sentry where the error happened. To get that in Signing Service we need to:
- Allow specifying environment name on the command line (`--sentry-environment` option).
- Hard-code component name to `signing-service`.

Environment name can be an arbitrary string. It's not mandatory - if not specified, the `environment` key  should simply not be present in `RAVEN_CONFIG`.

Add information in the `README` that this option exists and that it should match the Concent environment.

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/160092152

Issue #704, opened 2018-08-03
Update Concent and SigningService READMEs on middleman-protocol manual installation


rwrzesien:
Update Concent and SigningService READMEs on middleman-protocol manual installation for development.

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/159525874

PR #768, opened 2018-08-29
SigningService: Pass environment and component names to Raven client

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/762

### DIFF
--- a/signing_service/README.md
+++ b/signing_service/README.md
@@ -1,0 +1,43 @@
+# signing-service
+
+Stand-alone Python application used for signing Ethereum transactions received via MiddleMan protocol.
+
+
+### Install
+
+You can install this library:
+
+``` bash
+python setup.py install
+```
+
+
+### Install for development
+
+You can install this library for development:
+
+``` bash
+python setup.py develop
+```
+
+
+### Running tests
+
+You can run automated tests from the same  with:
+
+``` bash
+python -m pytest -p no:django .
+```
+
+
+### Running
+
+To run SigningService you need the following:
+
+- Address and port of a Concent cluster to connect to. If port is omitted, default `9055` will be used.
+- A private key for signing sent messages.
+- Concent's public key for verifying signatures of received messages.
+- Ethereum private key for signing Ethereum transactions.
+- [Sentry DNS](https://docs.sentry.io/quickstart/#configure-the-dsn) if you want to submit error reports to a
+  Sentry instance (optional but recommended).
+    - If you want to submit reports to the instance used by Concent (recommended) please also use `--sentry-environment` option to set the `environment` tag to `concent-staging`, `concent-testnet` or `concent-mainnet`, depending on which cluster you're connecting to. If you're submitting to any other instance, you can set this tag to anything you want or even simply omit it. Its only purpose is to let us discern reports from different environments in Sentry.
--- a/signing_service/signing_service/signing_service.py
+++ b/signing_service/signing_service/signing_service.py
@@ -319,6 +319,13 @@ def _parse_arguments() -> argparse.Namespace:
         type=int,
         help=f'Port on which Concent cluster is listening (default: {SIGNING_SERVICE_DEFAULT_PORT}).',
     )
+    parser.add_argument(
+        '-e',
+        '--sentry-environment',
+        dest='sentry_environment',
+        type=str,
+        help=f'Environment which will be set in Raven client config `environment` parameter.',
+    )
 
     ethereum_private_key_parser_group = parser.add_mutually_exclusive_group(required=True)
     ethereum_private_key_parser_group.add_argument(
@@ -390,7 +397,13 @@ if __name__ == '__main__':
     # Parse required arguments.
     args = _parse_arguments()
 
-    raven_client = Client(dsn=args.sentry_dsn)
+    raven_client = Client(
+        dsn=args.sentry_dsn,
+        environment=args.sentry_environment,
+        tags={
+            'component': 'signing-service',
+        },
+    )
     crash_logger.handlers[0].client = raven_client  # type: ignore
 
     arg_host = args.concent_cluster_host


signing_service/signing_service/signing_service.py
```diff
@@ -319,6 +319,13 @@ def _parse_arguments() -> argparse.Namespace:
         type=int,
         help=f'Port on which Concent cluster is listening (default: {SIGNING_SERVICE_DEFAULT_PORT}).',
     )
+    parser.add_argument(
+        '-e',
```
kbeker: Why do you add here `-e`? Will it be used in deployment?

cameel: No, we'll use the long version but it's a good idea to always provide short versions too - for convenience in cases where you just want to launch it manually and don't care about readability.

rwrzesien:
@dybi 
> Don't we need to reflect this new option in SS tests?

It never reaches `SigningService` itself, I think it is not necessary to check if `ArgParser` itself works as expected. I have started to adding some of those at the beginning of work on this but after time I think they are not necessary, especially if this is optional parameter. There are some tests for more complicated cases, like groups of options from which only one can be set (`sentry-dsn`, `signing-service-private-key` etc.).

rwrzesien:
@kbeker @pawelkisielewicz Sorry I forgot to mention this, there is no `README` file in SignignService right now, there is a separate issue for creating it, I think it should be added there. Also, for now there is help message from `ArgParser` for this parameter.

cameel:
> Sorry I forgot to mention this, there is no `README` file in SignignService right now, there is a separate issue for creating it,

Please create at least a rudimentary `README` with information needed here. Looking back, I think that putting this off was not a good idea because the issue about the `README` keeps getting bigger and bigger and will keep being pushed back indefinitely. I think that doing small updates with each pull request is a better approach.

rwrzesien:
@cameel @kbeker @pawelkisielewicz  done :+1:

signing_service/README.md
```diff
@@ -0,0 +1,68 @@
...
+python setup.py develop
+```
+
+
+### Running tests
+
+You can run automated tests with:
+
+``` bash
+python -m pytest -p no:django .
```
cameel: Please specify from which directory. Top-level or inside `signing_service`? Also, does it require the rest of Concent to work or can it be run if you just take the bits belonging to `signing_service` and `middleman_protocol` and put it in a separate directory (like we do in deployment)?

signing_service/README.md
```diff
@@ -0,0 +1,68 @@
...
+```
+
+
+### Running options
+
+positional arguments:
+  concent_cluster_host  Host or IP address of a service on Concent cluster, to
```
cameel: Please don't hard-code this here. No one will remember to update it when it changes in the code so it'll be always out of date.

Instead, please only describe parameters that need special attention. Like `--sentry-environment`. And just say in general what is optional and what is needed to run it (which keys, where the user is expected to get them from, etc). Basically the information that is not normally included in the `--help` output.

signing_service/README.md
rwrzesien: > Please specify from which directory. Top-level or inside signing_service?

I didn't mention this because it works both ways. 

> Also, does it require the rest of Concent to work or can it be run if you just take the bits belonging to signing_service and middleman_protocol and put it in a separate directory (like we do in deployment)?

It is described as `Stand-alone` at the beginning of the file.

signing_service/README.md
rwrzesien: > Please don't hard-code this here. No one will remember to update it when it changes in the code so it'll be always out of date.

I don't agree with this assumption. It sounds like: no matter what, we will do it wrong anyway. If we pay attention to this during reviews, it won't get outdated.

I have rephrased this part a little bit, but I think mentioning all parameters is ok. There are not so many of them, and I couldn't find a pattern in which some are more important than other.

signing_service/README.md
cameel: > I didn't mention this because it works both ways.

Great.

> It is described as `Stand-alone` at the beginning of the file.

I mean just the tests you're running here. Are these tests completely independent of Concent's? If so, it's fine.

signing_service/README.md
cameel: > I don't agree with this assumption. It sounds like: no matter what, we will do it wrong anyway.

Sorry, I didn't want to come off as offensive. I don't mean that they'll go out of sync due to someone's incompetence or bad will. I simply accept that if you have two identical pieces of code/text and nothing to enforce that they're identical, they'll go out of sync sooner or later. It just happens.

Having to pay attention doing reviews is fine but the best strategy is to simply avoid duplication if it's not necessary. I think that in this case the duplication is not worth it because you can easily get this information from `--help`.

Also, I assumed that you just did copy&paste of the output of the `--help` option. If you wrote it from scratch and it does not contain exactly the same information, that's a different story. I did not compare them so I don't know.

> I have rephrased this part a little bit, but I think mentioning all parameters is ok. There are not so many of them, and I couldn't find a pattern in which some are more important than other.

I would not include all the parameters here. I think a high level description is better because it does not rely as much on the implementation details:

```
To run Signing Service you need the following:
- Address and port of a Concent cluster to connect to.
- A private key for signing sent messages.
- Concent's public key for verifying signatures of received messages.
- [Sentry DNS](https://docs.sentry.io/quickstart/#configure-the-dsn) if you want to submit error reports to a Sentry instance (optional but recommended).
    - If you want to submit reports to the instance used by Concent (recommended) please also use `--sentry-environment` option to set the `environment` tag to `concent-staging`, `concent-testnet` or `concent-mainnet`, depending on which cluster you're connecting to. If you're submitting to any other instance, you can set this tag to anything you want or even simply omit it. Its only purpose is to let us discern reports from different environments in Sentry.
```

That's just a draft. Option names and explaining how we pass secrets should probably be added too.

I wouldn't really mention the reconnect delay unless there anything more to say than `--help` already does. You might just mention that there's more options and to use `--help` to see them all.

Anyway, that's my opinion. But the `README` update was meant to be just a small addition to this pull request. I don't want to hold the whole pull request back so if you still strongly disagree, please remove this section of the `README`, mention only `--sentry-environment` (because that's what this pull request is about) and let's move the discussion and the removed changes to #704.

signing_service/README.md
```diff
@@ -0,0 +1,44 @@
...
+``` bash
+python -m pytest -p no:django .
+```
+
+
+### Running
+
+To run SigningService you need the following:
+
+To run Signing Service you need the following:
```
cameel: This line is here twice.

signing_service/README.md
cameel: Looks fine now.

PR #777, opened 2018-09-03
Bugfix integration tests error codes fixes

rwrzesien:
Resolves:
1. There was a bug in `tests.utils._test_400_response` method that causes passed `error_code` parameter not being tested if `error_message` is None. This cause several tests to have wrong error code tested but they were still passing.
2. Golem messages no longer hat `TYPE` attribute so fix to `core.message_handlers.handle_unsupported_golem_messages_type` was required.
3. In `core.tests.test_core_views` provider and requestor keys were passed in wrong formats to TaskToCompute, which failed requests on different validations that were expected by tests, but tests were still passing.

### DIFF
--- a/concent_api/core/tests/test_auth_accept_or_reject_results.py
+++ b/concent_api/core/tests/test_auth_accept_or_reject_results.py
@@ -121,21 +121,25 @@ class AuthAcceptOrRejectIntegrationTest(ConcentIntegrationTestCase):
                         provider_public_key=self._get_diffrent_provider_hex_public_key(),
                         requestor_public_key=self._get_diffrent_requestor_hex_public_key(),
                         signer_private_key=self.DIFFERENT_REQUESTOR_PRIVATE_KEY,
+                        compute_task_def=task_to_compute.compute_task_def,
                     ),
                     signer_private_key=self.DIFFERENT_PROVIDER_PRIVATE_KEY,
                 ),
                 signer_private_key=self.DIFFERENT_REQUESTOR_PRIVATE_KEY,
-                task_to_compute=task_to_compute
             ),
             provider_private_key=self.DIFFERENT_PROVIDER_PRIVATE_KEY,
         )
 
-        with freeze_time("2018-02-05 10:00:31"):
-            response = self.client.post(
-                reverse('core:send'),
-                data                                = different_serialized_force_subtask_results,
-                content_type                        = 'application/octet-stream',
-            )
+        with mock.patch(
+            'core.message_handlers.payments_service.is_account_status_positive',
+            side_effect=self.is_account_status_positive_true_mock
+        ):
+            with freeze_time("2018-02-05 10:00:31"):
+                response = self.client.post(
+                    reverse('core:send'),
+                    data=different_serialized_force_subtask_results,
+                    content_type='application/octet-stream',
+                )
 
         self._test_response(
             response,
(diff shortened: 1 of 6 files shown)


concent_api/core/tests/test_auth_accept_or_reject_results.py
```diff
@@ -121,21 +121,25 @@ def test_provider_forces_subtask_results_for_task_which_was_already_submitted_co
...
-            )
+        with mock.patch(
+            'core.message_handlers.payments_service.is_account_status_positive',
+            side_effect=self.is_account_status_positive_true_mock
+        ):
+            with freeze_time("2018-02-05 10:00:31"):
+                response = self.client.post(
+                    reverse('core:send'),
+                    data                                = different_serialized_force_subtask_results,
+                    content_type                        = 'application/octet-stream',
```
kbeker: Please use our new style :)

concent_api/core/tests/test_auth_accept_or_reject_results.py
```diff
@@ -152,12 +156,16 @@ def test_provider_forces_subtask_results_for_task_which_was_already_submitted_co
...
-            )
+        with mock.patch(
+            'core.message_handlers.payments_service.is_account_status_positive',
+            side_effect=self.is_account_status_positive_true_mock
+        ):
+            with freeze_time("2018-02-05 10:00:31"):
+                response = self.client.post(
+                    reverse('core:send'),
+                    data                                = serialized_force_subtask_results,
+                    content_type                        = 'application/octet-stream',
```
kbeker: Same here

concent_api/core/tests/test_core_views.py
```diff
@@ -174,9 +172,8 @@ def test_send_should_return_http_400_if_data_is_incorrect(self):
         compute_task_def    = message.ComputeTaskDef()
         task_to_compute     = message.TaskToCompute(
             compute_task_def     = compute_task_def,
-            provider_public_key  = PROVIDER_PUBLIC_KEY,
-            requestor_public_key = REQUESTOR_PUBLIC_KEY,
-            price=0,
+            provider_public_key  = self._get_encoded_provider_public_key(),
+            requestor_public_key = self._get_encoded_requestor_public_key(),
```
kbeker: Same here

concent_api/core/tests/test_core_views.py
```diff
@@ -714,13 +707,13 @@ def test_receive_should_return_first_messages_in_order_they_were_added_to_queue_
         stored_ack_report_computed_task.save()
 
         client_provider = Client(
-            public_key_bytes = PROVIDER_PUBLIC_KEY
+            public_key_bytes = self.PROVIDER_PUBLIC_KEY
```

concent_api/core/tests/utils.py
```diff
@@ -169,8 +169,8 @@ def _get_serialized_force_get_task_result(
 
     def _get_deserialized_report_computed_task(
         self,
-        subtask_id: str = str(uuid.uuid4()),
-        task_id: str = str(uuid.uuid4()),
+        subtask_id: Optional[str] = None,
+        task_id: Optional[str] = None,
```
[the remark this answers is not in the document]
rwrzesien: It will not work, it is incorrect syntax.

concent_api/core/tests/test_core_views.py
```diff
@@ -195,10 +196,7 @@ def test_send_should_return_http_400_if_data_is_incorrect(self):
             ),
             content_type                        = 'application/octet-stream',
         )
-        self._test_400_response(
-            response,
-            error_code=ErrorCode.MESSAGE_VALUE_BLANK
-        )
+        self._test_400_response(response)
```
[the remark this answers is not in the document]
rwrzesien: There is no error code in this response. Check `common.decorators.require_golem_message` - there are cases when it is not passed to response. I have added checking beginning of `error_message`.

>  think that error_message and error_code should be obligatory, not optional. 

`error_code` is not always available, and we decided a long time ago that we don't want to check `error_message` until it is necessary - because they might change too often (but really they don't).

concent_api/core/tests/test_auth_accept_or_reject_results.py
rwrzesien: :+1: done

concent_api/core/tests/test_auth_accept_or_reject_results.py
rwrzesien: :+1: done

concent_api/core/tests/test_core_views.py

concent_api/core/tests/test_core_views.py

concent_api/core/tests/utils.py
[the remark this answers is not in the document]
pawelkisielewicz: ok, I didn't know :)

concent_api/core/tests/test_core_views.py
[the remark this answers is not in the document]
pawelkisielewicz: ok...It is quite strange, sometimes there are erorr_codes, sometimes not. But I agree with explanation.

Issue #770, opened 2018-08-30
Views sometimes overwrite a subtask when they're supposed to create it or fail


cameel:
The problem originally came up in this review: https://github.com/golemfactory/concent/pull/764#discussion_r213659010

Our views use `store_or_update_subtask()` to create or update a subtask, depending on whether it already exists or not. The problem is that subtask data passed to the function is fetched first and only then the function checks whether the subtask exists. This means that if another process creates the subtask after the data has been fetched but before the check, the subtask is updated even though the view should report that the subtask already exists and fail with HTTP 5xx.

To fix this, the code should decide whether the subtask needs to be created or updated only once and not change that decision. You can achieve that by splitting `store_or_update_subtask()` into separate `create_subtask()` and `update_subtask()` and moving the check to the view.

Also, remember to use `select_for_update()` when you fetch the data that will later be used to update the subtask and/or check subtask's existence. It's best if the check and fetching are actually a single query (i.e. use `get()` or `filter()` instead of `exists()` and then check existence by simply looking at the result/catching the exception).

### Example illustrating the problem
1. Client 1 submits message x with `subtask_id = A`.
2. Client 2 submits message y with `subtask_id = A`.
3. Process 1 checks if subtask with `subtask_id = A` exists. It does not.
4. Process 2 checks if subtask with `subtask_id = A` exists. It does not.
5. Process 1 prepares subtask X with `subtask_id = A` and data from message x.
6. Process 2 prepares subtask Y with `subtask_id = A` and data from message y,
7. Process 1 calls  `store_or_update_subtask()`
    1. `store_or_update_subtask()` checks if subtask with `subtask_id = A` exists. It does not.
    2. `store_or_update_subtask()` tries to save subtask X and succeeds.
8. Process 2 calls  `store_or_update_subtask()`
    1. `store_or_update_subtask()` checks if subtask with `subtask_id = A` exists. It does.
    2. `store_or_update_subtask()` updates subtask X with data from subtask Y and succeeds.
9. Result
    - Both clients get HTTP 2xx. Both succeeded.
    - Both clients may have triggered Irreversible actions for the same subtask.
    - Data from message x has been overwritten by data from message y. One subtask has been created but irreversible actions were executed twice.

PaweuB:
Pivotal: 
https://www.pivotaltracker.com/story/show/160169099

PR #778, opened 2018-09-04
Bugfix decide store or update subtask in only one place

pawelkisielewicz:
Resolves https://github.com/golemfactory/concent/issues/770

### DIFF
--- a/concent_api/core/message_handlers.py
+++ b/concent_api/core/message_handlers.py
@@ -44,6 +44,7 @@ from core.payments.backends.sci_backend import TransactionType
 from core.queue_operations import send_blender_verification_request
 from core.subtask_helpers import are_keys_and_addresses_unique_in_message_subtask_results_accepted
 from core.subtask_helpers import are_subtask_results_accepted_messages_signed_by_the_same_requestor
+from core.subtask_helpers import get_one_or_none
 from core.transfer_operations import store_pending_message
 from core.transfer_operations import create_file_transfer_token_for_golem_client
 from core.utils import calculate_additional_verification_call_time
@@ -189,7 +190,7 @@ def handle_send_ack_report_computed_task(client_message: message.tasks.AckReport
         except ConcentValidationError:
             new_report_computed_task = report_computed_task
 
-        subtask = update_subtask(
+        subtask = update_and_return_updated_subtask(
             subtask                     = subtask,
             state                       = Subtask.SubtaskState.REPORTED,
             next_deadline               = None,
@@ -308,7 +309,7 @@ def handle_send_reject_report_computed_task(client_message: message.tasks.Reject
 
     if client_message.reason == message.tasks.RejectReportComputedTask.REASON.SubtaskTimeLimitExceeded:
 
-        subtask = update_subtask(
+        subtask = update_and_return_updated_subtask(
             subtask                     = subtask,
             state                       = Subtask.SubtaskState.REPORTED,
             next_deadline               = None,
@@ -343,7 +344,7 @@ def handle_send_reject_report_computed_task(client_message: message.tasks.Reject
                 error_code=ErrorCode.SUBTASK_DUPLICATE_REQUEST,
             )
 
-        subtask = update_subtask(
+        subtask = update_and_return_updated_subtask(
             subtask                     = subtask,
             state                       = Subtask.SubtaskState.FAILED,
             next_deadline               = None,
@@ -392,16 +393,6 @@ def handle_send_force_get_task_result(
         task_to_compute,
     )
 
-    try:
-        subtask = Subtask.objects.select_for_update().get(subtask_id=task_to_compute.compute_task_def['subtask_id'])
-    except Subtask.DoesNotExist:
-        pass
-    else:
-        if subtask.state_enum == Subtask.SubtaskState.FORCING_RESULT_TRANSFER:
-            return message.concents.ServiceRefused(
-                reason=message.concents.ServiceRefused.REASON.DuplicateRequest,
-            )
-
     maximum_download_time = calculate_maximum_download_time(
         client_message.report_computed_task.size,
         settings.MINIMUM_UPLOAD_RATE,
@@ -423,22 +414,51 @@ def handle_send_force_get_task_result(
             reason=message.concents.ForceGetTaskResultRejected.REASON.AcceptanceTimeLimitExceeded,
         )
 
-    subtask = store_or_update_subtask(
-        task_id=task_to_compute.compute_task_def['task_id'],
+    subtask = get_one_or_none(
+        Subtask.objects.select_for_update(),
         subtask_id=task_to_compute.compute_task_def['subtask_id'],
-        provider_public_key=provider_public_key,
-        requestor_public_key=requestor_public_key,
-        state=Subtask.SubtaskState.FORCING_RESULT_TRANSFER,
-        next_deadline=(
-            int(task_to_compute.compute_task_def['deadline']) +
-            2 * maximum_download_time +
-            3 * settings.CONCENT_MESSAGING_TIME
-        ),
-        set_next_deadline=True,
-        report_computed_task=client_message.report_computed_task,
-        task_to_compute=task_to_compute,
-        force_get_task_result=client_message,
     )
+    if subtask is None:
+        subtask = store_subtask(
+            task_id=task_to_compute.compute_task_def['task_id'],
+            subtask_id=task_to_compute.compute_task_def['subtask_id'],
+            provider_public_key=provider_public_key,
+            requestor_public_key=requestor_public_key,
+            state=Subtask.SubtaskState.FORCING_RESULT_TRANSFER,
+            next_deadline=(
+                int(task_to_compute.compute_task_def['deadline']) +
+                2 * maximum_download_time +
+                3 * settings.CONCENT_MESSAGING_TIME
+            ),
+            task_to_compute=task_to_compute,
+            report_computed_task=client_message.report_computed_task,
+            force_get_task_result=client_message,
+        )
+
+    else:
+        if subtask.state_enum == Subtask.SubtaskState.FORCING_RESULT_TRANSFER:
+            return message.concents.ServiceRefused(
+                reason=message.concents.ServiceRefused.REASON.DuplicateRequest,
+            )
+        if task_to_compute is not None and subtask.task_to_compute is not None:
+            validate_all_messages_identical([
+                task_to_compute,
+                deserialize_message(subtask.task_to_compute.data.tobytes()),
+            ])
+        subtask = update_and_return_updated_subtask(
+            subtask=subtask,
+            state=Subtask.SubtaskState.FORCING_RESULT_TRANSFER,
+            next_deadline=(
+                int(task_to_compute.compute_task_def['deadline']) +
+                2 * maximum_download_time +
+                3 * settings.CONCENT_MESSAGING_TIME
+            ),
+            set_next_deadline=True,
+            task_to_compute=task_to_compute,
+            report_computed_task=client_message.report_computed_task,
+            force_get_task_result=client_message,
+        )
+
     store_pending_message(
         response_type       = PendingResponse.ResponseType.ForceGetTaskResultUpload,
         client_public_key   = provider_public_key,
@@ -470,16 +490,6 @@ def handle_send_force_subtask_results(
     )
 
     current_time = get_current_utc_timestamp()
-
-    try:
-        subtask = Subtask.objects.select_for_update().get(subtask_id=task_to_compute.compute_task_def['subtask_id'])
-    except Subtask.DoesNotExist:
-        pass
-    else:
-        if subtask.state_enum == Subtask.SubtaskState.FORCING_ACCEPTANCE:
-            return message.concents.ServiceRefused(
-                reason=message.concents.ServiceRefused.REASON.DuplicateRequest,
-            )
 
     if not payments_service.is_account_status_positive(  # pylint: disable=no-value-for-parameter
         client_eth_address=task_to_compute.requestor_ethereum_address,
@@ -526,18 +536,42 @@ def handle_send_force_subtask_results(
             reason=message.concents.ForceSubtaskResultsRejected.REASON.RequestPremature,
         )
 
-    subtask = store_or_update_subtask(
-        task_id                     = task_to_compute.compute_task_def['task_id'],
-        subtask_id                  = task_to_compute.compute_task_def['subtask_id'],
-        provider_public_key         = provider_public_key,
-        requestor_public_key        = requestor_public_key,
-        state                       = Subtask.SubtaskState.FORCING_ACCEPTANCE,
-        next_deadline               = forcing_acceptance_deadline + settings.CONCENT_MESSAGING_TIME,
-        set_next_deadline           = True,
-        ack_report_computed_task    = client_message.ack_report_computed_task,
-        task_to_compute             = client_message.ack_report_computed_task.report_computed_task.task_to_compute,
-        report_computed_task        = client_message.ack_report_computed_task.report_computed_task,
+    subtask = get_one_or_none(
+        Subtask.objects.select_for_update(),
+        subtask_id=task_to_compute.compute_task_def['subtask_id'],
     )
+
+    if subtask is None:
+        subtask = store_subtask(
+            task_id=task_to_compute.compute_task_def['task_id'],
+            subtask_id=task_to_compute.compute_task_def['subtask_id'],
+            provider_public_key=provider_public_key,
+            requestor_public_key=requestor_public_key,
+            state=Subtask.SubtaskState.FORCING_ACCEPTANCE,
+            next_deadline=forcing_acceptance_deadline + settings.CONCENT_MESSAGING_TIME,
+            task_to_compute=client_message.ack_report_computed_task.report_computed_task.task_to_compute,
+            report_computed_task=client_message.ack_report_computed_task.report_computed_task,
+            ack_report_computed_task=client_message.ack_report_computed_task,
+        )
+    else:
+        if subtask.state_enum == Subtask.SubtaskState.FORCING_ACCEPTANCE:
+            return message.concents.ServiceRefused(
+                reason=message.concents.ServiceRefused.REASON.DuplicateRequest,
+            )
+        if task_to_compute is not None and subtask.task_to_compute is not None:
+            validate_all_messages_identical([
+                task_to_compute,
+                deserialize_message(subtask.task_to_compute.data.tobytes()),
+            ])
+        subtask = update_and_return_updated_subtask(
+            subtask=subtask,
+            state=Subtask.SubtaskState.FORCING_ACCEPTANCE,
+            next_deadline=forcing_acceptance_deadline + settings.CONCENT_MESSAGING_TIME,
+            set_next_deadline=True,
+            task_to_compute=client_message.ack_report_computed_task.report_computed_task.task_to_compute,
+            report_computed_task=client_message.ack_report_computed_task.report_computed_task,
+            ack_report_computed_task=client_message.ack_report_computed_task,
+        )
     store_pending_message(
         response_type       = PendingResponse.ResponseType.ForceSubtaskResults,
         client_public_key   = requestor_public_key,
@@ -631,7 +665,7 @@ def handle_send_force_subtask_results_response(
         deserialize_message(subtask.task_to_compute.data.tobytes()),
     ])
 
-    subtask = update_subtask(
+    subtask = update_and_return_updated_subtask(
         subtask                     = subtask,
         state                       = state,
         next_deadline               = None,
@@ -1115,7 +1149,7 @@ def mark_message_as_delivered_and_log(undelivered_message: PendingResponse, log_
     )
 
 
-def update_subtask(
+def update_and_return_updated_subtask(
     subtask: Subtask,
     state: Subtask.SubtaskState,
     next_deadline: Optional[int] = None,
@@ -1236,67 +1270,6 @@ def set_subtask_messages(
             )
 
 
-def store_or_update_subtask(
-    task_id: str,
-    subtask_id: str,
-    provider_public_key: bytes,
-    requestor_public_key: bytes,
-    state: Subtask.SubtaskState,
-    next_deadline: Optional[int] = None,
-    set_next_deadline: Optional[bool] = False,
-    task_to_compute: Optional[message.TaskToCompute] = None,
-    report_computed_task: Optional[message.ReportComputedTask] = None,
-    ack_report_computed_task: Optional[message.tasks.AckReportComputedTask] = None,
-    reject_report_computed_task: Optional[message.tasks.RejectReportComputedTask] = None,
-    subtask_results_accepted: Optional[message.tasks.SubtaskResultsAccepted] = None,
-    subtask_results_rejected: Optional[message.tasks.SubtaskResultsRejected] = None,
-    force_get_task_result: Optional[message.concents.ForceGetTaskResult] = None,
-) -> Subtask:
-    try:
-        subtask = Subtask.objects.select_for_update().get(
-            subtask_id = subtask_id,
-        )
-    except Subtask.DoesNotExist:
-        subtask = None
-
-    if subtask is not None:
-        if task_to_compute is not None and subtask.task_to_compute is not None:
-            validate_all_messages_identical([
-                task_to_compute,
-                deserialize_message(subtask.task_to_compute.data.tobytes()),
-            ])
-        subtask = update_subtask(
-            subtask=subtask,
-            state=state,
-            next_deadline=next_deadline,
-            set_next_deadline=set_next_deadline,
-            task_to_compute=task_to_compute,
-            report_computed_task=report_computed_task,
-            ack_report_computed_task=ack_report_computed_task,
-            reject_report_computed_task=reject_report_computed_task,
-            subtask_results_accepted=subtask_results_accepted,
-            subtask_results_rejected= subtask_results_rejected,
-            force_get_task_result=force_get_task_result,
-        )
-    else:
-        subtask = store_subtask(
-            task_id=task_id,
-            subtask_id=subtask_id,
-            provider_public_key=provider_public_key,
-            requestor_public_key=requestor_public_key,
-            state=state,
-            next_deadline=next_deadline,
-            task_to_compute=task_to_compute,
-            report_computed_task=report_computed_task,
-            ack_report_computed_task=ack_report_computed_task,
-            reject_report_computed_task=reject_report_computed_task,
-            subtask_results_accepted=subtask_results_accepted,
-            subtask_results_rejected=subtask_results_rejected,
-            force_get_task_result=force_get_task_result,
-        )
-    return subtask
-
-
 def store_message(
     golem_message:          message.base.Message,
     task_id:                str,
@@ -1361,10 +1334,31 @@ def handle_send_subtask_results_verify(
             reason=message.concents.ServiceRefused.REASON.InvalidRequest,
         )
 
-    try:
-        subtask = Subtask.objects.select_for_update().get(subtask_id=compute_task_def['subtask_id'])
-    except Subtask.DoesNotExist:
-        pass
+    if not payments_service.is_account_status_positive(  # pylint: disable=no-value-for-parameter
+        client_eth_address=task_to_compute.requestor_ethereum_address,
+        pending_value=task_to_compute.price,
+    ):
+        return message.concents.ServiceRefused(
+            reason=message.concents.ServiceRefused.REASON.TooSmallRequestorDeposit,
+        )
+
+    subtask = get_one_or_none(
+        Subtask.objects.select_for_update(),
+        subtask_id=compute_task_def['subtask_id'],
+    )
+    if subtask is None:
+        store_subtask(
+            task_id=compute_task_def['task_id'],
+            subtask_id=compute_task_def['subtask_id'],
+            provider_public_key=provider_public_key,
+            requestor_public_key=requestor_public_key,
+            state=Subtask.SubtaskState.VERIFICATION_FILE_TRANSFER,
+            next_deadline=verification_deadline,
+            task_to_compute=task_to_compute,
+            report_computed_task=report_computed_task,
+            subtask_results_rejected=subtask_results_rejected,
+        )
+
     else:
         if subtask.state_enum in[
             Subtask.SubtaskState.VERIFICATION_FILE_TRANSFER,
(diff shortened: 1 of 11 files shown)


concent_api/core/tests/test_unit_subtask_helpers.py
```diff
@@ -0,0 +1,66 @@
...
+        self.compute_task_def = self._get_deserialized_compute_task_def()
+
+        self.task_to_compute = self._get_deserialized_task_to_compute(
+            compute_task_def=self.compute_task_def,
+        )
+        self.report_computed_task = self._get_deserialized_report_computed_task(
+            task_to_compute=self.task_to_compute,
+        )
+
+        store_subtask(
```
dybi: `self.subtask = store_subtask(`

concent_api/core/tests/test_unit_subtask_helpers.py
```diff
@@ -0,0 +1,66 @@
...
+            requestor_public_key=self.REQUESTOR_PUBLIC_KEY,
+            state=Subtask.SubtaskState.FORCING_REPORT,
+            next_deadline=int(self.task_to_compute.compute_task_def['deadline']) + settings.CONCENT_MESSAGING_TIME,
+            task_to_compute=self.task_to_compute,
+            report_computed_task=self.report_computed_task,
+        )
+
+    def test_that_if_only_subtask_id_given_and_subtask_exist_function_returns_subtask(self):
+        subtask = get_one_or_none_subtask_from_database(self.task_to_compute.compute_task_def['subtask_id'])
+        self.assertIsInstance(subtask, Subtask)
```
dybi: instead of those, use `self.assertEqual(self.subtask, subtask)`, here and below

concent_api/core/tests/test_unit_subtask_helpers.py
```diff
@@ -0,0 +1,66 @@
...
+        )
+        self.assertIsInstance(subtask, Subtask)
+        self.assertEqual(subtask.subtask_id, self.task_to_compute.compute_task_def['subtask_id'])
+
+    def test_that_if_subtask_id_and_optional_dict_given_and_subtask_does_not_exist_function_returns_none(self):
+        subtask = get_one_or_none_subtask_from_database(
+            subtask_id=self.task_to_compute.compute_task_def['subtask_id'],
+            optional_condition_dict={'state': Subtask.SubtaskState.REPORTED.name},  # pylint: disable=no-member
+        )
+        self.assertEqual(subtask, None)
```
dybi: please use `assertIsNone`, it is more readable

concent_api/core/tests/test_unit_subtask_helpers.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def test_that_if_incorrect_dict_given_function_raises_assert_error(self):
```
dybi: as discussed, please remove

concent_api/core/tests/test_unit_subtask_helpers.py
```diff
@@ -0,0 +1,66 @@
...
+        with self.assertRaises(AssertionError):
+            get_one_or_none_subtask_from_database(
+                subtask_id=self.task_to_compute.compute_task_def['subtask_id'],
+                optional_condition_dict={'state': Subtask.SubtaskState.REPORTED},
+            )
+
+    def test_that_if_incorrect_dict_key_given_function_raises_assert_error(self):
```
dybi: as discussed, please remove

concent_api/core/tests/test_unit_subtask_helpers.py
pawelkisielewicz: done

concent_api/core/tests/test_unit_subtask_helpers.py
pawelkisielewicz: done

concent_api/core/tests/test_unit_subtask_helpers.py
pawelkisielewicz: done

concent_api/core/tests/test_unit_subtask_helpers.py
pawelkisielewicz: done

concent_api/core/subtask_helpers.py
```diff
@@ -185,3 +187,28 @@ def are_subtask_results_accepted_messages_signed_by_the_same_requestor(subtask_r
         ) for subtask_results_accepted in subtask_results_accepted_list
     )
     return are_all_signed_by_requestor
+
+
+def get_one_or_none_subtask_from_database(
+    subtask_id: str,
+    optional_condition_dict: Dict[str, str] = None,
```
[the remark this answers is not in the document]

concent_api/core/subtask_helpers.py
```diff
@@ -185,3 +187,28 @@ def are_subtask_results_accepted_messages_signed_by_the_same_requestor(subtask_r
...
+    lock_returned_subtasks_in_database: bool = False,
+) -> Union[Subtask, None]:
+
+    condition_dict = {'subtask_id': subtask_id}
```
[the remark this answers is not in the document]

concent_api/core/subtask_helpers.py
```diff
@@ -185,3 +187,28 @@ def are_subtask_results_accepted_messages_signed_by_the_same_requestor(subtask_r
         ) for subtask_results_accepted in subtask_results_accepted_list
     )
     return are_all_signed_by_requestor
+
+
+def get_one_or_none_subtask_from_database(
+    subtask_id: str,
+    optional_condition_dict: Dict[str, str] = None,
+    lock_returned_subtasks_in_database: bool = False,
+) -> Union[Subtask, None]:
```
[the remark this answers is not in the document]

concent_api/core/subtask_helpers.py
```diff
@@ -185,3 +187,28 @@ def are_subtask_results_accepted_messages_signed_by_the_same_requestor(subtask_r
         ) for subtask_results_accepted in subtask_results_accepted_list
     )
     return are_all_signed_by_requestor
+
+
+def get_one_or_none_subtask_from_database(
+    subtask_id: str,
+    optional_condition_dict: Dict[str, str] = None,
+    lock_returned_subtasks_in_database: bool = False,
```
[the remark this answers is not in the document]
pawelkisielewicz: to resolve this problem this is not needed. I did it to make this function more general- to get subtask without obligatory locking, with this function. And False as default was, because I think locking database should be written explicitly, not by hidden default parametr. Removed

concent_api/core/message_handlers.py
```diff
@@ -42,6 +42,7 @@
 from core.payments.backends.sci_backend import TransactionType
 from core.queue_operations import send_blender_verification_request
 from core.subtask_helpers import are_keys_and_addresses_unique_in_message_subtask_results_accepted
+from core.subtask_helpers import get_one_or_none_subtasks_from_database
```
rwrzesien: Please keep alphabetic order.

concent_api/core/subtask_helpers.py
```diff
@@ -185,3 +186,22 @@ def are_subtask_results_accepted_messages_signed_by_the_same_requestor(subtask_r
         ) for subtask_results_accepted in subtask_results_accepted_list
     )
     return are_all_signed_by_requestor
+
+
+def get_one_or_none_subtasks_from_database(
+    subtask_id: str,
+    optional_condition_dict: Dict[str, str] = None,
+) -> Optional[Subtask]:
+
```
rwrzesien: I think this is a little bit to complicated to do a simple thing. Why not using `try/except` pattern here ? Btw. `conditions_dict` and its update is fine, I mean the `query` part.

concent_api/core/subtask_helpers.py
```diff
@@ -185,3 +186,22 @@ def are_subtask_results_accepted_messages_signed_by_the_same_requestor(subtask_r
         ) for subtask_results_accepted in subtask_results_accepted_list
     )
     return are_all_signed_by_requestor
+
+
+def get_one_or_none_subtasks_from_database(
+    subtask_id: str,
+    optional_condition_dict: Dict[str, str] = None,
```
rwrzesien: `optional_condition_dict: Dict[str, str]` - keys are fine but values are not always be strings.

concent_api/core/message_handlers.py
```diff
@@ -419,22 +411,45 @@ def handle_send_force_get_task_result(client_message: message.concents.ForceGetT
...
     )
+    if subtask is None:
+        subtask = store_subtask(
+            task_id=task_to_compute.compute_task_def['task_id'],
+            subtask_id=task_to_compute.compute_task_def['subtask_id'],
+            provider_public_key=provider_public_key,
+            requestor_public_key=requestor_public_key,
+            state=Subtask.SubtaskState.FORCING_RESULT_TRANSFER,
+            next_deadline=(
+                    int(task_to_compute.compute_task_def['deadline']) +
```
rwrzesien: Please fix indent.

concent_api/core/message_handlers.py
```diff
@@ -419,22 +411,45 @@ def handle_send_force_get_task_result(client_message: message.concents.ForceGetT
...
+                    2 * maximum_download_time +
+                    3 * settings.CONCENT_MESSAGING_TIME
+            ),
+            task_to_compute=task_to_compute,
+            report_computed_task=client_message.report_computed_task,
+            force_get_task_result=client_message,
+        )
+
+    else:
+        if subtask.state_enum==Subtask.SubtaskState.FORCING_RESULT_TRANSFER:
```
rwrzesien: Please fix spaces around `==`.

concent_api/core/message_handlers.py
```diff
@@ -465,18 +480,9 @@ def handle_send_force_subtask_results(client_message: message.concents.ForceSubt
...
-                reason=message.concents.ServiceRefused.REASON.DuplicateRequest,
-            )
-    except Subtask.DoesNotExist:
-        pass
-
     if not payments_service.is_account_status_positive(  # pylint: disable=no-value-for-parameter
-        client_eth_address=task_to_compute.requestor_ethereum_address,
-        pending_value=task_to_compute.price,
+            client_eth_address=task_to_compute.requestor_ethereum_address,
+            pending_value=task_to_compute.price,
```
kbeker: Please fix indent

concent_api/core/subtask_helpers.py
```diff
@@ -1,9 +1,10 @@
 from base64                     import b64encode
 from logging import getLogger
+from typing import Dict
 from typing import List
 from typing import Optional
 
-from django.db.models           import Q
+from django.db.models import Q
 from django.utils               import timezone
```
kbeker: We decided last time, that if we're changing code next to old style code we should change also block of code where is old style :)

concent_api/core/tests/test_auth_accept_or_reject_results.py
```diff
@@ -66,7 +66,7 @@ def test_provider_forces_subtask_results_for_task_which_was_already_submitted_co
             'core.message_handlers.payments_service.is_account_status_positive',
             side_effect=self.is_account_status_positive_true_mock
         ) as is_account_status_positive_true_mock_function:
-            with freeze_time("2018-02-05 10:00:30"):
+            with freeze_time("2018-02-05 10:00:29"):
```
kbeker: This change is not necessary if you change timestamp in below freeze_time

concent_api/core/tests/test_integration_accept_or_reject_results.py
```diff
@@ -1675,12 +1679,13 @@ def test_requestor_send_again_subtask_results_accepted_or_rejected_when_message_
...
         task_to_compute = deserialized_force_subtask_results.ack_report_computed_task.report_computed_task.task_to_compute  # pylint: disable=no-member,
 
-        subtask = store_or_update_subtask(
+        subtask = store_subtask(
             task_id=task_to_compute.compute_task_def['task_id'],
             subtask_id=task_to_compute.compute_task_def['subtask_id'],
             provider_public_key=self.PROVIDER_PUBLIC_KEY,
             requestor_public_key=self.REQUESTOR_PUBLIC_KEY,
             state=Subtask.SubtaskState.ACCEPTED,
+            next_deadline=None,
```
kbeker: I think better than adding `next_deadline=None` everywhere where you need this is just add default value for `next_deadline` in function.

concent_api/core/message_handlers.py
```diff
@@ -419,22 +411,45 @@ def handle_send_force_get_task_result(client_message: message.concents.ForceGetT
...
+    else:
+        if subtask.state_enum==Subtask.SubtaskState.FORCING_RESULT_TRANSFER:
+            return message.concents.ServiceRefused(
+                reason=message.concents.ServiceRefused.REASON.DuplicateRequest,
+            )
+        subtask = update_subtask(
+            subtask=subtask,
+            state=Subtask.SubtaskState.FORCING_RESULT_TRANSFER,
+            next_deadline=(
+                    int(task_to_compute.compute_task_def['deadline']) +
```
rwrzesien: Same here.

concent_api/core/message_handlers.py
```diff
@@ -465,18 +480,9 @@ def handle_send_force_subtask_results(client_message: message.concents.ForceSubt
...
-            return message.concents.ServiceRefused(
-                reason=message.concents.ServiceRefused.REASON.DuplicateRequest,
-            )
-    except Subtask.DoesNotExist:
-        pass
-
     if not payments_service.is_account_status_positive(  # pylint: disable=no-value-for-parameter
-        client_eth_address=task_to_compute.requestor_ethereum_address,
-        pending_value=task_to_compute.price,
+            client_eth_address=task_to_compute.requestor_ethereum_address,
```
rwrzesien: Same here.

concent_api/core/message_handlers.py
```diff
@@ -1134,6 +1158,12 @@ def update_subtask(
 
     next_deadline_datetime = parse_timestamp_to_utc_datetime(next_deadline) if next_deadline is not None else None
 
+    if task_to_compute is not None and subtask.task_to_compute is not None:
+        validate_all_messages_identical([
```
rwrzesien: Isn't that already done in every view handler ?

concent_api/core/tests/test_unit_subtask_helpers.py
```diff
@@ -0,0 +1,50 @@
+from django.conf import settings
```
rwrzesien: Add typings.

concent_api/core/message_handlers.py

concent_api/core/subtask_helpers.py

concent_api/core/tests/test_auth_accept_or_reject_results.py
pawelkisielewicz: right, fixed

concent_api/core/tests/test_integration_accept_or_reject_results.py
pawelkisielewicz: generally ok.. but I think in this case is not good idea.
1. This function saves object to database. I think all fields should be given explicitly, not by default.
2. In production code `None` is never passed, it is not good default value. None is only passed in tests. If you insist to change it... but i would leave it as it is now.

concent_api/core/message_handlers.py

concent_api/core/message_handlers.py

concent_api/core/message_handlers.py
pawelkisielewicz: it was. Done

concent_api/core/tests/test_unit_subtask_helpers.py
pawelkisielewicz: done

concent_api/core/tests/test_auth_accept_or_reject_results.py
cameel: Looks like it's still not fixed.

concent_api/core/message_handlers.py
cameel: Looks like it's still not done. Have you pushed the latest version of the code?

cameel:
@dybi 
> Removing `store_or_update_subtask` has resulted in significant Code Duplication. The design of the solution should be changed.

We could keep the method as long as the query itself stays in the view. `store_or_update_subtask()` could just get the subtask as a parameter. What do you think about a solution like this?

concent_api/core/tests/test_integration_subtask_results_verify.py
```diff
@@ -74,14 +74,18 @@ def test_that_concent_responds_with_service_refused_when_verification_for_this_s
...
-                reverse('core:send'),
-                data=serialized_subtask_results_verify,
-                content_type='application/octet-stream',
-                HTTP_CONCENT_CLIENT_PUBLIC_KEY=self._get_encoded_provider_public_key(),
-                HTTP_CONCENT_OTHER_PARTY_PUBLIC_KEY=self._get_encoded_requestor_public_key(),
-            )
+        with mock.patch(
+                "core.message_handlers.payments_service.is_account_status_positive",
+                return_value=True
+        ):
```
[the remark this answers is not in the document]
pawelkisielewicz: done

concent_api/core/tests/test_integration_subtask_results_verify.py
```diff
@@ -141,26 +150,31 @@ def test_that_concent_responds_with_http_400_when_verification_received_in_faile
...
```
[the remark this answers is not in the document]
pawelkisielewicz: done

concent_api/core/tests/test_auth_accept_or_reject_results.py
```diff
@@ -144,20 +147,23 @@ def test_provider_forces_subtask_results_for_task_which_was_already_submitted_co
...
-        with freeze_time("2018-02-05 10:00:31"):
-            response = self.client.post(
-                reverse('core:send'),
-                data                                = serialized_force_subtask_results,
-                content_type                        = 'application/octet-stream',
-            )
+        with mock.patch(
+                'core.message_handlers.payments_service.is_account_status_positive',
+                side_effect=self.is_account_status_positive_true_mock
+        ):
```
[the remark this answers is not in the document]
pawelkisielewicz: done

pawelkisielewicz:
I don't think it is a good approach. As it is now it is easy to see what is happening, one decision and two separated ways to each case. And there is no duplication, there is a few "similar lines" by passing arguments to `store_subtask` and `update_subtask`. But now it is easy to see and readable. I don't think we should make our code more complicated only to save a few lines...

Good way to make `handle...` functions shorter is move validations from this functions away (they are the same or very similar in all handle functions. But if we wanted this it is separate issue :)

concent_api/core/message_handlers.py
```diff
@@ -519,18 +525,36 @@ def handle_send_force_subtask_results(client_message: message.concents.ForceSubt
...
+        )
+    else:
+        if subtask.state_enum == Subtask.SubtaskState.FORCING_ACCEPTANCE:
+            return message.concents.ServiceRefused(
+                reason=message.concents.ServiceRefused.REASON.DuplicateRequest,
+            )
+        subtask = update_subtask(
+            subtask=subtask,
+            state=Subtask.SubtaskState.FORCING_ACCEPTANCE,
+            next_deadline=forcing_acceptance_deadline + settings.CONCENT_MESSAGING_TIME,
```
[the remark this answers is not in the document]
pawelkisielewicz: I think `update` suggests` function gets subtask, makes something with it-updated and returns. I need subtask returned to pass it to `store_pending_message`. It can be done without passing subtask, but in `store_pending_message` it had to been received from database. I don't think we want it.. 

If it is not obvious maybe leave the code as it is now and comment it.

concent_api/core/subtask_helpers.py
```diff
@@ -185,3 +186,22 @@ def are_subtask_results_accepted_messages_signed_by_the_same_requestor(subtask_r
...
+    subtask_id: str,
+    optional_condition_dict: Dict[str, str] = None,
+) -> Optional[Subtask]:
+
+    conditions_dict = {'subtask_id': subtask_id}
+
+    if optional_condition_dict is not None:
+        conditions_dict.update(optional_condition_dict)
+
+    query = Subtask.objects.select_for_update().filter(**conditions_dict)
```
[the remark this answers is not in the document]
pawelkisielewicz: done.

concent_api/core/subtask_helpers.py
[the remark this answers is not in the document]
pawelkisielewicz: If you say so.. I know `= None` shows that parameter is optional, i added it to show it more clearly. changed

concent_api/core/subtask_helpers.py
```diff
@@ -185,3 +186,22 @@ def are_subtask_results_accepted_messages_signed_by_the_same_requestor(subtask_r
...
     )
     return are_all_signed_by_requestor
+
+
+def get_one_or_none_subtasks_from_database(
+    subtask_id: str,
+    optional_condition_dict: Dict[str, str] = None,
+) -> Optional[Subtask]:
+
+    conditions_dict = {'subtask_id': subtask_id}
```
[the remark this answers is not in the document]
pawelkisielewicz: right, changed

concent_api/core/subtask_helpers.py
```diff
@@ -185,3 +186,22 @@ def are_subtask_results_accepted_messages_signed_by_the_same_requestor(subtask_r
         ) for subtask_results_accepted in subtask_results_accepted_list
     )
     return are_all_signed_by_requestor
+
+
+def get_one_or_none_subtasks_from_database(
+    subtask_id: str,
+    optional_condition_dict: Dict[str, str] = None,
+) -> Optional[Subtask]:
```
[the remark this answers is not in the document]
pawelkisielewicz: I did it similar way at the beginning, but it was the cause of lot, very lot of IDE warnings. IDE don't know what type it is (generic function returns `-> object`) and shows warnings (e.g. some_object.subtask_id - IDE sees it like this). 

I would leave it like it is now, it is good in mostly cases, and if we would do something different we can do it manually with `model.objects.filter(something)`.. 

What do you think??

cameel:
@dybi what do you think?

For me either way is fine, I suggested keeping `store_or_update...` to slightly reduce duplication though personally I too prefer having some of the logic clearly visible even if it needs to be repeated. And here the duplication mostly comes from the fact that the calls themselves are long and take many parameters.

As for extracting validations - I think they should stay. But if you wanted to deduplicate them you could just create bigger, combined validation methods that call other, more specific ones.

concent_api/core/subtask_helpers.py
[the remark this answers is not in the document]
cameel: I really think this function needs to be changed because it hides too much and is too much of a special case.

About warnings - surely there's some way around them. Can you show them to me? Python is a dynamic language. Doing things like returning arbitrary objects or duck typing by calling a method on the object whose type you don't know is nothing special.

concent_api/core/message_handlers.py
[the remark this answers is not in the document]
cameel: What is mean is that the function does not need to return `subtask` to update it. Returning it in addition to updating it only makes it confusing because that's not a typical construct.

concent_api/core/message_handlers.py
```diff
@@ -519,18 +531,42 @@ def handle_send_force_subtask_results(client_message: message.concents.ForceSubt
...
+            task_to_compute=client_message.ack_report_computed_task.report_computed_task.task_to_compute,
+            report_computed_task=client_message.ack_report_computed_task.report_computed_task,
+            ack_report_computed_task=client_message.ack_report_computed_task,
+        )
+    else:
+        if subtask.state_enum == Subtask.SubtaskState.FORCING_ACCEPTANCE:
+            return message.concents.ServiceRefused(
+                reason=message.concents.ServiceRefused.REASON.DuplicateRequest,
+            )
+        if subtask is not None:
```
rwrzesien: This condition is redundant - you are already in `else` part where `subtask` must exist.

concent_api/core/message_handlers.py
```diff
@@ -1356,50 +1331,64 @@ def handle_send_subtask_results_verify(
...
+            Subtask.SubtaskState.FAILED,
         ]:
             raise Http400(
                 "SubtaskResultsVerify is not allowed in current state",
                 error_code=ErrorCode.QUEUE_SUBTASK_STATE_TRANSITION_NOT_ALLOWED,
             )
 
-    except Subtask.DoesNotExist:
-        pass
+        if subtask is not None:
```
rwrzesien: Same here.

concent_api/core/subtask_helpers.py
```diff
@@ -185,3 +187,18 @@ def are_subtask_results_accepted_messages_signed_by_the_same_requestor(subtask_r
         ) for subtask_results_accepted in subtask_results_accepted_list
     )
     return are_all_signed_by_requestor
+
+
+def get_and_lock_subtask(
+    subtask_id: str,
+    conditions: Dict[str, Any] = None,
+) -> Optional[Subtask]:
+    assert 'subtask_id' not in conditions
```
rwrzesien: Maybe it would be better to
1. Remove `subtask_id` from parameters and this `assert`.
2. Assert that `subtask_id` key is in `conditions` and its value is a string.
@cameel What do you think ?

Or even better - I saw there was a discussion to make it more generic. How about maybe then making following parameters:
- `query_set_or_model` - as mentioned in @cameel comment
- `unique_id_field_name` - `subtask_id` in this case.
- `unique_id_field_type` - `str` in this case.
- `conditions` - `str` in this case.

It is just a proposal and please treat it as one. I don't want to hold this pull request due to all possible variations of how this function can be implemented (as it depends on how it will be used mostly).

concent_api/core/tests/test_auth_accept_or_reject_results.py
```diff
@@ -129,13 +129,16 @@ def test_provider_forces_subtask_results_for_task_which_was_already_submitted_co
...
         )
-
-        with freeze_time("2018-02-05 10:00:31"):
-            response = self.client.post(
-                reverse('core:send'),
-                data                                = different_serialized_force_subtask_results,
-                content_type                        = 'application/octet-stream',
-            )
+        with mock.patch(
+                'core.message_handlers.payments_service.is_account_status_positive',
```
rwrzesien: Please fix indentation.

@pawelkisielewicz I see this problem repeating in your pull requests. What IDE do you use?

concent_api/core/subtask_helpers.py
cameel: > Maybe it would be better to
> 1. Remove `subtask_id` from parameters and this `assert`.
> 2. Assert that `subtask_id` key is in `conditions` and its value is a string.
> @cameel What do you think ?

It's a good idea.

> Or even better - I saw there was a discussion to make it more generic. How about maybe then making following parameters:
> - `query_set_or_model` - as mentioned in @cameel comment
> - `unique_id_field_name` - `subtask_id` in this case.
> - `unique_id_field_type` - `str` in this case.
> - `conditions` - `str` in this case.

I still think that what I described in https://github.com/golemfactory/concent/pull/778#discussion_r215346045 is the way to go. It's very simple, flexible and reusable.

concent_api/core/message_handlers.py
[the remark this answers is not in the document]
pawelkisielewicz: If it is not typical I don't discuss any more. As you said at the beginning we will repair it in separately. For now as a quick repair i changed the name to `update_and_return_updated_subtask`. It is not nice name but shows that subtask is also returned.. Is it OK?

concent_api/core/message_handlers.py
pawelkisielewicz: right, done

dybi:
@cameel , @pawelkisielewicz 
1) as for duplication, I can accept the solution @pawelkisielewicz has proposed (but if the duplication starts to grow any larger it must be tackled)
2) as for validations, I would be for separating them: business logic should stay in the handle function, but more general validations (like proper signatures, etc) could be extracted and done just in one place (so the handle functions will be cleared out of the clutter)

concent_api/core/message_handlers.py

concent_api/core/subtask_helpers.py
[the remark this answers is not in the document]
pawelkisielewicz: as discussed, done

concent_api/core/subtask_helpers.py

concent_api/core/tests/test_auth_accept_or_reject_results.py
pawelkisielewicz: I use Pycharm. It fixes indentation, but I have to check settings to my Ide does it our style instead of it's style. Thanks for tip :)

concent_api/core/message_handlers.py
```diff
@@ -419,22 +411,52 @@ def handle_send_force_get_task_result(client_message: message.concents.ForceGetT
...
+            report_computed_task=client_message.report_computed_task,
+            force_get_task_result=client_message,
+        )
+
+    else:
+        if subtask.state_enum == Subtask.SubtaskState.FORCING_RESULT_TRANSFER:
+            return message.concents.ServiceRefused(
+                reason=message.concents.ServiceRefused.REASON.DuplicateRequest,
+            )
+        if subtask is not None:
```
cameel: This condition is always true.

concent_api/core/message_handlers.py
```diff
@@ -419,22 +411,52 @@ def handle_send_force_get_task_result(client_message: message.concents.ForceGetT
...
+            if task_to_compute is not None and subtask.task_to_compute is not None:
```
cameel: Can `subtask.task_to_compute` even be `None`? And can we even get this far in the handler without failing one of the validations earlier?

I think this whole condition should be an `assert` because it will never fail unless there's a serious bug somewhere. After you assert you can just call `validate_all_messages_identical()` without checking the condition.

concent_api/core/message_handlers.py
```diff
@@ -519,18 +532,42 @@ def handle_send_force_subtask_results(client_message: message.concents.ForceSubt
...
+            task_to_compute=client_message.ack_report_computed_task.report_computed_task.task_to_compute,
+            report_computed_task=client_message.ack_report_computed_task.report_computed_task,
+            ack_report_computed_task=client_message.ack_report_computed_task,
+        )
+    else:
+        if subtask.state_enum == Subtask.SubtaskState.FORCING_ACCEPTANCE:
+            return message.concents.ServiceRefused(
+                reason=message.concents.ServiceRefused.REASON.DuplicateRequest,
+            )
+        if task_to_compute is not None and subtask.task_to_compute is not None:
```
cameel: Same as above. This should be an `assert`.

concent_api/core/message_handlers.py
```diff
@@ -1356,50 +1332,64 @@ def handle_send_subtask_results_verify(
...
+            Subtask.SubtaskState.FAILED,
         ]:
             raise Http400(
                 "SubtaskResultsVerify is not allowed in current state",
                 error_code=ErrorCode.QUEUE_SUBTASK_STATE_TRANSITION_NOT_ALLOWED,
             )
 
-    except Subtask.DoesNotExist:
-        pass
+        if task_to_compute is not None and subtask.task_to_compute is not None:
```
cameel: This should be an `assert`.

concent_api/core/subtask_helpers.py
```diff
@@ -1,10 +1,13 @@
...
+from base64 import b64encode
 from logging import getLogger
+from typing import Any
 from typing import List
 from typing import Optional
+from typing import Union
 
-from django.db.models           import Q
-from django.utils               import timezone
+import django
```
cameel: Please use the `from django.db.models import Model` form. It's already used for other names imported from `django` in this module and importing stuff in two different ways in the same module can be confusing so avoid it unless there's no other way.

concent_api/core/subtask_helpers.py
```diff
@@ -185,3 +188,15 @@ def are_subtask_results_accepted_messages_signed_by_the_same_requestor(subtask_r
...
+def get_one_or_none(
+    query_set_or_model: Union[django.db.models.Model, django.db.models.QuerySet],
+    **conditions: Any
+)-> Optional[django.db.models.Model]:
+    if isinstance(query_set_or_model, django.db.models.base.ModelBase):
+        instances = query_set_or_model.objects.filter(**conditions)
+        return None if len(instances) == 0 else instances[0]
+    else:
+        instances = query_set_or_model.filter(**conditions)
+        return None if len(instances) == 0 else instances[0]
```
cameel: This function is meant to be used only in cases when there's at most one instance.

You should add an `assert` to enforce that.
``` python
assert len(instances) <= 1
```

BTW, any code that gets the first element from a collection (e.g. `instances[0]`) should be a big, red flag to you. If you take element `[0]` always ask yourself - what if there's more than one element? Usually the answer is that there can be only one element and you should add an `assert` to make that clear.

concent_api/core/tests/test_auth_accept_or_reject_results.py
cameel: Still not fixed here.

pawelkisielewicz: don't know why.. sorry for this, done

concent_api/core/message_handlers.py
pawelkisielewicz: fixed

concent_api/core/message_handlers.py
pawelkisielewicz: it can't be none, it was the whole block from other place, I moved it here, but forgot to remove unnecessary any more condition. Case ` if subtask is None:` is checked above and there is block for this situation. Here in else (subtask not none) assert is not necessary.

concent_api/core/message_handlers.py
pawelkisielewicz: the same situation as above: 
` if subtask is None:`
`.....`
`else:`
`... <- I think it is unnecessary to assert if subtask is not none in this place. It is obvious..`

concent_api/core/message_handlers.py
pawelkisielewicz: same as above

concent_api/core/subtask_helpers.py

concent_api/core/subtask_helpers.py
pawelkisielewicz: Right, I thought about this, what will happen if it will be more then one element.. But I agree with you, this function is to return one or none, so it should be used properly. Done

concent_api/core/message_handlers.py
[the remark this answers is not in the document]
cameel: It's OK for now.

But I think we should change it eventually. This function should not return anything.

Issue #746, opened 2018-08-21
Create shell script for generating requirements-development.lock


kbeker:
Script should create file `requirements-development.lock` with newest `requirements-development.txt`.

Script should do:
1. Create clean virtual environment.
2. Install requirements.lock
3. Install requirements-development.txt
4. Generate requirements-development.lock from `pip freeze`
5. Delete virtual environment.

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/159920388

cameel:
After talking with @kbeker, this needs to work a bit differently.

The script for generating `requirements-development.lock` file should work just like the one for `requirements.lock`. In fact, this should probably be a single script with path to the requirements file as a parameter. It should assume that the packages are already installed and just need to be frozen into the `.lock` file.

But we may want to have additional script that creates a clean virtualenv, installs the packages and calls the script described above. Basically, a script version of our procedure for regenerating the requirements files.

dybi:
@kbeker, please update the description, so it reflects what is really being done in this issue

cameel:
@kbeker What about the description?

PR #781, opened 2018-09-04
Change generate-requirements-lock script to generate .lock files from parameter

kbeker:
Resolves https://github.com/golemfactory/concent/issues/746

### DIFF
--- a/concent_api/generate-requirements-lock.sh
+++ b/concent_api/generate-requirements-lock.sh
@@ -1,11 +1,34 @@
 #!/bin/bash -e
 
+file_name="$1"
+
+
+display_help() {
+    echo "This script must be run within virtual environment"
+    echo "Script accepts exactly one parameter (requirements file)"
+    echo "Usage example: ./generate-requirements-lock.sh requirements.lock"
+}
+
+
+if [[ "$file_name" == "--help" || "$file_name" == "-h" || "$#" != 1 ]]
+then
+	display_help
+	exit 0
+fi
+
 if [[ "$VIRTUAL_ENV" != "" ]]
 then
-    echo "--find-links https://builds.golem.network/simple/pyelliptic" > ${BASH_SOURCE%/*}/requirements.lock
-    pip freeze >> ${BASH_SOURCE%/*}/requirements.lock
-    echo "Done"
+    if [ -s "$file_name" ] && [[ "$file_name" == *.lock ]]
+    then
+        additional_lines=$(cat $file_name | grep "^--find-links")
+        echo $additional_lines > $file_name
+        pip freeze >> $file_name
+        echo "Done"
+    else
+        echo "$file_name must be a python packages requirements .lock file"
+        exit 1
+    fi
 else
-    echo "Script should be run within virtual environment"
+    display_help
     exit 1
 fi


concent_api/generate-requirements-lock.sh
```diff
@@ -2,9 +2,21 @@
 
 if [[ "$VIRTUAL_ENV" != "" ]]
 then
-    echo "--find-links https://builds.golem.network/simple/pyelliptic" > ${BASH_SOURCE%/*}/requirements.lock
-    pip freeze >> ${BASH_SOURCE%/*}/requirements.lock
-    echo "Done"
+    if [ "$#" -ne 1 ]
```
bartoszbetka: This statement is not needed here. You are using the first argument, so you do not need to check the number of parameters.

concent_api/generate-requirements-lock.sh
```diff
@@ -2,9 +2,21 @@
...
+    then
+        echo "Illegal number of parameters"
+        exit 1
+    else
+        if ! [[ -f $1 ]]
```
bartoszbetka: Please add variable to first argument at the beginning of the file. variable=$1

concent_api/generate-requirements-lock.sh
michalpiasecki1: there could be no parameters ;)

concent_api/generate-requirements-lock.sh
```diff
@@ -2,9 +2,21 @@
 
 if [[ "$VIRTUAL_ENV" != "" ]]
 then
-    echo "--find-links https://builds.golem.network/simple/pyelliptic" > ${BASH_SOURCE%/*}/requirements.lock
-    pip freeze >> ${BASH_SOURCE%/*}/requirements.lock
-    echo "Done"
+    if [ "$#" -ne 1 ]
+    then
+        echo "Illegal number of parameters"
```
dybi: Please specify what parameters are needed/accepted

concent_api/generate-requirements-lock.sh
bartoszbetka: This statement is not correct. It should contain single square brackets. Also, you checked wrong condition. This should look like this 
```bash
if [ -f $1 ]
```

bartoszbetka: also, use option `-s` which is check that file is not empty.

concent_api/generate-requirements-lock.sh
bartoszbetka: This case can be covered by checking if the file is empty, your decision.

concent_api/generate-requirements-lock.sh
```diff
@@ -2,9 +2,21 @@
...
+    else
+        if ! [[ -f $1 ]]
+        then
+            additional_lines=$(cat $1 | grep "^--")
+            echo $additional_lines > $1
+            pip freeze >> $1
+            echo "Done"
+        else
+            echo "$1 must be a file"
+            exit 1
```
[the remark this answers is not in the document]
kbeker: Yea, it was my fault that I commit wrong version of code. Done

concent_api/generate-requirements-lock.sh
kbeker: I left it as it is for greater warnings readability.

bartoszbetka: Please, modify it for readability.
```bash
[[ "$#" != 1 ]]
```

concent_api/generate-requirements-lock.sh
```diff
@@ -1,10 +1,25 @@
...
-    pip freeze >> ${BASH_SOURCE%/*}/requirements.lock
-    echo "Done"
+    if [ "$#" -ne 1 ]
+    then
+        echo "Script should be called with exactly one parameter"
+        exit 1
+    else
+        if [ -s "$file_name" ] && [[ "$file_name" == *.lock ]]
+        then
+            additional_lines=$(cat $file_name | grep "^--")
```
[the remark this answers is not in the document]
kbeker: I would stay with `grep "^--"` because if we want to use this script with requirements files which would have different lines than `--find-links` it won't work.

concent_api/generate-requirements-lock.sh
kbeker: Done

concent_api/generate-requirements-lock.sh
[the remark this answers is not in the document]
cameel: Copying all those lines blindly is not a good idea. What if the file contains this:

``` bash
--editable git://github.com/golemfactory/golem-messages.git@v2.12.0#egg=golem_messages
```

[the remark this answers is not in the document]
kbeker: Ok, so changed to version:
```
additional_lines=$(cat $file_name | grep "^--find-links")
```

concent_api/generate-requirements-lock.sh
kbeker: Done

concent_api/generate-requirements-lock.sh
```diff
@@ -1,11 +1,42 @@
...
+
+
+display_help() {
+    echo "This script must be run within virtual environment"
+    echo "Script accepts exactly one parameter"
+    echo "Usage example: ./generate-requirements-lock.sh requirements.lock"
+}
+
+
+if [[  "$1" == "--help" ||  "$1" == "-h" ]]
```
michalpiasecki1: you can also print help, even when run without parameters ;)

kbeker: Added. Done :)

concent_api/generate-requirements-lock.sh
```diff
@@ -1,11 +1,42 @@
...
+
+
+display_help() {
+    echo "This script must be run within virtual environment"
+    echo "Script accepts exactly one parameter"
+    echo "Usage example: ./generate-requirements-lock.sh requirements.lock"
+}
+
+
+if [[  "$1" == "--help" ||  "$1" == "-h" || "$#" == 0 ]]
```
bartoszbetka: Use here also "$file_name".

kbeker: Done

Issue #103, opened 2018-02-08
Mypy doesn't catch wrong typing.


rwrzesien:
We have a function signature

```def handle_send_force_get_task_result(request, client_message: message.concents.ForceGetTaskResult) -> message.concents:```

which returns either `ForceGetTaskResultRejected` or `ForceGetTaskResultAck`. For some unknown reason, mypy is not complaining that returned type is not defined correctly. We should find out why.

The proper way of defining this signature according to mypy documentation should be

```def handle_send_force_get_task_result(request, client_message: message.concents.ForceGetTaskResult) -> Union[ForceGetTaskResultRejected, ForceGetTaskResultAck]:```

But for some reason, with that signature it doesn't complain if you put `return 11` somewhere in that function. So maybe the issue is not with signature itself.

It might be as well bug in mypy. I consider it not being mature library yet, but the most advanced from all currently available.

**Comment by kbeker** [comment]
I'm now reading docs from `mypy` and I'm thinking exactly which options do we want launch on? Or maybe what do we want to achieve by adding options to `mypy`? There is a lot of options and when I turned one of them on, console just got crazy from amount of warnings and errors from our code. 

And next thing is that during this task should I also fix all errors and warnings that will `mypy` show? 

Individually recommend a few settings to turn on:
`disallow_untyped_calls = True`
`disallow_untyped_defs = True`
`warn_no_return = True`
`warn_return_any = True`
`warn_unused_ignores = True`
`warn_redundant_casts = True`
`scripts_are_modules = True`

Just first two options returns **1056** errors and warnings. All of settings which are mentioned above gives in total **1076** warnings and errors.

Also I think that we should add to `full_check.sh` removing `.mypy_cache` because I saw that this makes  sometimes strange results of `mypy` work

All options I have found on [mypy docs page](http://mypy.readthedocs.io/en/latest/config_file.html)

**Comment by rwrzesien** [comment]
I would say we should discuss each one separately. For me those two look interesting and I would like to have them:
```
warn_unused_ignores = True
warn_redundant_casts = True
```

But I am against turning first two instantly, because we won't be able to tackle all those warnings for some time even if we put all our forces into it. Lets follow the current pattern in which we try to improve those places when we are doing some work around them or someone catches them during review.

[...]

PR #766, opened 2018-08-29
Feature middleman message routing

dybi:
Resolves: https://github.com/golemfactory/concent/issues/629
Resolves: https://github.com/golemfactory/concent/issues/679

PR #783, opened 2018-09-04
[WIP] Chore mypy fixes

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/103

TODO: Check and update MiddleMan after merging https://github.com/golemfactory/concent/pull/766

### DIFF
--- a/concent_api/api_testing_common.py
+++ b/concent_api/api_testing_common.py
@@ -1,3 +1,9 @@
+from typing import Any
+from typing import Callable
+from typing import Optional
+from typing import Sequence
+from typing import Tuple
+from typing import Union
 import argparse
 import datetime
 import http.client
@@ -12,6 +18,7 @@ from golem_messages.factories.tasks import ComputeTaskDefFactory
 from golem_messages.factories.tasks import TaskToComputeFactory
 from golem_messages.message import Message
 from golem_messages.message.concents import ClientAuthorization
+from golem_messages.message.tasks import TaskToCompute
 from golem_messages.shortcuts import dump
 from golem_messages.shortcuts import load
 from golem_messages.utils import encode_hex
@@ -40,13 +47,13 @@ class count_fails(object):
     instances = []  # type: ignore
     number_of_run_tests = 0
 
-    def __init__(self, function):
-        self._function     = function
+    def __init__(self, function: Callable) -> None:
+        self._function = function
         self.__name__ = function.__name__
-        self.failed  = False
+        self.failed = False
         count_fails.instances.append(self)
 
-    def __call__(self, *args, **kwargs):
+    def __call__(self, *args: Any, **kwargs: Any) -> Any:
         try:
             print("Running TC: " + self.__name__)
             count_fails.number_of_run_tests += 1
@@ -57,21 +64,21 @@ class count_fails(object):
             self.failed = True
 
     @classmethod
-    def get_fails(cls):
+    def get_fails(cls) -> int:
         return sum([instance.failed for instance in cls.instances])
 
     @classmethod
-    def print_fails(cls):
+    def print_fails(cls) -> None:
         print(f'Total failed tests : {cls.get_fails()} out of {cls.number_of_run_tests}')
 
 
-def assert_condition(actual, expected, error_message = None):
+def assert_condition(actual: Any, expected: Any, error_message: str=None) -> None:
     message = error_message or f"Actual: {actual} != expected: {expected}"
     if actual != expected:
         raise TestAssertionException(message)
 
 
-def print_golem_message(message, indent = 4):
+def print_golem_message(message: Message, indent: int=4) -> None:
     assert isinstance(message, Message)
     HEADER_FIELDS  = ['timestamp', 'encrypted', 'sig']
     PRIVATE_FIELDS = {'_payload', '_raw'}
@@ -90,7 +97,7 @@ def print_golem_message(message, indent = 4):
             print('{}{:30} = {}'.format(' ' * indent, field, value))
 
 
-def validate_response_status(actual_status_code, expected_status):
+def validate_response_status(actual_status_code: int, expected_status: Optional[int]=None) -> None:
     if expected_status is not None:
         assert_condition(
             actual_status_code,
@@ -99,7 +106,12 @@ def validate_response_status(actual_status_code, expected_status):
         )
 
 
-def validate_response_message(encoded_message, expected_message_type, private_key, public_key):
+def validate_response_message(
+    encoded_message: bytes,
+    expected_message_type: Message,
+    private_key: bytes,
+    public_key: bytes,
+) -> None:
     if expected_message_type is not None:
         decoded_message = try_to_decode_golem_message(private_key, public_key, encoded_message)
         assert_condition(
@@ -109,7 +121,7 @@ def validate_response_message(encoded_message, expected_message_type, private_ke
         )
 
 
-def validate_content_type(actual_content_type, expected_content_type):
+def validate_content_type(actual_content_type: str, expected_content_type: Optional[str]=None) -> None:
     if expected_content_type is not None:
         assert_condition(
             actual_content_type,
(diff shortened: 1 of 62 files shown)


concent_api/common/decorators.py
```diff
@@ -37,14 +41,14 @@
...
-def require_golem_auth_message(view):
+def require_golem_auth_message(view: Callable) -> Callable:
     """
     Decorator for authenticating golem clients
     Unpacks authorization message signed with the key it contains
     proof that the client indeed has the private part of that key
     """
     @wraps(view)
-    def wrapper(request, *args, **kwargs):
+    def wrapper(request: HttpRequest, *args: Any, **kwargs: Any) -> Union[HttpResponse, JsonResponse]:
```
pawelkisielewicz: this wrapper returns also `view`, so it should be `Union[HttpResponse, JsonResponse, Callable]:`

concent_api/common/decorators.py
```diff
@@ -100,14 +104,14 @@ def wrapper(request, *args, **kwargs):
...
-def require_golem_message(view):
+def require_golem_message(view: Callable) -> Callable:
     """
     Decorator for view accepting golem messages
     Unpacks golem message signed with the key it contains
     proof that the client indeed has the private part of that key
     """
     @wraps(view)
-    def wrapper(request, *args, **kwargs):
+    def wrapper(request: HttpRequest, *args: Any, **kwargs: Any) -> Union[HttpResponse, JsonResponse]:
```
pawelkisielewicz: I'm not sure in this place, you worked with decorators and wrapers longer than me :)

wrapper returns view, so it should be:
`-> Union[HttpResponse, JsonResponse, Callable]:` 

second: if wrapper inside returns `Union[...]`, shouldn't be returned the same by the  `require_golem_message`?

concent_api/common/helpers.py
```diff
@@ -74,7 +74,7 @@ def get_field_from_message(golem_message: message.base.Message, field_name: str)
     Returns None if field is not available.
     """
 
-    def check_task_id(golem_message):
+    def check_task_id(golem_message: message.Message) -> Union[str, None]:
```
pawelkisielewicz: shouldn't be `Optional[str]` ?

concent_api/common/tests/test_api_view.py
```diff
@@ -1,28 +1,30 @@
...
-from common.decorators               import handle_errors_and_responses
-from common.helpers                  import get_current_utc_timestamp
-from common.testing_helpers          import generate_ecc_key_pair
+from core.exceptions import Http400
+from core.models import Client
+from common.constants import ErrorCode
+from common.decorators import require_golem_message
+from common.decorators import handle_errors_and_responses
+from common.helpers import get_current_utc_timestamp
+from common.testing_helpers import generate_ecc_key_pair
```
pawelkisielewicz: Why are you reordering imports in this pull request? I don't say it is bad, but in future separate commit would be nice :)

concent_api/core/admin.py
```diff
@@ -67,21 +70,21 @@ class SubtaskAdmin(ModelAdminReadOnlyMixin, admin.ModelAdmin):
     ]
 
     @classmethod
-    def get_provider_public_key(cls, obj):
+    def get_provider_public_key(cls, obj: Subtask) -> str:
```
pawelkisielewicz: if it is Subtask instance, maybe change `obj -> subtask`. It is in few places

concent_api/core/admin.py
```diff
@@ -107,14 +110,14 @@ class PendingResponseAdmin(ModelAdminReadOnlyMixin, admin.ModelAdmin):
...
-    def get_subtask_subtask_id(cls, obj):
+    def get_subtask_subtask_id(cls, obj: PendingResponse) -> str:
         if obj.subtask is not None:
             return obj.subtask.subtask_id
         return '-not available-'
     get_subtask_subtask_id.short_description = 'Subtask id'  # type: ignore
 
     @classmethod
-    def get_client_public_key(cls, obj):
+    def get_client_public_key(cls, obj: PendingResponse) -> str:
```
pawelkisielewicz: similar situation `obj -> pending_response` ?

concent_api/core/tests/utils.py
```diff
@@ -400,12 +414,12 @@ def _test_last_stored_messages(self, expected_messages, task_id, subtask_id):
...
-        subtask_id,
-        client_public_key_out_of_band                   = None,
-        expected_pending_responses_receive              = None,
-        expected_pending_responses_receive_out_of_band  = None,
-    ):
+        client_public_key: str,
+        subtask_id: str,
+        client_public_key_out_of_band: Optional[str]=None,
+        expected_pending_responses_receive: list=None,
+        expected_pending_responses_receive_out_of_band: list=None,
```
pawelkisielewicz: shouldn't be `Optional[list] = None` ?

concent_api/core/admin.py
rwrzesien: :+1: done

concent_api/core/admin.py
rwrzesien: :+1: done

concent_api/common/helpers.py
rwrzesien: Works ok too :+1:

concent_api/common/decorators.py
rwrzesien: > second: if wrapper inside returns Union[...], shouldn't be returned the same by the require_golem_message?

The `require_golem_message` is the actual decorator, and it returns decorated function, that is why returned type is `Callable`.

> wrapper returns view, so it should be:

It returns result of the call to view function (please check the line before last `return view(...)`) so the returned type is either `HttpResponse` or `JsonResponse`.

concent_api/common/decorators.py
rwrzesien: Same as below, it returns what call to `view()` returns, which is `HttpResponse`.

concent_api/command_line_tool/tests/test_key_manager.py
```diff
@@ -14,45 +14,45 @@
...
 class TestKeyManager(TestCase):
-    def setUp(self):
+    def setUp(self) -> None:
         key_manager.REQUESTOR_PRIVATE_KEY = None
         key_manager.REQUESTOR_PUBLIC_KEY = None
         key_manager.PROVIDER_PRIVATE_KEY = None
         key_manager.PROVIDER_PUBLIC_KEY = None
 
-    def test_that_new_instance_generates_new_keys_for_requestor_when_globals_not_set(self):
+    def test_that_new_instance_generates_new_keys_for_requestor_when_globals_not_set(self) -> None:
```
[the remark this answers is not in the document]
rwrzesien: Already discussed, and we decided not to put typing in tests.

concent_api/command_line_tool/tests/test_key_manager.py
```diff
@@ -14,45 +14,45 @@
 
 
 class TestKeyManager(TestCase):
-    def setUp(self):
+    def setUp(self) -> None:
```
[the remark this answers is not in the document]
rwrzesien: Already discussed, and we decided not to put typing in tests.

concent_api/command_line_tool/key_manager.py
```diff
@@ -41,32 +42,32 @@ def are_keys_predifined(party):
...
+def _get_predefined_keys(party: str) -> Tuple[None, None]:
     formatted_party = party.upper()
     public_key = f"{formatted_party}_PUBLIC_KEY"
     private_key = f"{formatted_party}_PRIVATE_KEY"
     return CONFIG[private_key], CONFIG[public_key]
 
 
 class KeyManager(object):
-    def __init__(self):
+    def __init__(self) -> None:
```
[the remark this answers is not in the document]
rwrzesien: Actually mypy is creating warning for those, that is why I have added it. But you are right, maybe one day it will be fixed :)

concent_api/command_line_tool/key_manager.py
```diff
@@ -41,32 +42,32 @@ def are_keys_predifined(party):
     return is_pulbic_key_predefined and is_private_key_predefined
 
 
-def _get_predefined_keys(party):
+def _get_predefined_keys(party: str) -> Tuple[None, None]:
```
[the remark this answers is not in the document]
rwrzesien: Yes, as mypy reads it, it thinks that those are `None`. But this could be misleading, so it would be better to leave just `Tuple`. :+1:

concent_api/api_testing_common.py
```diff
@@ -120,17 +132,17 @@ def validate_content_type(actual_content_type, expected_content_type):
...
+    endpoint: str,
+    private_key: bytes,
+    public_key: bytes,
+    data: Optional[Union[bytes, str]]=None,
+    headers: Optional[dict]=None,
+    expected_status: Optional[int]=None,
+    expected_message_type: Optional[Message]=None,
+    expected_content_type: Optional[str]=None
+) -> Union[None, Message]:
+    def _prepare_data(data: Optional[Union[bytes, str]]=None) -> Union[str, bytes]:
```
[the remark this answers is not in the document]
rwrzesien: > I would do it different. In api_request signature data can be None, but in real it has no sense and we never use api_request without data. Without golem message such request will not pass validations.

It would require changing the signature of `api_request` which could be a candidate for a separate request, because it changes functionality (`None` was there already) and would like to avoid doing this in this pull request. But I think it should be still possible to send this kind of request to Concent. For example, we might want to check what will happen if `receive` is being hit without authentication message.

Issue #776, opened 2018-09-03
Update `task_id` and `subtask_id` in tests that are not uuid


rwrzesien:
There are still left a lot of `task_id` and `subtask_id` in tests which are defined as not uuid strings. Please fix them all.

Also, it is worth checking why tests were passing with wrong `task_id` and `subtask_id`. Maybe it would be good to add `error_code` checking in those places.

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/160233197

PR #784, opened 2018-09-04
Update `task_id` and `subtask_id` in tests that are not uuid

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/776

### DIFF
--- a/concent_api/command_line_tool/tests/test_message_extractor.py
+++ b/concent_api/command_line_tool/tests/test_message_extractor.py
@@ -8,6 +8,7 @@ from assertpy import assert_that
 from command_line_tool.message_extractor import convert_message_name
 from command_line_tool.message_extractor import MessageExtractor
 from command_line_tool.message_extractor import split_uppercase
+from core.tests.utils import generate_uuid
 
 
 class TestSplitUppercase(TestCase):
@@ -27,8 +28,8 @@ class TestMessageExtractor(TestCase):
         self.requestor_ethereum_public_key = "b'T8ER1hrI9hH0Zyu/m6u9H2K4rc1mC/dlCqK0PSzBSwvKBdHfysmrIIsQMyvecDU+TIkhDmq93Olfo5h2FSGRjw=='"
         self.requestor_public_key = "mrdOXP6Owe3i48G29RHPVU6wewvuUNuNxFB+kPTDnmI21+p5ShttCSUAbHOVm8OIUF9hEluc1+lICJZGkFSSOA=="
         self.deadline = 1510394400
-        self.subtask_id = "12"
-        self.task_id = "2"
+        self.subtask_id = generate_uuid()
+        self.task_id = generate_uuid()
 
     def test_that_exception_is_raised_when_wrong_dict_is_given(self):
         input_data = {
--- a/concent_api/common/tests/test_api_view.py
+++ b/concent_api/common/tests/test_api_view.py
@@ -16,13 +16,14 @@ from golem_messages.factories       import tasks
 from golem_messages.factories.tasks import ComputeTaskDefFactory
 from golem_messages.utils import encode_hex
 
-from core.exceptions                import Http400
-from core.models                    import Client
-from common.constants                import ErrorCode
-from common.decorators               import require_golem_message
-from common.decorators               import handle_errors_and_responses
-from common.helpers                  import get_current_utc_timestamp
-from common.testing_helpers          import generate_ecc_key_pair
+from common.constants import ErrorCode
+from common.decorators import require_golem_message
+from common.decorators import handle_errors_and_responses
+from common.helpers import get_current_utc_timestamp
+from common.testing_helpers import generate_ecc_key_pair
+from core.exceptions import Http400
+from core.models import Client
+from core.tests.utils import generate_uuid
 
 
 (CONCENT_PRIVATE_KEY,   CONCENT_PUBLIC_KEY)   = generate_ecc_key_pair()
@@ -38,13 +39,13 @@ class ApiViewTestCase(TestCase):
     def setUp(self):
         self.request_factory = RequestFactory()
         self.want_to_compute = message.WantToComputeTask(
-            node_name           = 1,
-            task_id             = 2,
-            perf_index          = 3,
-            price               = 4,
-            max_resource_size   = 5,
-            max_memory_size     = 6,
-            num_cores           = 7,
+            node_name=1,
+            task_id=generate_uuid(),
+            perf_index=3,
+            price=4,
+            max_resource_size=5,
+            max_memory_size=6,
+            num_cores=7,
         )
         self.message_to_view = {
             "node_name":            self.want_to_compute.node_name,             # pylint: disable=no-member
--- a/concent_api/core/tests/test_auth_force_payment.py
+++ b/concent_api/core/tests/test_auth_force_payment.py
@@ -5,11 +5,11 @@ from django.urls            import reverse
 from freezegun              import freeze_time
 from golem_messages         import message
 
+from common.testing_helpers import generate_ecc_key_pair
 from core.constants import ETHEREUM_PUBLIC_KEY_LENGTH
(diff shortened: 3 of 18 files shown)


concent_api/common/tests/test_api_view.py
```diff
@@ -39,7 +40,7 @@ def setUp(self):
         self.request_factory = RequestFactory()
         self.want_to_compute = message.WantToComputeTask(
             node_name           = 1,
-            task_id             = 2,
+            task_id             = generate_uuid(),
             perf_index          = 3,
             price               = 4,
             max_resource_size   = 5,
```
kbeker: Please also adapt this area of code to ours current style

concent_api/core/tests/test_core_views.py
```diff
@@ -57,7 +57,7 @@ def setUp(self):
 
         self.want_to_compute = message.WantToComputeTask(
             node_name           = 1,
-            task_id             = 2,
+            task_id             = self._get_uuid(),
             perf_index          = 3,
             price               = 4,
             max_resource_size   = 5,
```

concent_api/core/tests/test_integration_force_payment.py
```diff
@@ -36,7 +36,7 @@ def test_provider_send_force_payment_with_subtask_results_accepted_signed_by_dif
                 task_to_compute = self._get_deserialized_task_to_compute(
                     timestamp                       = "2018-02-05 10:00:00",
                     deadline                        = "2018-02-05 10:00:10",
-                    subtask_id='2',
+                    subtask_id=self._get_uuid(),
```
kbeker: And here

concent_api/common/tests/test_api_view.py
```diff
@@ -16,13 +16,14 @@
...
-from common.helpers                  import get_current_utc_timestamp
-from common.testing_helpers          import generate_ecc_key_pair
+from common.constants import ErrorCode
+from common.decorators import require_golem_message
+from common.decorators import handle_errors_and_responses
+from common.helpers import get_current_utc_timestamp
+from common.helpers import generate_uuid
+from common.testing_helpers import generate_ecc_key_pair
+from core.exceptions import Http400
+from core.models import Client
```
kbeker: This changes should be done in separate commit

concent_api/core/tests/test_auth_force_payment.py
```diff
@@ -32,23 +32,23 @@ def test_provider_send_force_payment_and_concent_should_return_it_to_requestor_w
...
-                    deadline                        = "2018-02-05 10:00:10",
-                    subtask_id='2',
-                    price                           = 15000,
+                timestamp="2018-02-05 10:00:15",
+                payment_ts="2018-02-05 11:55:00",
+                task_to_compute=self._get_deserialized_task_to_compute(
+                    timestamp="2018-02-05 10:00:00",
+                    deadline="2018-02-05 10:00:10",
+                    subtask_id=self._get_uuid(),
+                    price=15000,
```
kbeker: All style changes should be in separate commit :)

concent_api/command_line_tool/tests/test_message_extractor.py
```diff
@@ -27,8 +28,8 @@ def setUp(self):
         self.requestor_ethereum_public_key = "b'T8ER1hrI9hH0Zyu/m6u9H2K4rc1mC/dlCqK0PSzBSwvKBdHfysmrIIsQMyvecDU+TIkhDmq93Olfo5h2FSGRjw=='"
         self.requestor_public_key = "mrdOXP6Owe3i48G29RHPVU6wewvuUNuNxFB+kPTDnmI21+p5ShttCSUAbHOVm8OIUF9hEluc1+lICJZGkFSSOA=="
         self.deadline = 1510394400
-        self.subtask_id = "12"
-        self.task_id = "2"
+        self.subtask_id = str(uuid.uuid4())
+        self.task_id = str(uuid.uuid4())
```
pawelkisielewicz: why you don't use here `generate_uuid()`?

concent_api/common/tests/test_api_view.py
pawelkisielewicz: Agree, but in this situation I don't think it is necessary. For future- yes, let's do it in separate commits, for now- for me ok

concent_api/core/tests/utils.py
```diff
@@ -954,4 +954,4 @@ def _get_verification_deadline_as_timestamp(
 
     @staticmethod
     def _get_uuid() -> str:
-        return str(uuid.uuid4())
+        return generate_uuid()
```
pawelkisielewicz: I don't understand this method. Name _get... suggests me, you are asking for a particular field. For field, that is used multiple times (e.g. some Id or key, the same used a few times). This name is mistakenly for me, it suggest that you someway prepared Id and now get access to this particular Id. 

Please remove `_get_uuid()` and use `generate_uuid()` instead in all places. Unless you have good reason to keep it :)

concent_api/core/tests/test_auth_force_payment.py
pawelkisielewicz: as above- it should be, but for me it is readable and can stay..

concent_api/common/tests/test_api_view.py
rwrzesien: I think that if we want to force this kind of changes into separate commits people will do them less likely :) So maybe lets don't, but we may also discuss this with whole team if you want.

concent_api/core/tests/utils.py
rwrzesien: The reason for keeping it inside this class is that you can use this method directly in tests without a need to import `generate_uuid` in every file. 
About the name, we use `_get_something` all around in the tests to retrieve something, ie. `_get_serialized_force_payment` etc.

concent_api/common/tests/test_api_view.py
rwrzesien: :+1: done

concent_api/core/tests/test_core_views.py

concent_api/core/tests/test_integration_force_payment.py
rwrzesien: :+1: done

concent_api/command_line_tool/tests/test_message_extractor.py
```diff
@@ -27,8 +28,8 @@ def setUp(self):
         self.requestor_ethereum_public_key = "b'T8ER1hrI9hH0Zyu/m6u9H2K4rc1mC/dlCqK0PSzBSwvKBdHfysmrIIsQMyvecDU+TIkhDmq93Olfo5h2FSGRjw=='"
         self.requestor_public_key = "mrdOXP6Owe3i48G29RHPVU6wewvuUNuNxFB+kPTDnmI21+p5ShttCSUAbHOVm8OIUF9hEluc1+lICJZGkFSSOA=="
         self.deadline = 1510394400
-        self.subtask_id = "12"
-        self.task_id = "2"
+        self.subtask_id = str(uuid.uuid4())
```
dybi: please use `generate_uuid()` to enforce deterministic behavior in tests (here and below)

rwrzesien: :+1: done

concent_api/core/tests/utils.py
pawelkisielewicz: If you say so... I don't think it is good approach. Please change name to `_get_random_uuid`(would recommend) or leave as it is now

rwrzesien: Please check out this scenario;
Currently we have a `generate_uuid` function and test utils method `_get_uuid` for those tests which inherits from `ConcentIntegrationTestCase` class. Also, currently `_get_uuid` only calls `generate_uuid`. However, we can't be sure that in the future we will not be changing something or doing something extra in `_get_uuid`. With the current way we will not have to update anything in all tests if something changes only in this method.

About `_get_random_uuid` it is no longer random, but depends on the `last_char` passed (for the same `last_char` there will be the same returned value).

Issue #771, opened 2018-08-30
Failing unit tests when run in parallel


dybi:
### Problem
When our unit tests are run in parallel (ex. `./run_tests.sh -n=4`) they are failing:
```
❰dybowski❙~/GOLEM/concent(git≠master)❱✔≻ (concent) ./run_tests.sh -n=4
============================================================= test session starts ==============================================================
platform linux -- Python 3.6.6, pytest-3.6.2, py-1.5.4, pluggy-0.6.0
Django settings: concent_api.settings.testing (from ini file)
rootdir: /home/dybowski/GOLEM/concent/concent_api, inifile: pytest.ini
plugins: xdist-1.22.2, forked-0.2, django-3.3.2, cov-2.5.1, asyncio-0.8.0, celery-4.2.0
gw0 [488] / gw1 [488] / gw2 [488] / gw3 [488]
scheduling tests via LoadScheduling
collecting 0 items / 3 errors                                                                                                                  
==================================================================== ERRORS ====================================================================
_____________________________________________________________ ERROR collecting gw1 _____________________________________________________________
Different tests were collected between gw0 and gw1. The difference is:
--- gw0

+++ gw1

@@ -345,12 +345,12 @@

 core/tests/test_unit_validation.py::TestValidateAllMessagesIdentical::test_that_function_pass_when_in_list_are_two_same_report_computed_task
 core/tests/test_unit_validation.py::TestValidateAllMessagesIdentical::test_that_function_pass_when_in_list_is_one_item
 core/tests/test_unit_validation.py::TestValidateAllMessagesIdentical::test_that_function_raise_http400_when_any_slot_will_be_different_in_messages
-core/tests/test_unit_validation.py::TestValidateIdValue::()::test_that_function_should_pass_when_value_is_allowed[b866ffe85df545acb5ce42b8d84383f7]
-core/tests/test_unit_validation.py::TestValidateIdValue::()::test_that_function_should_pass_when_value_is_allowed[49477c0c-db53-44d7-a5df-5aac82c780ca]
-core/tests/test_unit_validation.py::TestValidateIdValue::()::test_that_function_should_raise_exception_when_value_is_not_allowed[1a1e700f-81b1-472d-a324-59cbf2b067f186e71861-22a3-4487-863b-a54014a47711]
[...]

PR #785, opened 2018-09-05
Change UUIDs to be deterministic in the tests

dybi:
So `py.test` collector stops failing.

resolves: https://github.com/golemfactory/concent/issues/771 (partialy, see update in the issue)

### DIFF
--- a/concent_api/core/tests/test_unit_validation.py
+++ b/concent_api/core/tests/test_unit_validation.py
@@ -1,4 +1,3 @@
-import uuid
 from unittest import TestCase
 from assertpy import assert_that
 import mock
@@ -25,6 +24,7 @@ from core.exceptions import HashingAlgorithmError
 from core.subtask_helpers import are_keys_and_addresses_unique_in_message_subtask_results_accepted
 from core.subtask_helpers import are_subtask_results_accepted_messages_signed_by_the_same_requestor
 from core.tests.utils import ConcentIntegrationTestCase
+from core.tests.utils import generate_uuid
 from core.validation import validate_all_messages_identical
 from core.validation import validate_compute_task_def
 from core.validation import validate_ethereum_addresses
@@ -116,11 +116,14 @@ class TestValidateAllMessagesIdentical(ConcentIntegrationTestCase):
             validate_all_messages_identical([self.report_computed_task, different_report_computed_task])
 
 
-class TestValidateIdValue():
+UUID: str = generate_uuid()
+
+
+class TestValidateIdValue:
 
     @pytest.mark.parametrize('id_', [
-        str(uuid.uuid4()).replace('-', ''),
-        str(uuid.uuid4()),
+        UUID.replace('-', ''),
+        UUID,
     ])  # pylint: disable=no-self-use
     def test_that_function_should_pass_when_value_is_allowed(self, id_):
 
@@ -130,10 +133,10 @@ class TestValidateIdValue():
             pytest.fail(f'{exception}')
 
     @pytest.mark.parametrize('id_', [
-        f'{str(uuid.uuid4())}{str(uuid.uuid4())}',
-        str(uuid.uuid4())[1:],
-        str(uuid.uuid4())[:-1],
-        str(uuid.uuid4()) + '1',
+        f'{UUID}{UUID}',
+        UUID[1:],
+        UUID[:-1],
+        UUID + '1',
         '',
     ])  # pylint: disable=no-self-use
     def test_that_function_should_raise_exception_when_value_is_not_allowed(self, id_):
--- a/concent_api/core/tests/utils.py
+++ b/concent_api/core/tests/utils.py
@@ -1,3 +1,4 @@
+import random
 import uuid
 from base64 import b64encode
 from typing import List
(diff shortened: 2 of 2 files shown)


concent_api/core/tests/test_unit_validation.py
```diff
@@ -24,7 +23,7 @@
 from core.exceptions import HashingAlgorithmError
 from core.subtask_helpers import are_keys_and_addresses_unique_in_message_subtask_results_accepted
 from core.subtask_helpers import are_subtask_results_accepted_messages_signed_by_the_same_requestor
-from core.tests.utils import ConcentIntegrationTestCase
+from core.tests.utils import ConcentIntegrationTestCase, generate_uuids
```
kbeker: You know ;)

pawelkisielewicz: please fix styling :)

dybi: sure, thing
done

concent_api/core/tests/utils.py
```diff
@@ -67,6 +68,14 @@ def add_time_offset_to_date(base_time: str, offset: int) -> str:
     )
 
 
+def generate_uuids(number_of_uuids: int) -> List[str]:
+    random.seed(123210912)
```
rwrzesien: I think this is a good candidate for comment(s)

concent_api/core/tests/utils.py
```diff
@@ -67,6 +68,14 @@ def add_time_offset_to_date(base_time: str, offset: int) -> str:
...
+    a = "%32x" % random.getrandbits(128)
+    rd = a[:12] + '4' + a[13:16] + 'a' + a[17:]
```
rwrzesien: Maybe it would be possible to choose some more meaningful names for those.

concent_api/core/tests/utils.py
```diff
@@ -67,6 +68,14 @@ def add_time_offset_to_date(base_time: str, offset: int) -> str:
     )
 
 
+def generate_uuids(number_of_uuids: int) -> List[str]:
```
rwrzesien: Can you give an example of where this might be used to generate more than one id at a time ? I can think of `task_id` and `subtask_id` which are sometimes generated line by line, but maybe it would be simpler to just call it twice ?

Also, why not returning tuple ?

dybi: quite frankly I don't see any, but it was @cameel 's idea. 
@cameel ?

As to returning tuple - what would be the advantage over list? Lists can be easily produced by means of list comprehension, that's way I have chosen it .

concent_api/core/tests/utils.py
dybi: done

concent_api/core/tests/utils.py
dybi: ok, I chenged this function so it now returns just a single UUID, with the possibility of changing the last character

cameel: My idea was a bit different. I agree with @rwrzesien that creating a list and having to choose one is not very convenient.

My idea was to have a function like this:
```
def generate_uuid(mnemonic: str) -> str
    ...
```

UUIDs are very hard to remember and recognize. Unless you know one by heart, you have to just copy and paste them every time. @dybi suggested the natural thing to do: create a constant. Since we may sometimes need more than one I suggested the natural extension: let's create a function that returns the same UUID given the same mnemonic. Then, if you need two UUIDs you could simply do `generate_uuid('provider'), generate_uuid('requestor')` instead of having to remember and copy long strings of characters.

Simple and elegant. Does not require you to create constants if you just need a predictable UUID for testing. And implementation is dead simple. I have originally suggested using a number as a mnemonic and simply replacing a part of the UUID and recalculating the checksum. But there's an even easier way. [UUID versions 3 and 5](https://en.wikipedia.org/wiki/Universally_unique_identifier#Versions_3_and_5_(namespace_name-based)) are meant exactly for that. Just use [`uuid5()`](https://docs.python.org/3.7/library/uuid.html#uuid.uuid5). We could even just use it directly without wrapping it in our own helper.

dybi:
@rwrzesien, I would either
 - create global constants `SUBTASK_ID = generate_uuid('s')`, `TASK_ID = generate_uuid('t')` and reuse them
or
- create fields in the `setUp()` method  and reuse them

concent_api/core/tests/utils.py
rwrzesien: So it will generate same UUID each time for a given value ?

cameel: Yeah. I that's how it's supposed to work:


``` python
from uuid import UUID, uuid5

>>> uuid5(UUID('00000000-0000-0000-0000-000000000000'), 'provider')
UUID('d81821ad-2d8e-58a5-a74a-05c272696d84')
>>> uuid5(UUID('00000000-0000-0000-0000-000000000000'), 'provider')
UUID('d81821ad-2d8e-58a5-a74a-05c272696d84')
>>> uuid5(UUID('00000000-0000-0000-0000-000000000000'), 'provider')
UUID('d81821ad-2d8e-58a5-a74a-05c272696d84')

>>> uuid5(UUID('00000000-0000-0000-0000-000000000000'), 'requestor')
UUID('894cf831-772d-5ec1-9c2f-ed0cc997b3d0')
>>> uuid5(UUID('00000000-0000-0000-0000-000000000000'), 'requestor')
UUID('894cf831-772d-5ec1-9c2f-ed0cc997b3d0')
>>> uuid5(UUID('00000000-0000-0000-0000-000000000000'), 'requestor')
UUID('894cf831-772d-5ec1-9c2f-ed0cc997b3d0')
```

PR #789, opened 2018-09-06
Mypy fixes after review

rwrzesien:
Resolves @pawelkisielewicz comments from https://github.com/golemfactory/concent/pull/783


rwrzesien:
@pawelkisielewicz Good catch, fixed.

PR #790, opened 2018-09-07
Fix asyncio Task scheduling in QueuePool and adapt unit tests accordingly

dybi:
(no issue for this PR)
This bug has been found during integration testing of MiidleMan /SS

### DIFF
--- a/concent_api/middleman/utils.py
+++ b/concent_api/middleman/utils.py
@@ -40,8 +40,8 @@ class QueuePool(dict):
             self._ensure_key_uniqueness(key)
         super().update(mapping, **kwargs)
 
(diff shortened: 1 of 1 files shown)


concent_api/middleman/utils.py
```diff
@@ -43,22 +43,22 @@ def update(self, mapping: Dict[int, asyncio.Queue], **kwargs: Any) -> None:  # t
     def pop(self, key: int, **kwargs: Any) -> asyncio.Queue:  # type: ignore
```
rwrzesien: Shouldn't there be also `*args` here ? I saw them in one of the temporary fixes.

dybi: good catch:)
done

Issue #787, opened 2018-09-05
Merge `receive/` and `receive-out-of-band/` into one endpoint


Jakub89:
Step 1(before any change in golem-app):
- Allow `receive/` endpoint to return `receive_out_of_band` messages.
- Adjust message handlers.
- Change tests to use `receive` instead of `receive_out_of_band` endpoint.

~~Step 2:~~
- ~~Wait for golem until they stop using `receive_out_of_band/` endpoint.~~

~~Step 3:~~
- ~~Delete `receive_out_of_band/` endpoint.~~

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/160283588

cameel:
1. Don't delete the view. It should still stay in the API for compatibility. Add a separate task for removing it later, when Golem no longer uses it.
2. Don't forget that we still need its functionality. The messages should still be returned, just from `receive/`, not `receive-out-of-band/`.

Jakub89:
@cameel issue updated

PaweuB:
Pivotal for Part 2 & 3 of #787: 
https://www.pivotaltracker.com/story/show/160384405

Keep this Issue open till fulfilling part 2 und 3.

cameel:
@Jakub89 When you merge this please drop a comment in #264 so that I remember I have to update it.

cameel:
@Jakub89 Is this done? I think that part 2 and 3 should be covered by separate issues.

Jakub89:
part2: #842

PR #791, opened 2018-09-07
Unify receive/ and receive_out_of_band/ endpoints

Jakub89:
Part 1 of: #787

### DIFF
--- a/concent_api/core/message_handlers.py
+++ b/concent_api/core/message_handlers.py
@@ -908,17 +908,13 @@ def store_subtask(
     return subtask
 
 
-def handle_messages_from_database(
-    client_public_key: bytes,
-    response_type: PendingResponse.Queue,
-) -> Union[message.Message, None]:
-    assert client_public_key    not in ['', None]
+def handle_messages_from_database(client_public_key: bytes) -> Union[message.Message, None]:
+    assert client_public_key not in ['', None]
 
     encoded_client_public_key = b64encode(client_public_key)
     pending_response = PendingResponse.objects.select_for_update().filter(
-        client__public_key = encoded_client_public_key,
-        queue              = response_type.name,
-        delivered          = False,
+        client__public_key=encoded_client_public_key,
+        delivered=False,
     ).order_by('created_at').first()
 
     if pending_response is None:
--- a/concent_api/core/tests/test_auth_report_computed_task.py
+++ b/concent_api/core/tests/test_auth_report_computed_task.py
@@ -920,16 +920,6 @@ class AuthReportComputedTaskIntegrationTest(ConcentIntegrationTestCase):
         self.assertEqual(response.status_code,        204)
         self.assertEqual(len(response.content),       0)
 
-        with freeze_time("2017-12-01 11:00:15"):
-            response = self.client.post(
-                reverse('core:receive'),
-                data                           = self._create_requestor_auth_message(),
-                content_type                   = 'application/octet-stream',
-            )
-
-        self.assertEqual(response.status_code,        204)
-        self.assertEqual(len(response.content),       0)
-
         # STEP 7: Concent overrides computed task rejection and sends acceptance message to the provider with correct key
 
         with freeze_time("2017-12-01 11:00:15"):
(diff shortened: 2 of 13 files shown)


concent_api/core/message_handlers.py
```diff
@@ -908,18 +908,21 @@ def store_subtask(
...
-    assert client_public_key    not in ['', None]
+def handle_messages_from_database(client_public_key: bytes) -> Union[message.Message, None]:
+    assert client_public_key not in ['', None]
 
     encoded_client_public_key = b64encode(client_public_key)
-    pending_response = PendingResponse.objects.select_for_update().filter(
-        client__public_key = encoded_client_public_key,
-        queue              = response_type.name,
-        delivered          = False,
+    pending_response = PendingResponse.objects.filter(
```
pawelkisielewicz: are you sure we should remove `select_for_update` in this place? What is the reason?

concent_api/core/message_handlers.py
```diff
@@ -908,18 +908,21 @@ def store_subtask(
...
+        client__public_key=encoded_client_public_key,
+        queue=PendingResponse.Queue.Receive.name,  # pylint: disable=no-member
+        delivered=False,
     ).order_by('created_at').first()
+    if pending_response is None:
+        pending_response = PendingResponse.objects.filter(
```
pawelkisielewicz: no `select_for_update` here?

concent_api/core/message_handlers.py
Jakub89: No reason, I did it by accident on rebase.

concent_api/core/message_handlers.py
```diff
@@ -908,18 +908,21 @@ def store_subtask(
...
+    assert client_public_key not in ['', None]
 
     encoded_client_public_key = b64encode(client_public_key)
-    pending_response = PendingResponse.objects.select_for_update().filter(
-        client__public_key = encoded_client_public_key,
-        queue              = response_type.name,
-        delivered          = False,
+    pending_response = PendingResponse.objects.filter(
+        client__public_key=encoded_client_public_key,
+        queue=PendingResponse.Queue.Receive.name,  # pylint: disable=no-member
```
rwrzesien: Do we still need `queue` field and `PendingResponse.Queue` enum ?

concent_api/core/tests/test_auth_report_computed_task.py
```diff
@@ -920,16 +920,6 @@ def test_provider_forces_computed_task_report_and_requestor_sends_rejection_due_
         self.assertEqual(response.status_code,        204)
         self.assertEqual(len(response.content),       0)
 
-        with freeze_time("2017-12-01 11:00:15"):
```
rwrzesien: Why is this removed ?

concent_api/core/message_handlers.py
Jakub89: I'm pretty sure there still need to be distinction between "standard" messages and "out-of-band".

concent_api/core/tests/test_auth_report_computed_task.py
Jakub89: Because this part return "out-of-band" message that we check later, so I don't think we need to replace it with different call.

concent_api/core/message_handlers.py
rwrzesien: As discussed, lets keep it.

concent_api/core/tests/test_auth_report_computed_task.py
rwrzesien: In other words, I think it should stay it if doesn't break anything and test one more corner case.

Jakub89: Okey, I'll switch it to other case.

rwrzesien: As discussed, it is ok to remove this as it is doing the same now as `# Step 9` of this test.

concent_api/core/message_handlers.py
```diff
@@ -908,18 +908,21 @@ def store_subtask(
...
-def handle_messages_from_database(
-    client_public_key: bytes,
-    response_type: PendingResponse.Queue,
-) -> Union[message.Message, None]:
-    assert client_public_key    not in ['', None]
+def handle_messages_from_database(client_public_key: bytes) -> Union[message.Message, None]:
+    assert client_public_key not in ['', None]
 
     encoded_client_public_key = b64encode(client_public_key)
     pending_response = PendingResponse.objects.select_for_update().filter(
```
rwrzesien: It would be the same and less redundant if you do:
```
pending_response = PendingResponse.objects.select_for_update().filter(
    client__public_key=encoded_client_public_key,
    delivered=False,
).order_by('created_at').first()
```

and remove the second query. This way it will return any `PendingResponse` it will find.

PR #794, opened 2018-09-07
Disable parallel test runs for non-Django apps

### LINKED ISSUE #793: Disable parallel pytest run for no-django apps
Parallel for non-Django apps do not work, due to a bug in pytest (it should be fixed in https://github.com/pytest-dev/pytest/issues/1618). 
The parallel run should be disabled for such apps, until the bug is fixed.

dybi:
resolves: https://github.com/golemfactory/concent/issues/793

### DIFF
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -49,8 +49,10 @@ cd middleman_protocol/
 pytest -p no:django                 \
     --cov-report term-missing       \
     --cov-config ../coverage-config \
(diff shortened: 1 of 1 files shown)


Jakub89 (approved):

kbeker (approved):

pawelkisielewicz (approved):

blueyed:
I think this is fixed (via pytest-xdist) in case you want to revert it.

PR #797, opened 2018-09-11
Add script to adapt data in database to current Concent version

kbeker:
No issue for that. Problem arose when staging cluster was updated.


concent_api/core/management/commands/concent_adapt_data_to_current_version.py
```diff
@@ -0,0 +1,29 @@
+from logging import getLogger
+
+from django.core.management.base import BaseCommand
+from typing import Any
```
rwrzesien: `typing` import should be after `logging`.

concent_api/core/management/commands/concent_adapt_data_to_current_version.py
```diff
@@ -0,0 +1,29 @@
...
+        for subtask in Subtask.objects.all():
+            try:
+                report_computed_task =  deserialize_message(subtask.report_computed_task.data.tobytes())
+                if not isinstance(report_computed_task.size, int):
+                    delete_unsupported_messages(subtask.subtask_id)
+            except:
+                delete_unsupported_messages(subtask.subtask_id)
+
+
+def delete_unsupported_messages(subtask_id: str) -> None:
```
rwrzesien: Why separated function for this? Those are only 3 lines.

concent_api/core/management/commands/concent_adapt_data_to_current_version.py
```diff
@@ -0,0 +1,29 @@
...
+    Subtask.objects.filter(subtask_id=subtask_id).delete()
```
rwrzesien: Instead of those two lines you can do (if you pass `subtask` instead of `subtask_id`):
```
StoredMessage.objects.filter(subtask_id=subtask.subtask_id).delete()
subtask.delete()
```

That would be one query less.

concent_api/core/management/commands/concent_adapt_data_to_current_version.py
```diff
@@ -0,0 +1,29 @@
...
+logger = getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = "Adapts whole database to current Concent's version"
+
+    def handle(self, *args: Any, **options: Any) -> None:
+        for subtask in Subtask.objects.all():
+            try:
+                report_computed_task =  deserialize_message(subtask.report_computed_task.data.tobytes())
```
rwrzesien: Extra space after `= `

concent_api/core/management/commands/concent_adapt_data_to_current_version.py
```diff
@@ -0,0 +1,29 @@
...
+                if not isinstance(report_computed_task.size, int):
+                    delete_unsupported_messages(subtask.subtask_id)
+            except:
```
rwrzesien: It should be golem `MessageError`. Or do we expect other exceptions?

concent_api/core/management/commands/concent_adapt_data_to_current_version.py
```diff
@@ -0,0 +1,29 @@
...
+
+from common.helpers import deserialize_message
+from core.models import StoredMessage
+from core.models import Subtask
+
+logger = getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = "Adapts whole database to current Concent's version"
```
rwrzesien: I would change it to `Removes Subtasks which related ReportComputedTask cannot be deserialized or its size field is not int.` or something more precise. 

`Adapts` may not make user aware that something might be deleted.

concent_api/core/management/commands/concent_adapt_data_to_current_version.py
kbeker: I didn't want to pass whole subtask to this function that's why I passed there only `subtask_id`. Do you want me to change it?

concent_api/core/management/commands/concent_adapt_data_to_current_version.py
```diff
@@ -0,0 +1,29 @@
...
+                if not isinstance(report_computed_task.size, int):
+                    delete_unsupported_messages(subtask.subtask_id)
+            except:
+                delete_unsupported_messages(subtask.subtask_id)
+
+
+def delete_unsupported_messages(subtask_id: str) -> None:
+    Subtask.objects.filter(subtask_id=subtask_id).delete()
+    StoredMessage.objects.filter(subtask_id=subtask_id).delete()
+    logger.info(f'Deleted subtask and all related messages with subtask_id: {subtask_id}')
```
pawelkisielewicz: We think there are **all** messages, but please be more specify and change to something like 'Deleted subtask with id:... and related StoredMessage. I also think there are all messages, but if not this log can be cause of mistakes..

concent_api/core/management/commands/concent_adapt_data_to_current_version.py
rwrzesien: Yes please, it will result in a one query less with each call of this function (you will not retrieve `Subtask` again from database but use the one which you already have) which is some optimization and good practice.

concent_api/core/management/commands/concent_adapt_data_to_current_version.py
kbeker: Done

concent_api/core/management/commands/concent_adapt_data_to_current_version.py
kbeker: Done

concent_api/core/management/commands/concent_adapt_data_to_current_version.py
kbeker: Ok, so in future we will need to change this comment to adapt :P this comment to newer functions which will be provided by this script

concent_api/core/management/commands/concent_adapt_data_to_current_version.py
pawelkisielewicz: I agree with Karol's approach to catch all exceptions, but I think we should log also reason of deleting: size - int or exception and in this case what kind of exception.

kbeker: Done

concent_api/core/management/commands/concent_adapt_data_to_current_version.py
kbeker: Done

concent_api/delete-unsupported-messages.py
```diff
@@ -0,0 +1,40 @@
...
+
+from logging import getLogger
+
+from common.helpers import deserialize_message
+
+from golem_messages.exceptions import MessageError
+
+sys.path.append('concent_api')
+os.environ['DJANGO_SETTINGS_MODULE'] = "concent_api.settings"
+django.setup()
```
cameel: Shouldn't `setup()` be in `main()`?

concent_api/delete-unsupported-messages.py
```diff
@@ -0,0 +1,40 @@
...
+
+def main() -> None:
+    for subtask in Subtask.objects.all():
+        try:
+            report_computed_task = deserialize_message(subtask.report_computed_task.data.tobytes())
+            if not isinstance(report_computed_task.size, int):
+                delete_unsupported_messages(subtask)
+        except MessageError as golem_messages_exception:
+            logger.info(f'During message deserialization exception raised: {golem_messages_exception}')
+            delete_unsupported_messages(subtask)
```
cameel: This looks very risky. How can you be sure what caused the `MessageError`?

If anything unexpected happens, you should fail, not ignore it and continue anyway.

concent_api/delete-unsupported-messages.py
```diff
@@ -0,0 +1,40 @@
...
+from core.models import StoredMessage  # noqa E402
+from core.models import Subtask  # noqa E402
+
+logger = getLogger(__name__)
+
+
+def main() -> None:
+    for subtask in Subtask.objects.all():
+        try:
+            report_computed_task = deserialize_message(subtask.report_computed_task.data.tobytes())
```
cameel: You're not running in a transaction. If the script fails in the middle of the operation the changes will not be roleld back.

concent_api/delete-unsupported-messages.py
```diff
@@ -0,0 +1,40 @@
...
+django.setup()
+
+from core.models import StoredMessage  # noqa E402
+from core.models import Subtask  # noqa E402
+
+logger = getLogger(__name__)
+
+
+def main() -> None:
+    for subtask in Subtask.objects.all():
```
cameel: You should lock all the subtasks you're processing with `select_for_update()`.

concent_api/delete-unsupported-messages.py
```diff
@@ -0,0 +1,40 @@
...
+            if not isinstance(report_computed_task.size, int):
+                delete_unsupported_messages(subtask)
+        except MessageError as golem_messages_exception:
+            logger.info(f'During message deserialization exception raised: {golem_messages_exception}')
+            delete_unsupported_messages(subtask)
+
+
+def delete_unsupported_messages(subtask: Subtask) -> None:
+    subtask_id = subtask.subtask_id
+    StoredMessage.objects.filter(subtask_id=subtask_id).delete()
```
rwrzesien: Use `subtask_id` here directly.

concent_api/delete-unsupported-messages.py
```diff
@@ -0,0 +1,40 @@
...
+    subtask.delete()
+    logger.info(f'Deleted subtask and all related messages with subtask_id: {subtask_id}')
```
rwrzesien: Now when I think about it.. do we really want to remove all related `StoredMessages` only because we were not able to deserialize `report_computed_task`? 

@cameel What do you think?

concent_api/delete-unsupported-messages.py
```diff
@@ -0,0 +1,40 @@
...
+                delete_unsupported_messages(subtask)
+        except MessageError as golem_messages_exception:
+            logger.info(f'During message deserialization exception raised: {golem_messages_exception}')
+            delete_unsupported_messages(subtask)
+
+
+def delete_unsupported_messages(subtask: Subtask) -> None:
+    subtask_id = subtask.subtask_id
+    StoredMessage.objects.filter(subtask_id=subtask_id).delete()
+    subtask.delete()
```
cameel: You don't need to do `StoredMessage.objects.filter(subtask_id=subtask_id).delete()` explicitly. All the objects that point at the subtask will be deleted when you delete it (unless it's overriden by the foreign key but it it's is, it's probably for a good reason).

See [QuerySet.delete()](https://docs.djangoproject.com/pl/1.11/ref/models/querysets/#delete)

concent_api/delete-unsupported-messages.py
```diff
@@ -0,0 +1,40 @@
+import django
```
rwrzesien: Django should be imported after std lib imports. Generally order should be:

1. std lib imports.
2. 3rd party imports.
3. Django imports.
4. application specific imports.

concent_api/delete-unsupported-messages.py
```diff
@@ -0,0 +1,40 @@
...
+import sys
+
+from logging import getLogger
+
+from common.helpers import deserialize_message
+
+from golem_messages.exceptions import MessageError
+
+sys.path.append('concent_api')
+os.environ['DJANGO_SETTINGS_MODULE'] = "concent_api.settings"
```
rwrzesien: Why is this not a `Command` anymore?

concent_api/delete-unsupported-messages.py
```diff
@@ -0,0 +1,40 @@
...
+            report_computed_task = deserialize_message(subtask.report_computed_task.data.tobytes())
+            if not isinstance(report_computed_task.size, int):
+                delete_unsupported_messages(subtask)
+        except MessageError as golem_messages_exception:
+            logger.info(f'During message deserialization exception raised: {golem_messages_exception}')
+            delete_unsupported_messages(subtask)
+
+
+def delete_unsupported_messages(subtask: Subtask) -> None:
+    subtask_id = subtask.subtask_id
```
cameel: Please don't create intermediate variables for short and simple stuff like this. It does not make it easier to read. It only adds an unnecessary level of indirection.

concent_api/delete-unsupported-messages.py
```diff
@@ -0,0 +1,40 @@
...
+
+logger = getLogger(__name__)
+
+
+def main() -> None:
+    for subtask in Subtask.objects.all():
+        try:
+            report_computed_task = deserialize_message(subtask.report_computed_task.data.tobytes())
+            if not isinstance(report_computed_task.size, int):
+                delete_unsupported_messages(subtask)
```
cameel: It would be useful to log `size` of the subtasks your're deleting.

concent_api/delete-unsupported-messages.py
cameel: Maybe print the whole subtask rather than just the ID?

concent_api/delete-unsupported-messages.py
cameel: We will not be merging it into master. The problem is that the code does not make sense in the release it would be merged into. It should always be executed on an older release (0.8.0 in this case) that still has the old schema and is useless on the current one. This would be confusing.

I have decided to create a standalone script and simply put it in release notes with instructions for the admin. See https://github.com/golemfactory/concent-deployment/issues/229

concent_api/delete-unsupported-messages.py
cameel: Yes. If we're removing the subtask we don't care about the messages either.

We we cared about them we should have kept running the old version until the client fetched them.

concent_api/core/management/commands/concent_adapt_data_to_current_version.py
cameel: We should not be catching all exceptions. The script must fail if anything goes wrong. It's better to fail than to do the wrong thing.

concent_api/delete-unsupported-messages.py
rwrzesien: > We we cared about them we should have kept running the old version until the client fetched them.

But those are not messages which are returned to client (those are `PendindResponse` objects), but those which we received from client.

concent_api/delete-unsupported-messages.py
rwrzesien: It could for example read or generate Concent version and don't start if the version is below 0.8.0/

concent_api/delete-unsupported-messages.py
rwrzesien: @cameel Important one:

It will not because `StoredMessage` objects don't point to `Subtask`. They are only fields on `Subtask`.

```
In [14]: s = Subtask.objects.get(subtask_id=st.subtask_id)
In [16]: s.task_to_compute
Out[16]: <StoredMessage: StoredMessage #61, type:2002, 2018-04-16 10:54:58.821811+00:00>
In [17]: s.delete()
Out[17]: (2, {'core.PaymentInfo': 0, 'core.PendingResponse': 1, 'core.Subtask': 1})
```

So it is actually opposite - if you remove `StoredMessage`, the related `Subtask` gets removed. Is that the behavior we want?

```
In [27]: st
Out[27]: <StoredMessage: StoredMessage #71, type:2002, 2018-04-16 11:03:40.500564+00:00>
In [28]: st.delete()
Out[28]: 
(4,
 {'core.PaymentInfo': 0,
  'core.PendingResponse': 2,
  'core.StoredMessage': 1,
  'core.Subtask': 1})
```

concent_api/delete-unsupported-messages.py
kbeker: @rwrzesien - exactly. 
@cameel In some cases messages could not be deserialize and if it happens we want to delete messages which are not supported any longer by Concent.

concent_api/delete-unsupported-messages.py
kbeker: We can, but then we need also move Model imports to this function. Do we want to do it this way?

concent_api/delete-unsupported-messages.py
kbeker: Done

concent_api/delete-unsupported-messages.py
kbeker: Done

concent_api/delete-unsupported-messages.py
kbeker: Do we really need this? Should we print whole Subtask with all nested messages?

concent_api/delete-unsupported-messages.py
kbeker: Done

concent_api/delete-unsupported-messages.py
```diff
@@ -0,0 +1,40 @@
...
+logger = getLogger(__name__)
+
+
+def main() -> None:
+    with transaction.atomic(using='control'):
+        for subtask in Subtask.objects.select_for_update().all():
+            try:
+                report_computed_task = deserialize_message(subtask.report_computed_task.data.tobytes())
+                if not isinstance(report_computed_task.size, int):
+                    logger.info(f'Size inside ReportComputedTask message inside subtask with subtask_id: {subtask.subtask_id} is: {report_computed_task.size}')
```
cameel: Logging each subtask twice makes it harder for the admin to see irregularities in the output. Please log one line per subtask and make it less verbose:

Why not:
``` python
f"Deleting subtask {subtask:5} and. REASON: ReportComputedTask.size not an int: {report_computed_task.size:10}"
```

After changing this please post a sample of the output in a comment here.

concent_api/delete-unsupported-messages.py
```diff
@@ -0,0 +1,40 @@
...
+                    delete_unsupported_messages(subtask)
+            except Exception as exception:
+                logger.info(f'During message deserialization exception raised: {exception}')
+                delete_unsupported_messages(subtask)
```
cameel: Why are you ignoring the error even though you don't even know what kind of error it is?

When you get an unknown error you should fail, not silence it and try to force the deletion. If you do it like this, you will delete any message that fails to decode which is definitely not something we want to do without first checking why.

concent_api/delete-unsupported-messages.py
cameel: Does `str(subtask)` print all nested messages? If it does then I think that `Subtask.__str__()` prints way too much.

concent_api/delete-unsupported-messages.py
cameel: Actually that's a really good point. But you should have added a comment explaining your intention because it's not obvious. As you can see, without a comment everyone thought that your intention was just readability.

Anyway, yeah - Django does not clear the ID if you delete the object so it still wasn't not necessary.

concent_api/delete-unsupported-messages.py
cameel: OK. Let's leave it as is.

concent_api/delete-unsupported-messages.py
rwrzesien: @cameel @kbeker 
>  is there a way to change this without changing the model, e.g. via a parameter to delete()?

Yes, by overriding `delete()` method on model and disconnecting related messages before calling `super().delete()`. BUT if we want to do this only for this purpose the same code can be placed here.

concent_api/delete-unsupported-messages.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def delete_unsupported_messages(subtask: Subtask) -> None:
+    StoredMessage.objects.filter(subtask_id=subtask.subtask_id).delete()
```
dybi: don't we have to delete the `Subtask` itself as well?

kbeker:
This PR is outdated. It was needed to use only once so I'm closing it

PR #799, opened 2018-09-11
Fix asyncio pending tasks not being cancelled properly

dybi:
No issue here.
This PR fixes issues in `MiddleMan` connected with task cancellation,  found during integration testing of `MiddleMan` and `SigninigService`

### DIFF
--- a/concent_api/middleman/middleman_server.py
+++ b/concent_api/middleman/middleman_server.py
@@ -2,9 +2,10 @@ import asyncio
 from asyncio.base_events import BaseEventLoop
 import traceback
 from collections import OrderedDict
+from contextlib import suppress
 from logging import getLogger
+from typing import Iterable
 from typing import Optional
-
 import signal
 
 from middleman.constants import CONNECTION_COUNTER_LIMIT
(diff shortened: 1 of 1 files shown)


concent_api/middleman/middleman_server.py
```diff
@@ -1,10 +1,11 @@
 import asyncio
-from asyncio.base_events import BaseEventLoop
+from asyncio import BaseEventLoop  # type: ignore
```
rwrzesien: Is there a difference with those two ? If yes and it causes some problems then I am sorry, because I have changed this to avoid the mypy ignore.

concent_api/middleman/middleman_server.py
```diff
@@ -148,21 +150,26 @@ def _close_middleman(self) -> None:
             await response_queue.join()  # 3. wait until items from response queue are processed (sent back to Concent)
             response_consumer_task.cancel()
 
+        except asyncio.CancelledError:
```
rwrzesien: Won't that be catch in the next `except`? Or we just don't want to extra log it?

concent_api/middleman/middleman_server.py
```diff
@@ -195,16 +202,31 @@ def _close_middleman(self) -> None:
                     exception_from_task = future.exception()
                     if exception_from_task is not None:
                         raise exception_from_task
+
+            except asyncio.CancelledError:
```
rwrzesien: Same here.

concent_api/middleman/middleman_server.py
dybi: quite frankly, I was playing with different options and eventually have left one. I will restore your solution.
done;

concent_api/middleman/middleman_server.py
dybi: it is added as a measure of avoiding sentry crash logs in case where some `asyncio.Task` is being interrupted by `CTRL-C`

concent_api/middleman/middleman_server.py
dybi: please see above

Issue #800, opened 2018-09-11
Don't import golem-messages in setup.py


cameel:
Currently our `setup.py` files import a global module called `version` which is a part of golem-messages. Since `pip install` executes `setup.py` from all packages before actually installing any of them, it's impossible to install our packages and `golem-messages` in a single operation. You have to install golem-messages first.

To fix this we need to stop importing the module in `setup.py`. We should read version from a file called `RELEASE-VERSION` instead. File should be located in the same directory as `setup.py` - make sure **not** to try to load it from the working directory. Each project in the repository should have it's own version file (though they'll all be the same).

This makes it a bit less convenient in development though because now everyone will need to generate the version file and keep it up to date. As a workaround I suggest to make the import conditional. I.e. import the module only if the version file does not exist.

PR #801, opened 2018-09-12
Add generating version in middleman_protocol setup.py

pawelkisielewicz:
Resolves https://github.com/golemfactory/concent/issues/800

### DIFF
--- a/middleman_protocol/setup.py
+++ b/middleman_protocol/setup.py
@@ -1,13 +1,22 @@
 #!/usr/bin/env python3
-
+from os import path
+from pathlib import Path
 from setuptools import setup
 
-from version import get_version
+
+def generate_version() -> str:
+    here = path.abspath(path.dirname(__file__))
+    version_file = Path(path.join(here, "RELEASE-VERSION"))
+    if version_file.is_file():
+        return open(version_file, "r").read()
+    else:
+        from version import get_version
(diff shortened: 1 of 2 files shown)


middleman_protocol/RELEASE-VERSION
```diff
@@ -0,0 +1 @@
+0.8.0+dev194.gce34fd0
```
cameel: We do not want to commit this file. Please remove it and add it to `.gitignore`.

In development you don't need it because `setup.py` can still import `version`.
In deployment we are going to generate it during the build.

middleman_protocol/setup.py
```diff
@@ -1,13 +1,22 @@
...
+        return open(version_file, "r").read()
+    else:
+        from version import get_version
+        return get_version(prefix='v')
 
 
 setup(
     name='Middleman-Protocol',
-    version=get_version(prefix='v'),
+    version=generate_version(),
```
cameel: This pull request changes only `middleman_protocol`. `signing_service` does this too so it also needs a fix (unless you're just planning to do it in a separate pull request).

pawelkisielewicz: I will do it in this pull request. Done

pawelkisielewicz:
I checked it before I created this pull request. It works

Issue #736, opened 2018-08-17
Installing middleman_protocol directly from a subdirectory


cameel:
Currently Signing Service and Concent just list `middleman_protocol` as their dependency but have no way to install it automatically. Similarly, `middleman_protocol` depends on golem-messages and can't install it (it's the parent application that decides on the version). This complicates deployment.

To solve this issue, we should start installing `middleman_protocol` automatically from the subdirectory in the repository. Both Concent and Signing Service should have a relative path to it in their requirements. This way everything will be installed automatically.

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/159850162

cameel:
We talked about it today with @pawelkisielewicz and I suggested to change the approach. Instead of assuming that `middleman_protocol` is in a subdirectory, just build the package and let pip find it.

But unfortunately I see a problem with specifying the right version in `requirements.lock` if we go that way. It would either have to be the the exact version, which on commits without tags includes commit ID - this will be hard to maintain - or no version at all which defeats the purpose of the lock file. The most precise way we can describe the version we want without using commit ID is simply "the one that's in `../middleman_protocol/` directory".

So I think it will be best to still do it that way after all. Simply always use `middleman_protocol` from the directly repository.

I think that to support running with a different version of `middleman_protocol` we should put it in a separate repository. We don't need it right now though.

I have added #800 which will solve the problem with version during installation.

Issue #711, opened 2018-08-06
Add shell script for supporting generating requirements.lock


rwrzesien:
Currently we have two add two things manually each time we create `requirements.lock` by freeze:
1. Add `--find-links https://builds.golem.network/simple/pyelliptic` at the beginning of the file.
2. Add `[database]` to django-constance to make it look like `django-constance[database]==2.2.0`.

The second one needs investigating if the issue is still valid. I remember that discovering this took me a long time, without it `django-constance` was not working correctly when installed, but it might be fixed already.

The first one needs to be added by shell script. Script should do freeze and add this line at the beginning of `requirements.lock` file.

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/159563086

PR #802, opened 2018-09-12
Chore installing middleman protocol directly from subdirectory

pawelkisielewicz:
Resolves https://github.com/golemfactory/concent/issues/736


concent_api/requirements.lock
```diff
@@ -15,13 +15,14 @@ cffi==1.11.5
...
 coincurve==7.1.0
+construct==2.9.45
 coverage==4.5.1
 cycler==0.10.0
 cytoolz==0.9.0.1
 dask==0.18.1
 decorator==4.3.0
 Django==1.11.13
-django-constance[database]==2.2.0
+django-constance==2.2.0
```
cameel: Why is this being changed?

concent_api/requirements.lock
```diff
@@ -43,6 +44,7 @@ kiwisolver==1.0.1
 kombu==4.2.1
 lru-dict==1.1.6
 matplotlib==2.2.2
+Middleman-Protocol==0.8.0+dev194.gce34fd0
```
cameel: We want to have a path in the .lock file too.

We definitely don't want to hard-code the commit ID here. It's always going to be out of date because you don't know the commit ID until you commit...

concent_api/requirements.txt
```diff
@@ -24,3 +24,4 @@ pytest-xdist
 pytest-cov
 pytest-asyncio
 web3
+..//middleman_protocol
```
cameel: Why two slashes? Is this a mistake or some special `pip` syntax?

concent_api/requirements.lock
pawelkisielewicz: I asked Karol about it, in the past it was necessary because of problems with database. Now everything is OK and it is not required any more

concent_api/requirements.txt
```diff
@@ -23,4 +23,5 @@ pytest-django
 pytest-xdist
 pytest-cov
 pytest-asyncio
-web3
+web3 == 4.2.1
```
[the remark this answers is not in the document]
pawelkisielewicz: done

concent_api/requirements.lock
pawelkisielewicz: right, done

concent_api/requirements.txt
pawelkisielewicz: it is fixed with `-e ../`

concent_api/requirements.lock
cameel: I know but I'm asking why it is being changed in this pull request. We added a script for generating `requirements.lock` always in the canonical format (#711) - why did its output suddenly change now and not back then?

install-development-requirements.sh
```diff
@@ -1,7 +1,9 @@
 #!/bin/bash -e
 
 printf "=================== INSTALL CONCENT_API REQUIREMENTS ====================\n"
-pip install -r concent_api/requirements.lock
+cd concent_api
+pip install -r requirements.lock
+cd ..
```
cameel: This commit description says:
> Change requirements installation path in script 

A good commit description should mention why you're doing this because it's not obvious at a glance. I.e.:

```
install-development-requirements.sh: Run pip in the directory containing requirements file because now it contains relative paths
```

Also, this definitely needs a comment in the code because these two ways of running pip are usually equivalent. If only one works, it should be explained why.

concent_api/requirements.lock
```diff
@@ -95,3 +94,4 @@ urllib3==1.23
 vine==1.1.4
 web3==4.2.1
 websockets==5.0.1
+-e ../middleman_protocol
```
cameel: It's not a part of this pull request but just as a general remark:

It would be more consistent to start using `--editable` instead of `-e`, just like we use the long name for `--find-links`.

concent_api/requirements.lock
kbeker: >why did its output suddenly change now and not back then?

Because back then I have only added script to generate `requirements.lock`. I have not generate new `requirements.lock` file in this -> https://github.com/golemfactory/concent/pull/754 PR.

cameel: OK. But in that case there should be a separate commit that does this (with its description saying what you said above) so that it's clear that it's just cleanup of a previous situation and not something that's a part of the solution.

install-development-requirements.sh
pawelkisielewicz: there were also relative paths before.. but comment is added

concent_api/requirements.lock
pawelkisielewicz: thanks, fixed

install-development-requirements.sh
cameel: Really? What else are we installing from a directory using `-e` with a relative path?

Issue #779, opened 2018-09-04
Timeouts are not always processed before starting work on a subtask


cameel:
In Concent subtask timeouts are enforced passively - i.e. when the time runs out the task is not actually updated until either Concent or a client actually needs to see that state.

For this to work well there are conditions that need to be satisfied:
1. The check and update must happen immediately before any read or write operation
2. The subtask must not be modified by any other process in the time between the check and the operation

Point 2. is covered by locking the subtask (#718) and keeping the transaction open until the end of the view so that all modifications are done while we're still holding the lock.

Point 1. is the problem. We're doing it only in our views. We need to add these calls to Celery tasks too.

In the views we're processing all the subtasks associated the client that made the request. We're doing this because in `receive/` we're not processing any particular subtask - the pending reponse may come from any of them. That's not the case in `send/` and in Celery tasks. So when adding this to Celery tasks, process timeouts only for the one particular subtask you're dealing with. And modify `send/` to do the same.

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/160252922

PR #820, opened 2018-09-13
Timeout only particular subtask in core-send and reduce number of database queries

pawelkisielewicz:
Resolves https://github.com/golemfactory/concent/issues/779
Additional reduced number of database queries

### DIFF
--- a/concent_api/core/subtask_helpers.py
+++ b/concent_api/core/subtask_helpers.py
@@ -10,134 +10,173 @@ from django.db.models import Model
 from django.db.models import Q
 from django.db.models import QuerySet
 from django.db.models.base import ModelBase
-from django.utils import timezone
+from golem_messages.message import Message
+from golem_messages.message.concents import ForcePayment
 from golem_messages.message.tasks import SubtaskResultsAccepted
 
-from core.models                import PendingResponse
-from core.models                import Subtask
+from core.models import PendingResponse
+from core.models import Subtask
 from core.payments import service as payments_service
-from core.transfer_operations   import store_pending_message
-from core.transfer_operations   import verify_file_status
+from core.transfer_operations import store_pending_message
+from core.transfer_operations import verify_file_status
 from core.validation import is_golem_message_signed_with_key
 from core.utils import hex_to_bytes_convert
-from common.helpers              import deserialize_message
-from common.helpers              import get_current_utc_timestamp
+from common.helpers import deserialize_message
+from common.helpers import get_current_utc_timestamp
 from common.helpers import parse_timestamp_to_utc_datetime
-from common                      import logging
+from common.logging import log_change_subtask_state_name
+from common.logging import log_string_message
 
 logger = getLogger(__name__)
 
 
-def update_timed_out_subtasks(client_public_key: bytes) -> None:
-    verify_file_status(client_public_key)
+def update_timed_out_subtask(subtask: Subtask) -> None:
+    """
+    Function called for timed out subtasks - checks state and changes it from one of actives to one of passives
+    """
+
+    subtasks_initial_state = subtask.state
+    if subtask.state == Subtask.SubtaskState.FORCING_REPORT.name:  # pylint: disable=no-member
+        update_subtask_state(
+            subtask=subtask,
+            state=Subtask.SubtaskState.REPORTED.name,  # pylint: disable=no-member
+        )
+        store_pending_message(
+            response_type=PendingResponse.ResponseType.ForceReportComputedTaskResponse,
+            client_public_key=subtask.provider.public_key_bytes,
+            queue=PendingResponse.Queue.Receive,
+            subtask=subtask,
+        )
+        store_pending_message(
+            response_type=PendingResponse.ResponseType.VerdictReportComputedTask,
+            client_public_key=subtask.requestor.public_key_bytes,
+            queue=PendingResponse.Queue.ReceiveOutOfBand,
+            subtask=subtask,
+        )
+    elif subtask.state == Subtask.SubtaskState.FORCING_RESULT_TRANSFER.name:  # pylint: disable=no-member
+        update_subtask_state(
+            subtask=subtask,
+            state=Subtask.SubtaskState.FAILED.name,  # pylint: disable=no-member
+        )
+        store_pending_message(
+            response_type=PendingResponse.ResponseType.ForceGetTaskResultFailed,
+            client_public_key=subtask.requestor.public_key_bytes,
+            queue=PendingResponse.Queue.Receive,
+            subtask=subtask,
+        )
+    elif subtask.state == Subtask.SubtaskState.FORCING_ACCEPTANCE.name:  # pylint: disable=no-member
+        update_subtask_state(
+            subtask=subtask,
+            state=Subtask.SubtaskState.ACCEPTED.name,  # pylint: disable=no-member
+        )
+        store_pending_message(
+            response_type=PendingResponse.ResponseType.SubtaskResultsSettled,
+            client_public_key=subtask.provider.public_key_bytes,
+            queue=PendingResponse.Queue.Receive,
+            subtask=subtask,
+        )
+        store_pending_message(
+            response_type=PendingResponse.ResponseType.SubtaskResultsSettled,
+            client_public_key=subtask.requestor.public_key_bytes,
+            queue=PendingResponse.Queue.ReceiveOutOfBand,
+            subtask=subtask,
+        )
+    elif subtask.state == Subtask.SubtaskState.VERIFICATION_FILE_TRANSFER.name:  # pylint: disable=no-member
+        update_subtask_state(
+            subtask=subtask,
+            state=Subtask.SubtaskState.FAILED.name,  # pylint: disable=no-member
+        )
+        store_pending_message(
+            response_type=PendingResponse.ResponseType.SubtaskResultsRejected,
+            client_public_key=subtask.provider.public_key_bytes,
+            queue=PendingResponse.Queue.ReceiveOutOfBand,
+            subtask=subtask,
+        )
+        store_pending_message(
+            response_type=PendingResponse.ResponseType.SubtaskResultsRejected,
+            client_public_key=subtask.requestor.public_key_bytes,
+            queue=PendingResponse.Queue.ReceiveOutOfBand,
+            subtask=subtask,
+        )
+    elif subtask.state == Subtask.SubtaskState.ADDITIONAL_VERIFICATION.name:  # pylint: disable=no-member
+        task_to_compute = deserialize_message(subtask.task_to_compute.data.tobytes())
+        # Worker makes a payment from requestor's deposit just like in the forced acceptance use case.
+        payments_service.make_force_payment_to_provider(  # pylint: disable=no-value-for-parameter
+            requestor_eth_address=task_to_compute.requestor_ethereum_address,
+            provider_eth_address=task_to_compute.provider_ethereum_address,
+            value=task_to_compute.price,
+            payment_ts=get_current_utc_timestamp(),
+        )
+
+        update_subtask_state(
+            subtask=subtask,
+            state=Subtask.SubtaskState.ACCEPTED.name,  # pylint: disable=no-member
+        )
+        store_pending_message(
+            response_type=PendingResponse.ResponseType.SubtaskResultsSettled,
+            client_public_key=subtask.provider.public_key_bytes,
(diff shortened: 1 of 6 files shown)


concent_api/core/subtask_helpers.py
```diff
@@ -38,106 +150,30 @@ def update_timed_out_subtasks(client_public_key: bytes) -> None:
...
-    )
+        update_timed_out_subtask(subtask)
+
+
+def _update_timed_out_subtask(
+    subtask_id: str,
+    client_public_key: bytes,
+) -> None:
+
+    try:
```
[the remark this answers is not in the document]
cameel: `with suppress(...):` looks nice but in this case we can't use it. This block should have an `else` clause:

``` python
    try:
        subtask = Subtask.objects.select_for_update().get(
            subtask_id=subtask_id,
            state__in=[state.name for state in Subtask.ACTIVE_STATES],
            next_deadline__lte=timezone.now()
        )
    except Subtask.DoesNotExist:
        pass
    else:
        verify_file_status(subtask_from_send_message=subtask, client_public_key=client_public_key)
        update_timed_out_subtask(subtask)
```

The `try ... except` block should only cover the operation that is expected to raise the exception. If for whatever reason `update_timed_out_subtask()` or `verify_file_status()` tries to fetch a subtask and raises `DoesNotExist`, it's a bug and should definitely not be simply ignored.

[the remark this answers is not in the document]
pawelkisielewicz: done, I left `try... except`

concent_api/core/subtask_helpers.py
```diff
@@ -11,133 +11,175 @@
...
+
+def update_timed_out_subtasks_in_message(client_message: Message, client_public_key: bytes) -> None:
+    if isinstance(client_message, ForcePayment):
+        for subtask_result_accepted in client_message.subtask_results_accepted_list:
+            _update_timed_out_subtask(subtask_result_accepted.subtask_id, client_public_key)
+    else:
+        _update_timed_out_subtask(client_message.subtask_id, client_public_key)
+
+
+def update_all_clients_timed_out_subtasks(client_public_key: bytes) -> None:
```
dybi: wrong function name: you update just 1 cleint's subtasks

concent_api/core/subtask_helpers.py
```diff
@@ -11,133 +11,175 @@
...
+
+    encoded_client_public_key = b64encode(client_public_key)
 
     clients_subtask_list = Subtask.objects.select_for_update().filter(
-        Q(requestor__public_key = b64encode(client_public_key)) | Q(provider__public_key = b64encode(client_public_key)),
-        state__in               = [state.name for state in Subtask.ACTIVE_STATES],
-        next_deadline__lte      = timezone.now()
+        Q(requestor__public_key=encoded_client_public_key) | Q(provider__public_key = encoded_client_public_key),
```
dybi: style

concent_api/core/subtask_helpers.py
```diff
@@ -11,133 +11,175 @@
...
-
-    logging.log_changes_in_subtask_states(
-        logger,
-        client_public_key,
-        clients_subtask_list.count(),
-    )
+        if subtask.state_enum == Subtask.SubtaskState.FORCING_RESULT_TRANSFER and subtask.requestor.public_key_bytes == client_public_key:
+            verify_file_status(subtask=subtask)
+
+# Subtask may change it's state in verify_file_status, so it is necessary to check the state again. If state will change
```
dybi: if this is to stay - move it, so it's aligned with code

concent_api/core/subtask_helpers.py
```diff
@@ -11,133 +11,175 @@
...
-                queue               = PendingResponse.Queue.ReceiveOutOfBand,
-                subtask             = subtask,
-            )
-
-    logging.log_changes_in_subtask_states(
-        logger,
-        client_public_key,
-        clients_subtask_list.count(),
-    )
+        if subtask.state_enum == Subtask.SubtaskState.FORCING_RESULT_TRANSFER and subtask.requestor.public_key_bytes == client_public_key:
```
dybi: this task might not be timed-out, but the function says "update **timed-out** subtasks"

concent_api/core/tasks.py
```diff
@@ -144,6 +149,9 @@ def verification_result(
         )
         return
 
+    if subtask.state_enum in Subtask.ACTIVE_STATES and subtask.next_deadline <= timezone.now():
```
dybi: please use `get_current_utc_timestamp()`

dybi: BTW: it should not be there - the handling is being done few lines below

concent_api/core/tasks.py
```diff
@@ -40,6 +42,9 @@ def upload_finished(subtask_id: str) -> None:
         logging.error(f'Task `upload_finished` tried to get Subtask object with ID {subtask_id} but it does not exist.')
         return
 
+    if subtask.state_enum in Subtask.ACTIVE_STATES and subtask.next_deadline <= timezone.now():
```
dybi: it should not be there - the handling is being done few lines below

concent_api/core/subtask_helpers.py

concent_api/core/subtask_helpers.py

concent_api/core/subtask_helpers.py
rwrzesien: It is still there, why changing the function name in the first place?

concent_api/core/transfer_operations.py
```diff
@@ -31,38 +31,33 @@
 logger = getLogger(__name__)
 
 
-def verify_file_status(client_public_key: bytes) -> None:
+def verify_file_status(
```
rwrzesien: Why splitting this? It has not more that 120 characters.

concent_api/core/views.py
```diff
@@ -32,11 +33,7 @@
 @log_communication
 def send(_request: HttpRequest, client_message: Message, client_public_key: bytes) -> Union[Message, HttpResponse]:
     assert isinstance(client_public_key, bytes) or client_public_key is None
-    if client_public_key is not None:
```
rwrzesien: Are we sure now key is always present? If yes please also remove it from `assert` above.

concent_api/core/subtask_helpers.py
```diff
@@ -11,133 +11,175 @@
...
+from common.logging import log_change_subtask_state_name
+from common.logging import log_string_message
 
 logger = getLogger(__name__)
 
 
-def update_timed_out_subtasks(client_public_key: bytes) -> None:
-    verify_file_status(client_public_key)
+def update_timed_out_subtask(subtask: Subtask) -> None:
+
```
rwrzesien: A comment here would be good.

concent_api/core/subtask_helpers.py
```diff
@@ -11,133 +11,175 @@
...
+            subtask=subtask,
+        )
+        store_pending_message(
+            response_type=PendingResponse.ResponseType.SubtaskResultsSettled,
+            client_public_key=subtask.requestor.public_key_bytes,
+            queue=PendingResponse.Queue.ReceiveOutOfBand,
+            subtask=subtask,
+        )
+    elif subtask.state == Subtask.SubtaskState.VERIFICATION_FILE_TRANSFER.name:  # pylint: disable=no-member
+
```
rwrzesien: This empty line should be removed.

concent_api/core/subtask_helpers.py
```diff
@@ -11,133 +11,175 @@
...
+            queue=PendingResponse.Queue.ReceiveOutOfBand,
+            subtask=subtask,
+        )
+        store_pending_message(
+            response_type=PendingResponse.ResponseType.SubtaskResultsSettled,
+            client_public_key=subtask.requestor.public_key_bytes,
+            queue=PendingResponse.Queue.ReceiveOutOfBand,
+            subtask=subtask,
+        )
+    log_string_message(
```
rwrzesien: Empty line should be placed above log function.

concent_api/core/views.py
```diff
@@ -53,7 +50,7 @@ def send(_request: HttpRequest, client_message: Message, client_public_key: byte
 @handle_errors_and_responses(database_name='control')
 def receive(_request: HttpRequest, message: Message, _client_public_key: bytes) -> Union[Message, HttpResponse]:
     assert isinstance(message.client_public_key, bytes)
-    update_timed_out_subtasks(
+    update_all_clients_timed_out_subtasks(
```
rwrzesien: Instead of splitting this into two functions, you could add `client_message` as an optional parameter to original `update_timed_out_subtasks` and limit the queryset depending on its presence.

concent_api/core/subtask_helpers.py
pawelkisielewicz: it was fixed, but not pushed :)

concent_api/core/subtask_helpers.py
pawelkisielewicz: done. I would keep this comment, this construction is not obvious, so I think if someone else will read this in future it will help

concent_api/core/tasks.py
pawelkisielewicz: right, removed

concent_api/core/tasks.py
pawelkisielewicz: removed

concent_api/core/transfer_operations.py
pawelkisielewicz: probably IDE 'did it'.. Fixed :)

concent_api/core/views.py
pawelkisielewicz: Right, it sometimes is None. Fixed and passed as `Optional`

pawelkisielewicz: :+1: for see this. Sometimes it is `None`. Fixed and passed as Oprional

concent_api/core/subtask_helpers.py

concent_api/core/subtask_helpers.py

concent_api/core/subtask_helpers.py
pawelkisielewicz: done, hope you will like it

concent_api/core/subtask_helpers.py
```diff
@@ -10,134 +10,173 @@
...
+        )
+
+    log_string_message(
+        logger,
+        f"Subtask with id: {subtask.subtask_id} changed it's state from: {subtasks_initial_state} to: {subtask.state}. "
+        f"Provider id: {subtask.provider_id}. Requestor id: {subtask.requestor_id}."
+    )
+
+
+def update_timed_out_subtasks_in_message(client_message: Message, client_public_key: bytes) -> None:
```
cameel: This function really should have a better name or at least a docstring that says what it's supposed to do.

pawelkisielewicz: sorry for this, didn't saw your comment before merge. I will fix it today in other PR.

cameel: I added it after merge :)

I wanted to review this pull request but it was merged before I managed to so - so I just took a brief look now. I did not find any problems and this is the only thing that stood out to me. There are two very similarly named functions and one checks upload status and the other does not but the name does not suggest such a difference at all.

@dybi @rwrzesien As a general observation - when a function has a good, descriptive name, docstring may be superfluous. But in situations like this where you have trouble finding a good name I think it's a good idea to have a docstring that explains it more verbosely.

pawelkisielewicz: Good naming is hard and comes with experience :) I think in this case docstrings are not necessary, I tried to make descriptive names in views:
`update_subtasks_from_incomming_message_if_timed_out` - send
`update_all_timed_out_subtasks_of_client` - receive
And both this functions select subtasks to update and call `_update_timed_out_subtask`  
Is it OK?

cameel: It's still confusing because by looking at the names of `update_subtasks_from_incomming_message_if_timed_out()` and `_update_timed_out_subtask()` I would never suspect that the first one checks if files have been uploaded and the other does not. It also does not indicate that one locks the subtask and the other does not.

If you can't or don't want include this information in the name, you should put it in a docstring.

BTW,
- `incomming` -> `incoming`
- `of_client` -> `of_a_client`

pawelkisielewicz: Let's move this discussion to https://github.com/golemfactory/concent/pull/841. I will fix it here. I tried to make good names, but if it is still confusing I will rename it again and add docstring. Everything should be obvious :)

Issue #837, opened 2018-09-19
Wrong dependency in system checks for MiddleMan


dybi:
There are two system checks:
- `check_middleman_port`
- `check_middleman_address`
where presence of specific settings is dependent on `'middleman' in settings.CONCENT_FEATURES`.

This is wrong, as  `CONCENT_FEATURES` is a declarations what features **are provided** by given pod and **not** what features are needed.

Both checks should be done when `settings.USE_SIGNING_SERVICE` is present and is set to `True`.

rwrzesien:
`USE_SIGNING_SERVICE` is a temporary solution, it will be removed at some point.

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/160616536

cameel:
Yeah, `USE_SIGNING_SERVICE` is going to be removed (#821) but @dybi has a point - we should not require Middleman port and address to be set when this setting is false.

But the main problem in this issue is simply depending on the wrong value of `CONCENT_FEATURES`. The fix is trivial - depend on `'concent-api'` and `'concent-worker'` rather than `'middleman'`. Those are the components that the SCI callback might be executed on.

PR #839, opened 2018-09-19
Change dependencies in system checks 

pawelkisielewicz:
resolves https://github.com/golemfactory/concent/issues/837


cameel:
This pull request and branch should have better names. It's not clear what it does without looking at the code.

Issue #630, opened 2018-07-11
Middleman: Signing Service authentication


cameel:
This is a part of the implementation of the "Concent Signing Service" blueprint (#597). This task depends on #618.

#### Overview
As implemented now Middleman simply accepts connections from the Signing service without asking any questions. While a rogue service won't be able to impersonate the service because its messages need to be signed with the right key, an attacker can still prevent the real service from connecting and sit quietly hogging the connection.

To prevent this Middleman should not start the normal communication with the service until it manages to respond correctly to an authentication challenge.

Connections with Concent do not need to be authenticated. We're not worried about a rogue connection blocking others because Middleman is prepared to maintain multiple connections with Concent at any given time. Also, they're never accepted on the external port and can come only from inside the cluster.

#### Main loop
Modify the logic for accepting new connections
1. Allow multiple connections as long as none of them is authenticated:
    - If there are no established connections or if no connection is authenticated yet, allow new connections to start the authentication handler.
    - If any connection successfully passes the authentication process, immediately terminate all other connections with the service
    - If a connection is established and authenticated, do not allow new connections until this one ends.
2. Once a new connections authenticates successfully, enter the connection handler.

#### Signing Service authentication handler:
1. Generate a random string.
    - Make sure that you use a cryptographically secure pseudo-random number generator. See what golem-messages uses internally for its encryption and use the same thing.
    - Use strings of the same length. Not too short but not too long.
2. Start listening for a response from the service.
3. If the service does not respond in a predefined time, terminate the connection.
    - This time should not be too long.
4. Validate the response. If the validation fails, terminate the connection:
    - The response must be a valid data frame as described in the protocol.
    - It must be the authentication response. Any other message type is treated as a failure to authenticate.
    - The signature must match the public key of the service.
5. If the connection was terminated, log the incident at `INFO` level and return an error.
6. Otherwise return a success.

PaweuB:
Pivotal: 
https://www.pivotaltracker.com/story/show/158999840

cameel:
@dybi Stuff like `ENABLE_AUTHENTICATION` should be mentioned before you close the issue.

Also why is this a constant and not a setting? How are we going to test authentication if it's permanently disabled?

PR #843, opened 2018-09-21
Feature middleman authentication

dybi:
resolves: https://github.com/golemfactory/concent/issues/630

I am not fully happy with this implementation, but is seems to work. I would do some refactor (extract additional classes to have better/easier unit tests) in a separate PR if it is ok.

### DIFF
--- a/concent_api/middleman/middleman_server.py
+++ b/concent_api/middleman/middleman_server.py
@@ -5,7 +5,9 @@ from collections import OrderedDict
 from contextlib import suppress
 from logging import getLogger
 from typing import Iterable
+from typing import List
 from typing import Optional
+from typing import Tuple
 import signal
 
 from middleman.constants import CONNECTION_COUNTER_LIMIT
@@ -14,6 +16,7 @@ from middleman.constants import DEFAULT_INTERNAL_PORT
 from middleman.constants import ERROR_ADDRESS_ALREADY_IN_USE
 from middleman.constants import LOCALHOST_IP
 from middleman.constants import PROCESSING_TIMEOUT
+from middleman.queue_operations import is_authenticated
 from middleman.queue_operations import request_consumer
 from middleman.queue_operations import request_producer
 from middleman.queue_operations import response_consumer
@@ -23,6 +26,9 @@ from middleman_protocol.constants import MAXIMUM_FRAME_LENGTH
 
 logger = getLogger(__name__)
 crash_logger = getLogger('crash')
+
+# TODO: remove together with corresponding flag in signing_service.py when testing on cluster is finished.
+ENABLE_AUTHENTICATION = False
 
 
 class MiddleMan:
@@ -44,6 +50,7 @@ class MiddleMan:
         self._request_queue: asyncio.Queue = asyncio.Queue(loop=self._loop)
         self._response_queue_pool = QueuePool(loop=self._loop)
         self._message_tracker: OrderedDict = OrderedDict()
+        self._ss_connection_candidates: List[Tuple[asyncio.Task, asyncio.StreamWriter]] = []
 
         # Handle shutdown signal.
         self._loop.add_signal_handler(signal.SIGTERM, self._terminate_connections)
@@ -175,9 +182,12 @@ class MiddleMan:
         if self._is_signing_service_connection_active:
             writer.close()
         else:
-            self._is_signing_service_connection_active = True
             tasks: list = []
             try:
+                successful = await self._authenticate_signing_service(reader, writer)
+                if not successful:
+                    writer.close()
+                    return
                 request_consumer_task = self._loop.create_task(
                     request_consumer(
                         self._request_queue,
@@ -230,3 +240,34 @@ class MiddleMan:
                 # Cancelled task raises asyncio.CancelledError that we can suppress:
                 with suppress(asyncio.CancelledError):
                     self._loop.run_until_complete(task)
+
+    async def _authenticate_signing_service(self, reader: asyncio.StreamReader, writer: asyncio.StreamWriter) -> bool:
+        if not ENABLE_AUTHENTICATION:
+            return True
+        logger.info("Signing Service candidate has connected, authenticating...")
+        authentication_task = self._loop.create_task(is_authenticated(reader, writer))
+        index = len(self._ss_connection_candidates)
+        self._ss_connection_candidates.append((authentication_task, writer))
+
+        await authentication_task
+        self._ss_connection_candidates.pop(index)
+
+        is_signing_service_authenticated = authentication_task.result()
+        if is_signing_service_authenticated:
+            logger.info("Authentication successful: Signing Service has connected.")
+            self._is_signing_service_connection_active = True
+            self._abort_ongoing_authentication()
+        else:
+            logger.info("Authentication unsuccessful, closing connection with candidate.")
+        return is_signing_service_authenticated
+
+    def _abort_ongoing_authentication(self) -> None:
+        counter = 0
+        length = len(self._ss_connection_candidates)
+        for task, writer in self._ss_connection_candidates:
+            logger.info(f"Canceling task {counter}/{length}...")
+            task.cancel()
+            writer.close()
+            logger.info("Canceled!")
+            counter += 1
+        self._ss_connection_candidates.clear()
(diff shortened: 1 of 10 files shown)


concent_api/middleman/queue_operations.py
```diff
@@ -100,19 +107,19 @@
...
     signing_service_reader: StreamReader,
     message_tracker: OrderedDict,
 ) -> None:
     while True:
         try:
             frame = await handle_frame_receive_async(signing_service_reader, settings.SIGNING_SERVICE_PUBLIC_KEY)
-        except IncompleteReadError:
+        except IncompleteReadError as exc:
+            logger.exception(str(exc))
             logger.info(f"Signing Service has closed the connection. Ending 'response_producer' coroutine")
```
pawelkisielewicz: do we still need it `logger.info` if you added `logger.exception`?

concent_api/middleman/queue_operations.py
```diff
@@ -162,6 +169,30 @@
...
+async def is_authenticated(reader: StreamReader, writer: StreamWriter) -> bool:
+    challenge = create_random_challenge()
+
+    request_id = randint(1, 7777)
+    frame = AuthenticationChallengeFrame(challenge, request_id)
+    await send_over_stream_async(frame, writer, settings.CONCENT_PRIVATE_KEY)
+    logger.info(f"Challenge has been sent for request ID: {request_id}.")
+    received_frame = await handle_frame_receive_async(reader, settings.SIGNING_SERVICE_PUBLIC_KEY)
+    logger.info(f"Response has been received for request ID: {request_id}.")
+    if received_frame.request_id != request_id:
```
pawelkisielewicz: Below you are check if `isinstance(received_frame, AuthenticationResponseFrame)`. Maybe it would be good to check it first and as second check  `if received_frame.request_id != request_id:`. If received_frame won't be `AuthenticationResponseFrame` instance it may don't have field `request_id` (received_frame.request_id ) and it may be cause of error.

concent_api/middleman/tests/test_middleman.py
```diff
@@ -185,3 +173,81 @@ def test_that_broken_connection_from_signing_service_is_reported_to_sentry(self)
...
+
+    def test_that_when_authentication_is_unsuccessful_then_connection_ends(self, event_loop):
+        self.middleman._is_signing_service_connection_active = False
+        mocked_reader = prepare_mocked_reader(b"some bytes")
+        mocked_writer = prepare_mocked_writer()
+
+        with mock.patch.object(
+            self.middleman,
+            "_authenticate_signing_servcie",
+            new=async_stream_actor_mock(return_value = False)
```
kbeker: Please fix style inside function

concent_api/middleman/tests/test_middleman.py
```diff
@@ -185,3 +173,81 @@ def test_that_broken_connection_from_signing_service_is_reported_to_sentry(self)
...
+
+    def test_that_when_authentication_is_successful_connection_lasts_until_its_end(self, event_loop):
+        with override_settings(
+            CONCENT_PRIVATE_KEY=CONCENT_PRIVATE_KEY,
+            CONCENT_PUBLIC_KEY=CONCENT_PUBLIC_KEY,
+            SIGNING_SERVICE_PUBLIC_KEY=SIGNING_SERVICE_PUBLIC_KEY,
+        ):
+            self.middleman._is_signing_service_connection_active = False
+            some_bytes = b"some bytes"
+            mocked_reader = prepare_mocked_reader(some_bytes, side_effect=asyncio.IncompleteReadError(some_bytes, 123))
```
kbeker: It might be more readable when we would break this function to separate lines

concent_api/middleman/middleman_server.py
```diff
@@ -44,6 +47,7 @@ def __init__(
         self._request_queue: asyncio.Queue = asyncio.Queue(loop=self._loop)
         self._response_queue_pool = QueuePool(loop=self._loop)
         self._message_tracker: OrderedDict = OrderedDict()
+        self._ss_connection_candidates: List[Tuple[asyncio.Task, asyncio.StreamWriter]] = []
```
rwrzesien: I think we should use full name here, `ss` => `signing_service`. Especially in this case ;)

concent_api/middleman/middleman_server.py
```diff
@@ -175,9 +179,12 @@ def _close_middleman(self) -> None:
         if self._is_signing_service_connection_active:
             writer.close()
         else:
-            self._is_signing_service_connection_active = True
             tasks: list = []
             try:
+                successful = await self._authenticate_signing_servcie(reader, writer)
```
rwrzesien: Typo `servcie` => `service`.

concent_api/middleman/middleman_server.py
```diff
@@ -230,3 +237,32 @@ def _cancel_pending_tasks(self, tasks: Iterable[asyncio.Task], await_cancellatio
                 # Cancelled task raises asyncio.CancelledError that we can suppress:
                 with suppress(asyncio.CancelledError):
                     self._loop.run_until_complete(task)
+
+    async def _authenticate_signing_servcie(self, reader: asyncio.StreamReader, writer: asyncio.StreamWriter) -> None:
+        logger.info("Signing Service candidate has connected, authenticating...")
+        authentication_task = self._loop.create_task(is_authenticated(reader, writer))
+        index = len(self._ss_connection_candidates)
```
rwrzesien: If this is used only once, maybe it is better to use it directly.

concent_api/middleman/queue_operations.py
```diff
@@ -162,6 +169,30 @@
         response_queue.task_done()
 
 
+async def is_authenticated(reader: StreamReader, writer: StreamWriter) -> bool:
+    challenge = create_random_challenge()
+
+    request_id = randint(1, 7777)
```
rwrzesien: This one looks suspicious. Shouldn't `request_id` be always unique? Maybe use the same method as for generating `request_id` for other frames. After all, authentication frames are frames too.

concent_api/middleman/queue_operations.py
rwrzesien: I also think that two checks should be combined together if they result in the same (here: `return False`).

dybi: done

concent_api/middleman/tests/test_middleman.py
dybi: done

concent_api/middleman/tests/test_middleman.py
dybi: done

concent_api/middleman/middleman_server.py
dybi: quite frankly, I would love to keep it that way - as a form of desperate protest against changes in naming in python community ;)

concent_api/middleman/middleman_server.py
dybi: done

concent_api/middleman/middleman_server.py
dybi: I need to get the index of the last element of the list, then I modify it.

concent_api/middleman/queue_operations.py
dybi: actually, during the authentication phase it does not matter: it doesn't have to be unique, it has to be unique for short moment of simultaneous authentication. It is only to map the request with response - randomly choosing it from pretty wide range should be good enough.

cameel: 7777 does not seem like a very wide range.

Maybe this is the cause of #844?

rwrzesien: I think this is only for authentication case.

concent_api/middleman/middleman_server.py
```diff
@@ -175,9 +182,12 @@ def _close_middleman(self) -> None:
         if self._is_signing_service_connection_active:
             writer.close()
         else:
-            self._is_signing_service_connection_active = True
             tasks: list = []
             try:
+                successful = await self._authenticate_signing_service(reader, writer)
+                if not successful:
+                    writer.close()
+                    return
```
Jakub89: We don't have a test for this part, when authentication is not successful. Just saying :P

PR #852, opened 2018-09-25
Release notes

cameel:
This pull request adds release notes for code that's been merged into `master` so far. Notes for old version were just copied from our tags. Notes for the upcoming 0.9.0 were written from scratch.

From now on every pull request that affects the final product in a way visible to the repository user will have to update this file.

See https://github.com/Code-Poets/coding-style/issues/19 for our rules on release notes.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,3 +1,70 @@
+### 0.9.0
+#### Signing Service and Middleman
+- Signing Service and Middleman are now fully implemented but disabled by default (#673, #633, #616, #629, #618, #625, #632).
+- Concent can now be configured to use Signing Service instead of signing transaction by itself (#635).
+- Authentication between Middleman and Signing Service - disabled by default (#631, #630).
+- More consistent command-line options for passing secrets to the Signing Service (#671).
+- Signing Service can now be configured to send crash reports to Sentry (#762).
+- Bugfix: The installation script (`setup.py`) no longer requires golem-messages to be already installed (#800).
+- Bugfix: Middleman was not properly cancelling asyncio tasks (#799).
+- Bugfix: Fixed error in task scheduling in `QueuePool` in Middleman (#790).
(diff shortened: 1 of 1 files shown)


RELEASE-NOTES.md
```diff
@@ -0,0 +1,281 @@
+### 0.9.0
+#### Signing Service and Middleman
+- Signing Service and Middleman are now fully implemented but disabled by default (#673, #633, #616, #629, #618, #625, #632).
+- Concent can now be configured to use Signing Service instead of signing transaction by itself (#635).
```
rwrzesien: Maybe, like for `Bugfix:`, we should add a prefix for other types, ie. `Feature: ` or `Improvement: `

rwrzesien:
In latest notes there is a division between different components, and before that there is not (I guess that is because there were not so many). So in next notes we want to keep them separated? What if some feature will touch many components?

RELEASE-NOTES.md
cameel: I think it's clear from context. I only added it for bugs not to constantly write "fixes bug xxx".

What do others think?

cameel:
> In latest notes there is a division between different components, and before that there is not (I guess that is because there were not so many).

Yes, when releases are big the list is very long and I think grouping things makes it easier to read and understand.

Also, I pasted the old notes with only very minor tweaks. I do not think redoing the old notes is worth the time now. Maybe at a later time when the conventions get more established.

> So in next notes we want to keep them separated? What if some feature will touch many components?

We can have a plain list and only group it just before tagging a new release.

There already are features that fit into more than one group. I solve it by just putting it in the one where it fits more or by creating a new group.

cameel:
I'm merging this since there are no objections. I've added two more lines to 0.9.0 for pull requests that have been merged in the meantime.

Please keep this list up to date and assign version to closed pull requests and issues from now on.

Issue #743, opened 2018-08-20
[Bankster Testing] MiddleMan


rwrzesien:
This is part of #647. It depends on #740.

- Create one wrapper which:
    - Creates a socket and connects it to given component.
    - Creates a given `Frame` with given payload.
    - Uses `MiddleMan` protocol for sending `Frame` over created socket.
    - Uses `MiddleMan` protocol for receiving response.
    - Verifies that response is as expected.

Create a test suite for testing communication flow: request to `MiddleMan` => `MiddleMan` should send request and get response from `SigningService` => Response from `MiddleMan`.

TODO: Specify test cases

dybi:
@rwrzesien: defined scanarios are here: https://github.com/golemfactory/concent/issues/724

rwrzesien:
@dybi I think the fun will start when you add routing and queues :) I see you have added the scenarios I was thinking of too, like sending several messages simultaneously, and mixing correct and incorrect messages. Lets see what else can we think of. 

If you want we can merge those two issues.

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/159927457

PR #863, opened 2018-10-01
Chore recovery cluster tests

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/743

### DIFF
--- a/concent_api/signing_service_integration_test_recovery.py
+++ b/concent_api/signing_service_integration_test_recovery.py
@@ -1,0 +1,279 @@
+#!/usr/bin/env python3
+
+from configparser import ConfigParser
+from time import sleep
+from typing import Any
+import os
+import socket
+import subprocess
+import sys
+
+from api_testing_common import assert_condition
+from api_testing_common import count_fails
+from common.helpers import get_current_utc_timestamp
+from middleman_protocol.message import GolemMessageFrame
+from middleman_protocol.concent_golem_messages.message import SignedTransaction
+from middleman_protocol.stream import send_over_stream
+from middleman_protocol.stream import unescape_stream
+from signing_service_testing_common import Components
+from signing_service_testing_common import create_golem_message_frame_with_correct_transaction_signing_request
+from signing_service_testing_common import run_tests
+from signing_service_testing_common import send_message_to_middleman_and_receive_response
+from signing_service_testing_common import send_message_to_middleman_without_response
+
+
+REQUIRED_COMPONENTS = []  # type: ignore
+
+SLEEP_TIME_AFTER_SPAWNING_PROCESS = 10
+SLEEP_TIME_AFTER_KILLING_PROCESS = 10
+
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "concent_api.settings")
+
+
+def run_middleman() -> subprocess.Popen:
+    return subprocess.Popen(
+        [
+            'python',
+            'manage.py',
+            'middleman'
+        ]
+    )
+
+
+def run_signing_service() -> subprocess.Popen:
+    return subprocess.Popen(
+        [
+            'python',
+            '-m',
+            'signing_service.signing_service',
+            '127.0.0.1',
+            '"85cZzVjahnRpUBwm0zlNnqTdYom1LF1P1WNShLg17cmhN2UssnPrCjHKTi5susO3wrr/q07eswumbL82b4HgOw=="',
+            '--signing-service-private-key',
+            '"GfkWzCRGL/Me2wk41mS0NRSch2XMpgIXYLRaasmcozw="',
+            '--ethereum-private-key',
+            '"M2ExMDc2YmY0NWFiODc3MTJhZDY0Y2NiM2IxMDIxNzczN2Y3ZmFhY2JmMjg3MmU4OGZkZDlhNTM3ZDhmZTI2Ng=="'
+        ],
+        cwd='../signing_service/'
+    )
+
+
+@count_fails
+def test_case_0_prove_of_concept(config: ConfigParser, **kwargs: Any) -> None:
+    """
+    1. Spawn MiddleMan and SigningService processes.
+    2. Client sends GolemMessageFrame with correct TransactionSigningRequest to MiddleMan
+       and receives response to prove that connection works.
+    """
+
+    try:
+        middleman_process = run_middleman()
+        signing_service_process = run_signing_service()
+
+        # Waiting for MiddleMan and SigningService to start.
+        sleep(SLEEP_TIME_AFTER_SPAWNING_PROCESS)
+
+        # Create GolemMessageFrame with correct TransactionSigningRequest.
+        golem_message_frame = create_golem_message_frame_with_correct_transaction_signing_request(
+            request_id=get_current_utc_timestamp(),
+        )
+
+        # Send message through wrapper and receive deserialized response.
+        response = send_message_to_middleman_and_receive_response(
+            message=golem_message_frame,
+            config=config,
+            concent_private_key=CONCENT_PRIVATE_KEY,
+            concent_public_key=CONCENT_PUBLIC_KEY,
+        )
+
+        # Check response.
+        assert_condition(
+            type(response),
+            GolemMessageFrame,
+            f'Deserialized response type is {type(response)} instead of GolemMessageFrame.'
+        )
+        assert_condition(
+            type(response.payload),
+            SignedTransaction,
+            f'Deserialized response payload type is {type(response.payload)} instead of SignedTransaction.'
+        )
+        assert_condition(
+            response.request_id,
+            golem_message_frame.request_id,
+            f'Deserialized response request_id is {response.request_id} instead of {golem_message_frame.request_id}.'
+        )
+
+    finally:
+        middleman_process.kill()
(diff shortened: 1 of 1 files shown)


concent_api/signing_service_integration_test_recovery.py
```diff
@@ -0,0 +1,246 @@
...
+            concent_private_key=CONCENT_PRIVATE_KEY,
+            concent_public_key=CONCENT_PUBLIC_KEY,
+        )
+
+        # Check response.
+        assert_condition(type(response), GolemMessageFrame, f'Deserialized response type is {type(response)} instead of GolemMessageFrame.')
+        assert_condition(type(response.payload), SignedTransaction, f'Deserialized response payload type is {type(response.payload)} instead of SignedTransaction.')
+        assert_condition(response.request_id, golem_message_frame.request_id, f'Deserialized response request_id is {response.request_id} instead of {golem_message_frame.request_id}.')
+
+    except Exception as exception:
```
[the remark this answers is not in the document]
Jakub89: I think @cameel would not be happy about putting such big portion of code in `try/except` block.

concent_api/signing_service_integration_test_recovery.py
```diff
@@ -0,0 +1,246 @@
...
+
+        middleman_process = run_middleman()
+
+        # Waiting for MiddleMan to start.
+        sleep(10)
+
+        receive_frame_generator = unescape_stream(connection=client_socket)
+        try:
+            next(receive_frame_generator)
+        except socket.error as exception:
```
Jakub89: `try/except` inside `try/except`, Bro. :P

rwrzesien: What is the problem with it ? ;)

concent_api/signing_service_integration_test_recovery.py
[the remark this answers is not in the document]
rwrzesien: Ok guys, let me explain the main problem with decorator - it would fit only for first test (which is not really a test but a prove of concept that MiddleMan and SigningService will communicate when spawn as processes inside test). In other tests, I am first creating instances 
```
middleman_process = run_middleman()
signing_service_process = run_signing_service()
```

but then in the middle of the test, I am killing and recreating one of them
```
middleman_process.kill()
...
middleman_process = run_middleman()
```

So sometimes there would be the same instance in `try` and `finally` blocks (if something goes wrong before recreating) but in most cases, including successful test, it would be a different one. Then it would try to kill already killed process, leaving another one not killed.
 
About re-raising exceptions you are right, I have removed those :+1: done

concent_api/signing_service_integration_test_recovery.py
```diff
@@ -0,0 +1,246 @@
...
+            concent_private_key=CONCENT_PRIVATE_KEY,
+            concent_public_key=CONCENT_PUBLIC_KEY,
+        )
+
+        # Check response.
+        assert_condition(type(response), GolemMessageFrame, f'Deserialized response type is {type(response)} instead of GolemMessageFrame.')
+        assert_condition(type(response.payload), SignedTransaction, f'Deserialized response payload type is {type(response.payload)} instead of SignedTransaction.')
+        assert_condition(response.request_id, golem_message_frame.request_id, f'Deserialized response request_id is {response.request_id} instead of {golem_message_frame.request_id}.')
+
+    except Exception as exception:
```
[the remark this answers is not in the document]
rwrzesien: :+1: done

concent_api/signing_service_integration_test_recovery.py
```diff
@@ -0,0 +1,242 @@
...
+    2. Client sends GolemMessageFrame with correct TransactionSigningRequest to MiddleMan
+       and receives response to prove that connection works.
+    """
+
+    try:
+        middleman_process = run_middleman()
+        signing_service_process = run_signing_service()
+
+        # Waiting for MiddleMan and SigningService to start.
+        sleep(SLEEP_TIME_AFTER_SPAWNING_OR_KILLING_PROCESS)
```
pawelkisielewicz: I don't know if it is possible but can you instead of waiting check (e.g. in loop with short sleep time) if MM and SS started and break the loop when they are? It will be much faster

concent_api/signing_service_integration_test_recovery.py
```diff
@@ -0,0 +1,242 @@
...
+
+        middleman_process.kill()
+
+        # Waiting for MiddleMan to finish.
+        sleep(SLEEP_TIME_AFTER_SPAWNING_OR_KILLING_PROCESS)
+
+        middleman_process = run_middleman()
+
+        # Waiting for MiddleMan to start.
+        sleep(SLEEP_TIME_AFTER_SPAWNING_OR_KILLING_PROCESS)
```
pawelkisielewicz: `SLEEP_TIME_AFTER_SPAWNING_OR_KILLING_PROCESS` is 10, so this test takes at least 30 seconds. Is it possible to do it faster?

concent_api/signing_service_integration_test_recovery.py
Jakub89: I know that @cameel is not here right now, but he's always pointing thinks like that out. `try/except` block should always catch exception that we expect to happen sometimes. I understand this part `except socket.error ...`, but I don't understand why the whole test is a `try/except` block.

concent_api/signing_service_integration_test_recovery.py
rwrzesien: No, it is not possible, you can't get info directly from the process, only get return code or parse `stdout` and `stderr` but for test purposes it would be overkill.

concent_api/signing_service_integration_test_recovery.py
rwrzesien: 10 is safe value, because this test might be run on different machines. For example when I was running it on debug mode in Pycharm, which always make it slower, 10 seconds was not enough.

concent_api/signing_service_integration_test_recovery.py
rwrzesien: But we do it all the time :) Consider this - we use view decorators, which does `try/except`, and they call view code in `try` block, and inside view we do more `try/except`.

Actually here the `except socket.error as exception` is part of the test flow, which checks if old connection is dead after `SigningService` restart. This is the only way to verify this I think. I know it looks weird, but the whole test case is not typical one.

> but I don't understand why the whole test is a try/except block.

If anything unexpected happens during the test, `SigningService` and `MiddleMan` processed must be killed, otherwise you would have to kill them manually in console. It is done for convenience mostly.

Jakub89: >  I know it looks weird, but the whole test case is not typical one.

True, okey then, I just wanted to be sure there was no other reasonable option to do this.

concent_api/signing_service_integration_test_recovery.py
pawelkisielewicz: ok, thanks

concent_api/signing_service_integration_test_recovery.py
pawelkisielewicz: ok, I didn't meant to decrease this time, only thought about a return message from app or check if process is started or killed. But you answered it above

concent_api/signing_service_integration_test_recovery.py
```diff
@@ -0,0 +1,242 @@
...
+from signing_service_testing_common import Components
+from signing_service_testing_common import create_golem_message_frame_with_correct_transaction_signing_request
+from signing_service_testing_common import run_tests
+from signing_service_testing_common import send_message_to_middleman_and_receive_response
+from signing_service_testing_common import send_message_to_middleman_without_response
+
+
+REQUIRED_COMPONENTS = []  # type: ignore
+
+SLEEP_TIME_AFTER_SPAWNING_OR_KILLING_PROCESS = 10
```
dybi: Please have two separate constants - one for SPAWNING_WAIT_TIME and one for KILLING_WAIT_TIME. Right now they are equal, but this might change and they are conceptually different, so we should have them separated.

concent_api/signing_service_integration_test_recovery.py
```diff
@@ -0,0 +1,242 @@
...
+            message=golem_message_frame,
+            config=config,
+            concent_private_key=CONCENT_PRIVATE_KEY,
+            concent_public_key=CONCENT_PUBLIC_KEY,
+        )
+
+        # Check response.
+        assert_condition(type(response), GolemMessageFrame, f'Deserialized response type is {type(response)} instead of GolemMessageFrame.')
+        assert_condition(type(response.payload), SignedTransaction, f'Deserialized response payload type is {type(response.payload)} instead of SignedTransaction.')
+        assert_condition(response.request_id, golem_message_frame.request_id, f'Deserialized response request_id is {response.request_id} instead of {golem_message_frame.request_id}.')
```
kbeker: I would break all this `assert_condition` to separate lines because in GitHub I can't see whole line.

concent_api/signing_service_integration_test_recovery.py
rwrzesien: :+1: done

concent_api/signing_service_integration_test_recovery.py
rwrzesien: :+1: done

PR #864, opened 2018-10-03
Chore change golem messages from v2.14.0 to v2.15.0

Jakub89:
Apparently there's no issue for this, @PaweuB ?


concent_api/common/tests/test_api_view.py
```diff
@@ -218,10 +219,23 @@ def setUp(self):
         message_timestamp = get_current_utc_timestamp() + deadline_offset
         compute_task_def = ComputeTaskDefFactory()
         compute_task_def['deadline'] = message_timestamp
+        want_to_compute_task = WantToComputeTaskFactory(
+            provider_public_key=encode_hex(PROVIDER_PUBLIC_KEY),
+        )
+        want_to_compute_task = load(
```
rwrzesien: Why not using `.sign_message()` to achieve this?

concent_api/common/tests/test_api_view.py
```diff
@@ -218,10 +219,23 @@ def setUp(self):
...
+            ),
+            settings.CONCENT_PRIVATE_KEY,
+            PROVIDER_PUBLIC_KEY,
+            check_time=False,
+        )
         task_to_compute = tasks.TaskToComputeFactory(
             compute_task_def=compute_task_def,
             requestor_public_key=encode_hex(REQUESTOR_PUBLIC_KEY),
-            provider_public_key=encode_hex(PROVIDER_PUBLIC_KEY),
+            want_to_compute_task=want_to_compute_task,
```
rwrzesien: `TaskToComputeFactory` have possibility to create nested `WantToComputeTask` with custom key like this:
```
task_to_compute = tasks.TaskToComputeFactory(
    ...
    want_to_compute_task__provider_public_key=encode_hex(PROVIDER_PUBLIC_KEY),
)
```

There is no need to create `WantToComputeTask` manually.

concent_api/core/models.py
```diff
@@ -379,6 +391,7 @@ class Meta:
 
     # Related messages
     task_to_compute = OneToOneField(StoredMessage, related_name='subtasks_for_task_to_compute')
+    want_to_compute_task = OneToOneField(StoredMessage, related_name='subtasks_for_want_to_compute_task')
     report_computed_task = OneToOneField(StoredMessage, related_name='subtasks_for_report_computed_task')
     ack_report_computed_task = OneToOneField(StoredMessage, blank=True, null=True, related_name='subtasks_for_ack_report_computed_task')
     reject_report_computed_task = OneToOneField(StoredMessage, blank=True, null=True, related_name='subtasks_for_reject_report_computed_task')
```
kbeker: I can't find changes to validation that `want_to_compute` stored separately is the same message as `want_to_compute` nested inside another messages. Please add this validation in the same way as it is now for `TTC` and `RCT`.

concent_api/core/tests/utils.py
```diff
@@ -299,6 +295,47 @@ def _get_deserialized_task_to_compute(
...
+                price=price,
+                concent_enabled=concent_enabled,
+                extra_data=extra_data,
+                provider_public_key=(
+                    provider_public_key if provider_public_key is not None else self._get_provider_hex_public_key()
+                ),
+                provider_ethereum_public_key=(
+                    provider_ethereum_public_key if provider_ethereum_public_key is not None else self._get_provider_ethereum_hex_public_key()
+                ),
+            )
```
kbeker: I think, that it should not be provided `None` everywhere. If there is `None` it should get values from Golem's Factories.

concent_api/core/tests/test_core_views.py
```diff
@@ -169,8 +172,10 @@ def test_send_should_return_http_400_if_data_is_incorrect(self):
         compute_task_def = message.ComputeTaskDef()
         task_to_compute = message.TaskToCompute(
             compute_task_def=compute_task_def,
-            provider_public_key=self._get_encoded_provider_public_key(),
             requestor_public_key=self._get_encoded_requestor_public_key(),
+            want_to_compute_task=WantToComputeTaskFactory(
```
rwrzesien: Same here.

concent_api/common/tests/test_api_view.py
Jakub89: I thought that it's a method from our `ConcentIntegrationTestCase`. I'm gone use our `sign_message()` from `common.helpers` instead of GM once, since we use it everywhere.

concent_api/core/tests/test_core_views.py
```diff
@@ -476,43 +477,35 @@ def test_send_should_return_http_202_if_task_to_compute_deadline_is_correct(self
...
 
     @freeze_time("2017-11-17 10:00:00")
     def test_send_task_to_compute_with_public_key_with_wrong_length_should_return_http_400(self):
         assert StoredMessage.objects.count() == 0
 
-        for field_name in [
-            'provider_public_key',
+        setattr(
+            self.task_to_compute,
             'requestor_public_key',
```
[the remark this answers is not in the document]
Jakub89: Fine, but that line have 121 characters right now :(

concent_api/common/tests/test_api_view.py
rwrzesien: There is a method in `TaskToCompute` which adds signature to `TaskToCompute`, so the nested messages inside it should be signed along. If not, keep the current code but report issue to golem-messages.

Jakub89: I don't think so, because `TaskToCompute` is sign by requestor, where `WantToComputeTask` must be sign by provider.

concent_api/core/models.py
Jakub89: As we spoke, validation of `want_to_compute` is redundant, because it's being done when we validate `task_to_compute`. I've try to implement it, and then I notice it can't be tested, because `ValidationError` is raised sooner.

concent_api/core/tests/utils.py
Jakub89: We don't do it like that in any other field in any other message I don't think so. We need to by able to create `WantToComputeTask` with custom keys. We had it like that in `TaskToCompute` creation, I just move it from one message to another because it moved in GM.

concent_api/common/tests/test_api_view.py
Jakub89: Done

concent_api/core/tests/test_core_views.py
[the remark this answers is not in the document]
Jakub89: Done

concent_api/core/tests/test_core_views.py
Jakub89: Done

PR #867, opened 2018-10-04
Add golem-messages 2.15.0 compatibility to release-notes.md

Jakub89:
Check if I should add something more from previous PR!


cameel:
Milestone should be added when you merge/close. Until then there should be no milestone.

Issue #870, opened 2018-10-08
Move reconnection_counter reset to "_authenticate" in Signing Service


dybi:
Nginx is keeping connection even when MiddleMan is down. This is causing `Signing Service` to reconnect, succeeds (reconnection counter is restarted),  then fail to authenticate (`MiddleMan` is down and is not sending authentication challenge within timeout limits), attempts reconnection, succeeds (reconnection counter is restarted), etc.

In the long run `Signing Service` attempts reconnection infinitely. To overcome this, we want to move `reconnection_counter` reset in "connect" part of the code after "_authenticate" part. It should be configurable from command line in the same way: `--max-reconnect-attempts`. Default value: `SIGNING_SERVICE_DEFAULT_RECONNECT_ATTEMPTS = 10`.

Counter should be reset after successful authentication.

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/161049287

PR #871, opened 2018-10-08
Move reconnection_counter reset after authentication in SigningService

Jakub89:
Resolve: #870


rwrzesien:
Looks good, maybe it would be possible to add test for this case?

PR #841, opened 2018-09-20
Create non-atomic views and separate transactions in core views

pawelkisielewicz:
Resolves https://github.com/golemfactory/concent/issues/782
It is updated and reopened https://github.com/golemfactory/concent/pull/834

PR #879, opened 2018-10-11
Bugfix processing timeouts in separate transaction

### LINKED ISSUE #782: Processing timeouts in a separate transaction in receive/
Currently `receive/` and `receive-out-of-band/` views do everything in a single transaction. First they use `update_timed_out_subtasks()` to ensure that all messages triggered by timeouts are already in the queue, then they pop a pending message from the queue, mark it as delivered and return it.

If the second phase fails, changes made in the first one are rolled back too. For example if a timed out task triggered a payment and which increased the nonce, failure later in the view invalidates those changes. Also, all the processed subtasks stay locked until the view ends which means that any parallel calls to `send/` will wait.

To fix this, we need to split the views into two parts, each running in a separate transaction.

Use [`@non_atomic_request`](https://docs.djangoproject.com/en/1.11/topics/db/transactions/#django.db.transaction.non_atomic_requests) on the view to disable the top-level transaction and inside use `@atomic` as a context manager instead. Then put an `@atomic` block over the part that delivers the message. It should not cover `update_timed_out_subtasks()` or any code that does not need to access the datatabase.

Modify `update_timed_out_subtasks()`. It should first find all subtasks that may need processing. Then process timeouts for each one in a separate transaction. We need a commit after each subtask so that a failure for one does not invalidate changes already made for another one.

**NOTE**: In the above I assume that #779 is already done. These tasks do not really depend on each other but doing #779 first will make this one slightly easier (both need a function for processing timeouts for a single subtask).

Also, currently `send/` also needs a change like this (putting timeouts and processing in separate transactions) but after #779 `send/` will only be processing timeouts for a single task so it won't be necessary. If #779 is never done, you should do the above for `send/` too.

pawelkisielewicz:
Resolves #782
It is updated and reopened #841


concent_api/common/tests/test_api_view.py
```diff
@@ -271,20 +303,32 @@ def test_api_view_should_rollback_changes_on_500_error(self):
                         PROVIDER_PRIVATE_KEY,
                         CONCENT_PUBLIC_KEY
                     ),
-                    content_type                        = 'application/octet-stream',
+                    content_type='application/octet-stream',
                 )
-            except TypeError:
+            except CustomException:
```
dybi: please change:
```python
try:
   # some code here
except CustomException:
   pass
```
to
```python
with suppres(CustomException):
   # some code here
```
throughout the code

concent_api/core/message_handlers.py
```diff
@@ -153,75 +154,74 @@ def handle_send_ack_report_computed_task(client_message: message.tasks.AckReport
...
-            queue               = PendingResponse.Queue.Receive,
-            subtask             = subtask,
-        )
-        logging.log_message_added_to_queue(
-            logger,
-            client_message,
-            requestor_public_key,
-        )
-        return HttpResponse("", status = 202)
+            new_report_computed_task = None
```
dybi: this part could be extracted to a function (i.e. `substitute_report_computed_task_if_needed`) and get some small comment describing why this is happening

concent_api/common/tests/test_api_view.py
rwrzesien: Why not `self.assertRaises(CustomException)` ?

concent_api/common/tests/test_api_view.py
```diff
@@ -254,15 +246,55 @@ def setUp(self):
...
+                    self.force_get_task_result,
+                    PROVIDER_PRIVATE_KEY,
+                    CONCENT_PUBLIC_KEY
+                ),
+                content_type='application/octet-stream',
+            )
+        _update_timed_out_subtask_correct_response_mock_function.assert_called()
+        _create_client_and_raise_http400_error_mock_function.assert_called()
+
+        self.assertEqual(Client.objects.count(), 1)
```
rwrzesien: There is a method `_assert_client_count_is_equal` in `ConcentIntegrationTestCase`.

concent_api/core/message_handlers.py
pawelkisielewicz: Done, hope you will like it

concent_api/common/tests/test_api_view.py
pawelkisielewicz: OK, but I would prefer to leave it as it is now. I think `self.assertEqual(Client.objects.count(), 1)` is short, easy to read and understand. Besides, this test extends `TransactionTestCase`, not `ConcentIntegrationTestCase`. I don't think this change will make code better or easier to understand

concent_api/core/message_handlers.py
```diff
@@ -193,14 +192,10 @@ def handle_send_ack_report_computed_task(client_message: message.tasks.AckReport
...
-            new_report_computed_task = None
-            try:
-                validate_all_messages_identical([
-                    report_computed_task,
-                    deserialize_message(subtask.report_computed_task.data.tobytes()),
-                ])
-            except ConcentValidationError:
-                new_report_computed_task = report_computed_task
+            new_report_computed_task = substitute_new_report_computed_task_if_needed(
+                report_computed_task=report_computed_task,
```
dybi: report_computed_task -> report_computed_task_from_acknowledgement
data (bytes) -> stored_report_computed_task (ReportComputedTask)

concent_api/core/validation.py
```diff
@@ -410,3 +413,20 @@ def validate_database_report_computed_task(
                 'Nested ReportComputedTask message must be the same as ReportComputedTask stored separately'
             )
         })
+
+
+def substitute_new_report_computed_task_if_needed(report_computed_task: ReportComputedTask, data: bytes) -> Optional[ReportComputedTask]:
+    """
+    If previous sent by requestor report_computed_task does not match report_computed_task sent by provider it
```
dybi: Apart from mixing Requestor with Provider ;), it needs rephrasing. I would suggest st like this
```python
"""
If stored ReportComputedTask (previously sent by Provider in ForceReportComputedTask message) is different than ReportComputedTask sent by Requestor (in AckReportComputedTask), it means that Provider had sent different  ReportComputedTask to Requestor and the message stored in the DB must be replaced.
"""
```

@rwrzesien. @kbeker, @Jakub89 what do you think guys?

concent_api/core/message_handlers.py
pawelkisielewicz: done except of changing type of `stored_report_computed_task`. Function `deserialize_message` expects `bytes`, not `ReportComputedTask`

pawelkisielewicz: deserialize moved to function called. It is passed two `ReportComputedTask` to function `substitute_new_report_computed_task_if_needed`

concent_api/common/tests/test_api_view.py
```diff
@@ -321,16 +363,14 @@ def test_non_api_view_should_rollback_changes_on_500_error(self):
 
         with mock.patch(
             'gatekeeper.views.gatekeeper_access_denied_response',
-            side_effect=gatekeeper_access_denied_response_500_mock
+            side_effect=_create_client_and_raise_http500_exception_mock
         ) as gatekeeper_access_denied_response_500_mock_function:
-            try:
+            with suppress(CustomException):
```
dybi: :+1:

concent_api/core/transfer_operations.py
```diff
@@ -28,6 +28,7 @@
 def verify_file_status(subtask: Subtask, client_public_key: bytes) -> None:
     """
     Function to verify existence of a file on cluster storage by checking `result_upload_finished` flag on Subtask.
+    It is checked only for requestors (client is requestor) and subtask's state is FORCING_RESULT_TRANSFER
```
dybi: Please change to:
```python
"""
 It is checked only when client is a requestor and subtask's state is FORCING_RESULT_TRANSFER.
"""
```

PR #880, opened 2018-10-12
Feature use one retry of message trying to store subtask

pawelkisielewicz:
Resolves https://github.com/golemfactory/concent/issues/759
Please see also https://github.com/golemfactory/concent/pull/879

### DIFF
--- a/concent_api/api-e2e-multi-threads-test.py
+++ b/concent_api/api-e2e-multi-threads-test.py
@@ -1,4 +1,4 @@
-# !/usr/bin/env python3
+#!/usr/bin/env python3
 import os
 import sys
 import time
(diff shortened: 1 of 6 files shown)


concent_api/api-e2e-multi-threads-test.py
```diff
@@ -171,6 +171,36 @@ def test_case_multiple_requests_concerning_one_subtask_will_be_processed_one_by_
...
+    )
+    # waiting for responses from all threads
+    end_time = time.time() + MAXIMUM_WAITING_TIME_FOR_ALL_RESPONSES
+    while len(responses_global) != NUMBER_OF_TESTING_THREADS:
+        time.sleep(0.1)
+        if time.time() >= end_time:
+            break
+
+    expected_responses = ['AckForceGetTaskResult', 'ServiceRefused', 'ServiceRefused']
+    assert len(expected_responses) == NUMBER_OF_TESTING_THREADS, 'Did you changed number of testing threads and forgot to change expected_responses?'
```
dybi: 'Did you changed' -> 'Did you **change**', 
'and forgot' -> 'and **forget**'

concent_api/core/message_handlers.py
```diff
@@ -58,8 +60,7 @@
 from core.utils import calculate_additional_verification_call_time
 from core.utils import calculate_maximum_download_time
 from core.utils import calculate_subtask_verification_time
-from core.validation import is_golem_message_signed_with_key
-from core.validation import substitute_new_report_computed_task_if_needed
+from core.validation import is_golem_message_signed_with_key, substitute_new_report_computed_task_if_needed
```
dybi: format

concent_api/core/message_handlers.py
```diff
@@ -918,51 +919,55 @@ def store_subtask(
...
+            requestor_public_key=requestor_public_key,
+            computation_deadline=computation_deadline,
+            result_package_size=result_package_size,
+            next_deadline=next_deadline,
+        )
 
-    return subtask
+        return subtask
+    except IntegrityError:
+        logger.info(f'IntegrityError when tried to store subtask with id {subtask_id}. Task_id: {task_id}. '
```
dybi: format

concent_api/core/transfer_operations.py
```diff
@@ -62,27 +64,30 @@ def store_pending_message(
...
-    if payment_message is not None:
-        payment_committed_message = PaymentInfo(
-            payment_ts=parse_timestamp_to_utc_datetime(payment_message.payment_ts),
-            task_owner_key=payment_message.task_owner_key,
-            provider_eth_account=payment_message.provider_eth_account,
-            amount_paid=payment_message.amount_paid,
-            recipient_type=payment_message.recipient_type.name,  # pylint: disable=no-member
-            amount_pending=payment_message.amount_pending,
-            pending_response=receive_queue
+    try:
```
dybi: is it really needed here? I thought that creating `Subtask` might be a problem...

concent_api/common/decorators.py
```diff
@@ -183,6 +184,10 @@ def wrapper(
                 response_from_view = view(request, client_message, client_public_key, *args, **kwargs)
                 if database_name is not None:
                     transaction.savepoint_commit(sid, using=database_name)
+
+            except CreateModelIntegrityError as exception:
+                logger.info(f'CreateModelIntegrityError occurred. View will be retried. Exception: {exception}.')
+                response_from_view = view(request, client_message, client_public_key, *args, **kwargs)
```
dybi: What will happen if `CreateModelIntegrityError` occurs a second time?

concent_api/core/transfer_operations.py
pawelkisielewicz: The task is not only for subtasks, @cameel wrote about other objects as well. I checked the code and generally other (not subtask) objects that may fail with IntegrityError are in this place.

concent_api/common/decorators.py
pawelkisielewicz: This means something went wrong. It can occur only one time, when two or more processes are trying to create the same object. If it happens, one process creates this object, all other fail with `CreateModelIntegrityError`, are restarted, validated one more time and they should't try to create object again (this time it already exists in DB). If `CreateModelIntegrityError` occurs a second time it means that we have bug in our code..

Issue #849, opened 2018-09-25
Chore update GSCI to v1.6.0


kbeker:
Please update Concent's requirements GSCI to newest version, where we can add contract's address.

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/160753052

PR #886, opened 2018-10-15
Chore update golem smart contract interface to v1.6.0

Jakub89:
Resolves: #849

### DIFF
--- a/concent_api/concent_api/settings/development.py
+++ b/concent_api/concent_api/settings/development.py
@@ -48,3 +48,5 @@ MIDDLEMAN_ADDRESS = 'localhost'
 MIDDLEMAN_PORT = DEFAULT_INTERNAL_PORT
 
 SIGNING_SERVICE_PUBLIC_KEY = b'\x19\xaew&P>\xce\xd7D\xfb\xbff)55\xc1%\xea\xebN\x0e\x05>\xf7v\x0c\x94I\xe4\xa3\x80\x1dl\xed\xb5j\xe9apZ\x8c\x86\xe2\xb2x^M}>8c\x90\x0e\xa2\xf4\x8cOb\xb1\x1er\x11\xd7\xee'
+
+GNT_DEPOSIT_CONTRACT_ADDRESS = '0xcfB81A6EE3ae6aD4Ac59ddD21fB4589055c13DaD'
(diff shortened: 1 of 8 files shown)


concent_api/concent_api/settings/development.py
```diff
@@ -48,3 +48,5 @@
 MIDDLEMAN_PORT = DEFAULT_INTERNAL_PORT
 
 SIGNING_SERVICE_PUBLIC_KEY = b'\x19\xaew&P>\xce\xd7D\xfb\xbff)55\xc1%\xea\xebN\x0e\x05>\xf7v\x0c\x94I\xe4\xa3\x80\x1dl\xed\xb5j\xe9apZ\x8c\x86\xe2\xb2x^M}>8c\x90\x0e\xa2\xf4\x8cOb\xb1\x1er\x11\xd7\xee'
+
+GNT_DEPOSIT = '0xcfB81A6EE3ae6aD4Ac59ddD21fB4589055c13DaD'
```
rwrzesien: I believe this is some fake value ? :)

concent_api/concent_api/system_check.py
```diff
@@ -719,3 +735,18 @@ def check_use_signing_service(app_configs: None=None, **kwargs: Any) -> list:  #
...
+def check_gntdeposit_adress(app_configs: None=None, **kwargs: Any) -> list:  # pylint: disable=unused-argument
+    if (
+        hasattr(settings, 'PAYMENT_BACKEND') and
+        settings.PAYMENT_BACKEND == 'core.payments.backends.sci_backend'
+    ):
+        if not hasattr(settings, 'GNT_DEPOSIT'):
+            return [create_error_56_gntdeposit_not_set()]
+
+        if not isinstance(settings.GNT_DEPOSIT, str):
+            return [create_error_57_gntdeposit_has_wrong_type(settings.GNT_DEPOSIT)]
```
pawelkisielewicz: Does this address have e.g permanent length which can be checked? Or something more that is important..?

concent_api/concent_api/settings/development.py
Jakub89: No, it's our contract address that we use on `dev` cluster. It's not some secret value like private key that someone can use to take advantage of.

concent_api/concent_api/system_check.py
Jakub89: Actually yes. It's like every address on ethereum blockchain, it starts with `0x` and should have length of 42 characters(I need to double check that though).

Jakub89: Ok, I've added value check after type check.

PR #891, opened 2018-10-15
Bugfix signig service doesnt detect when connection is dropped

dybi:
workaround for ngingx problem. AFAIK, no issue for that.

@bartoszbetka ? @PaweuB ?


dybi:
Sorry, wrong branch

Issue #878, opened 2018-10-11
Extend Signing Service with transaction sum storage


dybi:
This is a part of https://github.com/golemfactory/concent/issues/872

`Signing Service` should store its daily transaction amount in case it is restarted. 

The file should be called  `YYYY-mm-dd`, where `YYYY` stands for year, `mm` for month and `dd` for day, be in a text format and contain single number which would be the sum of  values of daily transactions made so far.

1. Upon creation, `Signing Service` should get current date and check whether there is a corresponding file.
1.1. If yes, it should try to read a single value out of it and set `self.daily_costs_so_far` accordingly
1.2 If no, or if reading from the file fails, `self.daily_costs_so_far` should be set to `0`. In case of failing to read the file, log the warning.
2. When `Signing Service` is asked to sign the transaction, it should check if the operation doesn't exceed daily limit (see: https://github.com/golemfactory/concent/issues/873)
3. Upon successfully signing the transaction, override the file contents.

Note: The files are not deleted - new day means a new file being created.
Note 2: The mechanism is very simple and when need arises could be transformed into DB storage.

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/161178998

Issue #873, opened 2018-10-08
Extend Signing Service with daily threshold checks


dybi:
This is part of https://github.com/golemfactory/concent/issues/872. 
Extend `Signing Service` with two checks:
1. If the sum of transaction exceeds `WARNING_DAILY_THRESHOLD` (1000 GNTB) - then issue a warning. Prepare a stub of utilty class `Notifier`, which will be responsible for notifications from `Signing Service`
2. If the sum of transactions so far + current transaction's cost exceeds `MAXIMUM_DAILY_THRESHOLD` (10000 GNTB) - then issue a warning (again, using `Notifier` stub) and stop signing transactions. Respond with `TransactionRejected` with reason: `DailyLimitExceeded` (NOTE: **the reason must be added**)
```
if transaction_to_be_signed.cost + daily_costs_so_far > MAXIMUM_DAILY_THRESHOLD:
     refuse_transaction()
```

Make sure that counter is reset daily (i.e. on midnight)

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/161054230

PR #900, opened 2018-10-17
Feature extend signing service with daily threshold checks

Jakub89:
Resolves: #878 
Resolves: #873

### DIFF
--- a/signing_service/signing_service/signing_service.py
+++ b/signing_service/signing_service/signing_service.py
@@ -54,6 +54,7 @@ from signing_service.exceptions import SigningServiceValidationError
 from signing_service.utils import is_private_key_valid
 from signing_service.utils import is_public_key_valid
 from signing_service.utils import make_secret_provider_factory
+from signing_service.utils import Notifier
 
 logger = logging.getLogger()
 crash_logger = logging.getLogger('crash')
@@ -78,6 +79,7 @@ class SigningService:
         'ethereum_private_key',
         'reconnection_counter',
         'maximum_reconnection_attempts',
+        'notifier',
         'signing_service_daily_transaction_sum_so_far',
         'daily_transactions_limit_file_name',
     )
@@ -91,6 +93,7 @@ class SigningService:
         signing_service_private_key: bytes,
         ethereum_private_key: str,
         maximum_reconnect_attempts: int,
+        notifier=Notifier(),
     ) -> None:
         assert isinstance(host, str)
         assert isinstance(port, int)
@@ -109,6 +112,7 @@ class SigningService:
         self.was_sigterm_caught: bool = False
         self.reconnection_counter = 0
         self.maximum_reconnection_attempts = maximum_reconnect_attempts
+        self.notifier = notifier
         self.signing_service_daily_transaction_sum_so_far = 0
         self.daily_transactions_limit_file_name = datetime.datetime.now().strftime('%Y-%m-%d')
 
@@ -281,20 +285,20 @@ class SigningService:
                 self._update_daily_transactions_limit_file_name()
                 golem_message_response = self._get_signed_transaction(middleman_message.payload)
                 if isinstance(golem_message_response, SignedTransaction):
-                    email_notifier = Notifier()
                     transaction_sum_combined = self.signing_service_daily_transaction_sum_so_far + middleman_message.payload.value
                     if transaction_sum_combined > MAXIMUM_DAILY_THRESHOLD:
                         logger.warning(
                             f'Signing Service is unable to transact more then {MAXIMUM_DAILY_THRESHOLD} GNTB today.'
                             f'Transaction from {middleman_message.payload.from_address} rejected.'
                         )
+                        self.notifier.send()
                         golem_message_response = TransactionRejected(
                             reason=TransactionRejected.REASON.DailyLimitExceeded,
                             nonce=middleman_message.payload.nonce,
                         )
                     elif transaction_sum_combined > WARNING_DAILY_THRESHOLD:
-                        email_notifier.send()
                         logger.warning(f'Signing Service has signed transactions worth {transaction_sum_combined} GNTB today.')
+                        self.notifier.send()
                         self._add_payload_value_to_daily_transactions_sum(transaction_sum_combined)
                     else:
                         self._add_payload_value_to_daily_transactions_sum(transaction_sum_combined)
--- a/signing_service/signing_service/signing_service.py
+++ b/signing_service/signing_service/signing_service.py
@@ -1,9 +1,12 @@
 import argparse
+import datetime
 import logging.config
 import os
 import signal
 import socket
 from contextlib import closing
+from pathlib import Path
+from pathlib import PosixPath
 from time import sleep
 from types import FrameType
 from typing import Iterator
@@ -38,11 +41,13 @@ from middleman_protocol.stream import send_over_stream
 from middleman_protocol.stream import unescape_stream
 
 from signing_service.constants import CONNECTION_TIMEOUT
+from signing_service.constants import MAXIMUM_DAILY_THRESHOLD
 from signing_service.constants import RECEIVE_AUTHENTICATION_CHALLENGE_TIMEOUT
 from signing_service.constants import SIGNING_SERVICE_DEFAULT_INITIAL_RECONNECT_DELAY
 from signing_service.constants import SIGNING_SERVICE_DEFAULT_PORT
 from signing_service.constants import SIGNING_SERVICE_DEFAULT_RECONNECT_ATTEMPTS
 from signing_service.constants import SIGNING_SERVICE_MAXIMUM_RECONNECT_TIME
+from signing_service.constants import WARNING_DAILY_THRESHOLD
 from signing_service.exceptions import SigningServiceMaximumReconnectionAttemptsExceeded
 from signing_service.exceptions import SigningServiceUnexpectedMessageError
 from signing_service.exceptions import SigningServiceValidationError
@@ -73,6 +78,8 @@ class SigningService:
         'ethereum_private_key',
         'reconnection_counter',
         'maximum_reconnection_attempts',
+        'signing_service_daily_transaction_sum_so_far',
+        'daily_transactions_limit_file_name',
     )
 
     def __init__(
@@ -102,6 +109,8 @@ class SigningService:
         self.was_sigterm_caught: bool = False
         self.reconnection_counter = 0
         self.maximum_reconnection_attempts = maximum_reconnect_attempts
+        self.signing_service_daily_transaction_sum_so_far = 0
+        self.daily_transactions_limit_file_name = datetime.datetime.now().strftime('%Y-%m-%d')
 
         self._validate_arguments()
 
@@ -165,6 +174,7 @@ class SigningService:
         # Reset delay and reconnection counter on successful authentication and set flag that connection is established.
         self.current_reconnect_delay = None
         self.reconnection_counter = 0
+        self.signing_service_daily_transaction_sum_so_far = self._get_signing_service_daily_transaction_sum_so_far()
         self._handle_connection(receive_frame_generator, tcp_socket)
 
     def _authenticate(
@@ -268,7 +278,26 @@ class SigningService:
                 middleman_message_response = self._prepare_error_response(ErrorCode.UnexpectedMessage, exception)
             # If received frame is correct, validate transaction.
             else:
+                self._update_daily_transactions_limit_file_name()
                 golem_message_response = self._get_signed_transaction(middleman_message.payload)
+                if isinstance(golem_message_response, SignedTransaction):
+                    email_notifier = Notifier()
+                    transaction_sum_combined = self.signing_service_daily_transaction_sum_so_far + middleman_message.payload.value
+                    if transaction_sum_combined > MAXIMUM_DAILY_THRESHOLD:
+                        logger.warning(
+                            f'Signing Service is unable to transact more then {MAXIMUM_DAILY_THRESHOLD} GNTB today.'
+                            f'Transaction from {middleman_message.payload.from_address} rejected.'
+                        )
+                        golem_message_response = TransactionRejected(
+                            reason=TransactionRejected.REASON.DailyLimitExceeded,
+                            nonce=middleman_message.payload.nonce,
+                        )
+                    elif transaction_sum_combined > WARNING_DAILY_THRESHOLD:
+                        email_notifier.send()
+                        logger.warning(f'Signing Service has signed transactions worth {transaction_sum_combined} GNTB today.')
+                        self._add_payload_value_to_daily_transactions_sum(transaction_sum_combined)
+                    else:
+                        self._add_payload_value_to_daily_transactions_sum(transaction_sum_combined)
                 golem_message_response.sign_message(private_key=self.signing_service_private_key)
                 middleman_message_response = GolemMessageFrame(
                     payload=golem_message_response,
@@ -347,6 +376,7 @@ class SigningService:
             # If not, rejection reason is InvalidTransaction
             return TransactionRejected(
                 reason=TransactionRejected.REASON.InvalidTransaction,
+                nonce=transaction_signing_request.nonce,
             )
 
         # If transaction is correct, sign it.
@@ -357,6 +387,7 @@ class SigningService:
             # If not, rejection reason is UnauthorizedAccount.
             return TransactionRejected(
                 reason=TransactionRejected.REASON.UnauthorizedAccount,
+                nonce=transaction_signing_request.nonce,
             )
 
         assert transaction.v is not None
@@ -384,6 +415,35 @@ class SigningService:
             self.signing_service_private_key,
             authentication_challenge,
         )
+
+    @staticmethod
+    def _get_daily_transaction_threshold_file_path() -> PosixPath:
+        thresholds_directory = (Path.cwd()).joinpath('daily_tresholds')  # pylint: disable=no-member
+        thresholds_directory.mkdir(exist_ok=True)  # pylint: disable=no-member
+        daily_file = datetime.datetime.now().strftime('%Y-%m-%d')
+        thresholds_directory.joinpath(daily_file).touch(exist_ok=True)  # pylint: disable=no-member
+        tresholds_file_path = thresholds_directory.joinpath(daily_file)
+        return tresholds_file_path
+
+    def _get_signing_service_daily_transaction_sum_so_far(self) -> int:
+        threshold_file = self._get_daily_transaction_threshold_file_path()
+        try:
+            transaction_sum = int(threshold_file.read_text())  # pylint: disable=no-member
+        except ValueError:
+            threshold_file.write_text('0')  # pylint: disable=no-member
+            return 0
+        return transaction_sum
+
+    def _add_payload_value_to_daily_transactions_sum(self, transactions_sum: int) -> None:
+        tresholds_file_path = self._get_daily_transaction_threshold_file_path()
+        tresholds_file_path.write_text(str(transactions_sum))  # pylint: disable=no-member
+
+    def _update_daily_transactions_limit_file_name(self) -> None:
+        """ Create new file with daily transaction sum if last file is outdated. """
+        current_day_string = datetime.datetime.now().strftime('%Y-%m-%d')
+        if self.daily_transactions_limit_file_name != current_day_string:
+            self.daily_transactions_limit_file_name = current_day_string
+            self.signing_service_daily_transaction_sum_so_far = self._get_signing_service_daily_transaction_sum_so_far()
 
 
 def _parse_arguments() -> argparse.Namespace:
--- a/.gitignore
+++ b/.gitignore
@@ -115,3 +115,6 @@ ENV/
 
 # PyCharm
 /.idea
+
+# Signing service daily tresholds
+daily_tresholds/
--- a/middleman_protocol/middleman_protocol/concent_golem_messages/message.py
+++ b/middleman_protocol/middleman_protocol/concent_golem_messages/message.py
@@ -152,6 +152,8 @@ class TransactionRejected(NonceAbstractMessage):
         # The service is not authorized to transfer funds from the account
         # specified in the transaction.
         UnauthorizedAccount = 'unauthorized_account'
+        # The message itself is valid but daily limit of GNTB that went through Signing Service has been exceeded
+        DailyLimitExceeded = 'daily_limit_exceeded'
 
     ENUM_SLOTS = {
         'reason': REASON,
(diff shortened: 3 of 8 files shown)


signing_service/signing_service/signing_service.py
```diff
@@ -268,6 +273,25 @@ def _handle_connection(
             # If received frame is correct, validate transaction.
             else:
                 golem_message_response = self._get_signed_transaction(middleman_message.payload)
+                if isinstance(golem_message_response, SignedTransaction):
+                    transaction_sum_so_far = self._get_signing_service_daily_transaction_sum_so_far()
+                    transaction_sum_combined = transaction_sum_so_far + middleman_message.payload.value
+                    email_notifier = Notifier()
```
rwrzesien: How about creating instance of `Notifier` in `__init__` ? This way it will be done only once.

signing_service/signing_service/signing_service.py
```diff
@@ -268,6 +273,25 @@ def _handle_connection(
             # If received frame is correct, validate transaction.
             else:
                 golem_message_response = self._get_signed_transaction(middleman_message.payload)
+                if isinstance(golem_message_response, SignedTransaction):
+                    transaction_sum_so_far = self._get_signing_service_daily_transaction_sum_so_far()
```
rwrzesien: How about changing it to work this way:
1. You read value only once, on `__init__` (and maybe when connection is dropped and SS is trying to reconnect too?).
2. You keep the value in class attribute like `self.signing_service_daily_transaction_sum_so_far`.
3. You only update this value and write to file.

This way we will decrease IO operations by half, and there should not be much risk of those two values to become different.

@dybi What do you think?

signing_service/signing_service/signing_service.py
```diff
@@ -384,6 +410,28 @@ def _get_authentication_challenge_signature(self, authentication_challenge: byte
...
+        daily_file = datetime.datetime.now().strftime('%Y-%m-%d')
+        thresholds_directory.joinpath(daily_file).touch(exist_ok=True)  # pylint: disable=no-member
+        thresholds_directory = thresholds_directory.joinpath(daily_file)
+        if thresholds_directory.read_text() == '':  # pylint: disable=no-member
+            thresholds_directory.write_text('0')  # pylint: disable=no-member
+            return 0
+        try:
+            transaction_sum = int(thresholds_directory.read_text())  # pylint: disable=no-member
+        except ValueError:
+            thresholds_directory.write_text('0')  # pylint: disable=no-member
```
rwrzesien: I think in this case you should also erase current file content (the one that caused `ValueError`).

signing_service/signing_service/signing_service.py
```diff
@@ -384,6 +410,28 @@ def _get_authentication_challenge_signature(self, authentication_challenge: byte
             authentication_challenge,
         )
 
+    @staticmethod
+    def _get_signing_service_daily_transaction_sum_so_far() -> int:
+        thresholds_directory = (Path.cwd()).joinpath('daily_tresholds')  # pylint: disable=no-member
```
rwrzesien: This path should be constant (or configurable with default?). What do you guys think @dybi @bartoszbetka

dybi: I would do this as simple as it can get:
```python
directory_path = os.path.join('.', 'daily_tresholds')
daily_file = os.path.join(directory_path, datetime.datetime.now().strftime('%Y-%m-%d'))
```

signing_service/signing_service/signing_service.py
```diff
@@ -384,6 +410,28 @@ def _get_authentication_challenge_signature(self, authentication_challenge: byte
...
+        thresholds_directory.mkdir(exist_ok=True)  # pylint: disable=no-member
+        daily_file = datetime.datetime.now().strftime('%Y-%m-%d')
+        thresholds_directory.joinpath(daily_file).touch(exist_ok=True)  # pylint: disable=no-member
+        thresholds_directory = thresholds_directory.joinpath(daily_file)
+        if thresholds_directory.read_text() == '':  # pylint: disable=no-member
+            thresholds_directory.write_text('0')  # pylint: disable=no-member
+            return 0
+        try:
```
dybi: Quite frankly, I would simplify this:
Just try to read from: `daily_file` and convert it to `int`. Catch any exception from the combined operation and set `self.daily_transactions_sum = 0`

signing_service/signing_service/signing_service.py
```diff
@@ -384,6 +410,28 @@ def _get_authentication_challenge_signature(self, authentication_challenge: byte
...
+            transaction_sum = int(thresholds_directory.read_text())  # pylint: disable=no-member
+        except ValueError:
+            thresholds_directory.write_text('0')  # pylint: disable=no-member
+            return 0
+        return transaction_sum
+
+    @staticmethod
+    def _add_payload_value_to_daily_transactions_sum(transactions_sum: int) -> None:
+        thresholds_directory = (Path.cwd()).joinpath('daily_tresholds').joinpath(datetime.datetime.now().strftime('%Y-%m-%d'))
+        thresholds_directory.write_text(str(transactions_sum))
```
dybi: you are writing to a directory. I assume, we would like to create a file first ;)

Jakub89: It is created in `_get_signing_service_daily_transaction_sum_so_far()`, that we always call before. I didn't want to copy same code. Or should I?

signing_service/signing_service/signing_service.py
Jakub89: If I we `os.path` instead of `pathlib.Path` to do all that, we are actually overcomplicating that function. I will have to write things like:
```
try:
    os.makedirs(directory)
except OSError as e:
    ...
```
when creating directory and something similar when creating file.
`pathlib.Path` makes all of that for us.

signing_service/signing_service/signing_service.py
Jakub89: I've just tested `.write_text()`, and it does erase the whole file content. I don't know if any special erasing is needed?

rwrzesien: If it is working like that than its OK. I have took a quick look into documentation of this method but couldn't find this info, thanks.

signing_service/signing_service/signing_service.py

signing_service/signing_service/signing_service.py

signing_service/signing_service/signing_service.py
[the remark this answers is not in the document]

signing_service/signing_service/signing_service.py
```diff
@@ -268,6 +273,25 @@ def _handle_connection(
...
+                        logger.warning(
+                            f'Signing Service is unable to transact more then {MAXIMUM_DAILY_THRESHOLD} GNTB today.'
+                        )
+                        email_notifier.send()
+                        golem_message_response = TransactionRejected(
+                            reason=TransactionRejected.REASON.DailyLimitExceeded,
+                            nonce=middleman_message.payload.nonce,
+                        )
+                    elif transaction_sum_combined > WARNING_DAILY_THRESHOLD:
+                        logger.warning(f'Signing Service transact {transaction_sum_combined} GNTB today so far.')
```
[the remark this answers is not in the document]

signing_service/signing_service/signing_service.py

signing_service/signing_service/signing_service.py
Jakub89: Ok, I made separate function `_get_daily_transaction_threshold_file_directory()` and I'm using it in both `_get_signing_service_daily_transaction_sum_so_far` and `_add_payload_value_to_daily_transactions_sum`

signing_service/signing_service/signing_service.py
```diff
@@ -384,6 +414,28 @@ def _get_authentication_challenge_signature(self, authentication_challenge: byte
             authentication_challenge,
         )
 
+    @staticmethod
+    def _get_daily_transaction_threshold_file_directory() -> PosixPath:
```
pawelkisielewicz: In this method you return the path with filename. I think it should be renamed (e.g. `_get_daily_transaction_threshold_file`). There is line in `_get_signing_service_daily_transaction_sum_so_far`: `transaction_sum = int(thresholds_directory.read_text()) ` - it suggests you are reading from directory..

BTW. :+1: for general conception of this method. Keeping actual amount in file named as date :)

signing_service/signing_service/signing_service.py
```diff
@@ -268,6 +279,23 @@ def _handle_connection(
             # If received frame is correct, validate transaction.
             else:
                 golem_message_response = self._get_signed_transaction(middleman_message.payload)
+                if isinstance(golem_message_response, SignedTransaction):
+                    transaction_sum_combined = self.signing_service_daily_transaction_sum_so_far + middleman_message.payload.value
+                    if transaction_sum_combined > MAXIMUM_DAILY_THRESHOLD:
+                        logger.warning(
+                            f'Signing Service is unable to transact more then {MAXIMUM_DAILY_THRESHOLD} GNTB today.'
```
pawelkisielewicz: Consider add client_key to identify which transaction was rejected

signing_service/signing_service/signing_service.py
```diff
@@ -268,6 +279,23 @@ def _handle_connection(
...
+                        )
+                        self.notifier.send()
+                        golem_message_response = TransactionRejected(
+                            reason=TransactionRejected.REASON.DailyLimitExceeded,
+                            nonce=middleman_message.payload.nonce,
+                        )
+                    elif transaction_sum_combined > WARNING_DAILY_THRESHOLD:
+                        logger.warning(f'Signing Service has signed transactions worth {transaction_sum_combined} GNTB today.')
```
pawelkisielewicz: same as above

signing_service/signing_service/signing_service.py

signing_service/signing_service/signing_service.py

signing_service/signing_service/signing_service.py
Jakub89: Above case send reject, where this one pass, only with warning. So as we discuss, I'm leaving it as it is.

signing_service/signing_service/signing_service.py
```diff
@@ -102,6 +112,9 @@ def __init__(
         self.was_sigterm_caught: bool = False
         self.reconnection_counter = 0
         self.maximum_reconnection_attempts = maximum_reconnect_attempts
+        self.notifier = notifier
+        self.signing_service_daily_transaction_sum_so_far = self._get_signing_service_daily_transaction_sum_so_far()
+        self.last_signed_transaction = datetime.datetime.now().strftime('%Y-%m-%d')
```
rwrzesien: This name is misleading, it is suggesting that you will have a `ethereum.Tranaction` object here, not a date.

signing_service/signing_service/signing_service.py
```diff
@@ -267,7 +281,26 @@ def _handle_connection(
                 middleman_message_response = self._prepare_error_response(ErrorCode.UnexpectedMessage, exception)
             # If received frame is correct, validate transaction.
             else:
+                self._update_last_signed_transaction_date(self.last_signed_transaction)
```
rwrzesien: There is no need to pass `self.last_signed_transaction` here, you can use it directly in this method.

signing_service/signing_service/signing_service.py
```diff
@@ -384,6 +419,33 @@ def _get_authentication_challenge_signature(self, authentication_challenge: byte
...
+        except ValueError:
+            threshold_file.write_text('0')  # pylint: disable=no-member
+            return 0
+        return transaction_sum
+
+    def _add_payload_value_to_daily_transactions_sum(self, transactions_sum: int) -> None:
+        thresholds_directory = self._get_daily_transaction_threshold_file_path()
+        thresholds_directory.write_text(str(transactions_sum))  # pylint: disable=no-member
+
+    def _update_last_signed_transaction_date(self, last_transaction_date: str) -> None:
```
rwrzesien: This function name describes something else that it is actually doing. The function is not updating the date but loading the value when date changes? Maybe apart from fixing the name there should be some comment here.

signing_service/signing_service/signing_service.py
rwrzesien: What I have meant is maybe we should make this path configurable in case if deployment would like to keep it elsewhere on the file system. @bartoszbetka Do you have some thoughts about this?

signing_service/signing_service/signing_service.py
```diff
@@ -102,6 +112,9 @@ def __init__(
         self.was_sigterm_caught: bool = False
         self.reconnection_counter = 0
         self.maximum_reconnection_attempts = maximum_reconnect_attempts
+        self.notifier = notifier
+        self.signing_service_daily_transaction_sum_so_far = self._get_signing_service_daily_transaction_sum_so_far()
```
rwrzesien: If it is in `_connect()` there is no need to call it here.

signing_service/signing_service/signing_service.py
```diff
@@ -267,7 +281,26 @@ def _handle_connection(
                 middleman_message_response = self._prepare_error_response(ErrorCode.UnexpectedMessage, exception)
             # If received frame is correct, validate transaction.
             else:
+                self._update_last_signed_transaction_date(self.last_signed_transaction)
                 golem_message_response = self._get_signed_transaction(middleman_message.payload)
+                if isinstance(golem_message_response, SignedTransaction):
+                    transaction_sum_combined = self.signing_service_daily_transaction_sum_so_far + middleman_message.payload.value
+                    if transaction_sum_combined > MAXIMUM_DAILY_THRESHOLD:
```
rwrzesien: Please reverse conditions here and below.

signing_service/signing_service/signing_service.py
Jakub89: @dybi comment:
> Also, since we want to move file reading to __init__ (separate function, i.e load_transactions_sum that is called in __init__), ...

It was `0` in `__init__` at first, but then I had to change it.

Jakub89: @dybi comment:
> Also, since we want to move file reading to __init__ (separate function, i.e load_transactions_sum that is called in __init__), ...

It was `0` in `__init__` at first, but then I had to change it.

Jakub89: @dybi comment:
> Also, since we want to move file reading to __init__ (separate function, i.e load_transactions_sum that is called in __init__), ...

It was `0` in `__init__` at first, but then I had to change it.

Jakub89: @dybi comment:
> Also, since we want to move file reading to __init__ (separate function, i.e load_transactions_sum that is called in __init__), ...

It was `0` in `__init__` at first, but then I had to change it.

Jakub89: @dybi comment:
> Also, since we want to move file reading to __init__ (separate function, i.e load_transactions_sum that is called in __init__), ...

It was `0` in `__init__` at first, but then I had to change it.

Jakub89: @dybi comment:
> Also, since we want to move file reading to __init__ (separate function, i.e load_transactions_sum that is called in __init__), ...

It was `0` in `__init__` at first, but then I had to change it.

signing_service/signing_service/signing_service.py

signing_service/signing_service/signing_service.py

signing_service/signing_service/signing_service.py

signing_service/signing_service/signing_service.py

signing_service/signing_service/signing_service.py
Jakub89: Done, back to `0`.

signing_service/signing_service/signing_service.py
```diff
@@ -295,7 +295,7 @@ def _handle_connection(
                             reason=TransactionRejected.REASON.DailyLimitExceeded,
                             nonce=middleman_message.payload.nonce,
                         )
-                    elif transaction_sum_combined > WARNING_DAILY_THRESHOLD:
+                    elif WARNING_DAILY_THRESHOLD < transaction_sum_combined:
```
dybi: same here:
`elif transaction_sum_combined > WARNING_DAILY_THRESHOLD:`

signing_service/signing_service/signing_service.py
```diff
@@ -81,7 +81,7 @@ class SigningService:
         'maximum_reconnection_attempts',
         'notifier',
         'signing_service_daily_transaction_sum_so_far',
-        'last_signed_transaction',
+        'last_transactions_limit_file',
```
dybi: why was it renamed? last name was better

signing_service/signing_service/signing_service.py
```diff
@@ -384,6 +419,34 @@ def _get_authentication_challenge_signature(self, authentication_challenge: byte
             authentication_challenge,
         )
 
+    @staticmethod
+    def _get_daily_transaction_threshold_file_path() -> PosixPath:
+        thresholds_directory = (Path.cwd()).joinpath('daily_tresholds')  # pylint: disable=no-member
+        thresholds_directory.mkdir(exist_ok=True)  # pylint: disable=no-member
+        daily_file = datetime.datetime.now().strftime('%Y-%m-%d')
+        thresholds_directory.joinpath(daily_file).touch(exist_ok=True)  # pylint: disable=no-member
+        thresholds_directory = thresholds_directory.joinpath(daily_file)
```
dybi: you are getting a `tresholds_file_path` - why overriding `tresholds_directory`?

signing_service/signing_service/signing_service.py
```diff
@@ -384,6 +419,34 @@ def _get_authentication_challenge_signature(self, authentication_challenge: byte
...
+        return thresholds_directory
+
+    def _get_signing_service_daily_transaction_sum_so_far(self) -> int:
+        threshold_file = self._get_daily_transaction_threshold_file_path()
+        try:
+            transaction_sum = int(threshold_file.read_text())  # pylint: disable=no-member
+        except ValueError:
+            threshold_file.write_text('0')  # pylint: disable=no-member
```
dybi: don't you need to `open` / `close` the file?

signing_service/signing_service/signing_service.py
```diff
@@ -384,6 +419,34 @@ def _get_authentication_challenge_signature(self, authentication_challenge: byte
...
+            return 0
+        return transaction_sum
+
+    def _add_payload_value_to_daily_transactions_sum(self, transactions_sum: int) -> None:
+        thresholds_directory = self._get_daily_transaction_threshold_file_path()
```
dybi: `directory`? wrong name here ;)

signing_service/signing_service/signing_service.py
```diff
@@ -281,11 +281,11 @@ def _handle_connection(
                 middleman_message_response = self._prepare_error_response(ErrorCode.UnexpectedMessage, exception)
             # If received frame is correct, validate transaction.
             else:
-                self._update_last_signed_transaction_date(self.last_signed_transaction)
+                self._update_transactions_limit_file()
                 golem_message_response = self._get_signed_transaction(middleman_message.payload)
                 if isinstance(golem_message_response, SignedTransaction):
                     transaction_sum_combined = self.signing_service_daily_transaction_sum_so_far + middleman_message.payload.value
-                    if transaction_sum_combined > MAXIMUM_DAILY_THRESHOLD:
+                    if MAXIMUM_DAILY_THRESHOLD < transaction_sum_combined:
```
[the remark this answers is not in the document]
Jakub89: @rwrzesien wrote:
> Please reverse conditions here and below.

You guys have to figure this out first, I can't just switch between two versions all the time xD

signing_service/signing_service/signing_service.py
Jakub89: same as above

signing_service/signing_service/signing_service.py
Jakub89: Because of comment above:
>This name is misleading, it is suggesting that you will have a ethereum.Tranaction object here, not a date string.

I do agree with @rwrzesien , it's just a name of the file we are storing.

signing_service/signing_service/signing_service.py
Jakub89: No reason, I'll fix it.

signing_service/signing_service/signing_service.py
Jakub89: No, because we use `pathlib.Path` instead of `os.path`.

signing_service/signing_service/signing_service.py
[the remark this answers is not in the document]
rwrzesien: @dybi We were trying to do conditions this way, I don't know if it is in coding style, but the reason is that it is more .. logical? readable? name it ;) to put lower value first, then bigger one.

signing_service/signing_service/signing_service.py
rwrzesien: Yea, the previous name suggested it is a `Transaction` object. But the new name suggest it is `file` object ;) I would add `last_transactions_limit_file_name`.

rwrzesien:
> What is more - there is a logical BUG in the new functionality

Please add tests to cover those cases.

Jakub89:
>What is more - there is a logical BUG in the new functionality:
>1. SS is started `self.last_transactions_limit_file ` gets current date
>2. SS connects, gets authenticated and immediately after that there is a request to it to sign a transaction.
>3. SS enters `_handle_connection()` and then `_update_transactions_limit_file()` 
>4. `current_day_string` is equal to `self.last_transactions_limit_file ` -> `signing_service_daily_transaction_sum_so_far` **will not get updated!**

>What is more, `self.signing_service_daily_transaction_sum_so_far` doesn't get updated when SS signs a transaction

@dybi `signing_service_daily_transaction_sum_so_far` is updated by calling `_add_payload_value_to_daily_transactions_sum(transactions_sum)` method. So I don't see that bug. 
>4. `current_day_string` is equal to `self.last_transactions_limit_file ` -> `signing_service_daily_transaction_sum_so_far` **will not get updated!**

Why would we want to update `transaction_sum_so_far` after checking dates? I'm confused.

signing_service/signing_service/signing_service.py
bartoszbetka: I see two different solution here:
- This directory and files within are temporary.
This solution is good.
- This directory and files within are persistent.
In this case @rwrzesien is right. We want to path to this directory are configurable. This solution also requires create and mount additional disk. We also make sure that directory was creating after disk mounted, so after the first connection.

bartoszbetka: By temporary, I mean for every deployment this directory will be removed.

Jakub89: I think that first solution is fine, but I'll wait for @dybi clarification.

signing_service/signing_service/signing_service.py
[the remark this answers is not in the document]
dybi: for me it is more logical what you want to express - in this case we want to check whether some variable is **above** a certain threshold.

signing_service/signing_service/signing_service.py
dybi: same as above

signing_service/signing_service/signing_service.py
dybi: or simpler - `last_signed_transaction_date`?

PR #919, opened 2018-10-30
Feature ttc script src replace with individual parameters

kbeker:
Before this will be merged complementary branch in Golem_Messages must be tagged https://github.com/golemfactory/golem-messages/pull/266


kbeker:
This is now so outdated. Closed

Issue #877, opened 2018-10-11
Extend Signing Service with email notifications


dybi:
this is a part of: https://github.com/golemfactory/concent/issues/872

Implement functionality in `Notifier` class, so it is able to send emails.
E-mail address should be added as a command line mandatory argument for running `Signing Service`:
`-e, --email, type=str`

**[Optional]** Some simple validation for email address could be added (like, it should be string of format `sth@domain.ext`).

`Notifier`'s instance should get this email address as argument and should be passed as argument for `Signing Service` instance.

PaweuB:
https://www.pivotaltracker.com/story/show/161179031

Jakub89:
`-e` is taken by `--sentry-environment` and `-m` is taken by `--max-reconnect-attempts`. I don't know if we need shorter version for `--email`, `-em` maybe?

PR #921, opened 2018-10-31
Feature extend signing service with email notifications

Jakub89:
Resolves: #877

### DIFF
--- a/signing_service/signing_service/signing_service.py
+++ b/signing_service/signing_service/signing_service.py
@@ -426,7 +426,7 @@ class SigningService:
 
     @staticmethod
     def _get_daily_transaction_threshold_file_path() -> Path:
-        thresholds_directory = (Path.cwd()).joinpath('daily_tresholds')  # pylint: disable=no-member
+        thresholds_directory = (Path.cwd()).joinpath('daily_thresholds')  # pylint: disable=no-member
         thresholds_directory.mkdir(exist_ok=True)  # pylint: disable=no-member
         daily_file = datetime.datetime.now().strftime('%Y-%m-%d')
         thresholds_directory.joinpath(daily_file).touch(exist_ok=True)  # pylint: disable=no-member
--- a/signing_service/signing_service/signing_service.py
+++ b/signing_service/signing_service/signing_service.py
@@ -50,10 +50,11 @@ from signing_service.constants import WARNING_DAILY_THRESHOLD
 from signing_service.exceptions import SigningServiceMaximumReconnectionAttemptsExceeded
 from signing_service.exceptions import SigningServiceUnexpectedMessageError
 from signing_service.exceptions import SigningServiceValidationError
+from signing_service.utils import ConsoleNotifier
+from signing_service.utils import EmailNotifier
 from signing_service.utils import is_private_key_valid
 from signing_service.utils import is_public_key_valid
 from signing_service.utils import make_secret_provider_factory
-from signing_service.utils import Notifier
 
 logger = logging.getLogger()
 crash_logger = logging.getLogger('crash')
@@ -92,7 +93,7 @@ class SigningService:
         signing_service_private_key: bytes,
         ethereum_private_key: str,
         maximum_reconnect_attempts: int,
-        notifier: Notifier = Notifier(),
+        notifier: Union[ConsoleNotifier, EmailNotifier],
     ) -> None:
         assert isinstance(host, str)
         assert isinstance(port, int)
@@ -290,14 +291,18 @@ class SigningService:
                             f'Signing Service is unable to transact more then {MAXIMUM_DAILY_THRESHOLD} GNTB today.'
                             f'Transaction from {middleman_message.payload.from_address} rejected.'
                         )
-                        self.notifier.send()
+                        self.notifier.send(
+                            f'Signing Service is unable to transact more then {MAXIMUM_DAILY_THRESHOLD} GNTB today.'
+                        )
                         golem_message_response = TransactionRejected(
                             reason=TransactionRejected.REASON.DailyLimitExceeded,
                             nonce=middleman_message.payload.nonce,
                         )
                     elif transaction_sum_combined > WARNING_DAILY_THRESHOLD:
                         logger.warning(f'Signing Service has signed transactions worth {transaction_sum_combined} GNTB today.')
-                        self.notifier.send()
+                        self.notifier.send(
+                            f'Signing Service has signed transactions worth {transaction_sum_combined} GNTB today.'
+                        )
                         self._add_payload_value_to_daily_transactions_sum(transaction_sum_combined)
                     else:
                         self._add_payload_value_to_daily_transactions_sum(transaction_sum_combined)
@@ -569,7 +574,7 @@ def _parse_arguments() -> argparse.Namespace:
     return parser.parse_args()
 
 
-if __name__ == '__main__':
+def main() -> None:
     logging.config.fileConfig(os.path.join(os.path.dirname(os.path.abspath(__file__)), '..', 'logging.ini'))
 
     # Parse required arguments.
@@ -584,6 +589,15 @@ if __name__ == '__main__':
     )
     crash_logger.handlers[0].client = raven_client  # type: ignore
 
+    if hasattr(args, "from_email_address"):
+        notifier = EmailNotifier(
+            args.from_email_address,
+            args.from_email_password,
+            args.to_email_addresses,
+        )
+    else:
+        notifier = ConsoleNotifier()  # type: ignore
+
     SigningService(
         args.concent_cluster_host,
         args.concent_cluster_port,
@@ -592,4 +606,9 @@ if __name__ == '__main__':
         args.signing_service_private_key,
         args.ethereum_private_key,
         args.max_reconnect_attempts,
+        notifier,
     ).run()
+
+
+if __name__ == '__main__':
+    main()
--- a/signing_service/signing_service/signing_service.py
+++ b/signing_service/signing_service/signing_service.py
@@ -491,6 +491,20 @@ def _parse_arguments() -> argparse.Namespace:
         help=f'Environment which will be set in Raven client config `environment` parameter.',
     )
 
+    subparsers = parser.add_subparsers(help='options for sending notifications')
+
+    # create optional parser for the email notifier
+    parser_for_cluster_notifier = subparsers.add_parser('smtp-notifier', help='smtp notifier with no harcoded values')
+    parser_for_cluster_notifier.add_argument('-fea', '--from-email-address', type=str, help='gmail login', required=True)
+    parser_for_cluster_notifier.add_argument('-fep', '--from-email-password', type=str, help='gmail password', required=True)
+    parser_for_cluster_notifier.add_argument(
+        '-tea',
+        '--to-email-addresses',
+        nargs='+',
+        help='list of addresses to send notifications to.',
+        required=True,
+    )
+
     ethereum_private_key_parser_group = parser.add_mutually_exclusive_group(required=True)
     ethereum_private_key_parser_group.add_argument(  # type: ignore
         '--ethereum-private-key',
--- a/signing_service/signing_service/tests/test_authenticate.py
+++ b/signing_service/signing_service/tests/test_authenticate.py
@@ -15,6 +15,7 @@ from middleman_protocol.stream import unescape_stream
 
 from signing_service.constants import SIGNING_SERVICE_DEFAULT_RECONNECT_ATTEMPTS
 from signing_service.signing_service import SigningService
+from signing_service.utils import ConsoleNotifier
 from .utils import SigningServiceIntegrationTestCase
 
 
@@ -125,6 +126,7 @@ class TestSigningServiceAuthenticate(SigningServiceIntegrationTestCase):
                         SIGNING_SERVICE_PRIVATE_KEY,
                         TEST_ETHEREUM_PRIVATE_KEY,
                         SIGNING_SERVICE_DEFAULT_RECONNECT_ATTEMPTS,
+                        ConsoleNotifier(),
                     )
 
                     # For test purposes we reverse roles, so signing service works as server.
--- a/signing_service/signing_service/tests/test_handle_connection.py
+++ b/signing_service/signing_service/tests/test_handle_connection.py
@@ -26,6 +26,7 @@ from signing_service.constants import SIGNING_SERVICE_DEFAULT_RECONNECT_ATTEMPTS
 from signing_service.constants import WARNING_DAILY_THRESHOLD
 from signing_service.exceptions import SigningServiceValidationError
 from signing_service.signing_service import SigningService
+from signing_service.utils import ConsoleNotifier
 from .utils import SigningServiceIntegrationTestCase
 
 
@@ -297,6 +298,7 @@ class TestSigningServiceHandleConnection(SigningServiceIntegrationTestCase):
                         SIGNING_SERVICE_PRIVATE_KEY,
                         TEST_ETHEREUM_PRIVATE_KEY,
                         SIGNING_SERVICE_DEFAULT_RECONNECT_ATTEMPTS,
+                        ConsoleNotifier(),
                     )
 
                     # For test purposes we reverse roles, so signing service works as server.
--- a/signing_service/signing_service/tests/test_run.py
+++ b/signing_service/signing_service/tests/test_run.py
@@ -8,7 +8,7 @@ import pytest
 from signing_service.constants import SIGNING_SERVICE_DEFAULT_RECONNECT_ATTEMPTS
 from signing_service.exceptions import SigningServiceMaximumReconnectionAttemptsExceeded
 from signing_service.signing_service import SigningService
-
+from signing_service.utils import ConsoleNotifier
 
 TEST_ETHEREUM_PRIVATE_KEY = '47a286230c8b3a1c3fa0282f6a65d1d57ffe5147dafaef7cd110d24ed51b462e'
 
@@ -34,6 +34,7 @@ class TestSigningServiceRun:
             SIGNING_SERVICE_PRIVATE_KEY,
             TEST_ETHEREUM_PRIVATE_KEY,
             SIGNING_SERVICE_DEFAULT_RECONNECT_ATTEMPTS,
+            ConsoleNotifier(),
         ]
 
     def test_that_signing_service_should_be_instantiated_correctly_with_all_parameters(self):
--- a/signing_service/signing_service/utils.py
+++ b/signing_service/signing_service/utils.py
@@ -1,8 +1,12 @@
+import logging.config
 import os
+import smtplib
 from argparse import Action
 from argparse import Namespace
 from argparse import ArgumentParser
 from base64 import b64decode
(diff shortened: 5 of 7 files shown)


signing_service/signing_service/utils.py
```diff
@@ -112,5 +115,41 @@ def __call__(  # type: ignore
 
 class Notifier(object):
```
rwrzesien: `object` is redundant.

signing_service/signing_service/utils.py
```diff
@@ -112,5 +115,41 @@ def __call__(  # type: ignore
...
+    ) -> None:
+        message_to_send = MIMEMultipart()
+        message_to_send['From'] = "Signing Service Notifier"
+        message_to_send['To'] = ','.join(self.to_email_addresses)
+        message_to_send['Subject'] = "Signing Service notification"
+        message_to_send.attach(MIMEText(message))
+        if self.from_email_address is not None and self.from_email_password is not None:
+            self.send_email(message_to_send)
+        print(message_to_send)
+        del message_to_send
```
rwrzesien: Is this necessary? I think garbage collector should handle this by default. If there is a reason please let me know.

signing_service/signing_service/utils.py
```diff
@@ -112,5 +115,41 @@ def __call__(  # type: ignore
...
 
-    def send(self) -> None:
-        pass
+    def __init__(
+        self,
+        from_email_address: str,
+        from_email_password: str,
+        to_email_addresses: list,
+    ) -> None:
+        self.from_email_address = from_email_address
```
rwrzesien: You may also use `__slots__` like in `SigningService` to make it faster.

signing_service/signing_service/utils.py
```diff
@@ -112,5 +115,41 @@ def __call__(  # type: ignore
...
+        message: str,
+    ) -> None:
+        message_to_send = MIMEMultipart()
+        message_to_send['From'] = "Signing Service Notifier"
+        message_to_send['To'] = ','.join(self.to_email_addresses)
+        message_to_send['Subject'] = "Signing Service notification"
+        message_to_send.attach(MIMEText(message))
+        if self.from_email_address is not None and self.from_email_password is not None:
+            self.send_email(message_to_send)
+        print(message_to_send)
```
rwrzesien: Is this leftover or on purpose? If on purpose why not using logging?

signing_service/signing_service/utils.py
```diff
@@ -112,5 +115,41 @@ def __call__(  # type: ignore
...
+        try:
+            server = smtplib.SMTP_SSL('smtp.gmail.com:465', timeout=3)
+            server.ehlo()
+            server.starttls()
+            server.login(self.from_email_address, self.from_email_password)
+            server.set_debuglevel(True)
+            server.send_message(message)
+            server.quit()
+        except smtplib.SMTPException as exception:
+            print(exception)
```
rwrzesien: Same here about using logging.

signing_service/signing_service/utils.py
```diff
@@ -112,5 +115,41 @@ def __call__(  # type: ignore
...
+            self.send_email(message_to_send)
+        print(message_to_send)
+        del message_to_send
+
+    def send_email(
+        self,
+        message: str,
+    ) -> None:
+        try:
+            server = smtplib.SMTP_SSL('smtp.gmail.com:465', timeout=3)
```
rwrzesien: How about starting server and login in `__init__()`, so it will be done once?
Same for `quit()`, I think it would be best to make a separate method for it and connect it to `SigningService` shutdown handlers (shutdown signal, exception handler etc.) so it will go down along with `SigningService`.

signing_service/signing_service/signing_service.py
```diff
@@ -578,4 +596,9 @@ def _parse_arguments() -> argparse.Namespace:
         args.signing_service_private_key,
         args.ethereum_private_key,
         args.max_reconnect_attempts,
+        Notifier(
```
rwrzesien: Won't it be better to start `Notifier` after connection to `MiddleMan` is established? @dybi what do you think?

signing_service/signing_service/utils.py
Jakub89: I've just seen it being use in some examples when `MIMEMultipart()` is used, I'm don't know what was the exact reason.

signing_service/signing_service/utils.py
Jakub89: Yes, on purpose. We want to see that message in console, when `Notifier()` have no email. I do realize that `print()` is not the prettiest, and need to be change. But we kind of use logging already in Signing Service with same text.

signing_service/signing_service/utils.py
Jakub89: I was concern that somehow we can lose connection to gmail `smtp`, when Signing Service is still up and running. But okey it's definitely possible to split it to separate method.

signing_service/signing_service/signing_service.py
Jakub89: The question is, are we planning to use email Notifications somewhere else in near future other then `_handle_connection`.

signing_service/signing_service/utils.py
rwrzesien: Now that you have mentioned it, if that happen and server is not able to reconnect by itself, it might be a problem. I am not sure if we should change it then. After all this email will be sent at most 2 times a day, right? 
@dybi what do you think?

signing_service/signing_service/signing_service.py
```diff
@@ -491,6 +496,20 @@ def _parse_arguments() -> argparse.Namespace:
...
+    parser_for_local_notifier = subparsers.add_parser('console-notifier', help='default console notifier')
+    parser_for_local_notifier.add_argument('--from-email-address', default=None)
+    parser_for_local_notifier.add_argument('--from-email-password', default=None)
+    parser_for_local_notifier.add_argument('--to-email-addresses', default=[''])
+
+    # create the parser for the email notifier
+    parser_for_cluster_notifier = subparsers.add_parser('smtp-notifier', help='smtp notifier with no harcoded values')
+    parser_for_cluster_notifier.add_argument('from-email-address', type=str, help='gmail login')
+    parser_for_cluster_notifier.add_argument('from-email-password', type=str, help='gmail password')
+    parser_for_cluster_notifier.add_argument('to-email-addresses', type=List[str], help='list of addresses to send norifications to.')
```
dybi: "norifications" -> "notifications"

signing_service/signing_service/signing_service.py
```diff
@@ -491,6 +496,20 @@ def _parse_arguments() -> argparse.Namespace:
         help=f'Environment which will be set in Raven client config `environment` parameter.',
     )
 
+    subparsers = parser.add_subparsers(help='sub-command help')
+
+    # create the parser for the console notifier
+    parser_for_local_notifier = subparsers.add_parser('console-notifier', help='default console notifier')
+    parser_for_local_notifier.add_argument('--from-email-address', default=None)
```
dybi: This and below arguments are unnecessary for `console-notifier` - please remove them. `ConsoleNotifier` doesn't need all that stuff.

signing_service/signing_service/signing_service.py
```diff
@@ -578,4 +597,9 @@ def _parse_arguments() -> argparse.Namespace:
         args.signing_service_private_key,
         args.ethereum_private_key,
         args.max_reconnect_attempts,
+        Notifier(
```
dybi: We need two `Notifier` classes:
1. `EmailNotifier` - which instance will be created when parser finds `smtp-notifier` option
2. `ConsoleNotifier` - which will be created when parser finds `console-notifier` option
3. `smtp-notifier` and `console-notifier` options must be mutually exclusive

`Notifier` is just a simple interface that provides a method `send(message)`. The above mentioned implementations do it like that:
1. `ConsoleNotifier` "sends" the message to stdout via logger (simply - logs the message)
2. `EmailNotifier` sends the message through gmail account.

signing_service/signing_service/tests/test_authenticate.py
```diff
@@ -125,6 +126,7 @@ def mocked_generator():
                         SIGNING_SERVICE_PRIVATE_KEY,
                         TEST_ETHEREUM_PRIVATE_KEY,
                         SIGNING_SERVICE_DEFAULT_RECONNECT_ATTEMPTS,
+                        Notifier(None, None, ['']),
```
dybi: In tests, `SigningService` should be fed with `ConsoleNotifier` instance

signing_service/signing_service/tests/test_handle_connection.py
```diff
@@ -297,6 +298,7 @@ def mocked_generator():
                         SIGNING_SERVICE_PRIVATE_KEY,
                         TEST_ETHEREUM_PRIVATE_KEY,
                         SIGNING_SERVICE_DEFAULT_RECONNECT_ATTEMPTS,
+                        Notifier(None, None, ['']),
```
dybi: In tests, `SigningService` should be fed with `ConsoleNotifier` instance

signing_service/signing_service/tests/test_run.py
```diff
@@ -34,6 +34,7 @@ def setUp(self, unused_tcp_port_factory):
             SIGNING_SERVICE_PRIVATE_KEY,
             TEST_ETHEREUM_PRIVATE_KEY,
             SIGNING_SERVICE_DEFAULT_RECONNECT_ATTEMPTS,
+            Notifier(None, None, ['']),
```
dybi: In tests, `SigningService` should be fed with `ConsoleNotifier` instance

signing_service/signing_service/utils.py
```diff
@@ -110,7 +113,43 @@ def __call__(  # type: ignore
         setattr(namespace, self.dest, self.const)
 
 
-class Notifier(object):
+class Notifier():
```
dybi: As stated above - we need two `Notifier` classes

signing_service/signing_service/utils.py
```diff
@@ -110,7 +113,43 @@ def __call__(  # type: ignore
...
+        message: str,
+    ) -> None:
+        message_to_send = MIMEMultipart()
+        message_to_send['From'] = "Signing Service Notifier"
+        message_to_send['To'] = ','.join(self.to_email_addresses)
+        message_to_send['Subject'] = "Signing Service notification"
+        message_to_send.attach(MIMEText(message))
+        if self.from_email_address is not None and self.from_email_password is not None:
+            self.send_email(message_to_send)
+        print(message_to_send)
```
dybi: This must be removed: in `EmailNotifier` we just send the message, in `ConsoleNotifier` we log it through a logger

signing_service/signing_service/utils.py
dybi: I think it shouldn't be deleted explicitly. Leave it to python ;)

signing_service/signing_service/utils.py
dybi: I agree - please use logger.

signing_service/signing_service/signing_service.py
dybi: @rwrzesien, I wouldn't be for tangling `Notifier` and `SigningService`classes. Right know we have loose coupling between them and `Notifier` isn't even aware of existence of `SignigService` and `SignigService` just gets something in its `__init__`  and knows that it use `send(message)` of thath something ;) Let it stay as it is.

signing_service/signing_service/signing_service.py
Jakub89: @dybi There are some troubles with that part. Because base on what am I suppose to chose which `Notifier` to use? `subparsers` is not a variable I can check. 

- Right now I can check if `from_email_address` and ` from_email_password` are not `None`, so that I can use `EmailNotifier`, else console.

- Or we could change it by adding additional default argument that specify what did we chose. Something like `--notifier`

dybi: yes, we should have an argument to specify which notifier to choose. And depending on its value use corresponding subparser.

signing_service/signing_service/utils.py
Jakub89: Done

signing_service/signing_service/utils.py
Jakub89: Done

signing_service/signing_service/utils.py
Jakub89: Done

signing_service/signing_service/utils.py
Jakub89: Done. Switch to logging.

signing_service/signing_service/utils.py
Jakub89: Done

signing_service/signing_service/utils.py
Jakub89: Ok. I've separated `connect_to_gmail_smtp_server()` and `end_smtp_server_connection()` function from `send_email()`, but adding `self.server` to the class "upset" `mypy` a little bit, so take another look at this.

signing_service/signing_service/signing_service.py

signing_service/signing_service/signing_service.py
Jakub89: As we discuss, I've deleted whole `console-notifier` subparser because it's our default and has no additional arguments.

signing_service/signing_service/signing_service.py
Jakub89: Done.
But like I wrote above, `console-notifier` is gone since it has no argument and is our default if `smtp-notifier` is not listed. It's impossible to make `subparsers` also a `add_mutually_exclusive_group`. They suppose to be like that by default but all `subparsers` need at least one argument.

signing_service/signing_service/tests/test_authenticate.py
Jakub89: Done

signing_service/signing_service/tests/test_handle_connection.py
Jakub89: Done

signing_service/signing_service/tests/test_run.py
Jakub89: Done

signing_service/signing_service/utils.py
Jakub89: Done

signing_service/signing_service/utils.py
Jakub89: Done

signing_service/signing_service/signing_service.py
```diff
@@ -491,6 +496,14 @@ def _parse_arguments() -> argparse.Namespace:
         help=f'Environment which will be set in Raven client config `environment` parameter.',
     )
 
+    subparsers = parser.add_subparsers(help='options for sending notifications')
+
+    # create optional parser for the email notifier
+    parser_for_cluster_notifier = subparsers.add_parser('smtp-notifier', help='smtp notifier with no harcoded values')
+    parser_for_cluster_notifier.add_argument('-fea', '--from-email-address', type=str, help='gmail login')
+    parser_for_cluster_notifier.add_argument('-fep', '--from-email-password', type=str, help='gmail password')
+    parser_for_cluster_notifier.add_argument('-tea', '--to-email-addresses', nargs='+', help='list of addresses to send notifications to.')
```
rwrzesien: :+1: tea

Jakub89: I like coffee more, but it would be to long :P

signing_service/signing_service/signing_service.py
```diff
@@ -570,6 +583,15 @@ def _parse_arguments() -> argparse.Namespace:
     )
     crash_logger.handlers[0].client = raven_client  # type: ignore
 
+    if any('smtp-notifier' in args for args in argparse.ArgumentParser().parse_known_args()):  # type: ignore
```
dybi: it is overcomplicated, you culd simply:
```suggestion
    if 'smtp-notifier' in args:
```

signing_service/signing_service/signing_service.py
```diff
@@ -570,6 +583,15 @@ def _parse_arguments() -> argparse.Namespace:
...
+        notifier_class = EmailNotifier(
```
dybi:
```suggestion
        notifier = EmailNotifier(
```

signing_service/signing_service/signing_service.py
```diff
@@ -570,6 +583,15 @@ def _parse_arguments() -> argparse.Namespace:
...
+            args.from_email_address,
+            args.from_email_password,
+            args.to_email_addresses,
+        )
+    else:
+        notifier_class = ConsoleNotifier()  # type: ignore
```
dybi:
```suggestion
        notifier = ConsoleNotifier()  # type: ignore
```
BTW: why ` # type: ignore`?

signing_service/signing_service/signing_service.py
Jakub89: I can't because it wouldn't work. I've been trying to explain it before. `subarsers` are not arguments, so `smtp-notifier` is never in `args`.

signing_service/signing_service/signing_service.py
Jakub89: Incompatible types in assignment (expression has type "ConsoleNotifier", variable has type "EmailNotifier")
I don't know how to fix this in `SigningService` I have `notifier: Union[ConsoleNotifier, EmailNotifier]`, but it doesn't seems to make difference.

signing_service/signing_service/signing_service.py
Jakub89: pylint is crying about that:
`************* Module signing_service.signing_service
W: 96, 8: Redefining name 'notifier' from outer scope (line 587) (redefined-outer-name)
`
Maybe `notifier_type` if not `notifier_class`

dybi: just make a `main` function and then it will become local scope of it.

signing_service/signing_service/signing_service.py
dybi: BTW: Make all those arguments required, with `required=True`

Jakub89: I just did it actually for `from-email-address` and `from-email-password`, but I'm not sure if `to-email-addresses` should be required too, because leaving it empty don't cause any errors.

signing_service/signing_service/signing_service.py
dybi: then just:
```suggestion
   if hasattr(args, "from_email_address"): 
```

signing_service/signing_service/signing_service.py
dybi: You could make an abstract class `Notifier`, from which `ConsoleNotifier` and `EmailNotifier` would be derived. And then use type `Notifier` in annotations.

Jakub89: That didn't change anything and now I've got:
```
signing_service/signing_service/signing_service.py:295: error: "Notifier" has no attribute "send"
signing_service/signing_service/signing_service.py:304: error: "Notifier" has no attribute "send"
signing_service/signing_service/signing_service.py:594: error: Incompatible types in assignment (expression has type "ConsoleNotifier", variable has type "EmailNotifier")
```
I thing I'm gone stick with #ignore

signing_service/signing_service/signing_service.py

Issue #896, opened 2018-10-16
Bankster: `finalize payment` operation changes


rwrzesien:
This is a part of the implementation of the "Bankster" blueprint (#479).

Requires https://github.com/golemfactory/concent/issues/894 and https://github.com/golemfactory/concent/issues/568 to be implemented first.

There should be some updates in the current implementation of `finalize payment` operation in Bankster to allow it to follow deposit state.

##### Input - replace current input with:
| Parameter                     | Type                              | Optional | Remarks                                                                                                                 |
|-------------------------------|-----------------------------------|----------|-------------------------------------------------------------------------------------------------------------------------|
| `deposit_claim`               | `DepositClaim`                    | no       | The claim to pay.

##### Output - replace current output with:
| Parameter                     | Type                              | Optional | Remarks                                                                                                                 |
|-------------------------------|-----------------------------------|----------|-------------------------------------------------------------------------------------------------------------------------|
| `tx_hash`                     | string                            | yes       | Hash of an Ethereum transaction that was created. Empty value if no transaction was actually submitted.

##### Sequence of operation - replace current sequence with:
1. **Deposit query**
    - Bankster asks SCI about the amount of funds available on the deposit account listed in the `DepositClaim`.
2. **Claim freeze**:
    - Bankster begins a database transaction.
    - Bankster puts a database lock on the `DepositAccount` object.
3. **Payment calculation**:
    - Bankster sums the `amount`s of all existing `DepositClaim`s that have the same payer as the one being processed.
    - Bankster subtracts that value from the amount of funds available in the deposit.
    - If the result is negative or zero, Bankster removes the `DepositClaim` object being processed.
    - Otherwise if the result is lower than `DepositAccount.amount`, Bankster sets this field to the amount that's actually available.
4. **Transaction**:
    - If the `DepositClaim` still exists at this point:
        - Bankster uses SCI to create an Ethereum transaction.
        - Bankster puts transaction ID in `DepositClaim.tx_hash`.
    - This part must be done while the `DepositAccount` is still locked to prevent two simultaneous operations from independently making two separate payments for the same claim.
6. **Unfreeze and result**: If everything goes well, the operation returns.
    - Bankster commits the database transaction.
        Database locks on `DepositAccount`s are released.
    - Bankster returns `DepositClaim.tx_hash` if a transaction has been created.

PaweuB:
Pivotal: 
https://www.pivotaltracker.com/story/show/161280988

Issue #895, opened 2018-10-16
Bankster: `claim deposit` operation changes


rwrzesien:
There should be some updates in the current implementation of `claim deposit` operation in Bankster to allow it to keep deposit state.

##### Input - modify current input by adding following parameters
| Parameter                       | Type                  | Optional | Remarks                                                                                                                 |
|---------------------------------|-----------------------|----------|-------------------------------------------------------------------------------------------------------------------------|
| `subtask`                       | `Subtask`             | no       | Subtask object.
| `requestor_public_key`          | string                | no       | Public key of the requestor. Comes from `TaskToCompute`.
| `provider_public_key`           | string                | no       | Public key of the provider. Comes from `TaskToCompute`.
| `requestor_ethereum_public_key` | string                | no       | Address of the Ethereum account belonging to the requestor. Cannot be the same as `provider_ethereum_public_key`. Comes from `TaskToCompute`.
| `provider_ethereum_public_key`  | string                | no       | Address of the Ethereum account belonging to the provider. Cannot be the same as `requestor_ethereum_public_key`. Comes from `TaskToCompute`.
| `subtask_cost`                  | decimal               | no       | The cost of performing the work on the subtask that the requestor has agreed to pay the provider. Comes from `TaskToCompute.price` multiplied by maximum task duration. Must be greater than zero.

##### Output - replace current output with:
| Parameter                       | Type                  | Optional | Remarks                                                                                                                 |
|---------------------------------|-----------------------|----------|-------------------------------------------------------------------------------------------------------------------------|
| `claim_against_requestor`       | `DepositClaim`        | no       | The `DepositClaim` object that was created to lock a part of requestor's deposit.
| `claim_against_provider`        | `DepositClaim`        | yes      | The `DepositClaim` object that was created to lock a part of provider's deposit. May be empty if nothing is claimed from the provider.

##### Sequence of operation - modify current sequence:
1. **Claim calculation**: Bankster determines the amount that needs to be claimed from each account (This should be already done, but please review and confirm):
    - In `ForcedAcceptance` use case:
        - Requestor may have to pay `subtask_cost`.
        - Provider does not pay anything.
    - In `AdditionalVerification` use case:
        - Requestor may have to pay `subtask_cost`.
        - Provider needs to pay verification cost (which is constant and determined by Concent's settings).
2. **Initialization**:
    - Bankster begins a database transaction.
[...]

PR #922, opened 2018-11-02
Feature bankster claim deposit operation changes

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/895
Resolves https://github.com/golemfactory/concent/issues/896


concent_api/core/tasks.py
```diff
@@ -229,6 +278,21 @@ def verification_result(
...
 
+        task_to_compute = deserialize_message(subtask.task_to_compute.data.tobytes())
+
+        delete_deposit_claim(
+            subtask_id=subtask_id,
+            concent_use_case=ConcentUseCase.ADDITIONAL_VERIFICATION,
+            ethereum_address=task_to_compute.requestor_ethereum_address,
+            public_key=task_to_compute.requestor_public_key,
+        )
+        delete_deposit_claim(
```
dybi: I am not sure if I understand correctly. If there's `VERIFICATION_MISMATCH`, why provider's claim is deleted? I thought that provider should pay for verification anyways.

concent_api/core/tests/test_payments_bankster.py
```diff
@@ -3,87 +3,135 @@
...
+            claim_against_provider.payee_ethereum_address,
+            ethereum_public_key_to_address(settings.CONCENT_ETHEREUM_PUBLIC_KEY)
+        )
+        self.assertEqual(claim_against_provider.amount, settings.ADDITIONAL_VERIFICATION_COST)
+        self.assertEqual(claim_against_provider.concent_use_case, ConcentUseCase.ADDITIONAL_VERIFICATION)
+
         self.assertEqual(get_deposit_value.call_count, 2)
 
-    def test_that_claim_deposit_return_both_true_if_requestor_has_enough_funds_and_it_is_not_additional_verification(self):
+    def test_that_claim_deposit_return_deposit_claim_if_requestor_has_enough_funds_and_it_is_not_additional_verification(self):
```
dybi: ..that_claim_deposit_returns_**only_requestors**_deposit_claim_if

concent_api/core/tests/test_payments_bankster.py
```diff
@@ -3,87 +3,135 @@
...
+        self.assertIsInstance(claim_against_requestor, DepositClaim)
+        self.assertEqual(claim_against_requestor.subtask_id, self.task_to_compute.subtask_id)
+        self.assertEqual(claim_against_requestor.payee_ethereum_address, self.task_to_compute.provider_ethereum_address)
+        self.assertEqual(claim_against_requestor.amount, self.subtask_cost)
+        self.assertEqual(claim_against_requestor.concent_use_case, ConcentUseCase.FORCED_ACCEPTANCE)
+
+        self.assertIsNone(claim_against_provider)
+        self.assertEqual(get_deposit_value.call_count, 1)
+
+    def test_that_claim_deposit_return_none_if_requestor_has_zero_funds(self):
```
dybi: return_**nones**

concent_api/core/tests/test_payments_bankster.py
```diff
@@ -92,176 +140,103 @@ class FinalizePaymentBanksterTest(ConcentIntegrationTestCase):
...
         )
         force_subtask_payment.assert_called_once_with(
-            requestor_eth_address=self.task_to_compute.requestor_ethereum_address,
-            provider_eth_address=self.task_to_compute.provider_ethereum_address,
-            value=self.subtask_cost,
-            subtask_id=self.task_to_compute.subtask_id,
+            requestor_eth_address=self.deposit_claim.payer_deposit_account.ethereum_address,
+            provider_eth_address=self.deposit_claim.payee_ethereum_address,
+            # 5 - (2 + 2) | deposit_value - sum_of_other_claims
+            value=1,
```
dybi: is `5` `TaskToCompute`'s cost? If so, I would suggest setting it explicitly in setup

concent_api/core/subtask_helpers.py
```diff
@@ -258,3 +269,39 @@ def get_one_or_none(
...
         return None if len(instances) == 0 else instances[0]
+
+
+def finalize_deposit_claim(
+    subtask_id: str,
+    concent_use_case: ConcentUseCase,
+    ethereum_address: str,
+    public_key: str,
+) -> None:
+    deposit_claim = get_one_or_none(
```
pawelkisielewicz: When `get_one_or_none` was created @cameel wanted it to be generic, not only for subtasks. It was changed in the meantime (it was used only for Subtasks) but it can be used also for others Models. Fixing this function is not this pull request, if you can do it to be generic again please do. But here at least please use:
`deposit_claim: DepositClaim = get_one_or_none(`

pawelkisielewicz: When `get_one_or_none` was created @cameel wanted it to be generic, not only for subtasks. It was changed in the meantime (it was used only for Subtasks) but it can be used also for others Models. Fixing this function is not this pull request, if you can do it to be generic again please do. But here at least please use:
`deposit_claim: DepositClaim = get_one_or_none(`

concent_api/core/payments/bankster.py
```diff
@@ -71,96 +81,165 @@ def claim_deposit(
...
+        elif deposit_claim.concent_use_case == ConcentUseCase.ADDITIONAL_VERIFICATION:
+            ethereum_transaction = service.cover_additional_verification_cost(  # pylint: disable=no-value-for-parameter
+                provider_eth_address=deposit_claim.payer_deposit_account.ethereum_address,
+                value=deposit_claim.amount,
+                subtask_id=deposit_claim.subtask_id,
+            )
+
+        # Bankster puts transaction ID in DepositClaim.tx_hash.
+        deposit_claim.tx_hash = encode_hex(ethereum_transaction.hash)
+        deposit_claim.save()
```
pawelkisielewicz: Why `save` without `full_clean`?

concent_api/core/models.py
```diff
@@ -702,7 +724,7 @@ def clean(self) -> None:
 
 
 class DepositClaim(Model):
-    subtask = ForeignKey(Subtask, blank=True, null=True)
+    subtask_id = CharField(max_length=MESSAGE_TASK_ID_MAX_LENGTH, blank=True, null=True)
```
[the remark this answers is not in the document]
rwrzesien: Please check `handle_send_force_subtask_results`. When we call `claim_deposit` with already created subtask, but there will be no funds, we are returning `ServiceRefused`. But the subtask is already created, and what I understand it shouldn't. So I think it is better to use `subtask_id` here, and if there are no funds fail without creating subtask.

concent_api/core/payments/bankster.py
```diff
@@ -71,96 +81,165 @@ def claim_deposit(
...
-    provider_has_enough_deposit: bool = (
-        concent_use_case != ConcentUseCase.ADDITIONAL_VERIFICATION or
-        service.get_deposit_value(client_eth_address=provider_ethereum_address) >= settings.ADDITIONAL_VERIFICATION_COST  # pylint: disable=no-value-for-parameter
+    validate_bytes_public_key(requestor_public_key, 'requestor_public_key')
+    validate_bytes_public_key(provider_public_key, 'provider_public_key')
+    validate_uuid(subtask_id)
+
+    is_claim_against_provider: bool = (
+        concent_use_case == ConcentUseCase.ADDITIONAL_VERIFICATION and
+        0 < settings.ADDITIONAL_VERIFICATION_COST  # pylint: disable=misplaced-comparison-constant
```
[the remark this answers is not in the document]
rwrzesien: Ok, but I thought we have it agreed to create conditions like `smaller < bigger`.

concent_api/core/payments/bankster.py
```diff
@@ -71,96 +81,165 @@ def claim_deposit(
...
+    # If the amount claimed from provider's deposit is non-zero,
+    # Bankster asks SCI about the amount of funds available in his deposit.
+    if is_claim_against_provider:
+        provider_deposit = service.get_deposit_value(client_eth_address=provider_ethereum_address)  # pylint: disable=no-value-for-parameter
+
+    # Bankster puts database locks on DepositAccount objects
+    # that will be used as payers in newly created DatabaseClaims.
+    with transaction.atomic(using='control'):
+        # Bankster sums the amounts of all existing DepositClaims that have the same payer as the one being processed.
+        sum_of_existing_requestor_claims = DepositClaim.objects.filter(
```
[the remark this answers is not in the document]

concent_api/core/payments/bankster.py
```diff
@@ -71,96 +81,165 @@ def claim_deposit(
...
+                    settings.CONCENT_ETHEREUM_PUBLIC_KEY
+                ),
+                amount=subtask_cost,
+                concent_use_case=concent_use_case,
+                payer_deposit_account=provider_deposit_account,
+            )
+            claim_against_provider.full_clean()
+            claim_against_provider.save()
+        else:
+            claim_against_provider = None  # type: ignore
```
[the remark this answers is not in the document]
rwrzesien: I think so, and personally I have no hard feelings about just silencing mypy instead of doing some gymnastics to make it happy ;)

concent_api/core/payments/bankster.py
```diff
@@ -71,96 +81,165 @@ def claim_deposit(
...
+            )
+        elif deposit_claim.concent_use_case == ConcentUseCase.ADDITIONAL_VERIFICATION:
+            ethereum_transaction = service.cover_additional_verification_cost(  # pylint: disable=no-value-for-parameter
+                provider_eth_address=deposit_claim.payer_deposit_account.ethereum_address,
+                value=deposit_claim.amount,
+                subtask_id=deposit_claim.subtask_id,
+            )
+
+        # Bankster puts transaction ID in DepositClaim.tx_hash.
+        deposit_claim.tx_hash = encode_hex(ethereum_transaction.hash)
```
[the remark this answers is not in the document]
rwrzesien: The only place where deposit claim is created is bankster `claim_deposit` function, and it has `assert concent_use_case in [ConcentUseCase.FORCED_ACCEPTANCE, ConcentUseCase.ADDITIONAL_VERIFICATION]`, but right, I will just add `assert False` there.

concent_api/core/tests/test_payments_bankster.py

concent_api/core/tests/test_payments_bankster.py

concent_api/core/tests/test_payments_bankster.py
rwrzesien: No, it is in mock in line 224.

concent_api/core/tasks.py
rwrzesien: Same as above, plus there was no payment here before.

concent_api/core/tasks.py
```diff
@@ -80,6 +86,20 @@ def upload_finished(subtask_id: str) -> None:
 
             return
 
+        else:
```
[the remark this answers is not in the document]
rwrzesien: The blueprint says 

> Concent Core is responsible for removing DepositClaim objects if the outcome of the use case indicates that they don't need to be paid (e.g. if additional verification confirms requestor's decision to reject the result).

I think it is the case here.

concent_api/core/subtask_helpers.py
rwrzesien: Same as above.

concent_api/core/tests/utils.py
```diff
@@ -975,11 +979,79 @@ def _get_number_of_eth_block(self):  # pylint: disable=no-self-use
     def _pass_rpc_synchronization(self, _rpc, _address, _tx_sign):  # pylint: disable=no-self-use
         return None
 
-    def claim_deposit_true_mock(self, concent_use_case, requestor_ethereum_address, provider_ethereum_address, subtask_cost):  # pylint: disable=unused-argument, no-self-use
-        return (True, True)
+    def claim_deposit_true_mock(self, subtask_id, concent_use_case, requestor_ethereum_address, provider_ethereum_address, subtask_cost, requestor_public_key, provider_public_key):  # pylint: disable=unused-argument, no-self-use
```
Jakub89: I know that we allow more characters in line, but that one should be divide.

concent_api/core/tests/utils.py
```diff
@@ -975,11 +979,79 @@ def _get_number_of_eth_block(self):  # pylint: disable=no-self-use
...
+            amount=subtask_cost,
+            concent_use_case=concent_use_case,
+            payer_deposit_account=provider_deposit_account,
+        )
+        claim_against_provider.full_clean()
+        claim_against_provider.save()
+
+        return (claim_against_requestor, claim_against_provider)
+
+    def claim_deposit_requestor_ok_provider_none_mock(self, subtask_id, concent_use_case, requestor_ethereum_address, provider_ethereum_address, subtask_cost, requestor_public_key, provider_public_key):  # pylint: disable=unused-argument, no-self-use
```
Jakub89: same here

concent_api/core/tests/utils.py
```diff
@@ -975,11 +979,79 @@ def _get_number_of_eth_block(self):  # pylint: disable=no-self-use
...
+            payer_deposit_account=requestor_deposit_account,
+        )
+        claim_against_requestor.full_clean()
+        claim_against_requestor.save()
+
+        return (claim_against_requestor, None)
 
-    def claim_deposit_false_mock(self, concent_use_case, requestor_ethereum_address, provider_ethereum_address, subtask_cost):  # pylint: disable=unused-argument, no-self-use
-        return (False, False)
+    def claim_deposit_false_mock(self, subtask_id, concent_use_case, requestor_ethereum_address, provider_ethereum_address, subtask_cost, requestor_public_key, provider_public_key):  # pylint: disable=unused-argument, no-self-use
```
Jakub89: and maybe here

concent_api/core/tests/utils.py
rwrzesien: :+1: done

concent_api/core/tests/utils.py
rwrzesien: :+1: done

concent_api/core/tests/utils.py
rwrzesien: :+1: done

concent_api/core/payments/bankster.py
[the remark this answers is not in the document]
dybi: @rwrzesien, as far as I remember, conditions should be written as a sentence, so:
* if something is **smaller** than threshold: `value < THRESHOLD`
* if something is **bigger** than threshold: `value > THRESHOLD`

concent_api/core/payments/bankster.py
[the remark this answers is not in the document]
dybi: personally, I would sleep better if I understand why mypy cries about that - otherwise why bother using it if we are indifferent to its pain ;)

pawelkisielewicz:
I approve, but I don't know why I cannot change the icon. For me OK, I approve

concent_api/core/tests/utils.py
```diff
@@ -975,11 +979,106 @@ def _get_number_of_eth_block(self):  # pylint: disable=no-self-use
...
+        claim_against_requestor.save()
+
+        return (claim_against_requestor, None)
+
+    def claim_deposit_false_mock(
+        self,
+        subtask_id,
+        concent_use_case, 
+        requestor_ethereum_address,
+        provider_ethereum_address,
```
Jakub89:
```
./concent_api/core/tests/utils.py:1074:26: W291 trailing whitespace
```

Jakub89:
`api-e2e-additional-verification-test.py` is failing for me locally, please check that out.

concent_api/common/helpers.py
```diff
@@ -204,6 +205,12 @@ def generate_ethereum_address_from_ethereum_public_key(ethereum_public_key: str)
     return generate_ethereum_address_from_ethereum_public_key_bytes(ethereum_public_key).hex()
 
 
+def ethereum_public_key_to_address(ethereum_public_key: str) -> str:
+    return to_checksum_address(
+        sha3(decode_hex(ethereum_public_key))[12:].hex()
+    )
```
cameel: Please provide at least a rudimentary unit test for this function on a known pair of a key and address.

concent_api/core/message_handlers.py
```diff
@@ -571,6 +550,28 @@ def handle_send_force_subtask_results(
...
+            requestor_public_key=requestor_public_key,
+            provider_public_key=provider_public_key,
+        )
+
+        assert claim_against_provider is None
+
+        if claim_against_requestor is None:
+            return message.concents.ServiceRefused(
+                reason=message.concents.ServiceRefused.REASON.TooSmallRequestorDeposit,
+            )
```
cameel: I think Bankster should return a flag from `claim_deposit()` to signal this case. Otherwise there's no way for that function to simply say there's nothing to pay.

This was not in the original blueprint, but we should change it. We can do it in a separate pull request though.

concent_api/core/message_handlers.py
```diff
@@ -1312,27 +1317,42 @@ def handle_send_subtask_results_verify(
...
+            requestor_ethereum_address=task_to_compute.requestor_ethereum_address,
+            provider_ethereum_address=task_to_compute.provider_ethereum_address,
+            subtask_cost=task_to_compute.price,
+            requestor_public_key=requestor_public_key,
+            provider_public_key=provider_public_key,
+        )
+
+        if claim_against_requestor is None:
+            if claim_against_provider is not None:
+                claim_against_provider.delete()
```
cameel: Please use `discard_claim()`. All operations on claims (and especially deleting them) must be performed inside a critical section.

concent_api/core/message_handlers.py
```diff
@@ -1312,27 +1317,42 @@ def handle_send_subtask_results_verify(
...
+            return message.concents.ServiceRefused(
+                reason=message.concents.ServiceRefused.REASON.TooSmallRequestorDeposit,
+            )
+        if claim_against_provider is None:
+            if claim_against_requestor is not None:
+                claim_against_requestor.delete()
```
cameel: Please use `discard_claim()`. All operations on claims (and especially deleting them) must be performed inside a critical section.

rwrzesien:
@Jakub89 What exactly is failing? Isn't it the same reason that was discussed on slack channel - lack of funds on providers deposit? And isn't the same test failing on master?

concent_api/core/message_handlers.py
```diff
@@ -1312,27 +1317,42 @@ def handle_send_subtask_results_verify(
...
+            Subtask.SubtaskState.ADDITIONAL_VERIFICATION,
+        ]:
+            return message.concents.ServiceRefused(
+                reason=message.concents.ServiceRefused.REASON.DuplicateRequest,
+            )
+
+        (claim_against_requestor, claim_against_provider) = bankster.claim_deposit(
+            subtask_id=task_to_compute.subtask_id,
+            concent_use_case=ConcentUseCase.ADDITIONAL_VERIFICATION,
+            requestor_ethereum_address=task_to_compute.requestor_ethereum_address,
```
cameel: This operation can't be performed inside an `atomic` block because it has to be able to commit its transactions (if a transaction already exists, all its commits become a part of the parent transaction).

concent_api/core/message_handlers.py
```diff
@@ -1312,27 +1317,42 @@ def handle_send_subtask_results_verify(
...
+        if claim_against_requestor is None:
+            if claim_against_provider is not None:
+                claim_against_provider.delete()
+            return message.concents.ServiceRefused(
+                reason=message.concents.ServiceRefused.REASON.TooSmallRequestorDeposit,
+            )
+        if claim_against_provider is None:
+            if claim_against_requestor is not None:
+                claim_against_requestor.delete()
+            return message.concents.ServiceRefused(
```
cameel: Why are we removing the claim against requestor's deposit if there's nothing to claim from the provider (and vice versa)? These claims are independent. If bankster says there's nothing to claim from one and does not raise an exception, it should have no influence on the other.

Jakub89:
@rwrzesien No, I get different fails. All e2e tests works on master for me. In server console I get:
```
  File "/home/jakub/Documents/work/projects/concent/concent_api/core/message_handlers.py", line 1454, in handle_message
    return handle_send_force_payment(client_message)
  File "/home/jakub/Documents/work/projects/concent/concent_api/core/message_handlers.py", line 781, in handle_send_force_payment
    requestor_public_key=requestor_public_key,
  File "/home/jakub/Documents/work/projects/concent/concent_api/core/payments/bankster.py", line 314, in settle_overdue_acceptances
    assert youngest_transaction_timestamp <= cut_off_time
UnboundLocalError: local variable 'youngest_transaction_timestamp' referenced before assignment
```
and on console from where I run tests, I get:
```
test_case_6_test_without_script_src_in: FAILED
Expected:<class 'golem_messages.message.concents.AckSubtaskResultsVerify'>, actual:<class 'golem_messages.message.concents.ServiceRefused'>
--------------------------------------------------------------------------------
Total failed tests : 3 out of 6
```

concent_api/core/message_handlers.py
```diff
@@ -1345,19 +1365,13 @@ def handle_send_subtask_results_verify(
...
-            ]:
-                return message.concents.ServiceRefused(
-                    reason=message.concents.ServiceRefused.REASON.DuplicateRequest,
-                )
             if subtask.state_enum in [
                 Subtask.SubtaskState.ACCEPTED,
                 Subtask.SubtaskState.FAILED,
             ]:
+                claim_against_requestor.delete()
+                claim_against_provider.delete()
```
cameel: Please use `discard_claim()`.

concent_api/common/helpers.py
rwrzesien: :+1: done

concent_api/core/tests/utils.py
rwrzesien: :+1: done

rwrzesien:
@Jakub89 I think you are testing a different branch.

concent_api/core/models.py
```diff
@@ -698,7 +698,29 @@ class PendingEthereumTransaction(Model):
     created_at = DateTimeField(auto_now_add=True)
 
 
+class DepositAccountManager(Manager):
+
+    def get_or_create_full_clean(self, client: bytes, ethereum_address: str) -> 'DepositAccount':
```
cameel: Maybe it would be better to pass all `args` and `kwargs` to the `DepositAccount` constructor instead of listing them explicitly here?

concent_api/core/message_handlers.py
rwrzesien: This is not what the blueprint says

> Concent Core is responsible for removing DepositClaim objects if it fails at any point after the operation if the operation has succeeded.
> Bankster removes DepositClaim objects only after they're successfully paid for.

From what I understand in this part, using `discard_claim()`, which is part of the bankster, should be used only after they're successfully paid for.

concent_api/core/models.py
```diff
@@ -721,11 +743,14 @@ class DepositClaim(Model):
     created_at = DateTimeField(auto_now_add=True)
     modified_at = DateTimeField(auto_now=True)
 
+    class Meta:
+        unique_together = ('subtask_id', 'concent_use_case', 'payee_ethereum_address')
+
     def clean(self) -> None:
         super().clean()
-        if self.subtask is None and self.concent_use_case != ConcentUseCase.FORCED_PAYMENT:
+        if self.subtask_id is None and self.concent_use_case != ConcentUseCase.FORCED_PAYMENT:
```
cameel: Can the ID be an empty string here?

concent_api/core/payments/bankster.py
```diff
@@ -51,11 +58,14 @@ def __init__(
...
 def claim_deposit(
+    subtask_id: str,
     concent_use_case: ConcentUseCase,
     requestor_ethereum_address: str,
     provider_ethereum_address: str,
     subtask_cost: int,
-) -> Tuple[bool, bool]:
+    requestor_public_key: bytes,
+    provider_public_key: bytes,
+) -> Tuple[Optional[DepositClaim], Optional[DepositClaim]]:
```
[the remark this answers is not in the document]
cameel: This needs to be updated in the blueprint. When you change stuff like this, please always remember to drop me a comment in the blueprint issue (and it would be a good idea to make a comment in the issue this pull request is solving too).

concent_api/core/models.py
rwrzesien: No, it is passed directly from `task_to_compute` in our code when `ConcentUseCase` is not `FORCED_PAYMENT`, so it can't happen. And we validate against that earlier.

concent_api/core/models.py
rwrzesien: Not in this case, because I then want to query by those two specific fields, and I need to check if they are set. In other words, in this case it is better to point out that exactly those two are necessary.

concent_api/core/message_handlers.py
rwrzesien: We need to discuss, isn't it mentioned in blueprint that Bankster only removes `DepositClaim` objects after they're successfully paid for?

concent_api/core/message_handlers.py
rwrzesien: :+1: done I thought this is considered as operation failure. After all this stops use case from proceeding. So isn't that claim never finalized?

concent_api/core/message_handlers.py
rwrzesien: :+1: done, I guess this should work the same in `handle_send_force_subtask_results`? I have changed it there too.

concent_api/core/payments/bankster.py
```diff
@@ -71,96 +81,165 @@ def claim_deposit(
...
+            )
+            claim_against_provider.full_clean()
+            claim_against_provider.save()
+        else:
+            claim_against_provider = None  # type: ignore
+
+    return (claim_against_requestor, claim_against_provider)
+
+
+def finalize_payment(deposit_claim: DepositClaim) -> Optional[str]:
```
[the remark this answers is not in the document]
cameel: >> BTW. Why is return type str? This function is called in only one place and return value is not assigned to any variable in production code. But in tests it is always assigned and checked. I think it should not be like this. Return type should be None and tests should be fixed to don't use return value..

> I agree with second part, I don't know where this transaction hash is going to be used later (maybe there is a master plan for this).

The fact that it's not used in the current implementation does not mean that it won't be used in the future. Bankster is an API (remember that it was originally going to be a separate service) and it should be a little more generic than the code that uses it. I actually think that our views are making too many assumptions about that it does. It should be treated as a black box.

About the return value: even if the operation does not fail, there are two possible outcomes. Either there is enough deposit to satisfy the claim (fully or in part) or there isn't. If there isn't, the claim is simply removed. I think that it should not be done silently. The caller should be informed about that and be able to take some action. Right now the view does not really care what happens with the payment after we initiate it but this is not good and should change in the future.

So the function should return something. It could return simply `True` or `False`. Returning the transaction ID serves the same purpose and gives more information so why not? We might want to log it or even return it to the caller in the future. It does not cost us anything to be more flexible and informative here.

> > According to #896 return type should be `str` and it shouldn't be Optional. If it is OK in code please comment in issue (or change).

> About first part, updated in issue +1

I'd like to clarify one thing. The types I'm using in the spec are not as specific as `str`, `bytes`. I simply call it a string and that's on purpose. I do this with other types too (e.g. "decimal" instead of `Decimal` or `float`, "list of" instead of `list` or `tuple`). The actual type is a low level detail. Choose one that is appropriate.

The reason is that there are several different type systems we're using (Python, JSON, SQL, Django models). The same data often has to be moved across several of them and it's hard to describe the type in a way that's both accurate and compatible with all of them. Even when using one, same data can sometimes be stored as `string` or `bytes` depending on various factors. When you try to keep track of it, you often run into small inconsistencies in the spec and trying to get them all right requires a lot of effort. I often find that this all effort is unnecessary and it's simply better to leave it all up to implementation. On top of that, trying to be too precise makes the spec less readable.

concent_api/core/payments/bankster.py
```diff
@@ -71,96 +81,168 @@ def claim_deposit(
...
-            subtask_id=subtask_id,
-        )
-        providers_claim_payment_info = ClaimPaymentInfo(
-            amount_paid=available_provider_claim,
-            amount_pending=settings.ADDITIONAL_VERIFICATION_COST - available_provider_claim,
-            tx_hash=transaction_hash,
-            payment_ts=current_time,
+        # Bankster sums the amounts of all existing DepositClaims that have the same payer as the one being processed.
+        aggregated_requestor_claims = DepositClaim.objects.filter(
+            payer_deposit_account=deposit_claim.payer_deposit_account
```
cameel: How do you know requestor is the payer here?

concent_api/core/payments/bankster.py
```diff
@@ -71,96 +81,168 @@ def claim_deposit(
...
+            ethereum_transaction = service.force_subtask_payment(  # pylint: disable=no-value-for-parameter
+                requestor_eth_address=deposit_claim.payer_deposit_account.ethereum_address,
+                provider_eth_address=deposit_claim.payee_ethereum_address,
+                value=deposit_claim.amount,
+                subtask_id=deposit_claim.subtask_id,
+            )
+        elif deposit_claim.concent_use_case == ConcentUseCase.ADDITIONAL_VERIFICATION:
+            ethereum_transaction = service.cover_additional_verification_cost(  # pylint: disable=no-value-for-parameter
+                provider_eth_address=deposit_claim.payer_deposit_account.ethereum_address,
+                value=deposit_claim.amount,
```
cameel: This is wrong. You're deciding whether to use `force_subtask_payment()` or `cover_additional_verification_cost()` based on the use case alone.

This means that for example in verification use case you will use `cover_additional_verification_cost()` for both requestor and the provider. Instead you should use `cover_additional_verification_cost()` to get verification fee from the provider and `force_subtask_payment()` to transfer subtask payment from the requestor to the provider.

You need to look at `deposit_claim.subtask` to know whether whether the payer is a requestor or a provider here. If that's a problem we might also add a flag to `DepositClaim` (that would require changes in the blueprint so drop me a comment if you do this).

concent_api/core/message_handlers.py
cameel: See my response for the other comment: https://github.com/golemfactory/concent/pull/922#discussion_r233003576

It looks like I did not write that clearly enough. It's not that no Bankster function can touch these objects. I just meant that Bankster should not do this automatically. It should not decide on its own to remove them as a part of another operation. But if the view tells it explicitly to remove them (by running `discard_claim()`) it's fine to do it.

concent_api/core/message_handlers.py
cameel: Only `(None, None)` is a failure. And even in this case I now think we should add an additional flag or something to discern failure from the situation where there's simply nothing to pay (I think I suggested this in one of the reviews recently).

If Bankster creates one claim successfully and then fails in a way that does not roll back the transaction automatically, it's supposed to remove the claim and return `(None, None)`.

View should just accept what it gets. If Bankster tells it that only one claim is necessary (and did not fail), then that's it.

concent_api/core/message_handlers.py
cameel: But if `None` means "the deposit is too small, refuse service" then what is the value for "there's nothing to pay, proceed"? It can't mean both.

concent_api/core/subtask_helpers.py
```diff
@@ -283,3 +294,39 @@ def get_one_or_none(
...
+        DepositClaim,
+        subtask_id=subtask_id,
+        concent_use_case=concent_use_case,
+        payer_deposit_account__ethereum_address=ethereum_address,
+        payer_deposit_account__client__public_key=public_key,
+    )
+
+    if deposit_claim:
```
cameel: Please compare explicitly, with `is not None`.

concent_api/core/tasks.py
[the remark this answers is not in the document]
cameel: I agree with @dybi here. We're at the point where the provider has successfully uploaded all files and Concent is about to give a signal to start verification. We don't know the outcome yet. Nothing should be deleted.

We know the outcome only if we're past the deadline (i.e. in the other `if` branch). In the `else` branch it's not known yet.

concent_api/core/tasks.py
```diff
@@ -106,6 +126,18 @@ def upload_finished(subtask_id: str) -> None:
         Subtask.SubtaskState.ACCEPTED,
         Subtask.SubtaskState.FAILED
     ]:
+        delete_deposit_claim(
```
cameel: You should not touch the claims here. It's a duplicate notification. If the verification timed out, succeeded or failed, the claims must have already been processed. If the verification is still going, the claims should stay as is and wait for the result.

concent_api/core/tasks.py
```diff
@@ -114,6 +146,18 @@ def upload_finished(subtask_id: str) -> None:
         )
     # If it's one of the states that can precede verification, report an error. Processing ends here.
     else:
+        delete_deposit_claim(
+            subtask_id=subtask_id,
+            concent_use_case=ConcentUseCase.ADDITIONAL_VERIFICATION,
```
cameel: Just like above. Something's wrong. This case should never happen unless there's a bug in the Conductor or Verifier and one of them has sent a notification even though there's no verification in progress. The claims should not even exist. Just leave them as is for later analysis by the admin.

concent_api/core/tasks.py
cameel: @dybi is right. If there's a mismatch:
- claim against requestor's account should be deleted (provider did not deliver the correct result; no need to pay)
- claim against provider's account should be finalized (Concent did deliver it's service by issuing a verdict and needs to be paid).

In general the claim against the provider (i.e. the verification fee) should never be discarded if verification ends with a verdict (i.e. when Concent performs its service successfully). Currently we give a verdict even if verification fails due to our own fault or times out so this means that provider always pays.

I remember that we talked about it with Łukasz at some point and he said that there should be a separate case for Concent's failure. I'll need to check if that's been changed in the spec or not. Nevertheless, Concent's failure or timeout should be a rare case so it does not matter all that much. The current approach is fine for the time being.

concent_api/core/subtask_helpers.py
cameel: Wasn't the issue that `get_or_none()` has `Subtask` as its argument type or something like that? I remember it from another review.

concent_api/core/subtask_helpers.py
```diff
@@ -283,3 +294,39 @@ def get_one_or_none(
...
```
cameel: Please compare explicitly, with `is not None`.

concent_api/core/subtask_helpers.py
```diff
@@ -283,3 +294,39 @@ def get_one_or_none(
...
+    concent_use_case: ConcentUseCase,
+    ethereum_address: str,
+    public_key: str,
+) -> None:
+    deposit_claim = get_one_or_none(
+        DepositClaim,
+        subtask_id=subtask_id,
+        concent_use_case=concent_use_case,
+        payer_deposit_account__ethereum_address=ethereum_address,
+        payer_deposit_account__client__public_key=public_key,
```
cameel: Don't we have a validation that an `ethereum_address` cannot be used by more than one client? If not, I think we should because a duplicate would mean that multiple clients are using the same deposit.

This would also make `public_key` unnecessary here.

concent_api/core/subtask_helpers.py
```diff
@@ -103,15 +111,18 @@ def _update_timed_out_subtask(subtask: Subtask) -> None:
         )
     elif subtask.state == Subtask.SubtaskState.ADDITIONAL_VERIFICATION.name:  # pylint: disable=no-member
         task_to_compute = deserialize_message(subtask.task_to_compute.data.tobytes())
-        # Worker makes a payment from requestor's deposit just like in the forced acceptance use case.
-        bankster.finalize_payment(
```
cameel: Could you extract this branch of the condition into a function and use it both here and in the Celery task? In both cases a verification timeout should be processed in exactly the same way. Having it in two separate places makes it very likely that it will get out of sync.

concent_api/core/subtask_helpers.py
```diff
@@ -103,15 +111,18 @@ def _update_timed_out_subtask(subtask: Subtask) -> None:
...
-            subtask_id=task_to_compute.subtask_id,
-            concent_use_case=ConcentUseCase.FORCED_ACCEPTANCE,
-            requestor_ethereum_address=task_to_compute.requestor_ethereum_address,
-            provider_ethereum_address=task_to_compute.provider_ethereum_address,
-            subtask_cost=task_to_compute.price,
+        finalize_deposit_claim(
+            subtask_id=subtask.subtask_id,
+            concent_use_case=ConcentUseCase.ADDITIONAL_VERIFICATION,
+            ethereum_address=task_to_compute.requestor_ethereum_address,
```
cameel: Isn't `_update_timed_out_subtask()` usually called inside a transaction? If so, `finalize_payment()` will not be able to commit its own transactions.

BTW, I think we should make it easier to detect situations like this automatically. We should have a wrapper over `transaction.atomic()` that fails if a transaction is already in progress and use it everywhere. We need to detect and eliminate any nested transactions.

cameel: Are these rules consistent with those in Celery tasks?

concent_api/core/tasks.py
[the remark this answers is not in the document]
rwrzesien: :+1: done

concent_api/core/subtask_helpers.py
rwrzesien: Arguments are fine, the returned type was `Optional[Subtask]`, I have changed it to be more generic.

concent_api/core/message_handlers.py
rwrzesien: This situation should not appear. Subtask cost should be always greater than 0 so there will be always something to pay for.

concent_api/core/message_handlers.py

concent_api/core/message_handlers.py
rwrzesien: As mentioned in another comment, claims are not removed here.

cameel:
One more thing. Could you list the differences between your implementation and the blueprint in a comment under it? From what I can see, there are some differences and they've only been added to the issue based on the blueprint, not to the blueprint itself.

concent_api/core/message_handlers.py

concent_api/core/message_handlers.py

concent_api/core/models.py
rwrzesien: I don't know what do you mean by "everywhere". Basically `subtask_id` comes from `task_to_compute` and must be always uuid (we validate against that). So I think the only place where it is accepted to be something else occasionally is this method.

concent_api/core/models.py
```diff
@@ -712,7 +734,7 @@ def clean(self) -> None:
 
 
 class DepositClaim(Model):
-    subtask = ForeignKey(Subtask, blank=True, null=True)
+    subtask_id = CharField(max_length=MESSAGE_TASK_ID_MAX_LENGTH, blank=True, null=True)
```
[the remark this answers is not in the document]
rwrzesien: Actually we have everywhere `blank = True, null = True`. I think it is ok, as it all depends on the use case. I think that `blank = True, null = False` is used more in web development, when you often print the value directly to template. We don't do this, and it is better in my opinion to compare against `None` than empty string.

concent_api/core/models.py
rwrzesien: But then it would be possible to use `created_at` there too, and in case when object does not exist, this `created_at` would be passed to new object.

concent_api/core/payments/bankster.py
```diff
@@ -71,96 +81,168 @@ def claim_deposit(
...
+    # Bankster asks SCI about the amount of funds available in requestor's deposit.
+    requestor_deposit = service.get_deposit_value(client_eth_address=requestor_ethereum_address)  # pylint: disable=no-value-for-parameter
+
+    # If the amount claimed from provider's deposit is non-zero,
+    # Bankster asks SCI about the amount of funds available in his deposit.
+    if is_claim_against_provider:
+        provider_deposit = service.get_deposit_value(client_eth_address=provider_ethereum_address)  # pylint: disable=no-value-for-parameter
+
+    # Bankster puts database locks on DepositAccount objects
+    # that will be used as payers in newly created DatabaseClaims.
```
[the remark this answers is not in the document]

concent_api/core/payments/bankster.py
```diff
@@ -71,96 +81,168 @@ def claim_deposit(
...
+    with transaction.atomic(using='control'):
+        # Bankster sums the amounts of all existing DepositClaims that have the same payer as the one being processed.
+        aggregated_requestor_claims = DepositClaim.objects.filter(
```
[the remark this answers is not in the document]

concent_api/core/payments/bankster.py
```diff
@@ -71,96 +81,168 @@ def claim_deposit(
...
+            amount=subtask_cost,
+            concent_use_case=concent_use_case,
+            payer_deposit_account=requestor_deposit_account,
+        )
+        claim_against_requestor.full_clean()
+        claim_against_requestor.save()
+
+        if is_claim_against_provider:
+            # Bankster sums the amounts of all existing DepositClaims where the provider is the payer.
+            sum_of_existing_provider_claims = DepositClaim.objects.filter(
```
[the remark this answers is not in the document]

concent_api/core/payments/bankster.py
```diff
@@ -71,96 +81,168 @@ def claim_deposit(
...
+                payer_deposit_account=provider_deposit_account
+            ).aggregate(
+                sum_of_existing_claims=Coalesce(Sum('amount'), 0)
+            )
+
+            # If the total of existing claims and the current claim is greater or equal to the current deposit,
+            # we can't add a new claim.
+            if provider_deposit <= sum_of_existing_provider_claims['sum_of_existing_claims'] + settings.ADDITIONAL_VERIFICATION_COST:
+                return (claim_against_requestor, None)
```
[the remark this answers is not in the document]
rwrzesien: Yes, but I have changed it make difference in Concent between lack of provider and requestor funds and return proper reason in `ServiceRefused`. You can check it in `message_handlers.py:1349`.

concent_api/core/payments/bankster.py
```diff
@@ -71,96 +81,168 @@ def claim_deposit(
...
+
+        # Deposit lock for provider.
+        if is_claim_against_provider:
+            claim_against_provider = DepositClaim(
+                subtask_id=subtask_id,
+                payee_ethereum_address=ethereum_public_key_to_address(
+                    settings.CONCENT_ETHEREUM_PUBLIC_KEY
+                ),
+                amount=subtask_cost,
```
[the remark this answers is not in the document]

concent_api/core/payments/bankster.py

concent_api/core/payments/bankster.py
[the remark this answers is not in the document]
cameel: So this is another case where I think we should add a flag (or an exception) to indicate this.

concent_api/core/models.py
cameel: > I don't know what do you mean by "everywhere".

I mean all the Concent code, not just Bankster.

concent_api/core/payments/bankster.py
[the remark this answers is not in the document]
cameel: OK. So just please remember about it when we close this pull request. I'm leaving this as unresolved as a reminder.

concent_api/core/message_handlers.py
cameel: Yeah, kinda. But `handle_send_force_subtask_results()` has problems too. It does not nest transactions but it does not put a lock on the subtask. This means that multiple instances could potentially create claims against the same subtask.

I think it's good enough as a workaround for now but we should have an issue for a correct solution. This is a race condition.

concent_api/core/models.py
cameel: Yeah, but it's the same when you call normal `create()`. If you give it `created_at`, it will save it.

cameel: Anyway, maybe what I propose does have other disadvantages. Never mind. What you have here is a valid solution too.

concent_api/core/message_handlers.py
cameel: I don't understand. Maybe you mean that they are removed but they are not claims that existed before? I still think this operation must be performed inside a critical section. Any manipulation of claims must be.

concent_api/core/message_handlers.py
rwrzesien: Yes, but we agreed it will be done in separate issue.

concent_api/core/message_handlers.py
rwrzesien: I was referring to this comment https://github.com/golemfactory/concent/pull/922#discussion_r232995160. This visible part of the code is now outdated, there is no claims delete there.

concent_api/core/models.py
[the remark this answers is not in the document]
rwrzesien: > OK. But are you sure we're using this consistently?

Yes, in every field which use those two. And I think it works best for our case, because we don't allow users to edit data, we don't do it through admin panel etc. Actually when I have checked it, in all those places we could have `blank=False, null=True` rather than the opposite.
And I think this recommendation which you have quoted is common for web development.

concent_api/core/payments/bankster.py
```diff
@@ -71,96 +81,168 @@ def claim_deposit(
...
 
-    return (requestor_has_enough_deposit, provider_has_enough_deposit)
+    # Bankster creates Client and DepositAccount objects (if they don't exist yet) for the requestor
+    # and also for the provider if there's a non-zero claim against his account.
+    # This is done in single database transaction.
+    with transaction.atomic(using='control'):
+        requestor_client = Client.objects.get_or_create_full_clean(
+            public_key=requestor_public_key,
+        )
+        requestor_deposit_account = DepositAccount.objects.get_or_create_full_clean(
```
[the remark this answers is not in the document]

concent_api/core/subtask_helpers.py
rwrzesien: :+1: done

concent_api/core/tasks.py
rwrzesien: :+1: done

concent_api/core/tasks.py
rwrzesien: :+1: done

concent_api/core/subtask_helpers.py
rwrzesien: :+1: done

concent_api/core/subtask_helpers.py
rwrzesien: Yes, the same as in `upload_finished` and `verification_result` apart from `MISMATCH` case.

rwrzesien: :+1: done

concent_api/core/subtask_helpers.py
rwrzesien: But now it will be something different after change from https://github.com/golemfactory/concent/pull/922#discussion_r233454766

concent_api/core/subtask_helpers.py
rwrzesien: :+1: done

concent_api/core/payments/bankster.py

concent_api/core/message_handlers.py
cameel: Unfortunately it's not completely race-condition-free. There's still a small window between transactions in which to processes can interfere. Consider this:
1. First process runs the view, subtask is in a passive state.
2. Second process runs the view, subtask is in a passive state.
3. First process creates claims.
4. Second process creates claims.
5. First process locks the subtask.
6. Second process tries to lock the subtask and waits.
7. First process changes subtask state and commits.
8. Second pocess can now continue. It also sets subtask state and commits.

We end up with two sets of claims for the same subtask. It would be fine the subtask was locked from the beginning but unfortunately it isn't.

As I said, this is probably not something that will happen often so this flow is fine as a temporary workaround but still, it's broken and should be reported so that we can do it properly if we have time later.

concent_api/core/models.py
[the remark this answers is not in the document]
cameel: I mean that it's common not only in web development. It's simply common in Django.

concent_api/core/payments/bankster.py
```diff
@@ -71,96 +84,214 @@ def claim_deposit(
...
+        # If the DepositClaim still exists at this point, Bankster uses SCI to create an Ethereum transaction.
+        if deposit_claim.concent_use_case == ConcentUseCase.FORCED_ACCEPTANCE:
+            ethereum_transaction = service.force_subtask_payment(  # pylint: disable=no-value-for-parameter
+                requestor_eth_address=deposit_claim.payer_deposit_account.ethereum_address,
+                provider_eth_address=deposit_claim.payee_ethereum_address,
+                value=deposit_claim.amount,
+                subtask_id=deposit_claim.subtask_id,
+            )
+        elif deposit_claim.concent_use_case == ConcentUseCase.ADDITIONAL_VERIFICATION:
+            subtask = Subtask.objects.filter(subtask_id=deposit_claim.subtask_id).first()  # pylint: disable=no-member
```
cameel: Why not `get_one_or_none()`?

concent_api/core/subtask_helpers.py
cameel: You could still extract functions like `finalize_claim_for_additional_verification_case()` into a common module and use them in both places.

But well, it does get pretty ugly that way...

concent_api/core/payments/bankster.py
rwrzesien: There is a problem with circular import. And this code does actually the same.

concent_api/core/tasks.py
```diff
@@ -56,12 +57,15 @@ def upload_finished(subtask_id: str) -> None:
         # If subtask is past the deadline, processes the timeout.
         if parse_datetime_to_timestamp(subtask.next_deadline) < get_current_utc_timestamp():
             # Worker makes a payment from requestor's deposit just like in the forced acceptance use case.
-            bankster.finalize_payment(
+            finalize_deposit_claim(
                 subtask_id=subtask_id,
                 concent_use_case=ConcentUseCase.ADDITIONAL_VERIFICATION,
```
cameel: This Celery task has `@atomic`. `finalize_deposit_claim()` won't be able to commit its transaction.

concent_api/core/tasks.py
```diff
@@ -201,12 +205,15 @@ def verification_result(
     if subtask.next_deadline < parse_timestamp_to_utc_datetime(get_current_utc_timestamp()):
         task_to_compute = deserialize_message(subtask.task_to_compute.data.tobytes())
         # Worker makes a payment from requestor's deposit just like in the forced acceptance use case.
-        bankster.finalize_payment(
+        finalize_deposit_claim(
             subtask_id=subtask_id,
             concent_use_case=ConcentUseCase.ADDITIONAL_VERIFICATION,
```
cameel: Again, this Celery task has `@atomic`. `finalize_deposit_claim()` won't be able to commit its transaction.

concent_api/core/subtask_helpers.py
cameel: Great. Just in case: I just checked and we do have `unique=True` on `DepositAccount.ethereum_address`.

concent_api/core/message_handlers.py
cameel: This is not a complete fix. `claim_deposit()` still has a case where it returns only one claim when it wants to indicate a problem (see https://github.com/golemfactory/concent/pull/922/files#r233395518). If you leave it like this then this claim does not get deleted as it should. You must remove it either here or inside `claim_deposit()`. I said that in my opinion the claim should be deleted inside `claim_deposit()`, not here.

cameel: It's broken now. `claim_deposit()` still has a case where it returns only one claim when it wants to indicate a problem (see https://github.com/golemfactory/concent/pull/922/files#r233395518). If you leave it like this then this claim does not get deleted as it should. You must remove it either here or inside `claim_deposit()`.

I meant that in my opinion the claim should be deleted inside `claim_deposit()`, not that it should not be deleted at all. The function should then return `(None, None)` because it has already cleaned up the mess.

If `claim_deposit()` returns `(None, None)` it might mean either that we don't require payment or that there was not enough deposit. It would be best to add a flag (or maybe an exception?) to make this really explicit. Returning a claim and having the caller delete is a really messy and unintuitive way to indicate this.

concent_api/core/message_handlers.py
```diff
@@ -695,6 +699,12 @@ def handle_send_force_subtask_results_response(
             deserialize_message(subtask.task_to_compute.data.tobytes()),
         ])
 
+        delete_deposit_claim(
+            subtask_id=task_to_compute.subtask_id,
+            concent_use_case=ConcentUseCase.FORCED_ACCEPTANCE,
```
cameel: Please use `discard_claim()`. `delete_deposit_claim()` does not lock the `DepositAccount`.

concent_api/core/message_handlers.py
```diff
@@ -1312,27 +1322,51 @@ def handle_send_subtask_results_verify(
...
-    if not provider_has_enough_deposit:
+    if claim_against_provider is None:
         return message.concents.ServiceRefused(
             reason=message.concents.ServiceRefused.REASON.TooSmallProviderDeposit,
         )
 
+    if subtask is not None and subtask.state_enum in [
+        Subtask.SubtaskState.ACCEPTED,
+        Subtask.SubtaskState.FAILED,
+    ]:
```
cameel: Here you're first creating claims and then immediately deleting them if we're in a state where verification is not allowed. Why not make this check before creating claims instead?

concent_api/core/payments/bankster.py
cameel: Hmm... That's a problem with our module organization. I don't think that `subtask_helpers` is the right module for a very generic function like `get_one_or_none()`. And `_update_timed_out_subtask()` does not seem to be a "helper" either. It's a critical part of our business logic and introduces a dependency on bankster which I wouldn't really expect in a simple helper.

Anyway, the alternative works so let's leave it as is for now. I only asked because using `first()` is always suspicious in database-related code.

concent_api/core/message_handlers.py

concent_api/core/message_handlers.py

concent_api/core/tasks.py
rwrzesien: :+1: done

concent_api/core/tasks.py
rwrzesien: :+1: done

concent_api/core/subtask_helpers.py
rwrzesien: Yes, I have added it yesterday :)

concent_api/core/message_handlers.py

Issue #899, opened 2018-10-16
Bankster: Blockchain event subscriptions


rwrzesien:
This is a part of the implementation of the "Bankster" blueprint (#479).

Requires #894 to be implemented first.

1. **Implement `remove_claim` function**
2. **Register `remove_claim` function to SCI subscription handler**

##### Sequence of operation
The assumption here is that the event is only reported when a transaction has been successfully included in the blockchain and has enough confirmations to make its reversal very unlikely.

1. **Claim identification**:
    - Bankster begins a database transaction.
    - Bankster checks if `DepositClaim` with `tx_hash` reported in the event exists in the database.
        - If it does not, the handler ends.
        - If it does, the handler locks it (to prevent it from being modified between this step and the next).
2. **Claim freeze**:
    - Bankster locks the `DepositAccount` object associated with the claim.
3. **Claim removal**:
    - Bankster removes the `DepositClaim` object.
4. **Unfreeze**:
    - Bankster commits the database transaction.
        Database locks on `DepositAccount`s are released.

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/161281432

cameel:
@rwrzesien Have you checked this assumption?

> The assumption here is that the event is only reported when a transaction has been successfully included in the blockchain and has enough confirmations to make its reversal very unlikely.

I'm not 100% sure when the event gets reported by SCI so I added this bit to ensure that we don't forget about this detail.

cameel:
@rwrzesien What about my question?

rwrzesien:
@cameel No, I didn't check, I thought it was known when the blueprint was created.

pawelkisielewicz:
There was small problem, it is tested and fixed in https://github.com/golemfactory/concent/pull/967

pawelkisielewicz:
This event is reported when transaction is successful in blockchain. It was tested manually and the assumption is true.

cameel:
OK. I've checked in SCI sources and it does report only after `SCIImplementation.REQUIRED_CONFS` blocks.

BTW, the functions that return lists of transactions do not have this check they always take the first and last block to search. When using those we must remember to subtract that number ourselves.

PR #927, opened 2018-11-08
Feature blockchain event subsciption

pawelkisielewicz:
Resolves https://github.com/golemfactory/concent/issues/899

Please check careful this pull request, the main part is to add callback in `finalize_deposit_claim`. It is not possible to test it local and without @kbeker I don't think we can test it with blockchain.  

There is also added test for function removing deposit_claim. But this function actually checks if object exists in database and calls other function, which is tested. I would remove this (my) test, but please comment

### DIFF
--- a/concent_api/core/payments/bankster.py
+++ b/concent_api/core/payments/bankster.py
@@ -479,7 +479,9 @@ def discard_claim(deposit_claim: DepositClaim) -> bool:
         if deposit_claim.tx_hash is None:
             claim_removed = False
         else:
-            deposit_claim.delete()
-            claim_removed = True
-
+            try:
+                deposit_claim.delete()
+                claim_removed = True
+            except DepositAccount.DoesNotExist:
+                claim_removed = False
     return claim_removed
--- a/concent_api/core/subtask_helpers.py
+++ b/concent_api/core/subtask_helpers.py
@@ -25,6 +25,8 @@ from core.models import DepositClaim
 from core.models import PendingResponse
 from core.models import Subtask
 from core.payments import bankster
+from core.payments.bankster import discard_claim
+from core.payments.payment_interface import PaymentInterface
 from core.transfer_operations import store_pending_message
 from core.transfer_operations import verify_file_status
 from core.utils import hex_to_bytes_convert
@@ -314,7 +316,7 @@ def finalize_deposit_claim(
     concent_use_case: ConcentUseCase,
     ethereum_address: str,
 ) -> None:
-    deposit_claim = get_one_or_none(
+    deposit_claim: DepositClaim = get_one_or_none(  # type: ignore
         DepositClaim,
         subtask_id=subtask_id,
         concent_use_case=concent_use_case,
@@ -322,7 +324,12 @@ def finalize_deposit_claim(
     )
 
     if deposit_claim is not None:
-        bankster.finalize_payment(deposit_claim)
+        tx_hash = bankster.finalize_payment(deposit_claim)
+        payment_interface = PaymentInterface()  # new object is NOT created, it is singleton and its instance is returned
+        payment_interface.on_transaction_confirmed(   # type: ignore  # pylint: disable=no-member
+            tx_hash=tx_hash,
+            cb=lambda: discard_claim(deposit_claim)
+        )
 
 
 def delete_deposit_claim(
--- a/concent_api/core/tests/test_payments_bankster.py
+++ b/concent_api/core/tests/test_payments_bankster.py
@@ -1,3 +1,4 @@
+
 from freezegun import freeze_time
 import mock
 
@@ -401,3 +402,15 @@ class DiscardClaimBanksterTest(ConcentIntegrationTestCase):
 
         self.assertTrue(claim_removed)
         self.assertFalse(DepositClaim.objects.filter(pk=self.deposit_claim.pk).exists())
+
+    def test_that_discard_claim_should_work_correctly_in_parallel(self):
+        self.deposit_claim.tx_hash = 64 * '0'
+        self.deposit_claim.clean()
+        self.deposit_claim.save()
+
+        from api_testing_common import call_function_in_threads
+        call_function_in_threads(
+            func=discard_claim,
+            number_of_threads=2,
+            deposit_claim=self.deposit_claim
+        )


dybi:
@pawelkisielewicz 
> There is also added test for function removing deposit_claim. But this function actually checks if object exists in database and calls other function, which is tested. I would remove this (my) test, but please comment

I do not understand. Could you please clarify what you mean?

concent_api/core/subtask_helpers.py
```diff
@@ -23,6 +23,8 @@
 from core.models import PendingResponse
 from core.models import Subtask
 from core.payments import bankster
+from core.payments.bankster import remove_deposit_claim_if_transaction_successed_in_blockchain
```
Jakub89: You can remove that import and use `bankster.remove_deposit_claim_if_transaction_successed_in_blockchain` in call below, or remove `bankster` import and import all used functions separately, but I don't know if that's possible.

pawelkisielewicz:
There is function:
```python
def remove_deposit_claim_if_transaction_successed_in_blockchain(deposit_claim: DepositClaim) -> None:
    if DepositClaim.objects.filter(pk=deposit_claim.pk).exists():
        discard_claim(deposit_claim)
```
I wrote test to this function, but it is actually duplication of test to function  `def discard_claim`. I don't think test for `remove_deposit_claim_if_transaction_successed_in_blockchain` is necessary. What do you think?

concent_api/core/subtask_helpers.py
```diff
@@ -277,7 +279,7 @@ def finalize_deposit_claim(
     ethereum_address: str,
     public_key: str,
 ) -> None:
-    deposit_claim = get_one_or_none(
+    deposit_claim: DepositClaim = get_one_or_none(  # type: ignore
```
dybi: something's wrong here. `DepositClaim` and `# type: ignore` at the same time?

concent_api/core/subtask_helpers.py
pawelkisielewicz: done, import removed and now function is called:
`bankster.remove_deposit_claim_if_transaction_successed_in_blockchain`

concent_api/core/payments/bankster.py
```diff
@@ -391,3 +391,8 @@ def discard_claim(deposit_claim: DepositClaim) -> bool:
             claim_removed = True
 
     return claim_removed
+
+
+def remove_deposit_claim_if_transaction_successed_in_blockchain(deposit_claim: DepositClaim) -> None:
+    if DepositClaim.objects.filter(pk=deposit_claim.pk).exists():
```
dybi: is there a possibility that this `deposit_claim` doesn't exist at this point?

dybi: BTW:
"successed" -> "succeeded"

concent_api/core/subtask_helpers.py
pawelkisielewicz: see `get_one_or_none` return type. It returns Optional[Subtask]. This function was created to be generic, for different objects, not only subtask. Signature was changed in the meantime (I don't think it was good change), but function still works for different objects. 

In this case `DepositClaim` is passed as Model, so its instance is return type. I added `: DepositClaim` to use real/correct type, but ignoring type was also necessary. 

BTW.  typing in `get_one_or_none` should be fixed. What do you think?

concent_api/core/payments/bankster.py
pawelkisielewicz: good idea :+1: . It is checked by function calling, so there is no need to check it again

pawelkisielewicz:
@dybi @rwrzesien can you please check if part with creating callback with passing function using lambda is OK. This is the most important part here!

concent_api/core/payments/bankster.py
```diff
@@ -145,7 +145,7 @@ def claim_deposit(
             # If the total of existing claims and the current claim is greater or equal to the current deposit,
             # we can't add a new claim.
             if provider_deposit <= sum_of_existing_provider_claims['sum_of_existing_claims'] + settings.ADDITIONAL_VERIFICATION_COST:
-                return (requestor_deposit, None)
+                return (None, None)
```
cameel: This looks like a bugfix for something that was broken before this pull request. It should be a separate commit.

concent_api/core/payments/bankster.py
```diff
@@ -239,6 +239,7 @@ def finalize_payment(deposit_claim: DepositClaim) -> Optional[str]:
 
         # Bankster puts transaction ID in DepositClaim.tx_hash.
         deposit_claim.tx_hash = encode_hex(ethereum_transaction.hash)
+        deposit_claim.full_clean()
```
cameel: Also a bugfix. Separate commit.

concent_api/core/subtask_helpers.py
cameel: `get_one_or_none` definitely should get fixed. It was meant to work with any model.

concent_api/core/subtask_helpers.py
```diff
@@ -311,7 +313,12 @@ def finalize_deposit_claim(
     )
 
     if deposit_claim:
```
cameel: It's a model, not a boolean. It's not a change from your pull request but the fix is trivial so please change it to `if deposit_claim is not None:`.

concent_api/core/subtask_helpers.py
```diff
@@ -311,7 +313,12 @@ def finalize_deposit_claim(
...
-        bankster.finalize_payment(deposit_claim)
+        tx_hash = bankster.finalize_payment(deposit_claim)
+        payment_interface = PaymentInterface()  # new object is NOT created, it is singleton and its instance is returned
+        payment_interface.on_transaction_confirmed(   # type: ignore  # pylint: disable=no-member
+            tx_hash=tx_hash,
+            cb=lambda: discard_claim(deposit_claim)
+        )
```
cameel: Please make sure that `discard_claim()` does not fail if two processes try to do it at the same time (i.e. first gets a lock, seconds waits on the lock, first runs `delete()` and releases the lock, second runs `delete()` on the same, already deleted object). I think it won't fail on the SQL level (the `WHERE` clause in the second `DELETE` will simply match no rows) but I'm not sure how Django will behave.

concent_api/core/payments/bankster.py
rwrzesien: I think it is now seen because branch is not rebased. It must `return (requestor_deposit, None)`

cameel: But why? It's supposed to return a claim instance. `requestor_deposit` is a number.

concent_api/core/subtask_helpers.py

Issue #898, opened 2018-10-16
Bankster: `settle overdue acceptances` operation changes


rwrzesien:
This is a part of the implementation of the "Bankster" blueprint (#479).

Requires #894 and https://github.com/golemfactory/concent/issues/565 to be implemented first.

There should be some updates in the current implementation of `settle overdue acceptances` operation in Bankster to allow it to follow deposit state.

##### Input - add following parameters:
| Parameter                       | Type                        | Optional | Remarks                                                                                                                 |
|---------------------------------|-----------------------------|----------|-------------------------------------------------------------------------------------------------------------------------|
| `requestor_public_key`          | string                      | no       | Public key of the requestor. Comes from `TaskToCompute`.
| `provider_public_key`           | string                      | no       | Public key of the provider. Comes from `TaskToCompute`.

##### Sequence of operation - modify current sequence:
1. **Initialization**:
    - Bankster creates `Client` and `DepositAccount` objects for the requestor if they don't exist yet.
        - The transaction is committed at this point to prevent these objects from being rolled back in case of failure.
2. **Deposit query** (This should be already done, but please review and confirm):
    - Bankster asks SCI about the amount of funds available in requestor's deposit.
3. **Claim freeze**:
    - Bankster begins a database transaction.
    - Bankster puts a database lock on the `DepositAccount` object.
4. **Requestor's spare deposit check** (This is partially done, so please apply only necessary updates):
    - Bankster sums the `amount`s of all existing `DepositClaim`s where the requestor is the payer.
    - If the existing claims against requestor's deposit are greater or equal to his current deposit, we can't add a new claim.
        - Bankster responds with `claim_against_requestor` being `None`.
    - Otherwise Bankster continues.
5. **Claim calculation** (This should be already done, but please review and confirm):
    - Bankster sums up the amount specified in acceptances.
    - Bankster asks SCI for the list of all relevant payments listed on the blockchain (both normal payments performed by the requestor and forced payments performed by Concent).
    - Bankster computes the amount that is still owed to the provider.
    - Bankster compares the amount with the available deposit minus the existing claims against requestor's account.
        - If the whole amount can't be paid, Concent lowers it to pay as much as possible.
        - The final amount must be non-zero - spare deposit check prevents this.
6. **Transaction** (This should be already done, but please review and confirm):
    - Bankster uses SCI to create an Ethereum transaction.
7. **Deposit lock**:
    - Bankster creates a `DepositClaim` object
[...]

PR #929, opened 2018-11-08
Feature bankster settle overdue acceptances operation changes

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/898

### DIFF
--- a/concent_api/core/constants.py
+++ b/concent_api/core/constants.py
@@ -83,3 +83,7 @@ class VerificationResult(IntEnum):
     MATCH       = 0
     MISMATCH    = 1
     ERROR       = 2
+
+
+# Defines a string that is returned from Bankster to Core views to indicate timestamp error case.
+BANKSTER_TIMESTAMP_ERROR = 'timestamp_error'
--- a/concent_api/core/message_handlers.py
+++ b/concent_api/core/message_handlers.py
@@ -41,6 +41,7 @@ from common.logging import log
 from common.validations import validate_secure_hash_algorithm
 from common import logging
 from conductor.tasks import result_transfer_request
+from core.exceptions import BanksterTimestampError
 from core.exceptions import CreateModelIntegrityError
 from core.exceptions import Http400
 from core.exceptions import TooSmallProviderDeposit
@@ -757,6 +758,16 @@ def handle_send_force_payment(
             reason = message.concents.ServiceRefused.REASON.DuplicateRequest
         )
 
+    for subtask_results_accepted in client_message.subtask_results_accepted_list:
+        validate_task_to_compute(subtask_results_accepted.task_to_compute)
+
+    if sum([subtask_results_accepted.task_to_compute.price
+            for subtask_results_accepted in client_message.subtask_results_accepted_list]) == 0:
+        return message.concents.ForcePaymentRejected(
+            force_payment=client_message,
+            reason=message.concents.ForcePaymentRejected.REASON.NoUnsettledTasksFound,
+        )
+
     task_to_compute = client_message.subtask_results_accepted_list[0].task_to_compute
     requestor_public_key = hex_to_bytes_convert(task_to_compute.requestor_public_key)
     (requestor_eth_address, provider_eth_address) = get_clients_eth_accounts(task_to_compute)
@@ -769,13 +780,14 @@ def handle_send_force_payment(
         *[subtask_results_accepted.task_to_compute for subtask_results_accepted in client_message.subtask_results_accepted_list],
     )
 
-    requestors_claim_payment_info = bankster.settle_overdue_acceptances(
-        requestor_ethereum_address=requestor_eth_address,
-        provider_ethereum_address=provider_eth_address,
-        acceptances=client_message.subtask_results_accepted_list,
-    )
-
-    if requestors_claim_payment_info.timestamp_error:
+    try:
+        claim_against_requestor = bankster.settle_overdue_acceptances(
+            requestor_ethereum_address=requestor_eth_address,
+            provider_ethereum_address=provider_eth_address,
+            acceptances=client_message.subtask_results_accepted_list,
+            requestor_public_key=requestor_public_key,
+        )
+    except BanksterTimestampError:
         return message.concents.ForcePaymentRejected(
             force_payment=client_message,
             reason=message.concents.ForcePaymentRejected.REASON.TimestampError,
@@ -787,7 +799,7 @@ def handle_send_force_payment(
         subtask_results_accepted.payment_ts for subtask_results_accepted in client_message.subtask_results_accepted_list
     )
 
-    if requestors_claim_payment_info.amount_pending == requestors_claim_payment_info.amount_paid == 0:
+    if claim_against_requestor is None:
         return message.concents.ForcePaymentRejected(
             force_payment=client_message,
             reason=message.concents.ForcePaymentRejected.REASON.NoUnsettledTasksFound,
@@ -797,8 +809,8 @@ def handle_send_force_payment(
             payment_ts              = payment_ts,
             task_owner_key          = requestor_ethereum_public_key,
             provider_eth_account    = provider_eth_address,
-            amount_paid             = requestors_claim_payment_info.amount_paid,
-            amount_pending          = requestors_claim_payment_info.amount_pending,
+            amount_paid             = claim_against_requestor.amount,
+            amount_pending          = 0,
             recipient_type          = message.concents.ForcePaymentCommitted.Actor.Provider,
         )
 
@@ -806,17 +818,18 @@ def handle_send_force_payment(
             payment_ts              = payment_ts,
             task_owner_key          = requestor_ethereum_public_key,
             provider_eth_account    = provider_eth_address,
-            amount_paid             = requestors_claim_payment_info.amount_paid,
-            amount_pending          = requestors_claim_payment_info.amount_pending,
+            amount_paid             = claim_against_requestor.amount,
+            amount_pending          = 0,
             recipient_type          = message.concents.ForcePaymentCommitted.Actor.Requestor,
         )
 
-        store_pending_message(
-            response_type=PendingResponse.ResponseType.ForcePaymentCommitted,
-            client_public_key=requestor_public_key,
-            queue=PendingResponse.Queue.ReceiveOutOfBand,
-            payment_message=requestor_force_payment_commited
-        )
+        with transaction.atomic(using='control'):
+            store_pending_message(
+                response_type=PendingResponse.ResponseType.ForcePaymentCommitted,
+                client_public_key=requestor_public_key,
+                queue=PendingResponse.Queue.ReceiveOutOfBand,
+                payment_message=requestor_force_payment_commited
+            )
 
         provider_force_payment_commited.sig = None
         return provider_force_payment_commited
--- a/concent_api/core/payments/bankster.py
+++ b/concent_api/core/payments/bankster.py
@@ -2,12 +2,13 @@ from typing import List
 from typing import Optional
 from typing import Tuple
 from typing import Union
+import logging
 
 from django.conf import settings
+from django.db import transaction
+from django.db import IntegrityError
 from django.db.models import Sum
 from django.db.models.functions import Coalesce
-from django.db import IntegrityError
-from django.db import transaction
 
 from golem_messages.message.tasks import SubtaskResultsAccepted
 from golem_messages.utils import encode_hex
@@ -19,6 +20,7 @@ from common.constants import ConcentUseCase
 from common.helpers import ethereum_public_key_to_address
 from common.helpers import get_current_utc_timestamp
 from core.constants import ETHEREUM_ADDRESS_LENGTH
+from core.exceptions import BanksterTimestampError
 from core.exceptions import TooSmallProviderDeposit
 from core.models import Client
 from core.models import DepositAccount
@@ -30,35 +32,7 @@ from core.validation import validate_bytes_public_key
 from core.validation import validate_uuid
 
 
-class ClaimPaymentInfo:
-
-    __slots__ = [
-        'amount_paid',
-        'amount_pending',
-        'tx_hash',
-        'payment_ts',
-        'timestamp_error',
-    ]
-
-    def __init__(
-        self,
-        amount_paid: int,
-        amount_pending: int,
-        tx_hash: Optional[str] = None,
-        payment_ts: Optional[int] = None,
-        timestamp_error: bool = False,
-    ) -> None:
-        assert isinstance(amount_paid, int) and amount_paid >= 0
-        assert isinstance(amount_pending, int) and amount_paid >= 0
-        assert isinstance(tx_hash, str) or (tx_hash is None and amount_paid == 0)
-        assert isinstance(payment_ts, int) or (payment_ts is None and tx_hash is None)
-        assert timestamp_error is False or (timestamp_error is True and amount_paid == amount_pending == 0)
-
-        self.amount_paid = amount_paid
-        self.amount_pending = amount_pending
-        self.tx_hash = tx_hash
-        self.payment_ts = payment_ts
-        self.timestamp_error = timestamp_error
+logger = logging.getLogger(__name__)
 
 
 def claim_deposit(
@@ -300,7 +274,8 @@ def settle_overdue_acceptances(
     requestor_ethereum_address: str,
     provider_ethereum_address: str,
     acceptances: List[SubtaskResultsAccepted],
-) -> ClaimPaymentInfo:
+    requestor_public_key: bytes,
+) -> Optional[DepositClaim]:
     """
     The purpose of this operation is to calculate the total amount that the requestor owes provider for completed
     computations and transfer that amount from requestor's deposit.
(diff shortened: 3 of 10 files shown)


concent_api/core/payments/bankster.py
```diff
@@ -264,80 +235,115 @@ def settle_overdue_acceptances(
...
-    # matches condition current_time < payment_ts + PAYMENT_DUE_TIME
-    acceptance_time_overdue = any(
-        current_time < subtask_results_accepted.payment_ts + settings.PAYMENT_DUE_TIME
-        for subtask_results_accepted in acceptances
-    )
+        # Concent defines time T1 equal to youngest timestamp from list of transactions.
+        if len(list_of_transactions) == 0:
+            t1_is_bigger_than_payments_ts = False
+        else:
+            youngest_transaction = max(transaction.closure_time for transaction in list_of_transactions)
```
pawelkisielewicz: please rename variable `transaction`. There is imported `transaction` from `django.db` and my IDE shows warning, that this variable shadows imported one. I know it works, but I think it should be renamed. BTW. Does your IDE shows warning in this place as well?

concent_api/core/message_handlers.py
```diff
@@ -781,7 +792,7 @@ def handle_send_force_payment(
         subtask_results_accepted.payment_ts for subtask_results_accepted in client_message.subtask_results_accepted_list
     )
 
-    if requestors_claim_payment_info.amount_pending == requestors_claim_payment_info.amount_paid == 0:
+    if claim_against_requestor is None:
```
pawelkisielewicz: if -> elif ?

and in last else maybe assert, that `claim_against_requestor` is `DepositClaim` instance. I know it is not necessary now, but if something will change.. Such a sanity check. If you don't want to do this it is OK for me, just asking what do you think :)

concent_api/core/payments/bankster.py
```diff
@@ -264,80 +235,115 @@ def settle_overdue_acceptances(
...
-    list_of_transactions = service.get_list_of_payments(  # pylint: disable=no-value-for-parameter
-        requestor_eth_address=requestor_ethereum_address,
-        provider_eth_address=provider_ethereum_address,
-        payment_ts=oldest_payments_ts,
-        current_time=current_time,
-        transaction_type=TransactionType.BATCH,
-    )
+        # Concent defines time T0 equal to oldest payment_ts from passed SubtaskResultAccepted messages from
+        # subtask_results_accepted_list.
+        oldest_payments_ts = min(subtask_results_accepted.payment_ts for subtask_results_accepted in acceptances)
```
pawelkisielewicz: :+1: I like to see "senior code", if I did the same I would do it in coupe more lines and not so nice :)

concent_api/core/message_handlers.py
```diff
@@ -751,6 +752,15 @@ def handle_send_force_payment(
             reason = message.concents.ServiceRefused.REASON.DuplicateRequest
         )
 
+    for subtask_results_accepted in client_message.subtask_results_accepted_list:
+        try:
+            validate_task_to_compute(subtask_results_accepted.task_to_compute)
```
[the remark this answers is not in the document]
cameel: @pawelkisielewicz All `subtask_results_accepted.task_to_compute` must go through full validation here. They come from the client and the field is specific to this particular use case so we're definitely not validating it in decorators before this point.

It seems to me that `NonPositivePriceTaskToComputeError` is being caught here because all prices being zero is a special case that's not an error. Instead we must tell the client that there's nothing to pay.

Any other validation errors are handled by `@common.decorators.require_golem_message`.

@rwrzesien But I agree with @pawelkisielewicz  that using this validation to detect the case of price being zero is not a good idea. It's too broad. I can see two bugs here:
- The handler returns `NoUnsettledTasksFound` when the price of only some tasks is zero. E.g. if prices are `[100, 200, 0, 300]`. We should proceed with the settlement rather than return in such a case.
- The handler returns `NoUnsettledTasksFound` when there are negative prices. E.g. if prices are `[100, 200, 5, -300]`. It should be a validation error.

concent_api/core/payments/bankster.py
rwrzesien: It is actually @kbeker code so good for him :)

concent_api/core/message_handlers.py
rwrzesien: I don't know what should be put in `elif` as condition. Actually it should be only `if`, without `else` because if `if` condition is not met, then we should simply continue the flow. But it is not my part of the code, and I think we have somewhere in coding style to use `else` even if it is not necessary.

concent_api/core/payments/bankster.py
rwrzesien: :+1: yes, it was showing it too.

concent_api/core/message_handlers.py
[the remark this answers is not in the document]
rwrzesien: >  I assume that if flow came to this point task_to_compute was previous validated and everything is OK, but I'm not sure.

No, it wasn't, that is why I have added this part.

> If it was not validated and there is risk that validation may fail because of other reason add handling of this exception

It is already handled in decorators, like for the others message handlers flow. So what I am doing here is 
1. Validating all nested `task_to_compute`.
2a. If `price==0` error is found, handle it in special way (which is done in a way to keep the same API as before those changes)
2b. If any other problem is found, handle it in regular way (through decorator).

> the validation itself should allow price == 0 because a zero price is not a validation error 

We discussed this with @shadeofblue and agreed that `task_to_compute` with `price==0` is wrong from Concent point of view and should be always refused, this validation is now in every use case.

concent_api/core/message_handlers.py
cameel: @rwrzesien You probably mean the discussion about `return` and this case in particular: https://github.com/Code-Poets/coding-style/issues/6#issuecomment-396263483.

That's not the case we have here. Your `if`s are not nested.

I don't have a preference here For me either version is fine.

concent_api/core/message_handlers.py
```diff
@@ -763,13 +773,14 @@ def handle_send_force_payment(
...
-    requestors_claim_payment_info = bankster.settle_overdue_acceptances(
+    claim_against_requestor = bankster.settle_overdue_acceptances(
         requestor_ethereum_address=requestor_eth_address,
         provider_ethereum_address=provider_eth_address,
         acceptances=client_message.subtask_results_accepted_list,
+        requestor_public_key=requestor_public_key,
     )
 
-    if requestors_claim_payment_info.timestamp_error:
+    if claim_against_requestor == 'timestamp_error':
```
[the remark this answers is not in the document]
rwrzesien: Why enum? Would not constant be enough?

concent_api/core/payments/bankster.py
```diff
@@ -264,80 +235,115 @@ def settle_overdue_acceptances(
...
+    with transaction.atomic(using='control'):
+        DepositAccount.objects.select_for_update().get(
+            pk=requestor_deposit_account.pk
+        )
+
+        # Bankster sums the amounts of all existing DepositClaims that have the same payer as the one being processed.
+        sum_of_existing_requestor_claims = DepositClaim.objects.filter(
+            payer_deposit_account=requestor_deposit_account
+        ).aggregate(
+            sum_of_existing_claims=Coalesce(Sum('amount'), 0)
```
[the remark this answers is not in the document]

concent_api/core/payments/bankster.py
```diff
@@ -264,80 +235,115 @@ def settle_overdue_acceptances(
...
+            list_of_payments=list_of_transactions,
+            subtask_results_accepted_list=acceptances,
+        )
+
+        # Bankster compares the amount with the available deposit minus the existing claims against requestor's account.
+        # If the whole amount can't be paid, Concent lowers it to pay as much as possible.
+        requestor_payable_amount = min(
+            amount_pending,
+            requestor_deposit_value - sum_of_existing_requestor_claims['sum_of_existing_claims'],
+        )
```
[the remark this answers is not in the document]

concent_api/core/payments/bankster.py
```diff
@@ -264,80 +235,115 @@ def settle_overdue_acceptances(
     assert len(provider_ethereum_address) == ETHEREUM_ADDRESS_LENGTH
     assert provider_ethereum_address != requestor_ethereum_address
 
+    with transaction.atomic(using='control'):
+        requestor_client = Client.objects.get_or_create_full_clean(
+            public_key=requestor_public_key,
+        )
```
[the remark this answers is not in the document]
rwrzesien: If it is already present in the database, it should be returned without error.

concent_api/core/message_handlers.py
```diff
@@ -791,17 +802,17 @@ def handle_send_force_payment(
...
         requestor_force_payment_commited = message.concents.ForcePaymentCommitted(
             payment_ts              = payment_ts,
             task_owner_key          = requestor_ethereum_public_key,
             provider_eth_account    = provider_eth_address,
-            amount_paid             = requestors_claim_payment_info.amount_paid,
-            amount_pending          = requestors_claim_payment_info.amount_pending,
+            amount_paid             = claim_against_requestor.amount,  # type: ignore
+            amount_pending          = 0,
             recipient_type          = message.concents.ForcePaymentCommitted.Actor.Requestor,
         )
```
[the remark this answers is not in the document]

concent_api/core/payments/bankster.py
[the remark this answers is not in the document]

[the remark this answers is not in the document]

concent_api/core/payments/bankster.py
```diff
@@ -264,80 +235,115 @@ def settle_overdue_acceptances(
...
-    # Concent defines time T1 equal to youngest timestamp from list of transactions.
-    if len(list_of_transactions) == 0:
-        t1_is_bigger_than_payments_ts = False
-    else:
-        youngest_transaction = max(transaction.closure_time for transaction in list_of_transactions)
-        # Concent checks if all passed SubtaskResultAccepted messages from subtask_results_accepted_list
-        # have payment_ts < T1
-        t1_is_bigger_than_payments_ts = any(
-            youngest_transaction > subtask_results_accepted.payment_ts for subtask_results_accepted in acceptances
+        current_time = get_current_utc_timestamp()
```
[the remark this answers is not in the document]

concent_api/core/payments/bankster.py
```diff
@@ -264,80 +235,115 @@ def settle_overdue_acceptances(
...
-        list_of_forced_payments=list_of_forced_payments,
-        list_of_payments=list_of_transactions,
-        subtask_results_accepted_list=acceptances,
-    )
+        # Concent gets list of forced payments from payment API where T0 <= payment_ts + PAYMENT_DUE_TIME.
+        list_of_forced_payments = service.get_list_of_payments(  # pylint: disable=no-value-for-parameter
+            requestor_eth_address=requestor_ethereum_address,
+            provider_eth_address=provider_ethereum_address,
+            payment_ts=oldest_payments_ts + settings.PAYMENT_DUE_TIME,  # Im not sure, check it please
+            current_time=current_time,
```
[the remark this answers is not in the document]
cameel: It might have been based on the fact that we don't don't let provider start the procedure earlier than `payments_ts + settings.PAYMENT_DUE_TIME`. But it's not correct as T0. Payment after T0 is fine. But a payment after T0 + `settings.PAYMENT_DUE_TIME` is late. Using this interval would make Concent not see any payment that was made on time.

concent_api/core/message_handlers.py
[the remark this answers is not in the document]
cameel: A constant is fine too.

But this looks like a flag that indicates a specific condition so an enum seems natural to me. Just like for error codes.

concent_api/core/payments/bankster.py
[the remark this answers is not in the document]
cameel: Not if it gets created between the `get` and the `create`. It's the race condition you always get when doing a non-atomic get or create. Same problem as in https://github.com/golemfactory/concent/issues/486

concent_api/core/message_handlers.py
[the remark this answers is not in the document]
cameel: Is it wrong from Concent's point of view or wrong in general? For me a validation error means that the message is malformed. It can't even be processed and should be immediately rejected. Message is either valid or not. It should not be valid for one client and invalid for another one.

If it's wrong only for Concent then in my opinion it's business logic and it should be checked separately.

> 2a. If price==0 error is found, handle it in special way (which is done in a way to keep the same API as before those changes)

Where does this special case come from?

concent_api/core/message_handlers.py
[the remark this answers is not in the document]

concent_api/core/message_handlers.py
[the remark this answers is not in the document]
rwrzesien: In previous code there was a case when total price of summed tasks was 0, then `ForcePaymentRejected` with reason `NoUnsettledTasksFound` was returned. I thought it would be good to extend it to catch any task with 0 price.

concent_api/core/payments/bankster.py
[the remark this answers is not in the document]
rwrzesien: I don't understand how it should be repeated then? This method should be called again in `except` block?

concent_api/core/payments/bankster.py
```diff
@@ -264,80 +235,115 @@ def settle_overdue_acceptances(
...
+        )
+
+        # Bankster compares the amount with the available deposit minus the existing claims against requestor's account.
+        # If the whole amount can't be paid, Concent lowers it to pay as much as possible.
+        requestor_payable_amount = min(
+            amount_pending,
+            requestor_deposit_value - sum_of_existing_requestor_claims['sum_of_existing_claims'],
+        )
+
+        assert requestor_payable_amount > 0
```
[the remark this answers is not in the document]

concent_api/core/payments/bankster.py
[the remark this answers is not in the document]
cameel: The block should be in a context manager. If an exception with the right postgresql error code is raised, the manager should run the block again.

Or you could put it in a local function, run it in a `try ... except` block and run it again if you get the right exception.

The idea is simply that if you try to create it and the creation fails because the object exists you should try to fetch it instead. The current code tries to fetch first and only then creates and this introduces a race condition. Note that the create may not actually fail and you may get an error when the transaction is committed instead. Retrying once fixes the problem but we don't want to rerun the whole view. Just the part that creates the objects. It's what @pawelkisielewicz has been fixing with regard to subtasks lately. Since that's done, I'd prefer to avoid introducing this race condition in any new code.

concent_api/core/message_handlers.py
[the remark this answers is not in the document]
rwrzesien: Done as a separate validation, but this part of code will never be reached .

concent_api/core/payments/bankster.py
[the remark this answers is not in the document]

concent_api/core/message_handlers.py
```diff
@@ -791,26 +803,27 @@ def handle_send_force_payment(
...
-        )
+        with transaction.atomic(using='control'):
+            store_pending_message(
+                response_type=PendingResponse.ResponseType.ForcePaymentCommitted,
+                client_public_key=requestor_public_key,
+                queue=PendingResponse.Queue.ReceiveOutOfBand,
+                payment_message=requestor_force_payment_commited
+            )
 
         provider_force_payment_commited.sig = None
```
dybi: Is there any reason we are doing this?

concent_api/core/payments/bankster.py
```diff
@@ -265,80 +241,135 @@ def settle_overdue_acceptances(
...
-        t1_is_bigger_than_payments_ts = False
-    else:
-        youngest_transaction = max(transaction.closure_time for transaction in list_of_transactions)
-        # Concent checks if all passed SubtaskResultAccepted messages from subtask_results_accepted_list
-        # have payment_ts < T1
-        t1_is_bigger_than_payments_ts = any(
-            youngest_transaction > subtask_results_accepted.payment_ts for subtask_results_accepted in acceptances
+        # If the existing claims against requestor's deposit are greater or equal to his current deposit,
+        # we can't add a new claim.
+        if requestor_deposit_value <= sum_of_existing_requestor_claims['sum_of_existing_claims']:
```
dybi: I would suggest changing the condition, so it sounds exactly like your comment above:
```suggestion
        if  sum_of_existing_requestor_claims['sum_of_existing_claims'] >=:
```

concent_api/core/payments/bankster.py
```diff
@@ -249,7 +224,8 @@ def settle_overdue_acceptances(
     requestor_ethereum_address: str,
     provider_ethereum_address: str,
     acceptances: List[SubtaskResultsAccepted],
-) -> ClaimPaymentInfo:
+    requestor_public_key: bytes,
+) -> Union[DepositClaim, str, None]:
```
dybi: I have quite mixed feelings about this. I mean, we are returning 3 quite different types of objects here. My biggest concern is returning `DepositClaim` (normal operation) or `str` to indicate a `timestamp_error`. The last part seems to be a bussiness logic that should be handle in `message_handlers`. Maybe it would be better to raise some exception and handle it in the caller?

concent_api/core/tests/test_integration_force_payment.py
```diff
@@ -381,8 +379,8 @@ def test_provider_send_correct_force_payment_concent_should_accept(self):
             fields       = {
                 'recipient_type': message.concents.ForcePaymentCommitted.Actor.Provider,
                 'timestamp': parse_iso_date_to_timestamp("2018-02-05 12:00:20"),
-                'amount_pending': amount_pending,
-                'amount_paid':    amount_paid,
+                'amount_pending': 0,
```
dybi: I would suggest extracting all those `0` and `10` to test class fields, i.e. `self.amount_pending` and `self.amount_paid`

concent_api/core/tests/test_payments_bankster.py
```diff
@@ -241,7 +241,7 @@ def test_that_when_there_are_other_deposit_claims_finalize_payment_substract_the
 
 class SettleOverdueAcceptancesBanksterTest(ConcentIntegrationTestCase):
 
-    def test_that_settle_overdue_acceptances_should_return_empty_claim_deposit_info_if_subtask_costs_where_already_paid(self):
+    def test_that_settle_overdue_acceptances_should_return_empty_claim_deposit_if_subtask_costs_where_already_paid(self):
```
dybi: it doesn't return empty claim deposit, but a string (with error message)

concent_api/core/tests/test_payments_bankster.py
```diff
@@ -259,23 +259,22 @@ def test_that_settle_overdue_acceptances_should_return_empty_claim_deposit_info_
...
+                requestor_public_key=hex_to_bytes_convert(task_to_compute.requestor_public_key),
             )
 
         get_list_of_payments_mock.assert_called()
 
-        self.assertIsNone(requestors_claim_payment_info.tx_hash)
-        self.assertIsNone(requestors_claim_payment_info.payment_ts)
-        self.assertEqual(requestors_claim_payment_info.amount_paid, 0)
-        self.assertEqual(requestors_claim_payment_info.amount_pending, 0)
+        self.assertEqual(
```
dybi: it looks a bit strange... Once again I would propose having an exception in this case (instead of returning string).
@cameel, what do you think ?

concent_api/core/tests/test_payments_bankster.py
```diff
@@ -259,23 +259,22 @@ def test_that_settle_overdue_acceptances_should_return_empty_claim_deposit_info_
...
+            claim_against_requestor, 'timestamp_error')
 
     @override_settings(
         PAYMENT_DUE_TIME=10
     )
-    def test_that_settle_overdue_acceptances_should_return_claim_deposit_info_with_amount_pending_if_requestor_deposit_value_is_zero(self):
+    def test_that_settle_overdue_acceptances_should_return_claim_deposit_with_amount_pending_if_requestor_deposit_value_is_zero(self):
```
dybi: you are not checking the `amount` field of the returned `DepositClaim` object, so either test name is wrong or some assertions are missing

concent_api/core/tests/utils.py
```diff
@@ -1080,6 +1082,28 @@ def claim_deposit_false_mock(
...
+        requestor_deposit_account = DepositAccount(
+            ethereum_address=requestor_ethereum_address,
+            client=requestor_client,
+        )
+        requestor_deposit_account.full_clean()
+        requestor_deposit_account.save()
+
+        claim_against_requestor = DepositClaim(
+            payee_ethereum_address=provider_ethereum_address,
+            amount=10,
```
dybi: could the amount be an (optional) param? Or maybe it could come from a const?

concent_api/core/payments/bankster.py
cameel: Yeah. I've already suggested an enum (https://github.com/golemfactory/concent/pull/929#discussion_r232216556). which should be an improvement over returning a bare string but but you're right, the `"timestamp_error"` does feel like something that should be an exception.

concent_api/core/tests/test_payments_bankster.py
cameel: I agree. Answered in https://github.com/golemfactory/concent/pull/929/files#r232952040

cameel: Also, looks like @rwrzesien has made it a constant after my comment (`BANKSTER_TIMESTAMP_ERROR`) but it looks like there are still a few instances where the bare string is used instead of that constant. This is one of these instances.

concent_api/core/message_handlers.py
cameel: I can't see any. We sometimes clear (or used to clear) the signature to have it recalculated. If you change something in the message and leave the signature as is, golem-messages won't update it when the message is serialized. But that's not the case here. We're creating a new instance of `ForcePaymentCommitted` so it should not be necessary.

concent_api/core/constants.py
```diff
@@ -83,3 +83,6 @@ class VerificationResult(IntEnum):
     MATCH       = 0
     MISMATCH    = 1
     ERROR       = 2
+
+# Defines a string that is returned from Bankster to Core views to indicate timestamp error case.
+BANKSTER_TIMESTAMP_ERROR = 'timestamp_error'
```
Jakub89: `./concent_api/core/constants.py:88:1: E305 expected 2 blank lines after class or function definition, found 1`

concent_api/core/tests/utils.py
rwrzesien: > could the amount be an (optional) param? Or maybe it could come from a const?

It could come from acceptances list, as it is calculated in bankster, but there is no need for this now.

concent_api/core/constants.py
rwrzesien: :+1: done

concent_api/core/payments/bankster.py

concent_api/core/tests/test_payments_bankster.py

concent_api/core/tests/test_payments_bankster.py

concent_api/core/tests/test_payments_bankster.py

concent_api/core/tests/test_integration_force_payment.py
rwrzesien: :+1: done

concent_api/core/payments/bankster.py
dybi: @rwrzesien, where was that rule? ;)

dybi: @rwrzesien , what I found is: https://github.com/golemfactory/concent/pull/929#discussion_r232908597 and it is contrary to what you have stated above: 
`if value > 0` is used, rather than `if 0 < value`

concent_api/core/payments/bankster.py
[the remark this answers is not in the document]
cameel: Looks OK now. I think you should remove the `# Im not sure, check it please` comment now.

concent_api/core/payments/bankster.py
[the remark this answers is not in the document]
cameel: Since there are two separate "get or create" operations and each one should be in a block of its own. So you need two blocks.

Also, I think catching just `IntegrityError` is too broad. You should check the error code from postgresql. See https://github.com/golemfactory/concent/issues/759#issuecomment-417295320

BTW, just duplicating it is fine for now but I think we should have a context manager for this because this case will come up very often (I've already found it in the other pull request).

concent_api/core/payments/bankster.py
```diff
@@ -225,7 +225,7 @@ def settle_overdue_acceptances(
     provider_ethereum_address: str,
     acceptances: List[SubtaskResultsAccepted],
     requestor_public_key: bytes,
-) -> Union[DepositClaim, str, None]:
+) -> Union[DepositClaim, None]:
```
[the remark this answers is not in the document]

concent_api/core/payments/bankster.py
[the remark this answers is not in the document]

concent_api/core/message_handlers.py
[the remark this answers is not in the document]
cameel: Yes, and I what I'm saying does not contradict that. I'm simply saying that the function that validates `TaskToCompute` is not the right place to refuse it because a message with price=0 is not malformed. It's rejected for a different reason.

Issue #887, opened 2018-10-15
Error in Golem Message


jiivan:
**Symptom**:
```python
[...]

PR #932, opened 2018-11-13
Bugfix error in golem message more descriptive

pawelkisielewicz:
Resolves https://github.com/golemfactory/concent/issues/887

### DIFF
--- a/concent_api/core/decorators.py
+++ b/concent_api/core/decorators.py
@@ -32,6 +32,7 @@ from common.shortcuts import load_without_public_key
 from core.exceptions import CreateModelIntegrityError
 from core.exceptions import UnsupportedProtocolVersion
 from core.exceptions import NonPositivePriceTaskToComputeError
+from core.utils import generate_uuid
 from core.utils import is_given_golem_messages_version_supported_by_concent
 from core.validation import get_validated_client_public_key_from_client_message
 from core.validation import is_golem_message_signed_with_key
@@ -122,8 +123,9 @@ def require_golem_auth_message(view: Callable) -> Callable:
                 log(logger, 'Error:', exception.__class__.__name__)
                 return JsonResponse({'error': f'{exception}'}, status = 400)
             except MessageError as exception:
-                log(logger, ERROR_IN_GOLEM_MESSAGE, exception.__class__.__name__)
-                return JsonResponse({'error': join_messages(ERROR_IN_GOLEM_MESSAGE, str(exception))}, status = 400)
+                request_id = generate_uuid()
+                log(logger, f'uuid: {request_id}', ERROR_IN_GOLEM_MESSAGE, exception.__class__.__name__, logging_level=logging.LoggingLevel.EXCEPTION)
+                return JsonResponse({'error': join_messages(f'uuid: {request_id}', ERROR_IN_GOLEM_MESSAGE, exception.__class__.__name__)}, status=400)
         else:
             log(logger, 'error: Concent supports only application/octet-stream.')
             return JsonResponse({'error': "Concent supports only application/octet-stream."}, status = 415)
@@ -184,8 +186,9 @@ def require_golem_message(view: Callable) -> Callable:
                 log(logger, 'Error:', exception.__class__.__name__)
                 return JsonResponse({'error': f'{exception}'}, status = 400)
             except MessageError as exception:
-                log(logger, ERROR_IN_GOLEM_MESSAGE, exception.__class__.__name__)
-                return JsonResponse({'error': join_messages(ERROR_IN_GOLEM_MESSAGE, str(exception))}, status = 400)
+                request_id = generate_uuid()
+                log(logger, f'uuid: {request_id}', ERROR_IN_GOLEM_MESSAGE, exception.__class__.__name__, logging_level=logging.LoggingLevel.EXCEPTION)
+                return JsonResponse({'error': join_messages(f'uuid: {request_id}', ERROR_IN_GOLEM_MESSAGE, exception.__class__.__name__)}, status=400)
         else:
             log(logger, 'error: Concent supports only application/octet-stream.')
             return JsonResponse({'error': "Concent supports only application/octet-stream."}, status = 415)
--- a/concent_api/core/tests/test_core_views.py
+++ b/concent_api/core/tests/test_core_views.py
@@ -203,7 +203,7 @@ class CoreViewSendTest(ConcentIntegrationTestCase):
                 CONCENT_PUBLIC_KEY),
         )
         self._test_400_response(response)
-        self.assertTrue(response.json()['error'].startswith('Error in Golem Message'))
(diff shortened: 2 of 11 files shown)


concent_api/core/utils.py
```diff
@@ -140,3 +143,15 @@ def is_given_golem_messages_version_supported_by_concent(
         if not is_protocol_version_compatible(golem_message_version):
             return False
         return True
+
+
+def generate_uuid(last_char: Optional[str] = None) -> str:
+    random.seed(0)
+    random_bits = "%32x" % random.getrandbits(128)
```
cameel: I you initialize the generator every time, you'll get the same random number every time:

``` python
>>> import random
>>> random.seed(0)
>>> random.getrandbits(128)
302934307671667531413257853548643485645
>>> random.seed(0)
>>> random.getrandbits(128)
302934307671667531413257853548643485645
>>> random.seed(0)
>>> random.getrandbits(128)
302934307671667531413257853548643485645
>>> random.seed(0)
>>> random.getrandbits(128)
302934307671667531413257853548643485645
>>> random.seed(0)
>>> random.getrandbits(128)
302934307671667531413257853548643485645
>>> random.seed(0)
>>> random.getrandbits(128)
302934307671667531413257853548643485645
```

We want this instead:
``` python
>>> import random
>>> random.seed(0)
>>> random.getrandbits(128)
302934307671667531413257853548643485645
>>> random.getrandbits(128)
328530677494498397859470651507255972949
>>> random.getrandbits(128)
312458951358162775936600420074706364002
>>> random.getrandbits(128)
329325272347819381495408738591077500473
>>> random.getrandbits(128)
303416399459424055589361154128267592196
```

You should initialize it globally, but in such a way that it's only done when tests run (we don't want random numbers to be predictable when we're not running tests). @dybi maybe you know a good place for global test initialization?

pawelkisielewicz: Wow, right, good point.
What do you think if instead of function we create class as singleton with one method to generate random uuid. This way there is no need to initialize generator with other place, everything will be nice and in one class. And this singleton only once, when created will call `random.seed(0)` and then only `random.getrandbits(128)`. What do you think?

concent_api/core/tests/test_core_views.py
```diff
@@ -488,6 +488,16 @@ def test_send_task_to_compute_with_public_key_with_wrong_length_should_return_ht
         self._assert_stored_message_counter_not_increased()
         self._assert_client_count_is_equal(0)
 
+    @freeze_time("2017-11-17 10:00:00")
+    def test_error_in_golem_messages_should_be_logged_and_return_http_400_with_the_same_uuid(self):
+        setattr(self.task_to_compute, 'requestor_public_key', getattr(self.task_to_compute, 'requestor_public_key')[:-1])
+        with mock.patch('core.decorators.log') as log:
+            response = self._send_force_report_computed_task()
+
+        uuid = log.call_args[0][1]
```
dybi: it's more of `uuid_entry` than just uuid ;)

concent_api/core/utils.py
dybi: let's change this functions signature and add an optional parameter `seed = None`. If 
seed is `None`:
```python
if seed is None:
     seed = get_current_utc_timestamp()
random.seed(seed)
```

Also this function was intended to be a helper test function. Right now, it has grown too big to be a single function. So, core functionality with generating uuid based on seed should stay here in `core.utils`, but the part with substituting last character should be moved to another function, used in tests only. The latter could be `generate_uuid_for_tests` or `create_uuid_for tests` or sth similar and it should be placed in `core.tests.utils`

concent_api/core/decorators.py
```diff
@@ -122,8 +124,9 @@ def wrapper(request: HttpRequest, *args: Any, **kwargs: Any) -> Union[HttpRespon
                 log(logger, 'Error:', exception.__class__.__name__)
                 return JsonResponse({'error': f'{exception}'}, status = 400)
             except MessageError as exception:
-                log(logger, ERROR_IN_GOLEM_MESSAGE, exception.__class__.__name__)
-                return JsonResponse({'error': join_messages(ERROR_IN_GOLEM_MESSAGE, str(exception))}, status = 400)
+                request_id = generate_uuid()
+                log(logger, f'uuid: {request_id}', ERROR_IN_GOLEM_MESSAGE, exception.__class__.__name__, str(traceback.format_exc()))
```
dybi:
```suggestion
                log(logger, f'uuid: {request_id}', ERROR_IN_GOLEM_MESSAGE, str(exception), log_level=exception
```

this would require adding new log level to `LoggingLevel` and adapt `log` function

concent_api/core/decorators.py
```diff
@@ -122,8 +124,9 @@ def wrapper(request: HttpRequest, *args: Any, **kwargs: Any) -> Union[HttpRespon
...
+                return JsonResponse({'error': join_messages(f'uuid: {request_id}', ERROR_IN_GOLEM_MESSAGE, exception.__class__.__name__)}, status=400)
```
dybi: why `exception.__class__.__name__` instead of `str(exception)`?

concent_api/core/utils.py
pawelkisielewicz: done

concent_api/core/tests/test_core_views.py
```diff
@@ -488,6 +488,16 @@ def test_send_task_to_compute_with_public_key_with_wrong_length_should_return_ht
         self._assert_stored_message_counter_not_increased()
         self._assert_client_count_is_equal(0)
 
+    @freeze_time("2017-11-17 10:00:00")
+    def test_error_in_golem_messages_should_be_logged_and_return_http_400_with_the_same_uuid(self):
+        setattr(self.task_to_compute, 'requestor_public_key', getattr(self.task_to_compute, 'requestor_public_key')[:-1])
```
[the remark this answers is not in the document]
pawelkisielewicz: done

concent_api/core/tests/test_core_views.py
```diff
@@ -488,6 +488,16 @@ def test_send_task_to_compute_with_public_key_with_wrong_length_should_return_ht
         self._assert_stored_message_counter_not_increased()
         self._assert_client_count_is_equal(0)
 
+    @freeze_time("2017-11-17 10:00:00")
+    def test_error_in_golem_messages_should_be_logged_and_return_http_400_with_the_same_uuid(self):
```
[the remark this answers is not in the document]
pawelkisielewicz: done

concent_api/core/tests/test_core_views.py
pawelkisielewicz: done

concent_api/core/decorators.py
pawelkisielewicz: `str(exception)` is empty string, name of exception is  `exception.__class__.__name__ `. It is really don't necessary now because of stacktrace, but I would leave it as more readable

Issue #925, opened 2018-11-05
End-to-end tests for multiple golem-messages routing


rwrzesien:
There should be some end-to-end tests for testing how Concent network handles multiple golem-messages versions.

As it is client's resposibility to provide correct golem-messages version, the test needs to know it. The Concent instances should be named in the test, and for each of them there should parameter with their golem-messages version, ie:
```
> test-golem-messages-version.sh --concent_1_version=2.15.0 --concent_2_version=2.15.1 ...
```

Off course they will run under the same host, as we will be testing messages routing.

What we will be really testing is how Concent's view decorators reacts to HTTP header `HTTP_CONCENT_GOLEM_MESSAGES_VERSION`, so it doesn't matter which golem-messages message will be used. For simplicity, it should be the one that will change most unlikely, like `Ping` message.

The response for message which is not handled by Concent is `Http400` with error code `ErrorCode.MESSAGE_UNKNOWN`, so this response should be threated as correct. It means that a right version of Concent, handling the same version of golem-messages as in the header, got the request. Otherwise it would be `Http200` with message `ServiceRefused`. It would also have `reason` set to `UNSUPPORTED_PROTOCOL_VERSION` after https://github.com/golemfactory/golem-messages/issues/275 is implemented.

Tools and utils from `api_testing_common` should be reused if possible.

Possible test scenarios proposals:
1. Test if number of requests equal to known Concent instances with different golem-messages versions will be handled correctly.
- Create requests for each golem-messages version supported in the network.
- Check if response is `Http400` with error code `ErrorCode.MESSAGE_UNKNOWN`.

2. Test if unsupported golem-messages version will be handled correctly.
- Create request for several golem-messages versions not supported in the network.
- Check if response is `Http200` with message `ServiceRefused`.

3. Corner-cases of each of the above. 
- There are some corner rules on which version is treated as support or not. The corner-cases of those should be tested for both support and not-support
- The rules are defined here https://github.com/golemfactory/concent/issues/915

4. `Mayhem` scenario.
- Create a lot of requests with either supported or unsupported version. Send those requests in pseudo-random order.
- Test if each response is correct.
----------------------------------
# EDIT
After discussion, we have come to conclusion that all we need are simple ("sanity") tests that would check how `nginx-router` behaves.
Four scenarios have been identified:
1. _A request with `latest` version of `Golem Messages` is received._
[...]

PR #933, opened 2018-11-13
Chore end to end tests for multiple golem messages routing

Jakub89:
Resolves: #925

### DIFF
--- a/concent_api/api-e2e-cluster-multiple-golem-messages-routing.py
+++ b/concent_api/api-e2e-cluster-multiple-golem-messages-routing.py
@@ -1,0 +1,105 @@
+#!/usr/bin/env python3
+
+import os
+import sys
+import requests
+
+from golem_messages.message import Ping
+
+from api_testing_common import api_request
+from api_testing_common import count_fails
+from api_testing_common import PROVIDER_PRIVATE_KEY
+from api_testing_common import run_tests
+from protocol_constants import ProtocolConstants
+
+
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "concent_api.settings")
+
+
+@count_fails
+def test_case_ping_message_send_to_concent_with_supported_golem_messages_should_get_http400_response_with_message_unknown_error(
+    cluster_url: str,
+    cluster_consts: ProtocolConstants,
+    concent_1_golem_messages_version: str,
+    concent_2_golem_messages_version: str,
+) -> None:
+    # Sending Ping message to concent with supported versions of Golem Messages.
+    # Expected: nginx-router will forward it to the cluster with that version, user will receive Http400,
+    # with error code ErrorCode.MESSAGE_UNKNOWN, because concent doesn't handle Ping message.
+    api_request(
+        cluster_url,
+        'send',
+        PROVIDER_PRIVATE_KEY,
+        CONCENT_PUBLIC_KEY,
+        Ping(),
+        expected_status=400,
+        expected_golem_version=concent_1_golem_messages_version,
+        expected_error_code='message.unknown',
+    )
+    api_request(
+        cluster_url,
+        'send',
+        PROVIDER_PRIVATE_KEY,
+        CONCENT_PUBLIC_KEY,
+        Ping(),
+        expected_status=400,
+        expected_golem_version=concent_2_golem_messages_version,
+        expected_error_code='message.unknown',
+    )
+
+
+@count_fails
+def test_case_ping_message_send_to_concent_with_unsupported_golem_messages_should_get_http404_response_with_not_found_error(
+    cluster_url: str,
+    cluster_consts: ProtocolConstants,
+    concent_1_golem_messages_version: str,
+    concent_2_golem_messages_version: str,
+) -> None:
+    # Sending Ping message to concent with unsupported version of Golem Messages.
+    # Expected: nginx-router will respond with Http404.
+    api_request(
+        cluster_url,
+        'send',
+        PROVIDER_PRIVATE_KEY,
+        CONCENT_PUBLIC_KEY,
+        Ping(),
+        headers={
+            'Content-Type': 'application/octet-stream',
+            'X-Golem-Messages': '1.0.0'
+        },
+        expected_status=404,
+        expected_error_code='not-found',
+    )
+
+
+@count_fails
+def test_case_ping_message_send_to_concent_with_malformed_golem_messages_should_get_http400_response_with_bad_request_error(
+    cluster_url: str,
+    cluster_consts: ProtocolConstants,
+    concent_1_golem_messages_version: str,
+    concent_2_golem_messages_version: str,
+) -> None:
+    # Sending Ping message to concent with malformed header(Golem Messages version has a wrong, non-semver format).
+    # Expected: ngingx-router will respond with Http400 with error code "bad-request".
+    api_request(
+        cluster_url,
+        'send',
+        PROVIDER_PRIVATE_KEY,
+        CONCENT_PUBLIC_KEY,
+        Ping(),
+        headers={
+            'Content-Type': 'application/octet-stream',
+            'X-Golem-Messages': 'X.X.X'
+        },
+        expected_status=400,
+        expected_error_code='bad-request',
+    )
+
+
+if __name__ == '__main__':
+    try:
+        from concent_api.settings import CONCENT_PUBLIC_KEY
+        run_tests(globals())
+    except requests.exceptions.ConnectionError as exception:
+        print("\nERROR: Failed connect to the server.\n", file=sys.stderr)
+        sys.exit(str(exception))
--- a/concent_api/api_testing_common.py
+++ b/concent_api/api_testing_common.py
@@ -13,7 +13,6 @@ import datetime
 import http.client
 import json
 import requests
-import sys
 
 from freezegun import freeze_time
 
@@ -266,17 +265,6 @@ def create_client_auth_message(client_priv_key: bytes, client_public_key: bytes,
     client_auth = ClientAuthorization()
     client_auth.client_public_key = client_public_key
     return dump(client_auth, client_priv_key, concent_public_key)
-
-
-def parse_command_line(command_line: list) -> str:
-    if len(command_line) <= 1:
-        sys.exit('Not enough arguments')
-
-    if len(command_line) >= 3:
-        sys.exit('Too many arguments')
-
-    cluster_url = command_line[1]
-    return cluster_url
 
 
 def parse_arguments() -> Tuple:
--- a/concent_api/api_testing_common.py
+++ b/concent_api/api_testing_common.py
@@ -152,15 +152,35 @@ def validate_content_type(actual_content_type: str, expected_content_type: Optio
         )
 
 
+def validate_golem_version(actual_golem_version: str, expected_golem_version: str) -> None:
+    assert_condition(
+        '.'.join(actual_golem_version.split('.')[:-1]),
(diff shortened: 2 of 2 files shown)


concent_api/api_testing_common.py
```diff
@@ -153,15 +152,37 @@ def validate_content_type(actual_content_type: str, expected_content_type: Optio
         )
 
 
+def validate_golem_version(actual_golem_version: str, expected_golem_version: Optional[str]=None) -> None:
```
rwrzesien: Why is `expected_golem_version` optional? Is there any case where it will be `None`? If not, please make it non-optional.

concent_api/api_testing_common.py
```diff
@@ -153,15 +152,37 @@ def validate_content_type(actual_content_type: str, expected_content_type: Optio
...
+    if expected_golem_version is not None:
+        assert_condition(
+            '.'.join(actual_golem_version.split('.')[:-1]),
+            '.'.join(expected_golem_version.split('.')[:-1]),
+            f"Expected Golem Message version: {'.'.join(expected_golem_version.split('.')[:-1])}.X, "
+            f"actual version: {actual_golem_version}"
+        )
+
+
+def validate_error_code(actual_error_code: str, expected_error_code: Optional[str]=None) -> None:
```
rwrzesien: Same here.

concent_api/api_testing_common.py
```diff
@@ -153,15 +152,37 @@ def validate_content_type(actual_content_type: str, expected_content_type: Optio
...
+    if expected_error_code is not None:
+        assert_condition(
+            actual_error_code,
+            expected_error_code,
+            f"Wrong content type for Golem Message: {actual_error_code}"
```
rwrzesien: I think this message is not related with this function purpose.

concent_api/api_testing_common.py
```diff
@@ -185,21 +206,24 @@ def _print_data(data: Optional[Union[bytes, str]], url: str) -> None:
...
     url = "{}/api/v1/{}/".format(host, endpoint)
-
     _print_data(data, url)
-    response = requests.post("{}".format(url), headers=REQUEST_HEADERS, data=_prepare_data(data), verify=False)
+    response = requests.post("{}".format(url), headers=headers, data=_prepare_data(data), verify=False)
     _print_response(private_key, public_key, response)
     validate_response_status(response.status_code, expected_status)
     validate_content_type(response.headers['Content-Type'], expected_content_type)
     validate_response_message(response.content, expected_message_type, private_key, public_key)
+    if 'Concent-Golem-Messages-Version' in response.headers:
```
rwrzesien: Maybe we should we print something when this header is missing?
And I see now below that `expected_golem_version` can be None with this flow. I think it would be better to add to this `if `:
```
if expected_golem_version is not None and 'Concent-Golem-Messages-Version' in response.headers:
```
and the make it non-optional in function.

concent_api/api_testing_common.py
```diff
@@ -185,21 +206,24 @@ def _print_data(data: Optional[Union[bytes, str]], url: str) -> None:
...
+        validate_golem_version(response.headers['Concent-Golem-Messages-Version'], expected_golem_version)
     print()
     content_type = response.headers['Content-Type']
     if 'text/html' in content_type:
         return None
-    elif content_type == 'json':
+    elif content_type == 'application/json':
+        validate_error_code((json.loads(response.content))['error_code'], expected_error_code)
```
rwrzesien: Same here.

concent_api/api_testing_common.py
```diff
@@ -268,23 +292,24 @@ def create_client_auth_message(client_priv_key: bytes, client_public_key: bytes,
     return dump(client_auth, client_priv_key, concent_public_key)
 
 
-def parse_command_line(command_line: list) -> str:
```
rwrzesien: :+1: for refactoring this.

concent_api/api_testing_common.py
```diff
@@ -268,23 +292,24 @@ def create_client_auth_message(client_priv_key: bytes, client_public_key: bytes,
...
-    return cluster_url
-
-
 def parse_arguments() -> Tuple:
     parser = argparse.ArgumentParser()
     parser.add_argument("cluster_url")
     parser.add_argument("tc_patterns", nargs='*')
+    parser.add_argument(
+        '-c1',
+        '--concent_1_version',
```
rwrzesien: I think it should be named `--concent_1_golem_messages_version`. Same for second one.

concent_api/api_testing_common.py
```diff
@@ -308,7 +333,10 @@ def execute_tests(tests_to_execute: list, objects: dict, **kwargs: Any) -> None:
 def run_tests(objects: dict, additional_arguments: Optional[dict]=None) -> None:
     if additional_arguments is None:
         additional_arguments = {}
-    (cluster_url, patterns) = parse_arguments()
+    (cluster_url, patterns, concent_1, concent_2) = parse_arguments()
```
rwrzesien: Same here, longer names but more meaningful.

concent_api/api-e2e-cluster-multiple-golem-messages-routing.py
```diff
@@ -0,0 +1,107 @@
...
+
+from golem_messages.message import Ping
+
+from api_testing_common import api_request
+from api_testing_common import count_fails
+from api_testing_common import PROVIDER_PRIVATE_KEY
+from api_testing_common import run_tests
+from protocol_constants import ProtocolConstants
+
+import requests
```
rwrzesien: 3rd party lib should be before application specific imports (but pylint will not show it here as it is not checked by it).

concent_api/api_testing_common.py
```diff
@@ -13,7 +13,6 @@
 import http.client
 import json
 import requests
-import sys
```
rwrzesien: :+1:

concent_api/api-e2e-cluster-multiple-golem-messages-routing.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "concent_api.settings")
+
+
+@count_fails
+def test_case_ping_message_send_to_concent_with_supported_golem_messages_should_get_http400_response_with_message_unknown_error(
```
rwrzesien: Maybe we should write a short multi line comment (each line starting with #) why we expect 400 for supported version? It might be confusing for someone not really familiar with the case. The text can be copied from the issue (with little modification).

concent_api/api_testing_common.py
```diff
@@ -308,7 +333,10 @@ def execute_tests(tests_to_execute: list, objects: dict, **kwargs: Any) -> None:
 def run_tests(objects: dict, additional_arguments: Optional[dict]=None) -> None:
     if additional_arguments is None:
         additional_arguments = {}
-    (cluster_url, patterns) = parse_arguments()
+    (cluster_url, patterns, concent_1, concent_2) = parse_arguments()
+    if concent_1 is not None and concent_2 is not None:
```
rwrzesien: I think there should be a check that golem-messages versions are not the same.

concent_api/api-e2e-cluster-multiple-golem-messages-routing.py
```diff
@@ -0,0 +1,107 @@
...
+        'send',
+        PROVIDER_PRIVATE_KEY,
+        CONCENT_PUBLIC_KEY,
+        Ping(),
+        headers={
+            'Content-Type': 'application/octet-stream',
+            'X-Golem-Messages': '1.0.0'
+        },
+        expected_status=404,
+        expected_error_code='not-found',
```
rwrzesien: Why this error code? I can't find in `common/constants.py`.

concent_api/api_testing_common.py
Jakub89: Well yes, and it's the only time we can get `json` response from concent, so I thought it's all good?

concent_api/api_testing_common.py
Jakub89: If they going to be the same, test will work just fine :P but I guess I could add that sanity check.

concent_api/api_testing_common.py
Jakub89: This header is missing when we get http404 responses for example, so what do You suggest we should print about that?

concent_api/api_testing_common.py
Jakub89: Done

concent_api/api_testing_common.py
Jakub89: Done

concent_api/api-e2e-cluster-multiple-golem-messages-routing.py
Jakub89: Done

concent_api/api-e2e-cluster-multiple-golem-messages-routing.py
Jakub89: Added

concent_api/api_testing_common.py
Jakub89: Done

concent_api/api_testing_common.py
rwrzesien: You can now remove `if expected_golem_version is not None:` from the `validate_golem_version`.

Jakub89: Right! Done.

concent_api/api_testing_common.py
rwrzesien: Please add `and expected_error_code is not None` to condition, that is what I have meant by 'same here'. Then please remove `if` from `validate_error_code` function.

About the error codes, I couldn't find a place when we set something else than our custom error codes in response, that is why I was asking about it. If you are sure there is something like `bad-response` in `error_code` field it would be good to know why ;) I currently don't know.

Jakub89: Now that You mention that, `bad-request` and `not-found` might be some default strings.
``` python
Running TC: test_case_ping_message_send_to_concent_with_malformed_golem_messages_should_get_http400_response_with_bad_request_error
SEND (http://devel.concent.golem.network/api/v1/send/)
MESSAGE:
    type                           = Ping
    timestamp                      = 1542196028
    encrypted                      = False
    sig                            = None
    header                         = MessageHeader(type_=1001, timestamp=1542196028, encrypted=False)
STATUS: 400 Bad Request
{'error_message': 'Bad Request', 'error_code': 'bad-request'}

Running TC: test_case_ping_message_send_to_concent_with_unsupported_golem_messages_should_get_http404_response_with_not_found_error
SEND (http://devel.concent.golem.network/api/v1/send/)
MESSAGE:
    type                           = Ping
    timestamp                      = 1542196028
    encrypted                      = False
    sig                            = None
    header                         = MessageHeader(type_=1001, timestamp=1542196028, encrypted=False)
STATUS: 404 Not Found
{'error_message': 'Not Found', 'error_code': 'not-found'}
```

Jakub89:
@rwrzesien It looks like `bad-request` and `not-found` are sent by `nginx` that's why we can't find that in our code. @bartoszbetka ?

concent_api/api_testing_common.py
Jakub89: Done

rwrzesien:
@Jakub89 Then it was far beyond my jurisdiction ;)

Jakub89:
@dybi I see right now that it can't be done, because tests for `not-found` and `bad-request` don't return `golem-messages-version` in header, so we can't give them `expected_golem_version` to put in header.

PR #934, opened 2018-11-15
Change pylint version in requirements.txt

pawelkisielewicz:
No issue for that, pylint tools are installed in higher versions then our code is prepared for. There should be exactly in versions that we need


rwrzesien:
@pawelkisielewicz Please keep it indented, this is not Python code and it would be more readable to keep it that way.

Jakub89:
I agree with @rwrzesien, although this is the only requirement file with indentations. So should we change all of them?

PR #941, opened 2018-11-18
Fix for bankster public key matching issue

rwrzesien:
Fixes error in Bankster with matching `DepositClaim.DepositAccount.Client` and `Subtask.Client` public keys.

### DIFF
--- a/concent_api/core/payments/bankster.py
+++ b/concent_api/core/payments/bankster.py
@@ -17,6 +17,7 @@ from golem_sci.events import ForcedPaymentEvent
 from psycopg2 import errorcodes as pg_errorcodes
 
 from common.constants import ConcentUseCase
+from common.helpers import deserialize_message
 from common.helpers import ethereum_public_key_to_address
 from common.helpers import get_current_utc_timestamp
 from core.constants import ETHEREUM_ADDRESS_LENGTH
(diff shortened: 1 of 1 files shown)


concent_api/core/payments/bankster.py
```diff
@@ -246,14 +247,15 @@ def finalize_payment(deposit_claim: DepositClaim) -> Optional[str]:
         elif deposit_claim.concent_use_case == ConcentUseCase.ADDITIONAL_VERIFICATION:
             subtask = Subtask.objects.filter(subtask_id=deposit_claim.subtask_id).first()  # pylint: disable=no-member
             if subtask is not None:
-                if subtask.requestor.public_key == deposit_claim.payer_deposit_account.client.public_key:
+                task_to_compute = deserialize_message(subtask.task_to_compute.data.tobytes())
+                if task_to_compute.requestor_ethereum_address == deposit_claim.payer_deposit_account.ethereum_address:
```
cameel: Could you explain what this pull request is actually fixing? Neither the pull request nor the commit description gives any clue and there's no issue for it.

The condition should be fine no matter whether you compare the public key, the ethereum address or even just the IDs of the client object. They should be equivalent here.

concent_api/core/payments/bankster.py
```diff
@@ -246,14 +247,15 @@ def finalize_payment(deposit_claim: DepositClaim) -> Optional[str]:
...
                     ethereum_transaction = service.force_subtask_payment(  # pylint: disable=no-value-for-parameter
                         requestor_eth_address=deposit_claim.payer_deposit_account.ethereum_address,
                         provider_eth_address=deposit_claim.payee_ethereum_address,
                         value=deposit_claim.amount,
                         subtask_id=deposit_claim.subtask_id,
                     )
-                elif subtask.provider.public_key == deposit_claim.payer_deposit_account.client.public_key:
+                elif task_to_compute.provider_ethereum_address == deposit_claim.payer_deposit_account.ethereum_address:
                     ethereum_transaction = service.cover_additional_verification_cost(  # pylint: disable=no-value-for-parameter
                         provider_eth_address=deposit_claim.payer_deposit_account.ethereum_address,
```
cameel: The `else` branch should have `assert False`. The claim is always supposed to refer to either requestor or provider as payer.

concent_api/core/payments/bankster.py
rwrzesien: I agree, it should work the same for both the same for the cases which you have described, but my guess (I wasn't able to reproduce exception described on slack) is that it might be a problem with test, that we are using same `subtask_id` with different clients. And as we have `deposit_claim.payer_deposit_account.ethereum_address` unique, it might fix the original issue.

cameel: Please first test and make sure that it really fixes anything.

I really don't like magical changes like that. Especially if there're random tweaks just to safisfy the tests. If our tests are reusing IDs, that's what we should fix - please submit an issue.

If the problem is not reproducible, I'd prefer not to merge this change.

concent_api/core/payments/bankster.py

PR #942, opened 2018-11-19
Fix for Bankster that SCI calls returns transaction hash instead of t…

rwrzesien:
…ransaction object

Fix for Bankster that SCI calls returns transaction hash instead of transaction object.

### DIFF
--- a/concent_api/core/payments/bankster.py
+++ b/concent_api/core/payments/bankster.py
@@ -11,7 +11,6 @@ from django.db.models import Sum
 from django.db.models.functions import Coalesce
 
 from golem_messages.message.tasks import SubtaskResultsAccepted
-from golem_messages.utils import encode_hex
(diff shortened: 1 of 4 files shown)


concent_api/core/payments/bankster.py
```diff
@@ -263,7 +263,7 @@ def finalize_payment(deposit_claim: DepositClaim) -> Optional[str]:
             assert False
 
         # Bankster puts transaction ID in DepositClaim.tx_hash.
-        deposit_claim.tx_hash = encode_hex(ethereum_transaction.hash)
+        deposit_claim.tx_hash = ethereum_transaction_hash
```
pawelkisielewicz: Is it no need to use `encode_hex` any more?

rwrzesien:
> OK, please change also signature in:

:+1: done, good catch.

Issue #953, opened 2018-11-21
Remove all positional arguments in `./signing_service.sh`  and replace them with keywords.


Jakub89:
`./signing_service.sh "85cZzVjahnRpUBwm0zlNnqTdYom1LF1P1WNShLg17cmhN2UssnPrCjHKTi5susO3wrr/q07eswumbL82b4HgOw==" "127.0.0.1"` cause error because position arguments are reverse. Changing to keywords arguments will make it easier to use for random people trying to deploy Signing Service.

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/162127921

cameel:
@Jakub89 This task changes Signing Service parameters in non-backwards compatible way. Before merging it you should talk with @bartoszbetka about how it will be deployed. Our README with instructions needs to be updated, our Docker files need to be changed and a new package with Signing Service needs to be built and distributed.

PaweuB:
@cameel Would You provide an Issue for actions that is important to manage with in deployment?

kbeker:
Issue is already closed

PR #971, opened 2018-11-26
Change underscore to dashes in signing_service script values

Jakub89:
fix to #953


signing_service/signing_service/signing_service.py
```diff
@@ -458,14 +458,14 @@ def _parse_arguments() -> argparse.Namespace:
 
     parser = argparse.ArgumentParser()
     parser.add_argument(
-        '-host',
-        '--concent_cluster_host',
+        '-ch',
```
dybi: I am for `-c` if it is not already taken

cameel: The single-dash flags should be single-letter. It does not make sense to make them longer (and does it even work?). That's what the full flags are for.

Jakub89:
@cameel okey, I made separated commits.

Issue #939, opened 2018-11-16
Allow setting verification cost to zero


cameel:
Currently Bankster assumes that verification always has non-zero cost. We need to allow setting it to zero (in that case no claim against provider's account should be created).

When done, set it to zero in deployment.

Remove the default from base settings. It should have no default (except maybe in development settings).

cameel:
Is this done?

PR #973, opened 2018-11-26
Remove ADDITIONAL_VERIFICATION_COST setting default from base and tes…

rwrzesien:
…ting and add it to development

Resolves https://github.com/golemfactory/concent/issues/939

After examining the issue it seems that is just enough to modify the settings, the implementation already handles this issue correctly, and tests cover that case.

### DIFF
--- a/concent_api/concent_api/settings/base.py
+++ b/concent_api/concent_api/settings/base.py
@@ -311,7 +311,7 @@ VERIFIER_MIN_SSIM = 0.94
 ADDITIONAL_VERIFICATION_TIME_MULTIPLIER = 2.0
 
 # A global constant defining the cost of additional verification if provider requested it.
-ADDITIONAL_VERIFICATION_COST = 0
+# ADDITIONAL_VERIFICATION_COST = 0
 
 # A global constant defining deadline for submitting SubtaskResultsVerify.
 ADDITIONAL_VERIFICATION_CALL_TIME = int(constants.AVCT.total_seconds())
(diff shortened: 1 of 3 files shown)


concent_api/concent_api/settings/testing.py
```diff
@@ -84,8 +84,6 @@
 
 SIGNING_SERVICE_PUBLIC_KEY = b'\x19\xaew&P>\xce\xd7D\xfb\xbff)55\xc1%\xea\xebN\x0e\x05>\xf7v\x0c\x94I\xe4\xa3\x80\x1dl\xed\xb5j\xe9apZ\x8c\x86\xe2\xb2x^M}>8c\x90\x0e\xa2\xf4\x8cOb\xb1\x1er\x11\xd7\xee'
 
-ADDITIONAL_VERIFICATION_COST = 1
```
cameel: Do we have any unit tests for the case of `ADDITIONAL_VERIFICATION_COST = 0`?

concent_api/concent_api/settings/base.py
```diff
@@ -310,9 +310,6 @@
 # completing verification.
 ADDITIONAL_VERIFICATION_TIME_MULTIPLIER = 2.0
 
-# A global constant defining the cost of additional verification if provider requested it.
```
dybi: shouldn't it be left here? At least in commented out form?

rwrzesien: :+1: done

Issue #972, opened 2018-11-26
Move calculation of acceptance time overdue from Bankster to force payment handler


rwrzesien:
Currently calculation of acceptance time overdue is being done in Bankster. It has no reason for being there, as all required information are already in force payment handler.

As it is in current flow, when this happens the view should return `ForcePaymentRejected` message with reason `TimestampError`.

PaweuB:
Pivotal: 
https://www.pivotaltracker.com/story/show/162198463

PR #975, opened 2018-11-26
Chore move calculation of acceptance time overdue from bankster to force payment handler

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/972

### DIFF
--- a/concent_api/core/payments/bankster.py
+++ b/concent_api/core/payments/bankster.py
@@ -367,9 +367,7 @@ def settle_overdue_acceptances(
         )
 
         # Concent defines time T1 equal to youngest timestamp from list of transactions.
-        if len(list_of_transactions) == 0:
-            t1_is_bigger_than_payments_ts = False
-        else:
+        if len(list_of_transactions) > 0:
             youngest_transaction_timestamp = max(
                 ethereum_transaction.closure_time for ethereum_transaction in list_of_transactions
             )
@@ -378,19 +376,12 @@ def settle_overdue_acceptances(
 
             # Concent checks if all passed SubtaskResultAccepted messages from subtask_results_accepted_list
             # have payment_ts < T1
-            t1_is_bigger_than_payments_ts = any(
+            if any(
                 youngest_transaction_timestamp > subtask_results_accepted.payment_ts
                 for subtask_results_accepted in acceptances
-            )
-
-        # Any of the items from list of overdue acceptances
-        # matches condition current_time < payment_ts + PAYMENT_DUE_TIME
-        acceptance_time_overdue = any(
-            cut_off_time < subtask_results_accepted.payment_ts + settings.PAYMENT_DUE_TIME
-            for subtask_results_accepted in acceptances
-        )
-
-        if t1_is_bigger_than_payments_ts or acceptance_time_overdue:
+            ):
+                raise BanksterTimestampError
+        else:
             raise BanksterTimestampError
 
         # Concent gets list of forced payments from payment API where T0 <= payment_ts + PAYMENT_DUE_TIME.
--- a/concent_api/core/tests/test_integration_force_payment.py
+++ b/concent_api/core/tests/test_integration_force_payment.py
@@ -161,22 +161,11 @@ class ForcePaymentIntegrationTest(ConcentIntegrationTestCase):
             subtask_results_accepted_list = subtask_results_accepted_list
         )
 
-        with mock.patch(
-            'core.message_handlers.bankster.settle_overdue_acceptances',
-            side_effect=BanksterTimestampError
-        ) as settle_overdue_acceptances:
-            with freeze_time("2018-02-05 12:00:09"):
-                response =self.send_request(
-                    url='core:send',
-                    data                                = serialized_force_payment,
-                )
-
-        settle_overdue_acceptances.assert_called_with(
-            requestor_ethereum_address=task_to_compute.requestor_ethereum_address,
-            provider_ethereum_address=task_to_compute.provider_ethereum_address,
-            acceptances=subtask_results_accepted_list,
-            requestor_public_key=hex_to_bytes_convert(task_to_compute.requestor_public_key),
-        )
+        with freeze_time("2018-02-05 12:00:09"):
+            response =self.send_request(
+                url='core:send',
+                data                                = serialized_force_payment,
+            )
 
         self._test_response(
             response,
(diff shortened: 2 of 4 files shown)


concent_api/core/payments/bankster.py
```diff
@@ -373,20 +371,11 @@ def settle_overdue_acceptances(
 
             # Concent checks if all passed SubtaskResultAccepted messages from subtask_results_accepted_list
             # have payment_ts < T1
-            t1_is_bigger_than_payments_ts = any(
+            if any(
```
dybi: If I understand correctly, this is protection against double spending? If so, exception name seems wrong (and the returned message in the exception handler)

concent_api/core/tests/test_payments_bankster.py
```diff
@@ -360,6 +360,35 @@ def test_that_settle_overdue_acceptances_should_return_claim_deposit_with_amount
 
         self.assertEqual(claim_against_requestor.amount, 1000)
 
+    def test_that_settle_overdue_acceptances_should_raise_exception_if_any_transaction_from_acceptances_list_is_before_youngest_transaction_timestamp(self):
```
dybi: how this scenario is different from `test_that_settle_overdue_acceptances_should_return_none_if_subtask_costs_where_already_paid`? I thought that if any of transactions from `SRA` list is before the youngest transaction timestamp, this is an indicator that subtask costs have been already paid.

concent_api/core/payments/bankster.py
rwrzesien: If you are adding a comment that something is wrong it would be good to write what should be the correct version to avoid circulation.

concent_api/core/tests/test_payments_bankster.py
rwrzesien: The effect is different (return None vs raising exception) so it tests the different flow.

> I thought that if any of transactions from SRA list is before the youngest transaction timestamp, this is an indicator that subtask costs have been already paid.

Not necessary, client may try to manipulate the list to ie. try to get payment for already payed transaction.

concent_api/core/payments/bankster.py
```diff
@@ -362,9 +362,7 @@ def settle_overdue_acceptances(
         )
 
         # Concent defines time T1 equal to youngest timestamp from list of transactions.
-        if len(list_of_transactions) == 0:
-            t1_is_bigger_than_payments_ts = False
-        else:
+        if len(list_of_transactions) > 0:
             youngest_transaction_timestamp = max(
                 ethereum_transaction.closure_time for ethereum_transaction in list_of_transactions
             )
```
cameel: What if there are no transactions?

rwrzesien: Then we will never reach this point because view checks that.

concent_api/core/payments/bankster.py
```diff
@@ -373,20 +371,11 @@ def settle_overdue_acceptances(
...
-        # matches condition current_time < payment_ts + PAYMENT_DUE_TIME
-        acceptance_time_overdue = any(
-            cut_off_time < subtask_results_accepted.payment_ts + settings.PAYMENT_DUE_TIME
-            for subtask_results_accepted in acceptances
-        )
-
-        if t1_is_bigger_than_payments_ts or acceptance_time_overdue:
-            raise BanksterTimestampError
+            ):
+                raise BanksterTimestampError
```
cameel: @dybi asked if this part is correct and after looking more closely, it seems to make no sense to me either.

T0 is the timestamp of the oldest acceptance.
T1 is the timestamp of the youngest payment.

In the implementation above Concent will reject provider's case when `t1_is_bigger_than_payments_ts` is `True`, i.e. when there's any acceptance that's older than the youngest payment. So Concent will refuse any case where T0 <= T1. For example a case like this will be rejected outright:

```
     -100 GNT    1 GNT         -100 GNT
       |         |               |
------SRA-------PAY-------------SRA------> [time]
       |         |
       T0        T1
```
even though the right decision is for Concent to pay 199 GNT to the provider.

It might make sense if `PAY` in the above was a forced payment made by Concent (NOTE: I mean forced payment issued in use case 5, not a forced **subtask** payment from use case 3 or 4) because such a payment automatically settles all tasks completed before it (even if it does not cover the whole cost). That's not the case with normal payments.

rwrzesien: Well, it was like this since first implementation. This is interesting that nobody noticed this before. Please create a separate issue then.

concent_api/core/payments/bankster.py
cameel: > Then we will never reach this point

So this means that if the requestor has never paid anything (the list of transactions is empty) bankster will not pay anything either? That's wrong. The whole concept of Concent is that it can pay if requestor does not.

> because view checks that.

That's wrong too. Everything that touches the blockchain is supposed to be inside Bankster. Bankster is an API that wraps SCI and Concent should never communicate with SCI directly now that it's been introduced.

rwrzesien: Sorry I was wrong, I thought we are talking about client message here. So if there are no transactions there should be an exception raised as it is in current logic, which I have now added. If this is wrong please describe how it should work then.

cameel: Here's what Lukasz's spec says:

> Usługa Concent tworzy listę transakcji LT składającą się z listy transakcji płatności wykonanych przez Requestora, dla których podany w nich stempel czasu jest nie starszy niż T0. Lista może być pusta. (patrz punkt 4).
> Usługa Concent określa czas T1 równy najmłodszemu stemplowi czasu podanemu w transakcjach z listy LT (patrz punkt 5).

Obviously if there were no payments from the requestor, Concent should charge him with the full cost of all subtasks for which provider has submitted acceptances (minus the amount covered by forced payments made by Concent if any).

I think it would be best to move this discussion to a separate issue.

cameel: #982. #983, #984

concent_api/core/tests/test_integration_force_payment.py
```diff
@@ -161,22 +161,11 @@ def test_provider_send_force_payment_beyond_payment_time_concent_should_reject(s
...
-        settle_overdue_acceptances.assert_called_with(
-            requestor_ethereum_address=task_to_compute.requestor_ethereum_address,
-            provider_ethereum_address=task_to_compute.provider_ethereum_address,
-            acceptances=subtask_results_accepted_list,
-            requestor_public_key=hex_to_bytes_convert(task_to_compute.requestor_public_key),
-        )
+        with freeze_time("2018-02-05 12:00:09"):
+            response =self.send_request(
+                url='core:send',
+                data                                = serialized_force_payment,
```
pawelkisielewicz: styling :)

Issue #974, opened 2018-11-26
Fix discard_claim operation error handling


rwrzesien:
Currently `discard_claim` operation has some misplaced exception catching. The code looks like:

```
 DepositAccount.objects.select_for_update().get(
            pk=deposit_claim.payer_deposit_account_id
        )

        if deposit_claim.tx_hash is None:
            claim_removed = False
        else:
            try:
                deposit_claim.delete()
                claim_removed = True
            except DepositAccount.DoesNotExist:
                claim_removed = False
```

1. `DepositAccount.objects.select_for_update().get()` is the method which raises `DoesNotExist` exception, so it should be wrapped in `try/except`.
2. `delete()` does not raise that exception, so it shouldn't be wrapped in catching it.

PaweuB:
Pivotal: 
https://www.pivotaltracker.com/story/show/162205946

PR #976, opened 2018-11-26
Fix discard_claim operation error handling

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/974

### DIFF
--- a/concent_api/core/payments/bankster.py
+++ b/concent_api/core/payments/bankster.py
@@ -482,18 +482,19 @@ def discard_claim(deposit_claim: DepositClaim) -> bool:
     assert isinstance(deposit_claim, DepositClaim)
 
     with transaction.atomic(using='control'):
-        DepositAccount.objects.select_for_update().get(
-            pk=deposit_claim.payer_deposit_account_id
-        )
+        try:
+            DepositAccount.objects.select_for_update().get(
+                pk=deposit_claim.payer_deposit_account_id
+            )
+        except DepositAccount.DoesNotExist:
+            assert False
 
         if deposit_claim.tx_hash is None:
             claim_removed = False
         else:
-            try:
-                deposit_claim.delete()
-                claim_removed = True
-            except DepositAccount.DoesNotExist:
-                claim_removed = False
+            deposit_claim.delete()
+            claim_removed = True
+
     log(
         logger,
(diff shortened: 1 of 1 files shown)


concent_api/core/payments/bankster.py
```diff
@@ -477,18 +477,19 @@ def discard_claim(deposit_claim: DepositClaim) -> bool:
...
 
     with transaction.atomic(using='control'):
-        DepositAccount.objects.select_for_update().get(
-            pk=deposit_claim.payer_deposit_account_id
-        )
+        try:
+            DepositAccount.objects.select_for_update().get(
+                pk=deposit_claim.payer_deposit_account_id
+            )
+        except DepositAccount.DoesNotExist:
```
dybi: If this account doesn't exist, how did we get so far?

rwrzesien: It exists, in the first version of my code there was no such check, but as is was added later in wrong place I have moved it into correct one. It can be here or not, it should not make much difference. I would keep it as sanity check and good practice.

rwrzesien: It can always happen that it was removed in some other way, through admin panel, manually in database etc. So there is a case. 

How should `assert` looks here then? It can't be verified without making a query.

cameel: > It can always happen that it was removed in some other way, through admin panel, manually in database etc. So there is a case.

We don't support manual database modifications. And if this happens without such modifications, it's a bug. It's fine to crash then. If we wanted to work with an assumption that the database can be modified in an arbitrary way, we'd have to stuff all our code with such defensive checks and I it would make the code impossible to work with.

Our basic assumption is that the database is over Concent's total control.

We might want to handle manual modifications in some cases where they're likely but this is not one of them.

> How should `assert` looks here then? It can't be verified without making a query.

What's wrong about doing a query?

But anyway, I did not mean that an `assert` is needed here. I wanted to say that if you're worried about the object not being there, you should check it with an `assert` because it makes your intentions clear (you assume that this is never supposed to happen) rather than by adding a handler (which implies that it's something that can happen and we need to handle it).

cameel: You could do
``` python
except DepositAccount.DoesNotExist:
    assert False
```

Issue #950, opened 2018-11-21
Signing Service fails instead of printing a friendly error message when it gets an invalid base64 value


bartoszbetka:
There is a case when someone make mistake when running signing service by changing the order of `concent_cluster_host` and `concent_public_key` variables. By running:

```
./signing_service.sh "85cZzVjahnRpUBwm0zlNnqTdYom1LF1P1WNShLg17cmhN2UssnPrCjHKTi5susO3wrr/q07eswumbL82b4HgOw==" "127.0.0.1" --ethereum-private-key-from-env --signing-service-private-key-from-env
```
The stacktrace look like this:
```
Traceback (most recent call last):
  File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/vagrant/concent/signing_service/signing_service/signing_service.py", line 614, in <module>
    main()
  File "/home/vagrant/concent/signing_service/signing_service/signing_service.py", line 581, in main
    args = _parse_arguments()
  File "/home/vagrant/concent/signing_service/signing_service/signing_service.py", line 574, in _parse_arguments
    return parser.parse_args()
  File "/usr/lib/python3.6/argparse.py", line 1743, in parse_args
    args, argv = self.parse_known_args(args, namespace)
  File "/usr/lib/python3.6/argparse.py", line 1775, in parse_known_args
    namespace, args = self._parse_known_args(args, namespace)
  File "/usr/lib/python3.6/argparse.py", line 1963, in _parse_known_args
    positionals_end_index = consume_positionals(start_index)
  File "/usr/lib/python3.6/argparse.py", line 1940, in consume_positionals
    take_action(action, args)
  File "/usr/lib/python3.6/argparse.py", line 1849, in take_action
    action(self, namespace, argument_values, option_string)
  File "/home/vagrant/concent/signing_service/signing_service/utils.py", line 113, in __call__
    self.const = b64decode(self.const)
  File "/home/vagrant/virtualenv/lib/python3.6/base64.py", line 87, in b64decode
    return binascii.a2b_base64(s)
binascii.Error: Incorrect padding
```
The error message indicates fault in base64 variables, but it's actually tries to decode "127.0.0.1".

cameel:
[...]

PR #979, opened 2018-11-27
Add binascii.Error handel to signing service script when unable to de…

Jakub89:
…code base64 value

Resolves: #950

### DIFF
--- a/signing_service/signing_service/tests/test_signing_service_methods.py
+++ b/signing_service/signing_service/tests/test_signing_service_methods.py
@@ -18,6 +18,7 @@ from signing_service.constants import SIGNING_SERVICE_DEFAULT_PORT
 from signing_service.constants import SIGNING_SERVICE_DEFAULT_INITIAL_RECONNECT_DELAY
 from signing_service.constants import SIGNING_SERVICE_DEFAULT_RECONNECT_ATTEMPTS
 from signing_service.constants import SIGNING_SERVICE_MAXIMUM_RECONNECT_TIME
+from signing_service.exceptions import Base64DecodeError
 from signing_service.exceptions import SigningServiceValidationError
 from signing_service.signing_service import _parse_arguments
 from signing_service.signing_service import SigningService
@@ -274,6 +275,19 @@ class SigningServiceParseArgumentsTestCase(TestCase):
         os.remove(sentry_tmp_file)
         os.remove(ethereum_private_key_tmp_file)
         os.remove(signing_service_private_key_tmp_file)
+
+    def test_that_argument_parses_should_fail_gracefully_when_unable_to_decode_base64_value(self):
+        sys.argv += [
+            '--concent-cluster-host', '127.0.0.1',
+            '--concent-public-key', self.concent_public_key_encoded[:-1],
+        ]
+        with self.assertRaises(Base64DecodeError) as error:
+            _parse_arguments()
+
+        self.assertIn(
+            f'Unable to decode "{self.concent_public_key_encoded[:-1]}"',
+            str(error.exception)
+        )
 
 
 class SigningServiceValidateArgumentsTestCase(TestCase):
--- a/signing_service/signing_service/utils.py
+++ b/signing_service/signing_service/utils.py
@@ -1,3 +1,4 @@
+import binascii
 import logging.config
 import os
 import smtplib
@@ -16,6 +17,7 @@ from golem_messages.cryptography import verify_pubkey
 from golem_messages.exceptions import InvalidKeys
 
 from signing_service.constants import ETHEREUM_PRIVATE_KEY_REGEXP
+from signing_service.exceptions import Base64DecodeError
 
 logger = logging.getLogger()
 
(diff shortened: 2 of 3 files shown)


signing_service/signing_service/utils.py
```diff
@@ -16,6 +17,7 @@
 from golem_messages.exceptions import InvalidKeys
 
 from signing_service.constants import ETHEREUM_PRIVATE_KEY_REGEXP
+from signing_service.exceptions import SigningServiceValidationError
```
rwrzesien: I think it would be good to add a separate exception class for that case, as this was meant to be used to validate values inside SigningService.

signing_service/signing_service/signing_service.py
```diff
@@ -581,7 +582,10 @@ def main() -> None:
     logging.config.fileConfig(os.path.join(os.path.dirname(os.path.abspath(__file__)), '..', 'logging.ini'))
 
     # Parse required arguments.
-    args = _parse_arguments()
+    try:
```
dybi: this seems unneeded, please remove

signing_service/signing_service/utils.py
Jakub89: `InvalidValueError`, `Base64DecodeError` ?

signing_service/signing_service/signing_service.py
Jakub89: I understood form @cameel comment on the issue, that we don't want stacktrace, but it might not be the case?

signing_service/signing_service/utils.py
rwrzesien: `Base64DecodeError` sounds good, `InvalidValueError` is too broad I think.

signing_service/signing_service/signing_service.py

signing_service/signing_service/utils.py
```diff
@@ -110,7 +112,11 @@ def __call__(  # type: ignore
             assert False
         if self.base64_convert:
             assert isinstance(self.const, str)
-            self.const = b64decode(self.const)
+            try:
+                self.const = b64decode(self.const)
+            except binascii.Error as exception:
+                logger.error(f'Unable to decode "{self.const}", {exception}')
+                raise Base64DecodeError(f'Unable to decode "{self.const}", {exception}')
```
cameel: Would it be possible to report the error using the mechanisms built into argparse? What happens if you raise `argparse.ArgumentTypeError` or `ValueError` here?

It should work for [argparse.type](https://docs.python.org/3/library/argparse.html#type), not sure about a class inheriting from `argparse.Action`.

signing_service/signing_service/utils.py
```diff
@@ -96,21 +98,26 @@ def __call__(  # type: ignore
...
+            with open(value) as file:
                 self.const = file.read()
-        elif values is not None and self.read_command_line:
-            self.const = values
+        elif self.read_command_line:
+            self.const = value
         elif self.env_variable_name is not None:
             self.const = os.environ.get(self.env_variable_name)
         else:
             assert False
```
pawelkisielewicz: this is not your change, but consider instead of `assert False` raise some exception, may be custom. And even better last `elif` change to `else` and add `assert self.env_variable_name is not None`

```python
 else:
    assert self.env_variable_name is not None:
    self.const = os.environ.get(self.env_variable_name)
```

signing_service/signing_service/tests/test_signing_service_methods.py
```diff
@@ -275,6 +276,14 @@ def test_that_argument_parser_should_parse_parameters_if_passed_files_exist(self
...
         os.remove(signing_service_private_key_tmp_file)
 
+    def test_that_argument_parses_should_fail_gracefully_when_unable_to_decode_base64_value(self):
+        sys.argv += [
+            '--concent-cluster-host', '127.0.0.1',
+            '--concent-public-key', self.concent_public_key_encoded[:-1],
+        ]
+        with self.assertRaises(Base64DecodeError):
+            _parse_arguments()
+
```
pawelkisielewicz: please mock ` logger.error`  in `utils` and ensure, that `Base64DecodeError` with **right message** is called. For now `Base64DecodeError` is used in only one place, but in future in may change and we should be 100% sure we are expecting exception from right place

signing_service/signing_service/utils.py
Jakub89: Personally I would change it to what You propose, but as said, it's not my code originally, so some second opinion is needed. @rwrzesien ? @cameel ?

rwrzesien: I personally don't see much difference in functionality and logic between those two so up to you.

I also don't really understand what should be done here. First the suggestion is about raising exception, then using assert, so I don't know what is the final solution you want to apply here.

signing_service/signing_service/utils.py
Jakub89: I'm not sure what You mean. Raising `argparse.ArgumentTypeError` or `ValueError` in this line instead of `Base64DecodeError` work exactly the same. Same stacktrace with different error at the end.
But if You want me to add `type` checking to `SecretProvider(Action)` class it's going to require some significant refactoring, and will take me a lot of time, that can be spend on something more important.

signing_service/signing_service/utils.py
Jakub89: Okey, I've change it.

signing_service/signing_service/tests/test_signing_service_methods.py
Jakub89: Done

Issue #937, opened 2018-11-15
Prevent transaction nesting when using @atomic


cameel:
When we create transactions with `atomic()`, we usually rely on the fact that exiting the block commits the transaction. Only an actual commit releases locks and reveals `IntegrityErrors` caused by attempts to create two identical records in parallel.

Unfortunately one of the features provided by Django is that a call to `atomic()` when there's already a transaction in progress creates a savepoint. When such a nested `atomic()` block exits, the savepoint is simply removed and there is no commit. This ensures atomicity of the outer transaction but at the cost of making the nested one non-durable.

For that reason we want to avoid nesting transactions. It is however easy to nest transactions by mistake without realizing it when they're hidden in helper functions. The goal of this task is to implement a wrapper over `atomic()` (let's call it `non_nesting_atomic()`) that raises an exception if there's already a transaction in progress and calls `atomic()` only if there isn't. You can check that with `get_connection(using).in_atomic_block`.

After implementing it, replace all uses of `atomic()` in our code with the wrapper and run tests. It it uncovers any cases of undesirable nesting, fix them by reorganizing the code.

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/162008119

cameel:
@rwrzesien noticed that `non_nesting_atomic()` does not work in unit tests because they use transactions to clean the database. We could switch to `TransactionTestCase` as a base class but that would make our tests about 8 times slower. We don't want that.

As a workaround my suggestion is to add a setting called `DETECT_NESTED_TRANSACTIONS`. `non_nesting_atomic()` should allow nesting if this setting is `False`. Set it to `False` in `settings/testing.py` and to `True` in `settings/base.py`. This way we won't be able to detect nesting when running unit tests (which would be really useful) but we'll still catch it in end-to-end tests and on the cluster. That's better than nothing because nesting is really hard to notice and the code breaks silently.

PR #987, opened 2018-11-28
Chore prevent transaction nesting when using atomic

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/937

### DIFF
--- a/concent_api/common/decorators.py
+++ b/concent_api/common/decorators.py
@@ -54,10 +54,10 @@ def log_task_errors(task: Callable) -> Callable:
     return wrapper
 
 
-def non_nesting_atomic(using: Optional[Union[str, Callable]]=None, savepoint: bool=True) -> Atomic:
+def non_nesting_atomic(using: Union[str, Callable], savepoint: bool=True) -> Atomic:
     if (
         settings.DETECT_NESTED_TRANSACTIONS and
-        get_connection(using if using is not None else DEFAULT_DB_ALIAS).in_atomic_block
+        get_connection(using).in_atomic_block
     ):
         raise ConcentPendingTransactionError
     if callable(using):
--- a/concent_api/common/decorators.py
+++ b/concent_api/common/decorators.py
@@ -2,11 +2,16 @@ from functools import wraps
 from logging import getLogger
 from typing import Any
 from typing import Callable
+from typing import Union
 import traceback
 
 from django.conf import settings
+from django.db import DEFAULT_DB_ALIAS
+from django.db.transaction import Atomic
+from django.db.transaction import get_connection
 
 from common.exceptions import ConcentFeatureIsNotAvailable
+from common.exceptions import ConcentPendingTransactionError
 from common.logging import LoggingLevel
 from common.logging import log
 
@@ -47,3 +52,15 @@ def log_task_errors(task: Callable) -> Callable:
                 logging_level=LoggingLevel.ERROR)
             raise
     return wrapper
+
+
+def non_nesting_atomic(using: Optional[Union[str, Callable]]=None, savepoint: bool=True) -> Atomic:
+    if (
+        settings.DETECT_NESTED_TRANSACTIONS and
+        get_connection(using if using is not None else DEFAULT_DB_ALIAS).in_atomic_block
+    ):
+        raise ConcentPendingTransactionError
+    if callable(using):
+        return Atomic(DEFAULT_DB_ALIAS, savepoint)(using)
+    else:
+        return Atomic(using, savepoint)
--- a/concent_api/common/tests/test_decorators.py
+++ b/concent_api/common/tests/test_decorators.py
@@ -4,8 +4,9 @@ from django.conf import settings
 from django.http import HttpResponse
 from django.http import HttpResponseNotAllowed
 from django.http import JsonResponse
-from django.test import RequestFactory
 from django.test import override_settings
+from django.test import RequestFactory
+from django.test import TransactionTestCase
 from freezegun import freeze_time
 from golem_messages import dump
 from golem_messages import load
@@ -13,6 +14,8 @@ from golem_messages import message
 
 from common.constants import ErrorCode
 from common.decorators import log_task_errors
+from common.decorators import non_nesting_atomic
+from common.exceptions import ConcentPendingTransactionError
 from common.testing_helpers import generate_ecc_key_pair
 from core.decorators import handle_errors_and_responses
 from core.decorators import require_golem_auth_message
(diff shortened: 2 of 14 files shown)


concent_api/common/tests/test_decorators.py
```diff
@@ -278,3 +281,20 @@ def task():
...
+)
+class NonNestingAtomicDecoratorTestCase(TransactionTestCase):
+
+    def test_that_non_nesting_atomic_decorator_will_raise_exception_if_transaction_is_nested_using_same_database(self):
+        with non_nesting_atomic(using='control'):
+            with self.assertRaises(ConcentPendingTransactionError):
+                with non_nesting_atomic(using='control'):
+                    pass
+
+    def test_that_non_nesting_atomic_decorator_will_not_raise_exception_if_transaction_is_nested_using_different_databases(self):
```
Jakub89:
```
************* Module common.tests.test_decorators
R:297, 4: Method could be a function (no-self-use)
```
I'm guessing that's because there is no `assert` in this test. Can we check at least that there are no `Exception` being raised?

concent_api/core/tests/test_unit_subtask_helpers.py
```diff
@@ -63,20 +63,20 @@ def test_that_if_more_conditions_given_and_object_does_not_exist_function_return
...
+            Subtask.objects.all(),
             subtask_id=self.task_to_compute.compute_task_def['subtask_id'],
             state=Subtask.SubtaskState.FORCING_REPORT.name,  # pylint: disable=no-member
         )
         self.assertEqual(self.subtask, subtask)
 
     def test_that_if_more_conditions_given_and_queryset_does_not_exist_function_returns_none(self) -> None:
         subtask = get_one_or_none(
-            Subtask.objects.select_for_update(),
+            Subtask.objects.all(),
```
pawelkisielewicz: why is it changed? `get_one_or_none` should work with `select_for_update` as well. I think it should stay at least one test with `select_for_update`

concent_api/common/decorators.py
```diff
@@ -47,3 +51,15 @@ def wrapper(*args: Any, **kwargs: Any) -> None:
                 logging_level=LoggingLevel.ERROR)
             raise
     return wrapper
+
+
+def non_nesting_atomic(using: str=None, savepoint: bool=True) -> Atomic:
```
pawelkisielewicz: :+1: typing

dybi:
```suggestion
def non_nesting_atomic(using: str, savepoint: bool=True) -> Atomic:
```
it seems that all calls to the decorator contain a value for `using`. In other words it is never called with `None`, so both the signature and `get_connection()` call should be changed.

concent_api/common/decorators.py
```diff
@@ -47,3 +51,15 @@ def wrapper(*args: Any, **kwargs: Any) -> None:
...
+    if (
+        settings.DETECT_NESTED_TRANSACTIONS and
+        get_connection(using if using is not None else DEFAULT_DB_ALIAS).in_atomic_block
```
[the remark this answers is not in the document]
dybi: @rwrzesien, what would be the case when there is no `using`? Should we even bother with the default value here?

concent_api/common/decorators.py
```diff
@@ -47,3 +51,15 @@ def wrapper(*args: Any, **kwargs: Any) -> None:
...
+    ):
+        raise ConcentPendingTransactionError
+    if callable(using):
```
dybi: could you please elaborate what this `if` is in fact checking and what happens when it is `True`?

concent_api/common/tests/test_decorators.py
```diff
@@ -278,3 +281,20 @@ def task():
 
         mock_format_exc.assert_called_once()
         mock_error.assert_called_once()
+
+
+@override_settings(
+    DETECT_NESTED_TRANSACTIONS=True,
+)
+class NonNestingAtomicDecoratorTestCase(TransactionTestCase):
```
dybi: could you please add more tests, i.e for cases when `DETECT_NESTED_TRANSACTIONS=False`.

concent_api/common/decorators.py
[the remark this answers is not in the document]
pawelkisielewicz: I thought about it as well. If there is no `using` `DEFAULT_DB_ALIAS` is taken and it is `str: 'default'`. We are not using `default` database, only `storage` and `control`. 

But in django, if there is no `using` it means that this concerns `default` database.  So i think we will not need this now, but it is logic and coherent with django.

concent_api/common/tests/test_decorators.py
rwrzesien: > Can we check at least that there are no Exception being raised?

Unfortunately not, so I guess the only way is to silence it.

concent_api/common/decorators.py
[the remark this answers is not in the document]
rwrzesien: > And generally big +1 for this decorator, it took me a while to understand senior's code.

Go ahead and check `transaction.atomic`, I have the idea from there ;)

> what would be the case when there is no using? Should we even bother with the default value here?

Exactly like @pawelkisielewicz wrote, we shouldn't bother because we use it like `transaction.atomic`, in which we have always specified the database, but in case when we don't want to, we don't have to add anything extra to the decorator's code itself.

concent_api/common/decorators.py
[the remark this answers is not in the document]
rwrzesien: > using is database name and is str. Why are you checking if it is callable? str is not

Just like in `transaction.atomic`, you can use it as both decorator and context manager. In the first case you get a function as first parameter. I have updated the typing :+1: done.

concent_api/core/tests/test_unit_subtask_helpers.py
rwrzesien: But this test doesn't check that, so it should work for any queryset. And here there was a problem that `select_for_update()` works only in transaction, so when I have switched to `TransactionTestCase` (which, opposite to its name, is not run in transaction) there was an exception here.

concent_api/common/decorators.py
rwrzesien: It is, but it is also not required to work properly, so if someone wants to use it without it, he can. Please check `transaction.atomic` code for reference.

concent_api/common/decorators.py
rwrzesien: As mentioned above, just like in `transaction.atomic`, you can use it as both decorator and context manager. In the first case you get a function as first parameter. I have updated the typing :+1: done.

concent_api/common/tests/test_decorators.py
rwrzesien: :+1: done

concent_api/common/tests/test_decorators.py
Jakub89: Okey, I'm fine with that.

concent_api/common/decorators.py
```diff
@@ -47,3 +52,15 @@ def wrapper(*args: Any, **kwargs: Any) -> None:
                 logging_level=LoggingLevel.ERROR)
             raise
     return wrapper
+
+
+def non_nesting_atomic(using: Union[str, Callable]=None, savepoint: bool=True) -> Atomic:
```
dybi: the typing is still not ok here... the default value for `using` is neither `str`, nor `Callable`

concent_api/common/decorators.py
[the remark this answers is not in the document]
dybi: Quite frankly, I would still be for not having default value here, i.e. require that DB name is explicitly provided to the function. I think we don't have to exactly copy Django's logic here, as it is more confusing than just simply requiring parameter.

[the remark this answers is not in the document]
rwrzesien: What is better - having more options to use this decorator, or less options? Because in the solution you propose you limit options only to usages with explicitly provided database name.

concent_api/common/decorators.py
rwrzesien: :+1: done

concent_api/common/decorators.py
rwrzesien: :+1: done

concent_api/common/decorators.py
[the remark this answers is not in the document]
rwrzesien: :+1: done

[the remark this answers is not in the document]
cameel: Looking at it again now - I see that my biggest issue and the reason I agreed with @dybi was the use of a default value. I think that default values are overused in general and it's often better to force the user of the function to specify a value explicitly than have a default that is not obvious or might change, silently breaking other code.

For me forcing the programmer to spell out the name of the database is a feature for us - because we have two databases and we have to always be very careful about which one we use.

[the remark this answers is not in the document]
rwrzesien: > For me forcing the programmer to spell out the name of the database is a feature for us - because we have two databases and we have to always be very careful about which one we use.

That makes sense, and it makes even more that neither of those two is a default one. But I always lean towards the approach that developers are conscious when using the code which have defaults like this, and know when to use it or not. But I will not argue if this approach is always correct.

Issue #982, opened 2018-11-27
Wrong list of forced payments used in the payment use case


cameel:
There are two kinds of forced payments supported by `golem-smart-contracts-interface`:
- **forced subtask payments** (listed by `get_forced_subtask_payments()`). Payments performed by Concent in use cases 3 and 4 in order to pay for a specific subtask. If such a payment does not cover the cost of a subtask completely (e.g. because the requestor did not have enough deposit at the time), provider can still try to get the missing amount in use case 5. For that reason these payments are taken together with normal payments in that use case.
- **forced payments** (listed by `get_forced_payments()`). Payments performed by Concent in use case 5 to settle all unpaid subtasks in one transaction. After such a payment the balance is considered settled even if it does not cover the cost of all subtasks that preceded it. Thanks to this Concent never has to consider anything that happened before it.

Currently Concent checks `get_forced_payments()` in use case 5. It should be using the other function.

Łukasz's specification says:
> Usługa Concent tworzy listę LF (list of forced payments) wymuszonych płatności na konto Ethereum EA ze stemplem czasu nie starszym niż T0. Lista LF może być pusta.

In this part LF is actually the list of subtask payments since using the other list does not make sense.

PaweuB:
Pivotal: 
https://www.pivotaltracker.com/story/show/162242082

cameel:
Unfortunately looks like I was wrong here. After discussion with Łukasz it turns out that `LF` should contain settlement payments after all. We'll need to revert this. But please keep any tests (if any). Just change the expected results.

PR #988, opened 2018-11-29
Bugfix wrong list of forced payments used in the payment use case

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/982

I have also added tests for `sci_backend` because there were none so far for such important module.

### DIFF
--- a/concent_api/core/tests/test_payments_sci_backend.py
+++ b/concent_api/core/tests/test_payments_sci_backend.py
@@ -1,0 +1,257 @@
+from django.test import override_settings
+
+import mock
+from web3 import Web3
+
+from common.testing_helpers import generate_ecc_key_pair
+from common.helpers import get_current_utc_timestamp
+from core.constants import MOCK_TRANSACTION_HASH
+from core.payments.backends import sci_backend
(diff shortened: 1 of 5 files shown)


concent_api/core/tests/test_payments_sci_backend.py
```diff
@@ -0,0 +1,259 @@
...
+            )
+
+        self.assertEqual(deposit_value, self.deposit_value)
+
+        new_sci_rpc.return_value.get_deposit_value.assert_called_with(
+            self.task_to_compute.requestor_ethereum_address,
+        )
+
+    def test_that_sci_backend_force_subtask_payment_should_return_transaction_hash(self):
+        PaymentInterface.__instance = None
```
dybi: Is it needed? If so, why?

rwrzesien: No, it is leftover for my previous, not-working solution for mocking singleton object. It seems that if I have mocked it once during the test, then in other tests I couldn't overwrite it. I ended up mocking the whole `__new__` function instead of inner function which was called only if the instance does not already exist.

Issue #938, opened 2018-11-16
Tweak operation order to guard against irreversible operations being executed twice


cameel:
Currently there are many places in our code where we mix database changes with irreversible side-effects like blockchain operations, adding tasks to work queue or making external HTTP requests. The problem is that database changes can be rolled back in case of a failure while those side effects cannot. There's no way to make the whole operation atomic (i.e. all or nothing).

Consider the situation where we need to update subtask state in the database and issue a payment. We can do either of these things first and the order determines what's left if the operation fails in the middle. If we pay first then a failure means that a payment was performed but we have no record of it - we risk that we'll pay twice. If we commit database changes first then a failure means that we allowed a state transition without payment. While both outcomes are undesirable, the former is worse because it lets the client retry and potentially trigger more than one payment.

The goal of the task is to ensure that we always update and commit database state first. Only then we should trigger the irreversible side effects.

A complete solution for the problem of interrupted operations would be to keep a journal that would tell us which of the planned operations were actually performed. This is something to consider in the future but it's out of scope of this task.

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/162197386

rwrzesien:
@cameel What do you think about `handle_send_force_payment` handler and its usage of `bankster.settle_overdue_acceptances`? It is currently being done before other database changes, but the logic after depends of the outcome of that call.

cameel:
It's fine because in this case Concent first checks the blockchain to see if there are any settlement payments made before. If there are, it won't accept `SubtaskResultsAccepted` messages from before that.

Also, as long as `settle_overdue_acceptances()` creates and commits `DatabaseClaim` before the payment, Concent won't touch the deposit reserved in that use case.

The database changes related to creating out-of-band responses are a separate thing. It's fine if we don't create them (or maybe not really **fine** but better than the alternative anyway; this task is only about damage control, not fixing the underlying issue).

Actually there is one big issue: Concent will see the payment on the blockchain only after some delay. But it's out of the scope of this issue. This problem existed before and it's really about how Concent tracks forced payments. I proposed fixing it on Rocket and it's no up to Golem team to decide if and when to fix it.

PR #992, opened 2018-12-02
Chore tweak operation order to guard against irreversible operations being executed twice

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/938

### DIFF
--- a/concent_api/conductor/service.py
+++ b/concent_api/conductor/service.py
@@ -2,6 +2,7 @@ import logging
 from typing import List
 from typing import Optional
 
+from django.db import transaction
 from django.db.models import Q
 
 from common.constants import ErrorCode
@@ -58,7 +59,10 @@ def update_upload_report(file_path: str, result_transfer_request: ResultTransfer
         result_transfer_request.full_clean()
         result_transfer_request.save()
 
-        result_upload_finished.delay(result_transfer_request.subtask_id)
+        def call_result_upload_finished() -> None:
+            result_upload_finished.delay(result_transfer_request.subtask_id)
+
+        transaction.on_commit(call_result_upload_finished)
 
 
 def store_verification_request_and_blender_subtask_definition(
--- a/concent_api/conductor/service.py
+++ b/concent_api/conductor/service.py
@@ -1,11 +1,16 @@
 import logging
+from typing import List
+from typing import Optional
 
 from django.db.models import Q
 
 from common.constants import ErrorCode
+from common.helpers import parse_timestamp_to_utc_datetime
 from common.logging import log
 from common.logging import LoggingLevel
 from conductor.exceptions import VerificationRequestAlreadyInitiatedError
+from conductor.models import BlenderSubtaskDefinition
+from conductor.models import Frame
 from conductor.models import ResultTransferRequest
 from conductor.models import UploadReport
 from conductor.models import VerificationRequest
@@ -54,3 +59,50 @@ def update_upload_report(file_path: str, result_transfer_request: ResultTransfer
         result_transfer_request.save()
 
         result_upload_finished.delay(result_transfer_request.subtask_id)
+
+
+def store_verification_request_and_blender_subtask_definition(
+    subtask_id: str,
+    source_package_path: str,
+    result_package_path: str,
+    output_format: str,
+    scene_file: str,
+    verification_deadline: int,
+    blender_crop_script: Optional[str],
+) -> tuple:
+    verification_request = VerificationRequest(
+        subtask_id=subtask_id,
+        source_package_path=source_package_path,
+        result_package_path=result_package_path,
+        verification_deadline=parse_timestamp_to_utc_datetime(verification_deadline),
+    )
+    verification_request.full_clean()
+    verification_request.save()
+
+    blender_subtask_definition = BlenderSubtaskDefinition(
+        verification_request=verification_request,
+        output_format=BlenderSubtaskDefinition.OutputFormat[output_format].name,
+        scene_file=scene_file,
+        blender_crop_script=blender_crop_script,
+    )
+    blender_subtask_definition.full_clean()
+    blender_subtask_definition.save()
+
+    return (verification_request, blender_subtask_definition)
+
+
+def store_frames(
+    blender_subtask_definition: BlenderSubtaskDefinition,
+    frame_list: List[int],
+) -> None:
+    for frame in frame_list:
+        store_frame = Frame(
+            blender_subtask_definition=blender_subtask_definition,
+            number=frame,
+        )
+        store_frame.full_clean()
+        store_frame.save()
+
+
+def filter_frames_by_blender_subtask_definition(blender_subtask_definition: BlenderSubtaskDefinition) -> list:
+    return list(Frame.objects.filter(blender_subtask_definition=blender_subtask_definition).values_list('number', flat=True))
--- a/concent_api/conductor/views.py
+++ b/concent_api/conductor/views.py
@@ -53,12 +53,15 @@ def report_upload(_request: HttpRequest, file_path: str) -> HttpResponse:
     ):
         assert file_path in [verification_request.source_package_path, verification_request.result_package_path]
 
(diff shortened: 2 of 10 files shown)


concent_api/core/subtask_helpers.py
```diff
@@ -82,42 +98,41 @@ def finalize_claim_for_acceptance_case() -> None:
             finalize_claim_for_acceptance_case,
             using='control',
         )
-
+    elif subtask.state == Subtask.SubtaskState.VERIFICATION_FILE_TRANSFER.name:  # pylint: disable=no-member
         update_subtask_state(
             subtask=subtask,
-            state=Subtask.SubtaskState.ACCEPTED.name,  # pylint: disable=no-member
```
pawelkisielewicz: It looks like lot of important things would be changed. It is not, if somebody will review this please use PyCharm to see pull request. Github did it very mistakable, in PyCharm pull request you can see it much better

concent_api/core/subtask_helpers.py
```diff
@@ -70,6 +70,22 @@ def _update_timed_out_subtask(subtask: Subtask) -> None:
...
+            subtask=subtask,
+        )
+        store_pending_message(
+            response_type=PendingResponse.ResponseType.SubtaskResultsSettled,
+            client_public_key=subtask.requestor.public_key_bytes,
+            queue=PendingResponse.Queue.ReceiveOutOfBand,
+            subtask=subtask,
+        )
 
         def finalize_claim_for_acceptance_case() -> None:
```
pawelkisielewicz: @rwrzesien you moved this section. Why? I don't say it is wrong, but I thought that there is no difference where you put `transaction.on_commit(....`. Does it have to be at the end of `if` or `elif`?

concent_api/core/tasks.py
```diff
@@ -58,6 +58,20 @@ def upload_finished(subtask_id: str) -> None:
         if parse_datetime_to_timestamp(subtask.next_deadline) < get_current_utc_timestamp():
             # Worker makes a payment from requestor's deposit just like in the forced acceptance use case.
 
+            update_subtask_state(
```
pawelkisielewicz: same as above- you moved this block to get `transaction.on_commit(` at the end of function. Is it necessary?

concent_api/conductor/service.py
```diff
@@ -54,3 +59,50 @@ def update_upload_report(file_path: str, result_transfer_request: ResultTransfer
         result_transfer_request.save()
 
         result_upload_finished.delay(result_transfer_request.subtask_id)
+
+
+def store_verification_request_and_blender_subtask_definition(
```
dybi: is there a reason to do both things in one function? Or can it be done sequentially in two separate functions?

rwrzesien: For me there is no reason for this function to exist as it is used only once :)

rwrzesien:
@dybi 

> Side note - I know this code was just moved from one place to another, but I have realized, that some functions don't have unit tests. 

Which ones?

concent_api/core/subtask_helpers.py
rwrzesien: You are right, now it doesn't make a difference, but I have moved it to make it more readable in a way of sentence of operation chain. But if you think it is not necessary I can drop this change.

concent_api/core/tasks.py
rwrzesien: Same here, I wanted to make it consistent.

concent_api/core/subtask_helpers.py
rwrzesien: Yes I agree, when I was looking on it in Github I wasn't sure if I am looking at the correct pull request :)

concent_api/conductor/views.py
```diff
@@ -53,13 +53,13 @@ def report_upload(_request: HttpRequest, file_path: str) -> HttpResponse:
...
 
-        # If all expected files have been uploaded, the app sends upload_finished task to the work queue.
-        upload_finished.delay(verification_request.subtask_id)
-
         verification_request.upload_finished = True
         verification_request.full_clean()
         verification_request.save()
 
+        # If all expected files have been uploaded, the app sends upload_finished task to the work queue.
+        upload_finished.delay(verification_request.subtask_id)
```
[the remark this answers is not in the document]
rwrzesien: Done, good catch :+1:

rwrzesien:
@pawelkisielewicz 

> Please check line 488 in message_handlers, there is celery task in atomic block. It should be moved away from this block

Done, good catch :+1:

concent_api/core/subtask_helpers.py
pawelkisielewicz: It is OK, thanks for explanation. I was just asking :)

concent_api/conductor/views.py
cameel: What about `update_upload_report()` that's called later? It makes database changes too.

BTW, this is not the subject of this pull request but the structure of this function is really confusing and we should fix it, possibly in a separate pull request:
- `update_upload_report()` does not really modify `UploadReport`. It updates `ResultTransferRequest.upload_finished`
- Why is `ResultTransferRequest` modification hidden in a function while for `VerificationRequest` it's all out in the open? These to cases are symmetric: we either deal with an upload request or with a verification request; in both cases we have an upload report, use it to find and update the request, then fire a Celery task. The code should reflect that. I had to spend a few minutes to understand what's actually happening here.

concent_api/core/message_handlers.py
```diff
@@ -989,7 +989,8 @@ def handle_messages_from_database(client_public_key: bytes) -> Union[message.Mes
         elif pending_response.response_type == PendingResponse.ResponseType.ForceReportComputedTaskResponse.name:  # pylint: disable=no-member
             if pending_response.subtask.ack_report_computed_task is not None:
                 ack_report_computed_task = deserialize_message(
-                    pending_response.subtask.ack_report_computed_task.data.tobytes())
+                    pending_response.subtask.ack_report_computed_task.data.tobytes()
+                )
```
cameel: All the changes in this file seem unrelated to the commit at all.

concent_api/core/subtask_helpers.py
cameel: That's what happens when you move around a lot of very similar blocks of code. Neither github nor PyCharm is actually right or wrong here. This change can be done both by swapping the whole blocks (as Pycharm probably tells you) or by swapping just the values github shows. They can't know which way you actually did it. That would require understanding the meaning of this code. There are many situations where you'll see the opposite: you changed only a few values but the diff tells you that whole blocks were moved.

concent_api/conductor/service.py
```diff
@@ -54,3 +59,50 @@ def update_upload_report(file_path: str, result_transfer_request: ResultTransfer
         result_transfer_request.save()
 
         result_upload_finished.delay(result_transfer_request.subtask_id)
```
cameel: This task is being posted still inside `@atomic` while the changes have not been committed yet.

concent_api/core/message_handlers.py
rwrzesien: This is one of the small one-line non-related changes that we discussed on latest retro.

concent_api/conductor/views.py
rwrzesien: @cameel 
> update_upload_report() does not really modify UploadReport. It updates ResultTransferRequest.upload_finished

it does with this code:

```
UploadReport.objects.select_for_update().filter(
        path=file_path
    ).update(
        result_transfer_request=result_transfer_request
    )
```

> Why is ResultTransferRequest modification hidden in a function while for VerificationRequest it's all out in the open?

They were done in two different issues with some time interval between them, and `update_upload_report` is reused twice in both view and task.

> What about update_upload_report() that's called later? It makes database changes too.

Yes, but the transaction will still exist when those changes from `update_upload_report` are being done, and after they are all done the task will be scheduled.

concent_api/conductor/service.py
rwrzesien: :+1: done

Issue #984, opened 2018-11-27
Concent should reject force payment request if it includes acceptances from before the last settlement


cameel:
As described in #982, there are two types of forced payments. The one that Concent does in use case 5 settles everything done before it, even if it does not cover all the claims.

Currently Concent does not check when was the last time such a forced payment was made. Concent should find the last such payment and if it exists, compare its `closure_time` with the oldest `payment_ts` in submitted acceptances. If the's an acceptance older than the closure time, it should reject provider's request.

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/162264942

rwrzesien:
@cameel 

> Concent should find the last such payment and if it exists

How can this be achieved? If this should be taken from result of call of `get_forced_subtask_payments`, then this can be done this way, because elements returned from `get_forced_subtask_payments` doesn't have `closure_time` attribute.

cameel:
Please see the description of #982 I mentioned above. There are two types of forced payments and each has separate functions that create and list them. The one you need here is `get_forced_payments()`.

rwrzesien:
@cameel In #982 I have replaced usage of `get_forced_payments` to `get_forced_subtask_payments` for the `TransactionType.FORCE` in `get_list_of_payments()` call. As I understand it is correct?

So that new call should happen in Bankster too, but before getting list of forced and batch transactions?

cameel:
Yes. It's one of the checks before doing the calculations but since it needs information from the blockchain, it must be inside Bankster. You can define a new exception class to report it.

You need to look for any settlement payments (I'm going to start calling forced payments returned by `get_forced_payments()` _settlement payments_ from now on to make it clearer which ones I mean) that are after the earliest submitted acceptance. If there are any, refuse to run the algorithm and return `TimestampError`.

PaweuB:
test

cameel:
Update: After discussion with Łukasz it turned out that the acceptances from before the last settlement should be allowed after all. That's because the list of forced payments we take into account is supposed to contain settlement payments, not forced subtask payments.

PR #1001, opened 2018-12-06
Concent should reject force payment request if it includes acceptances from before the last settlement

pawelkisielewicz:
Resolves https://github.com/golemfactory/concent/issues/984

### DIFF
--- a/concent_api/core/exceptions.py
+++ b/concent_api/core/exceptions.py
@@ -77,5 +77,5 @@ class TooSmallProviderDeposit(Exception):
     pass
 
 
-class BanksterTimestampError(Exception):
+class PaymentTimestampError(Exception):
     pass
--- a/concent_api/core/message_handlers.py
+++ b/concent_api/core/message_handlers.py
@@ -41,9 +41,9 @@ from common.logging import log
 from common.validations import validate_secure_hash_algorithm
 from common import logging
 from conductor.tasks import result_transfer_request
-from core.exceptions import PaymentTimestampError
 from core.exceptions import CreateModelIntegrityError
 from core.exceptions import Http400
+from core.exceptions import PaymentTimestampError
 from core.exceptions import TooSmallProviderDeposit
 from core.models import Client
 from core.models import PaymentInfo
--- a/concent_api/core/message_handlers.py
+++ b/concent_api/core/message_handlers.py
@@ -51,7 +51,6 @@ from core.models import PendingResponse
 from core.models import StoredMessage
 from core.models import Subtask
 from core.payments import bankster
-from core.payments.service import are_all_payment_ts_younger_than_last_payment_closure_time_if_payment_exists
 from core.queue_operations import send_blender_verification_request
 from core.subtask_helpers import are_keys_and_addresses_unique_in_message_subtask_results_accepted
 from core.subtask_helpers import are_subtask_results_accepted_messages_signed_by_the_same_requestor
@@ -803,16 +802,6 @@ def handle_send_force_payment(
         *client_message.subtask_results_accepted_list,
         *[subtask_results_accepted.task_to_compute for subtask_results_accepted in client_message.subtask_results_accepted_list],
     )
-
-    if not are_all_payment_ts_younger_than_last_payment_closure_time_if_payment_exists(  # pylint: disable=no-value-for-parameter
-        requestor_eth_address=requestor_eth_address,
-        provider_eth_address=provider_eth_address,
-        subtask_results_accepted_list=client_message.subtask_results_accepted_list
-    ):
-        return message.concents.ForcePaymentRejected(
-            force_payment=client_message,
-            reason=message.concents.ForcePaymentRejected.REASON.TimestampError,
-        )
 
     try:
         claim_against_requestor = bankster.settle_overdue_acceptances(
--- a/concent_api/core/message_handlers.py
+++ b/concent_api/core/message_handlers.py
@@ -41,7 +41,7 @@ from common.logging import log
 from common.validations import validate_secure_hash_algorithm
 from common import logging
 from conductor.tasks import result_transfer_request
-from core.exceptions import BanksterTimestampError
+from core.exceptions import PaymentTimestampError
 from core.exceptions import CreateModelIntegrityError
 from core.exceptions import Http400
 from core.exceptions import TooSmallProviderDeposit
@@ -810,7 +810,7 @@ def handle_send_force_payment(
             acceptances=client_message.subtask_results_accepted_list,
             requestor_public_key=requestor_public_key,
         )
-    except BanksterTimestampError:
+    except PaymentTimestampError:
         return message.concents.ForcePaymentRejected(
             force_payment=client_message,
             reason=message.concents.ForcePaymentRejected.REASON.TimestampError,
--- a/concent_api/core/message_handlers.py
+++ b/concent_api/core/message_handlers.py
@@ -51,6 +51,7 @@ from core.models import PendingResponse
 from core.models import StoredMessage
 from core.models import Subtask
 from core.payments import bankster
+from core.payments.service import are_all_payment_ts_younger_than_last_payment_closure_time_if_payment_exists
 from core.queue_operations import send_blender_verification_request
 from core.subtask_helpers import are_keys_and_addresses_unique_in_message_subtask_results_accepted
 from core.subtask_helpers import are_subtask_results_accepted_messages_signed_by_the_same_requestor
@@ -802,6 +803,16 @@ def handle_send_force_payment(
         *client_message.subtask_results_accepted_list,
         *[subtask_results_accepted.task_to_compute for subtask_results_accepted in client_message.subtask_results_accepted_list],
     )
+
+    if not are_all_payment_ts_younger_than_last_payment_closure_time_if_payment_exists(  # pylint: disable=no-value-for-parameter
+        requestor_eth_address=requestor_eth_address,
+        provider_eth_address=provider_eth_address,
+        subtask_results_accepted_list=client_message.subtask_results_accepted_list
+    ):
+        return message.concents.ForcePaymentRejected(
+            force_payment=client_message,
+            reason=message.concents.ForcePaymentRejected.REASON.TimestampError,
+        )
 
     try:
         claim_against_requestor = bankster.settle_overdue_acceptances(
--- a/concent_api/core/payments/backends/mock.py
+++ b/concent_api/core/payments/backends/mock.py
@@ -58,7 +58,7 @@ def call_on_confirmed_transaction(
     pass
 
 
-def validate_that_there_is_no_younger_payment_then_any_of_closure_times(
+def validate_that_last_closure_time_is_older_than_oldest_payment(
     requestor_eth_address: str,  # pylint: disable=unused-argument
     provider_eth_address: str,  # pylint: disable=unused-argument
     oldest_payments_ts: int,  # pylint: disable=unused-argument
--- a/concent_api/core/payments/backends/mock.py
+++ b/concent_api/core/payments/backends/mock.py
@@ -56,3 +56,11 @@ def call_on_confirmed_transaction(
     _callback: Callable,
 ) -> None:
     pass
+
+
+def are_all_payment_ts_younger_than_last_payment_closure_time_if_payment_exists(
+    _requestor_eth_address: str,
+    _provider_eth_address: str,
+    _subtask_results_accepted_list: list,
+) -> bool:
+    return True
--- a/concent_api/core/payments/backends/mock.py
+++ b/concent_api/core/payments/backends/mock.py
@@ -59,7 +59,7 @@ def call_on_confirmed_transaction(
     pass
 
 
-def validate_that_all_payment_ts_are_younger_than_last_payment_closure_time_if_payment_exists(
+def validate_that_last_closure_time_is_older_than_oldest_payment(
     _requestor_eth_address: str,
     _provider_eth_address: str,
     _oldest_payments_ts: Optional[int],
--- a/concent_api/core/payments/backends/mock.py
+++ b/concent_api/core/payments/backends/mock.py
@@ -56,11 +56,3 @@ def call_on_confirmed_transaction(
     callback: Callable,  # pylint: disable=unused-argument
 ) -> None:
     pass
-
-
-def validate_that_last_closure_time_is_older_than_oldest_payment(
-    requestor_eth_address: str,  # pylint: disable=unused-argument
-    provider_eth_address: str,  # pylint: disable=unused-argument
-    oldest_payments_ts: int,  # pylint: disable=unused-argument
-) -> None:
-    pass
--- a/concent_api/core/payments/backends/mock.py
+++ b/concent_api/core/payments/backends/mock.py
@@ -1,5 +1,6 @@
 import uuid
 from typing import Callable
+from typing import Optional
 
 from core.constants import CLIENT_ETH_ADDRESS_WITH_0_DEPOSIT
 from core.payments.backends.sci_backend import TransactionType
@@ -58,9 +59,9 @@ def call_on_confirmed_transaction(
     pass
 
 
-def are_all_payment_ts_younger_than_last_payment_closure_time_if_payment_exists(
+def validate_that_all_payment_ts_are_younger_than_last_payment_closure_time_if_payment_exists(
     _requestor_eth_address: str,
     _provider_eth_address: str,
-    _subtask_results_accepted_list: list,
-) -> bool:
-    return True
+    _oldest_payments_ts: Optional[int],
+) -> None:
+    pass
--- a/concent_api/core/payments/backends/mock.py
+++ b/concent_api/core/payments/backends/mock.py
@@ -1,6 +1,5 @@
 import uuid
 from typing import Callable
-from typing import Optional
 
 from core.constants import CLIENT_ETH_ADDRESS_WITH_0_DEPOSIT
 from core.payments.backends.sci_backend import TransactionType
@@ -53,15 +52,15 @@ def cover_additional_verification_cost(
 
 
 def call_on_confirmed_transaction(
-    _tx_hash: str,
-    _callback: Callable,
+    tx_hash: str,  # pylint: disable=unused-argument
+    callback: Callable,  # pylint: disable=unused-argument
 ) -> None:
     pass
 
 
-def validate_that_last_closure_time_is_older_than_oldest_payment(
-    _requestor_eth_address: str,
-    _provider_eth_address: str,
-    _oldest_payments_ts: Optional[int],
+def validate_that_there_is_no_younger_payment_then_any_of_closure_times(
+    requestor_eth_address: str,  # pylint: disable=unused-argument
+    provider_eth_address: str,  # pylint: disable=unused-argument
+    oldest_payments_ts: int,  # pylint: disable=unused-argument
 ) -> None:
     pass
--- a/concent_api/core/payments/backends/sci_backend.py
+++ b/concent_api/core/payments/backends/sci_backend.py
@@ -4,6 +4,7 @@ from enum import Enum
 
 from typing import Callable
 from typing import List
+from typing import Optional
 
 from golem_sci import ForcedPaymentEvent
 from golem_sci import SmartContractsInterface
@@ -11,6 +12,7 @@ from golem_sci.blockshelper import BlocksHelper
 from web3 import Web3
 
 from core.constants import ETHEREUM_ADDRESS_LENGTH
+from core.exceptions import BanksterTimestampError
 from core.payments.payment_interface import PaymentInterface
 from core.validation import validate_uuid
 
@@ -53,31 +55,27 @@ def get_list_of_payments(
             from_block      = last_block_before_payment,
             to_block        = payment_interface.get_block_number(),  # pylint: disable=no-member
         )
-
     return payments_list
 
 
-def are_all_payment_ts_younger_than_last_payment_closure_time_if_payment_exists(
+def validate_that_all_payment_ts_are_younger_than_last_payment_closure_time_if_payment_exists(
     requestor_eth_address: str,
     provider_eth_address: str,
-    subtask_results_accepted_list: List[ForcedPaymentEvent],
-) -> bool:
-
-    oldest_payments_ts = _get_oldest_payment_timestamp_from_subtask_results_accepted_list(subtask_results_accepted_list)
-
+    oldest_payment_ts: Optional[int],
+) -> None:
+    if oldest_payment_ts is None:
+        return
     forced_payment_event_list = _get_list_of_forced_payment_events(
         requestor_eth_address=requestor_eth_address,
         provider_eth_address=provider_eth_address,
-        search_payments_since_ts=oldest_payments_ts,
+        search_payments_since_ts=oldest_payment_ts,
     )
-
     for forced_payment_event in forced_payment_event_list:
-        if oldest_payments_ts < forced_payment_event.closure_time:
-            return False
-    return True
+        if oldest_payment_ts < forced_payment_event.closure_time:
+            raise BanksterTimestampError
 
 
-def _get_oldest_payment_timestamp_from_subtask_results_accepted_list(subtask_results_accepted_list: list) -> int:
+def get_oldest_payment_timestamp_from_subtask_results_accepted_list(subtask_results_accepted_list: list) -> Optional[int]:
     return min(subtask_results_accepted.payment_ts for subtask_results_accepted in subtask_results_accepted_list)
 
 
--- a/concent_api/core/payments/backends/sci_backend.py
+++ b/concent_api/core/payments/backends/sci_backend.py
@@ -8,7 +8,6 @@ from golem_sci.blockshelper import BlocksHelper
 from web3 import Web3
 
 from core.constants import ETHEREUM_ADDRESS_LENGTH
-from core.exceptions import BanksterTimestampError
 from core.payments.payment_interface import PaymentInterface
 from core.validation import validate_uuid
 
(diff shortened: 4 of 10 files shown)


concent_api/core/message_handlers.py
```diff
@@ -803,6 +804,16 @@ def handle_send_force_payment(
         *[subtask_results_accepted.task_to_compute for subtask_results_accepted in client_message.subtask_results_accepted_list],
     )
 
+    if not are_all_payment_ts_younger_than_last_payment_closure_time_if_payment_exists(  # pylint: disable=no-value-for-parameter
```
rwrzesien: Only Bankster should call SCI methods. This has to be moved to Bankster and raise exception catched by handler, then return `ForcePaymentRejected`.

In the issue description there is
> Yes. It's one of the checks before doing the calculations but since it needs information from the blockchain, it must be inside Bankster.

concent_api/core/payments/backends/sci_backend.py
```diff
@@ -55,6 +57,46 @@ def get_list_of_payments(
...
+    )
+
+    for forced_payment_event in forced_payment_event_list:
+        if oldest_payments_ts < forced_payment_event.closure_time:
+            return False
+    return True
+
+
+def _get_oldest_payment_timestamp_from_subtask_results_accepted_list(subtask_results_accepted_list: list) -> int:
+    return min(subtask_results_accepted.payment_ts for subtask_results_accepted in subtask_results_accepted_list)
```
rwrzesien: There is no need to create as simple one liner as a separate function if it is used only once.

dybi: @rwrzesien, have you watched Uncle Bob's presentation? ;) 

It is ok (or even: is is _desired_), to extract some logical functionality to a separate function, regardless of how many times it is called

dybi:
One generall remark - please adhere to our branch naming rules: as this is a bug fix, branch name should start with: `bugfix-`

concent_api/core/payments/backends/sci_backend.py
```diff
@@ -55,6 +57,46 @@ def get_list_of_payments(
...
+
+    oldest_payments_ts = _get_oldest_payment_timestamp_from_subtask_results_accepted_list(subtask_results_accepted_list)
+
+    forced_payment_event_list = _get_list_of_forced_payment_events(
+        requestor_eth_address=requestor_eth_address,
+        provider_eth_address=provider_eth_address,
+        search_payments_since_ts=oldest_payments_ts,
+    )
+
+    for forced_payment_event in forced_payment_event_list:
```
rwrzesien: This can be changed to 
```
return all([oldest_payments_ts < forced_payment_event.closure_time for forced_payment_event in forced_payment_event_list])
```

concent_api/core/payments/backends/sci_backend.py
```diff
@@ -55,6 +57,46 @@ def get_list_of_payments(
     return payments_list
 
 
+def are_all_payment_ts_younger_than_last_payment_closure_time_if_payment_exists(
```
dybi: This name should be changed as well:
`is_last_closure_time_older_than_oldest_payment` and if yes we should reject. It seems simpler that way, at least to me

concent_api/core/payments/backends/sci_backend.py
```diff
@@ -55,6 +57,46 @@ def get_list_of_payments(
...
+    provider_eth_address: str,
+    subtask_results_accepted_list: List[ForcedPaymentEvent],
+) -> bool:
+
+    oldest_payments_ts = _get_oldest_payment_timestamp_from_subtask_results_accepted_list(subtask_results_accepted_list)
+
+    forced_payment_event_list = _get_list_of_forced_payment_events(
+        requestor_eth_address=requestor_eth_address,
+        provider_eth_address=provider_eth_address,
+        search_payments_since_ts=oldest_payments_ts,
```
dybi: you should call with with the `youngest_payment_ts` from my PoV.

concent_api/core/payments/backends/sci_backend.py
dybi: but we are not interested in all `force_payment_events`, but it the last one ;)

concent_api/core/payments/backends/sci_backend.py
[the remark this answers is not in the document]
rwrzesien: Opposite ;)

I was born in 1901, you were born in 1902, who is older? So it should be `min`, unless it should be "younger".

concent_api/core/payments/backends/sci_backend.py
cameel: @dybi I agree.
It would be wrong because the comprehension returns `False` only if all the settlement payments found have `closure_time` after the oldest acceptance. Concent must reject the settlement request if even one of them satisfies this condition.

But I also agree with @rwrzesien that it should be checked with a compact list comprehension:
``` python
return not any([forced_payment_event.closure_time >= oldest_payments_ts for forced_payment_event in forced_payment_event_list])
```

Also, note that it should use `>=`, not `>`. The settlement payment settles acceptances up to `closure_time`, including that time.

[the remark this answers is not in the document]
cameel: I don't see a bug here. The code fetches all settlement payments with block timestamp later or equal to `oldest_payments_ts` which means all the ones with `closure_time >= oldest_payments_ts` (the ones we're actually interested in) and possibly some with earlier `closure_time`. We can be sure that all the existing settlement payments with `closure_time >= oldest_payments_ts` are on our list because the contract validates that `closure_time` <= block timestamp. Then the loop goes over all the events and checks if any of them actually has `closure_time >= oldest_payments_ts`. If so, returns `False`.

Note that if `closure_time ==` block timestamp was always true, we wouldn't even need a loop - we could just check if the list is empty or not. Unfortunately that's not the case the `closure_time` does not have to match block timestamp.

BTW, @pawelkisielewicz it's called `payment_ts` not `payments_ts` in golem-messages. Please use that name for consistency.

concent_api/core/payments/backends/sci_backend.py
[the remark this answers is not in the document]
cameel: Yeah, @rwrzesien is right.

I agree that it's really confusing though, especially when writing or reading a spec. I try to use "earliest" where I can but unfortunately it sounds weird in some contexts.

concent_api/core/message_handlers.py

concent_api/core/payments/backends/sci_backend.py
pawelkisielewicz: Implementation is changed, now function 'validates', so nothing is returned. It might only raise `BanksterTimestampError`

concent_api/core/payments/backends/sci_backend.py
```diff
@@ -55,6 +57,46 @@ def get_list_of_payments(
     return payments_list
 
 
+def are_all_payment_ts_younger_than_last_payment_closure_time_if_payment_exists(
+    requestor_eth_address: str,
+    provider_eth_address: str,
+    subtask_results_accepted_list: List[ForcedPaymentEvent],
```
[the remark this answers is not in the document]
pawelkisielewicz: done

concent_api/core/payments/backends/sci_backend.py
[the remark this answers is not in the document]
pawelkisielewicz: renamed

concent_api/core/payments/backends/sci_backend.py
pawelkisielewicz: done, it is validation now so I called it `validate_that_last_closure_time_is_older_than_oldest_payment`

concent_api/core/payments/backends/sci_backend.py
pawelkisielewicz: good catch :+1: done

concent_api/core/payments/backends/sci_backend.py
```diff
@@ -58,25 +58,25 @@ def get_list_of_payments(
...
     )
     for forced_payment_event in forced_payment_event_list:
-        if oldest_payment_ts < forced_payment_event.closure_time:
+        if youngest_payment_ts < forced_payment_event.closure_time:
             raise BanksterTimestampError
 
 
-def get_oldest_payment_timestamp_from_subtask_results_accepted_list(subtask_results_accepted_list: list) -> Optional[int]:
-    return min(subtask_results_accepted.payment_ts for subtask_results_accepted in subtask_results_accepted_list)
+def get_youngest_payment_timestamp_from_subtask_results_accepted_list(subtask_results_accepted_list: list) -> Optional[int]:
```
dybi: why is the return value `Optional`? From what I saw in the code, the list cannot be empty (because the code would have failed already), but even if it was, `max()`  would raise `ValueError`

concent_api/core/payments/backends/sci_backend.py
```diff
@@ -58,25 +58,25 @@ def get_list_of_payments(
     return payments_list
 
 
-def validate_that_all_payment_ts_are_younger_than_last_payment_closure_time_if_payment_exists(
+def validate_that_last_closure_time_is_older_than_last_payment(
     requestor_eth_address: str,
     provider_eth_address: str,
-    oldest_payment_ts: Optional[int],
+    youngest_payment_ts: Optional[int],
```
dybi: How could this be `None`? From my PoV in the current flow it cannot.

concent_api/core/payments/backends/sci_backend.py
```diff
@@ -58,25 +58,25 @@ def get_list_of_payments(
...
         return
     forced_payment_event_list = _get_list_of_forced_payment_events(
         requestor_eth_address=requestor_eth_address,
         provider_eth_address=provider_eth_address,
-        search_payments_since_ts=oldest_payment_ts,
+        search_payments_since_ts=youngest_payment_ts,
     )
     for forced_payment_event in forced_payment_event_list:
-        if oldest_payment_ts < forced_payment_event.closure_time:
+        if youngest_payment_ts < forced_payment_event.closure_time:
```
dybi: again, the function says something and than does something else: you are not comparing `last_payment` (`youngest_payment_ts`) with **last** `closure_time`, by with a **list** of closure times... Please change the implementation

concent_api/core/payments/backends/mock.py
```diff
@@ -56,3 +57,11 @@ def call_on_confirmed_transaction(
     _callback: Callable,
 ) -> None:
     pass
+
+
+def validate_that_last_closure_time_is_older_than_oldest_payment(
+    _requestor_eth_address: str,
```
rwrzesien: Please keep names the same. If someone will use it with naming parameters he will get an exception. Please also fix the function above.

concent_api/core/payments/backends/sci_backend.py
```diff
@@ -51,10 +55,46 @@ def get_list_of_payments(
...
     return payments_list
 
 
+def validate_that_last_closure_time_is_older_than_last_payment(
+    requestor_eth_address: str,
+    provider_eth_address: str,
+    youngest_payment_ts: Optional[int],
+) -> None:
+    if youngest_payment_ts is None:
+        return
```
rwrzesien: After fixing @dybi comment above this should be removed, but either way we `return None` explicitly.

concent_api/core/payments/backends/sci_backend.py
rwrzesien: Why it is raising exception from bankster? It should have it own, if any. I think it should return universal `True/False` than depend on exception.

concent_api/core/payments/backends/sci_backend.py
```diff
@@ -51,10 +55,46 @@ def get_list_of_payments(
...
+    for forced_payment_event in forced_payment_event_list:
+        if youngest_payment_ts < forced_payment_event.closure_time:
+            raise BanksterTimestampError
+
+
+def get_youngest_payment_timestamp_from_subtask_results_accepted_list(subtask_results_accepted_list: list) -> Optional[int]:
+    return max(subtask_results_accepted.payment_ts for subtask_results_accepted in subtask_results_accepted_list)
+
+
+def _get_list_of_forced_payment_events(
```
rwrzesien: This function does exactly the same as `get_list_of_payments` called with `transaction_type=TransactionType.FORCE`.

concent_api/core/payments/service.py
```diff
@@ -100,3 +101,15 @@ def call_on_confirmed_transaction(
...
+
+@_add_backend
+def validate_that_last_closure_time_is_older_than_oldest_payment(
+    backend: Any,
+    requestor_eth_address: str,
+    provider_eth_address: str,
+    search_payments_since_ts: Optional[int],
+) -> bool:
+    return backend.validate_that_last_closure_time_is_older_than_oldest_payment(
+        requestor_eth_address, provider_eth_address, search_payments_since_ts
```
rwrzesien: Split arguments into separate lines, please also fix function above.

concent_api/core/tests/test_payments_sci_backend.py
```diff
@@ -254,3 +268,27 @@ def test_that_sci_backend_cover_additional_verification_cost_should_return_trans
...
+        last_payment_closure_time_timestamp = get_current_utc_timestamp() - 3600 * 24 * 10  # 10 days ago
+        with mock.patch('core.payments.backends.sci_backend._get_list_of_forced_payment_events', return_value=[]):
+            sci_backend.validate_that_last_closure_time_is_older_than_last_payment(
+                'any_key', 'any_key', last_payment_closure_time_timestamp
+            )
+
+    def test_that_if_there_is_no_payment_younger_then_oldest_payment_timestamp_in_subtask_results_accepted_validation_should_pass(self):  # pylint: disable=no-self-use
+        last_payment_closure_time_timestamp = get_current_utc_timestamp() - 3600 * 24 * 10  # 10 days ago
+        with mock.patch('core.payments.backends.sci_backend._get_list_of_forced_payment_events',
+                        return_value=_prepare_forced_payment_event_list_without_younger_payments(last_payment_closure_time_timestamp)):
```
rwrzesien: This line breaking is incorrect, same below.

concent_api/core/payments/backends/sci_backend.py
pawelkisielewicz: right, changed

concent_api/core/payments/backends/sci_backend.py
pawelkisielewicz: I think it is `None` during first contact of pair requestor and provider with concent. If there is no previous payment, there is no it's timestamp as well. Do you agree?

concent_api/core/payments/backends/sci_backend.py
pawelkisielewicz: renamed, is is ok now?

concent_api/core/payments/backends/mock.py
pawelkisielewicz: done

concent_api/core/payments/backends/sci_backend.py
pawelkisielewicz: OK, thanks

concent_api/core/payments/backends/sci_backend.py
dybi: quite frankly I would prefer having the previous name and the code changed:
why not just
```python
if youngest_payment_ts < max([event.closure_time for events in forced_payment_event_list]):
    raise BanksterTimestampError
```
instead of the `for` loop? This way name of the function stays more meaningful and the implementation better reflects the nature of a problem.

concent_api/core/payments/backends/sci_backend.py
```diff
@@ -3,12 +3,12 @@
 from enum import Enum
 
 from typing import Callable
-
 from golem_sci import SmartContractsInterface
 from golem_sci.blockshelper import BlocksHelper
 from web3 import Web3
 
 from core.constants import ETHEREUM_ADDRESS_LENGTH
+from core.exceptions import BanksterTimestampError
```
Jakub89: I've deleted that exception in my PR, because it wasn't use anywhere anymore, but I see it now here. I just want to point that out. It will cause some conflicts for me or You, depended on which PR is going to be merged first.

concent_api/core/payments/backends/sci_backend.py
pawelkisielewicz: ok, done

concent_api/core/payments/validation.py
```diff
@@ -0,0 +1,20 @@
...
+    provider_eth_address: str,
+    youngest_payment_ts: int,
+) -> None:
+    forced_payment_event_list = get_list_of_payments(
+        requestor_eth_address=requestor_eth_address,
+        provider_eth_address=provider_eth_address,
+        payment_ts=youngest_payment_ts,
+        transaction_type=TransactionType.FORCE,
+    )
+    if forced_payment_event_list == []:
```
dybi: please change it to:
```python
if forced_payment_event_list != []:
 if youngest_payment_ts < max([event.closure_time for event in forced_payment_event_list]):
        raise PaymentTimestampError
```
 you will not have to add unnecessary `return` statements and disable pylint :>

concent_api/core/exceptions.py
```diff
@@ -77,5 +77,5 @@ class TooSmallProviderDeposit(Exception):
     pass
 
 
-class BanksterTimestampError(Exception):
+class PaymentTimestampError(Exception):
```
dybi: I like the new name better :+1:

concent_api/core/payments/backends/mock.py
```diff
@@ -52,7 +52,15 @@ def cover_additional_verification_cost(
...
 def call_on_confirmed_transaction(
-    _tx_hash: str,
-    _callback: Callable,
+    tx_hash: str,  # pylint: disable=unused-argument
+    callback: Callable,  # pylint: disable=unused-argument
+) -> None:
+    pass
+
+
+def validate_that_last_closure_time_is_older_than_oldest_payment(
```
rwrzesien: It shouldn't be here anymore.

concent_api/core/payments/service.py
```diff
@@ -99,4 +99,21 @@ def call_on_confirmed_transaction(
...
 ) -> None:
-    backend.call_on_confirmed_transaction(tx_hash, callback)
+    backend.call_on_confirmed_transaction(
+        tx_hash,
+        callback,
+    )
+
+
+@_add_backend
+def validate_that_last_closure_time_is_older_than_oldest_payment(
```
rwrzesien: Same here.

concent_api/core/payments/validation.py
rwrzesien: And also in @dybi version those two conditions can be combined into one.

pawelkisielewicz: done

concent_api/core/payments/backends/mock.py
pawelkisielewicz: I think it should stay. This function calls `get_list_of_payments` and in unit tests we don't want to call `get_list_of_payments`, just mock it and pass validation

concent_api/core/payments/service.py
pawelkisielewicz: Same as above

concent_api/core/payments/validation.py
dybi: @rwrzesien why do you insist on combining the conditions? Isn't it readable the way it is? If we combine those two, the line will become really long and/or will require weird formatting

concent_api/core/message_handlers.py
```diff
@@ -41,7 +41,7 @@
 from common.validations import validate_secure_hash_algorithm
 from common import logging
 from conductor.tasks import result_transfer_request
-from core.exceptions import BanksterTimestampError
+from core.exceptions import PaymentTimestampError
```
kbeker: Alphabetical order

concent_api/core/payments/bankster.py
```diff
@@ -21,14 +21,16 @@
 from common.helpers import get_current_utc_timestamp
 from common.logging import log
 from core.constants import ETHEREUM_ADDRESS_LENGTH
-from core.exceptions import BanksterTimestampError
+from core.exceptions import PaymentTimestampError
```
kbeker: Alphabetical order

concent_api/core/payments/bankster.py
```diff
@@ -21,14 +21,16 @@
...
 from core.exceptions import TooSmallProviderDeposit
 from core.models import Client
 from core.models import DepositAccount
 from core.models import DepositClaim
 from core.models import Subtask
 from core.payments import service
+from core.payments.backends.sci_backend import get_youngest_payment_timestamp_from_subtask_results_accepted_list
 from core.payments.backends.sci_backend import TransactionType
+from core.payments.service import validate_that_last_closure_time_is_older_than_oldest_payment
```
kbeker: You can use this function from import which is 3 lines higher

concent_api/core/payments/backends/sci_backend.py
```diff
@@ -141,3 +139,7 @@ def _hexencode_uuid(value: str) -> bytes:
     validate_uuid(value)
 
     return binascii.hexlify(uuid.UUID(value).bytes)
+
+
+def get_youngest_payment_timestamp_from_subtask_results_accepted_list(subtask_results_accepted_list: list) -> int:
+    return max(subtask_results_accepted.payment_ts for subtask_results_accepted in subtask_results_accepted_list)
```
kbeker: As we spoke. This function should be in bankster, because `sci_backend` is a place for functions which are related just with `PaymentInterface`

concent_api/core/payments/backends/mock.py
rwrzesien: But the purpose of this backend is different. This is mocked backend, it must be consistent with sci backend. If you want to create a reusable mock place it in `tests/utils.py`.

concent_api/core/payments/validation.py
```diff
@@ -0,0 +1,18 @@
...
+    provider_eth_address: str,
+    search_payments_since_ts: int,
+) -> None:
+    forced_payment_event_list = get_list_of_payments(
+        requestor_eth_address=requestor_eth_address,
+        provider_eth_address=provider_eth_address,
+        payment_ts=search_payments_since_ts,
+        transaction_type=TransactionType.FORCE,
+    )
+    if forced_payment_event_list != [] and search_payments_since_ts < max([event.closure_time for event in forced_payment_event_list]):
```
rwrzesien: Please split this line.

concent_api/core/payments/service.py
rwrzesien: Each of this functions must also exists in each backends. As it doesn't exist in `sci_backend`, it can't exist here.

concent_api/core/payments/backends/mock.py
pawelkisielewicz: removed

concent_api/core/payments/service.py
pawelkisielewicz: removed, this function simply uses other 'backend' function and this function is mocked

concent_api/core/message_handlers.py

concent_api/core/payments/bankster.py
pawelkisielewicz: ok  now..

concent_api/core/payments/bankster.py
pawelkisielewicz: ok, but now it is imported from `core.payments.validation`

concent_api/core/payments/backends/sci_backend.py
pawelkisielewicz: done

concent_api/core/payments/validation.py
pawelkisielewicz: done

cameel:
@pawelkisielewicz Please note that the issue is closed now. This fix should not be merged.

There's still some useful code here (code refactoring, new tests) and it's fine to merge that but please do it in a new pull request. The name of this one would be misleading otherwise.

pawelkisielewicz:
@cameel  I checked this code and actually there is nothing to merge now.. I didn't refactor in this PR and tests were only for function, which won't be used no more. But I will keep this branch locally, maybe after some change this code will help in future.

cameel:
- What about renaming `BanksterTimeoutError`?
- The three new tests are just fine. You only need to change `test_that_if_there_is_payment_younger_then_oldest_payment_timestamp_in_subtask_results_accepted_validation_should_raise_exception()` so that it does not expect any error. We do need tests that check what happens if there are payments earlier than T0.
- What about warnings disabled in the mock backend?

concent_api/core/payments/validation.py
```diff
@@ -0,0 +1,19 @@
...
+    search_payments_since_ts: int,
+) -> None:
+    forced_payment_event_list = get_list_of_payments(  # pylint: disable=no-value-for-parameter
+        requestor_eth_address=requestor_eth_address,
+        provider_eth_address=provider_eth_address,
+        payment_ts=search_payments_since_ts,
+        transaction_type=TransactionType.FORCE,
+    )
+    if forced_payment_event_list != [] and search_payments_since_ts < max([event.closure_time
+                                                                           for event in forced_payment_event_list]):
```
cameel: The pull request is now closed but I just spotted it so I'll post anyway.

This wrapping is against our coding style. In my opinion this way of wrapping things is not very readable and indenting code like this is not very convenient. I'd do it as:

``` python
if forced_payment_event_list != [] and search_payments_since_ts < max(
    [event.closure_time for event in forced_payment_event_list]
):
```

or, better:
``` python
closure_times = [event.closure_time for event in forced_payment_event_list]

if len(closure_times) > 0 and search_payments_since_ts < max(closure_times):
```

You also generally should not compare with `[]`. What if it was not a `list` but some type inherited from it? Or a `tuple`? A set? `len(forced_payment_event_list) > 0` is way more robust and just as readable.

concent_api/core/tests/test_payments_sci_backend.py
```diff
@@ -254,3 +267,31 @@ def test_that_sci_backend_cover_additional_verification_cost_should_return_trans
...
+
+    def test_that_if_there_is_payment_younger_then_oldest_payment_timestamp_in_subtask_results_accepted_validation_should_raise_exception(self):
+        last_payment_closure_time_timestamp = get_current_utc_timestamp() - 3600 * 24 * 10  # 10 days ago
+        with mock.patch(
+            'core.payments.validation.get_list_of_payments',
+            return_value=_prepare_forced_payment_event_list_with_younger_payments(
+                last_payment_closure_time_timestamp)):
+            with self.assertRaises(PaymentTimestampError):
+                validate_that_last_closure_time_is_older_than_oldest_payment(
+                    'any_key', 'any_key', last_payment_closure_time_timestamp
```
cameel: Same here. It's hard to see where the `patch()` ends and where the code inside the `with` starts. According to our style it should be:

``` python
        with mock.patch(
            'core.payments.validation.get_list_of_payments',
            return_value=_prepare_forced_payment_event_list_with_younger_payments(
                last_payment_closure_time_timestamp
            )
        ):
            with self.assertRaises(PaymentTimestampError):
                validate_that_last_closure_time_is_older_than_oldest_payment(
                    'any_key',
                    'any_key',
                    last_payment_closure_time_timestamp
                )
```

Even though the difference is pretty minimal, you can immediately see where one nesting level ends and another begins.

Issue #1006, opened 2018-12-07
The settlement algorithm does not accept acceptances signed by Concent


cameel:
As explained in the #996 blueprint, the provider must be able to submit acceptances signed not only by the requestor but also by Concent. Otherwise he won't be able to prove that he performed the work paid by forced subtask payments and Concent will count those payments towards other subtasks which is wrong.

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/162505388

cameel:
Probably. But please give me some time to update the spec and issues. I don't want to confirm anything until I get a clear picture of the situation.

cameel:
I'm closing it. Concent does not issue acceptances in UC3 and UC4 after all. It's not necessary because UC5 is not supposed to take forced subtask payments into account as I previously thought.

PR #1010, opened 2018-12-11
Bugfix concent should accept SubtaskResultsAccepted signed by concent

dybi:
resolves: https://github.com/golemfactory/concent/issues/1006


concent_api/core/subtask_helpers.py
```diff
@@ -280,17 +280,24 @@ def are_keys_and_addresses_unique_in_message_subtask_results_accepted(
     )
 
 
-def are_subtask_results_accepted_messages_signed_by_the_same_requestor(
+def are_subtask_results_accepted_messages_properly_signed(
     subtask_results_accepted_list: List[SubtaskResultsAccepted]
 ) -> bool:
     requestor_public_key = subtask_results_accepted_list[0].task_to_compute.requestor_public_key
```
pawelkisielewicz: Only one small question- is it not good place for assert, that all `subtask_results_accepted` are for one requestor? I know they should be and you hate assertions, but if we get this key from one message I suggest to make 100% that all keys are the same. I can avoid hours of debugging

concent_api/core/tests/test_unit_validation.py
```diff
@@ -49,7 +51,7 @@
 (DIFFERENT_REQUESTOR_PRIV_ETH_KEY, DIFFERENT_REQUESTOR_PUB_ETH_KEY) = generate_priv_and_pub_eth_account_key()
 
 
-class TestValidateGolemMessageSubtaskResultsRejected(TestCase):
+class TestValidateGolemMessageSubtaskResultsRejected(unittest.TestCase):
```
pawelkisielewicz: why is it changed?

concent_api/core/tests/test_unit_validation.py
```diff
@@ -266,11 +272,21 @@ def test_that_if_messages_are_signed_by_different_requestors_method_should_retur
             self.task_to_compute_2,
             subtask_2_signed_by=DIFFERENT_REQUESTOR_PRIVATE_KEY,
         )
-        result = are_subtask_results_accepted_messages_signed_by_the_same_requestor(subtask_results_accepted_list)
+        result = are_subtask_results_accepted_messages_properly_signed(subtask_results_accepted_list)
         assert_that(result).is_false()
 
+    def test_that_if_messages_are_signed_by_concent_method_should_return_true(self):
```
pawelkisielewicz: Maybe also for case, when one message is signed by concent and second by requestor?

concent_api/core/message_handlers.py
```diff
@@ -805,12 +805,6 @@ def handle_send_force_payment(
     validate_ethereum_addresses(requestor_eth_address, provider_eth_address)
     requestor_ethereum_public_key = hex_to_bytes_convert(task_to_compute.requestor_ethereum_public_key)
 
-    validate_that_golem_messages_are_signed_with_key(
```
rwrzesien: Why are you removing this? This is not mentioned in issue.

concent_api/core/message_handlers.py
```diff
@@ -765,7 +765,7 @@ def handle_send_force_payment(
 
     if not (
         are_keys_and_addresses_unique_in_message_subtask_results_accepted(client_message.subtask_results_accepted_list) and
-        are_subtask_results_accepted_messages_signed_by_the_same_requestor(client_message.subtask_results_accepted_list)
+        are_subtask_results_accepted_messages_properly_signed(client_message.subtask_results_accepted_list)
```
rwrzesien: This function name would be more describing if called
```
are_subtask_results_accepted_messages_signed_by_the_same_requestor_or_concent
```

concent_api/core/tests/test_integration_force_payment.py
```diff
@@ -519,95 +527,61 @@ def test_provider_send_force_payment_with_same_subtasks_id_concent_should_refuse
...
                 'timestamp': parse_iso_date_to_timestamp("2018-02-05 12:00:20"),
                 'amount_pending': self.amount_pending,
                 'amount_paid': self.amount_paid,
             }
         )
         self._assert_stored_message_counter_not_increased()
-
+        last_pending_message = PendingResponse.objects.filter(delivered=False).order_by('created_at').last()
+        self.assertEqual(last_pending_message.response_type,
+                         PendingResponse.ResponseType.ForcePaymentCommitted.name)  # pylint: disable=no-member
```
rwrzesien: Please fix indent.

concent_api/core/tests/test_unit_validation.py
```diff
@@ -1,7 +1,9 @@
-from unittest import TestCase
```
rwrzesien: This change is not necessary and I think also against our coding style.

concent_api/core/subtask_helpers.py
cameel: Yeah. You're making an assumption that all the keys are the same (and the function will not give you the right answer if they are not) but you're not checking that. Probably some other part of the code already checks that but since this one depends on it, there should be an `assert` here.

concent_api/core/message_handlers.py
dybi: ok, I will change it.

concent_api/core/tests/test_unit_validation.py
dybi: In this module both `unittest.TestCase` and `django.test.TestCase` are used

concent_api/core/tests/test_unit_validation.py
dybi: In this module both `unittest.TestCase` and `django.test.TestCase` are used.

concent_api/core/subtask_helpers.py
dybi: it **is** checked:
```python
if not (
        are_keys_and_addresses_unique_in_message_subtask_results_accepted(client_message.subtask_results_accepted_list) and
        are_subtask_results_accepted_messages_signed_by_the_same_requestor_or_concent(client_message.subtask_results_accepted_list)
    ):
```

We don't want to have code duplication.

concent_api/core/tests/test_unit_validation.py
dybi: ok, I will add

concent_api/core/message_handlers.py
dybi: this validation was wrong and was causing newly added tests to fail. Ownership is already checked at the begining og the message handler:
```python
    try:
        for subtask_results_accepted in client_message.subtask_results_accepted_list:
            subtask_results_accepted.validate_ownership_chain(settings.CONCENT_PUBLIC_KEY)
    except InvalidSignature:
        return ServiceRefused(
            reason=ServiceRefused.REASON.InvalidRequest
        )
```

cameel: Yeah.

BTW, It's great that it's in a separate commit so it's not mixed with other changes and pretty easy to review.

The commit description could be better though. `Remove wrong validation` does not say why it's wrong and this a crucial information when reviewing a change or even trying to describe it in a changelog.

concent_api/core/message_handlers.py
rwrzesien: What about `_or_concent` part?

concent_api/core/tests/test_integration_force_payment.py
rwrzesien: Still not fixed, should be
```
self.assertEqual(
    last_pending_message.response_type,
    PendingResponse.ResponseType.ForcePaymentCommitted.name
)
```

concent_api/core/tests/test_unit_validation.py
rwrzesien: Just for other reviewers, you could always used `import X as Y`. I know you know that ;)

concent_api/core/message_handlers.py
rwrzesien: Just to clarify, what I have meant was rather "This is not mentioned in the issue, and you are removing existing validation, that look suspicious" ;)

concent_api/core/subtask_helpers.py
cameel: 1) Avoiding duplication with `assert` makes no sense. It's as if you said that having unit tests is unnecessary duplication.

    There's a simple way to tell that it's not duplication. If you have two duplicates then you should be able to use one in place of the other. Otherwise they're not really the same. So we should be able to remove the validation and leave the `assert` while keeping the code functionally the same. That's not the case here. We can't because they do different things. Validation checks your input. `assert` checks your code.

    The purpose of this `assert` is to tell the programmer (and the Python interpreter) "this function will not give you the correct answer unless this condition is true". Adding this kind of meta-information to your function does not help you validate the input. It helps you detect that someone is using it wrong. Or that someone broke some external mechanism that this function cannot properly work without (i.e. the input validation).

2) Actually, with the new name the assertion is no longer necessary in my opinion.

    When it was called `are_subtask_results_accepted_messages_properly_signed()` I would be confused if it told me that messages are properly signed even though every one has a different requestor key inside. You could avoid that by checking it inside the function with an `if`. But **that** would be duplication. The third alternative is to just say "this case is not supported; never call this function with that input" with `assert`. Problem solved.

    But now that it's called `are_subtask_results_accepted_messages_signed_by_the_same_requestor_or_concent()`, it's fine because it does exactly what it says. It only checks if the signatures are the same but does not guarantee that each message is signed with the key inside it.

3) So `are_keys_and_addresses_unique_in_message_subtask_results_accepted()` checks if the addresses and keys in all messages are the same? That's confusing. If all elements on a list are unique, it means that they're all different and that there are no duplicates. That's how `unique` works in in Django models for example. It was not a part of this pull request, but I think we should rename it (and the local variables inside it).

concent_api/core/tests/test_unit_validation.py
cameel: You could, but most of the time I prefer the way @dybi did it. Unless the new name is very close to the old one, renaming makes it harder to find it with normal text search and is confusing when you're reading just a part of the code rather than the whole file from the beginning to end.

concent_api/core/tests/test_integration_force_payment.py
```diff
@@ -1,4 +1,5 @@
 import mock
+from django.conf import settings
 from django.test            import override_settings
 from freezegun              import freeze_time
 from golem_messages         import message
```
kbeker: Please also fix style in all imports here :)

concent_api/core/tests/test_unit_validation.py
```diff
@@ -1,7 +1,9 @@
...
+import unittest
```
kbeker: Why not pytest?

Issue #1007, opened 2018-12-07
The settlement algorithm does not treat acceptances covered by an existing payment that was too low as overdue


cameel:
As explained in the #996 blueprint, an acceptance should be treated as overdue if one of the following is true:
- It's been more than `PDT` since `payment_ts`.
- The requestor has already made a regular payment with `closure_time >= payment_ts`.

Currently Concent allows only the first condition which makes provider unable to submit acceptances that must be submitted because they won't be taken into account in the next settlement.

PaweuB:
Pivotal: 
https://www.pivotaltracker.com/story/show/162505420

PR #1012, opened 2018-12-12
Add logs to handle_send_force_payment function

kbeker:
Resolves https://github.com/golemfactory/concent/issues/997

PR #1015, opened 2018-12-12
Bugfix acceptances covered by an existing payment that was too low should overdue

Jakub89:
Resolves: #1007 
First commit resolves different bug that I found(someone make sure I'm correct).


concent_api/core/message_handlers.py
```diff
@@ -820,7 +820,7 @@ def handle_send_force_payment(
 
     # Concent defines time T2 (end time) equal to youngest payment_ts from passed SubtaskResultAccepted messages from
     # subtask_results_accepted_list.
-    payment_ts = min(
+    youngest_payment_ts = max(
```
cameel: Yeah, this is a bug and a pretty serious one. It allows the provider to run the settlement algorithm as many times as there are acceptances, getting paid for **all** the acceptances each time.

The fix looks good. If you had a separate pull request, but I can't because I have not looked at the rest of this pull request.

cameel: Oh, actually it's not that bad. It's the timestamp in the message so that most it can do is to confuse the provider. I confused it for a moment with `closure_time` in the settlement payment.

Still, it's a bug. Good catch.

Please make sure that `closure_time` in the payment is exactly the same as this `payment_ts`. Actually, I may be a good idea to move the calculation of this timestamp to bankster to have it in one place.

concent_api/core/payments/bankster.py
```diff
@@ -474,3 +477,18 @@ def discard_claim(deposit_claim: DepositClaim) -> bool:
         subtask_id=deposit_claim.subtask_id,
     )
     return claim_removed
+
+
+def validate_list_of_transaction_timestamp(
+        list_of_transactions: List[BatchTransferEvent],
```
rwrzesien: Please fix indentation.

concent_api/core/message_handlers.py
rwrzesien: Please rephrase the comment above by removing `T2` from it, as we no longer use it anywhere in the code

concent_api/core/payments/bankster.py
```diff
@@ -474,3 +477,18 @@ def discard_claim(deposit_claim: DepositClaim) -> bool:
...
+        acceptances: List[SubtaskResultsAccepted],
+) -> None:
+    """ Validate if cut_off_time < T2 + PDT, unless requestor already made some payment. """
```
rwrzesien: Same here with the comment.

concent_api/core/payments/bankster.py
```diff
@@ -474,3 +477,18 @@ def discard_claim(deposit_claim: DepositClaim) -> bool:
...
+
+    cut_off_time = get_current_utc_timestamp()
+
+    youngest_payments_ts = max(subtask_results_accepted.payment_ts for subtask_results_accepted in acceptances)
+
+    if not any(youngest_payments_ts <= closure_time for closure_time in list_of_transactions):
```
rwrzesien: Those two conditions should be combined into single.

concent_api/core/payments/bankster.py
```diff
@@ -366,6 +367,8 @@ def settle_overdue_acceptances(
             transaction_type=TransactionType.BATCH,
         )
 
+        validate_list_of_transaction_timestamp(list_of_transactions, acceptances)
```
rwrzesien: Split into multiple lines.

concent_api/core/tests/test_payments_bankster.py
```diff
@@ -268,20 +268,22 @@ def test_that_settle_overdue_acceptances_should_return_none_if_subtask_costs_whe
...
-                self._get_list_of_force_transactions(),
-            ]
-        ) as get_list_of_payments_mock:
-            claim_against_requestor = settle_overdue_acceptances(
-                requestor_ethereum_address=task_to_compute.requestor_ethereum_address,
-                provider_ethereum_address=task_to_compute.provider_ethereum_address,
-                acceptances=subtask_results_accepted_list,
-                requestor_public_key=hex_to_bytes_convert(task_to_compute.requestor_public_key),
-            )
+        with mock.patch('core.payments.bankster.validate_list_of_transaction_timestamp', return_value=None) as validate_list_of_transaction_mock:
```
rwrzesien: Is there any test that tests this newly created function `validate_list_of_transaction_timestamp`?

concent_api/core/message_handlers.py
Jakub89: > Please rephrase the comment above by removing `T2` from it, as we no longer use it anywhere in the code.

I did use it also in next commit, where I add function in bankster. I would like to leave it. It can be helpful for someone who know documentation.

concent_api/core/payments/bankster.py
dybi: please reverse the condition, so it's easier to understand
```suggestion
    if any(youngest_payments_ts > closure_time for closure_time in list_of_transactions):
```

concent_api/core/tests/test_payments_bankster.py
dybi: you doesn't have to provide explicit `return_value` in case you don't expect to get anything out of the function. It is even better not to, because when you set it, it looks as if it was important (an here - it is irrelevant)

concent_api/core/tests/test_payments_bankster.py
```diff
@@ -333,21 +335,23 @@ def test_that_settle_overdue_acceptances_should_return_claim_deposit_with_amount
...
-                        self._get_list_of_force_transactions(),
-                    ]
-                ) as get_list_of_payments_mock:
-                    claim_against_requestor = settle_overdue_acceptances(
-                        requestor_ethereum_address=task_to_compute.requestor_ethereum_address,
-                        provider_ethereum_address=task_to_compute.provider_ethereum_address,
-                        acceptances=subtask_results_accepted_list,
-                        requestor_public_key=hex_to_bytes_convert(task_to_compute.requestor_public_key),
-                    )
+                with mock.patch('core.payments.bankster.validate_list_of_transaction_timestamp', return_value=None) as validate_list_of_transaction_mock:
```
dybi: same here

concent_api/core/tests/test_payments_bankster.py
```diff
@@ -424,21 +428,23 @@ def test_that_settle_overdue_acceptances_should_return_claim_deposit_with_amount
...
```
dybi: and here

concent_api/core/tests/test_payments_bankster.py
Jakub89: To my surprise there is in `test_integration_force_payment.py`: `test_provider_send_force_payment_beyond_payment_time_concent_should_reject()`

Jakub89: There's no corner case being tested where there is `closure_time` > `youngest_payments_ts`, but then it's business as usual :P

concent_api/core/payments/bankster.py
```diff
@@ -474,3 +477,18 @@ def discard_claim(deposit_claim: DepositClaim) -> bool:
         subtask_id=deposit_claim.subtask_id,
     )
     return claim_removed
+
+
+def validate_list_of_transaction_timestamp(
```
dybi: one more thing - please move it to `validation` module

concent_api/core/payments/bankster.py
```diff
@@ -474,3 +477,18 @@ def discard_claim(deposit_claim: DepositClaim) -> bool:
...
+) -> None:
+    """ Validate if cut_off_time < T2 + PDT, unless requestor already made some payment. """
+
+    cut_off_time = get_current_utc_timestamp()
+
+    youngest_payments_ts = max(subtask_results_accepted.payment_ts for subtask_results_accepted in acceptances)
+
+    if not any(youngest_payments_ts <= closure_time for closure_time in list_of_transactions):
+        if cut_off_time < youngest_payments_ts + settings.PAYMENT_DUE_TIME:
+            raise BanksterTimestampError
```
kbeker: TODO from https://github.com/golemfactory/concent/pull/1012. Please add logging in this function

concent_api/core/payments/bankster.py
Jakub89: ehh, okey. I will need to change commit name then. So `push --force` here we go! xD

concent_api/core/message_handlers.py
Jakub89: Done #1020

concent_api/core/payments/bankster.py
Jakub89: Done

concent_api/core/payments/bankster.py
Jakub89: Done

concent_api/core/payments/bankster.py
Jakub89: This line change because of [this comment](https://github.com/golemfactory/concent/pull/1015#discussion_r241346283)

Jakub89: Done

concent_api/core/tests/test_payments_bankster.py
Jakub89: Done

concent_api/core/tests/test_payments_bankster.py
Jakub89: Done

concent_api/core/tests/test_payments_bankster.py
Jakub89: Done

concent_api/core/payments/bankster.py
Jakub89: Done

concent_api/core/payments/bankster.py
Jakub89: Done

concent_api/core/payments/bankster.py
rwrzesien: > They can, but it made more logical sense to me like that, because we don't want to touch the second if, when first isn't True.

This is how Python works, if you have two conditions in `if` connected with `and` and first fails, second is not being run.

concent_api/core/tests/test_payments_bankster.py
rwrzesien: @dybi Do you think we should have separate unit test for this function? If not, please resolve this conversation.

concent_api/core/payments/bankster.py
Jakub89: > I have meant the first condition, should it really raise that exception and stop processing if the list is empty?

I think there's some misunderstanding. First condition doesn't raise any exception. If `list_of_transactions` is empty, we still need to check if `youngest_payments_ts` is not too young. And we raise exception only when it is.

rwrzesien: > I think there's some misunderstanding. First condition doesn't raise any exception. If list_of_transactions is empty, we still need to check if youngest_payments_ts is not too young. And we raise exception only when it is.

Makes sense, but it still should be combined into single `if`.

Jakub89: Done

PR #1019, opened 2018-12-13
[WIP] Fix for databases used in conductor in `transaction.on_commit` calls

rwrzesien:
Fix for databases used in conductor in `transaction.on_commit` calls.

Awaiting for @dybi proposal on how to test better.

### DIFF
--- a/concent_api/conductor/tests/test_integration_conductor.py
+++ b/concent_api/conductor/tests/test_integration_conductor.py
@@ -155,15 +155,16 @@ class ConductorVerificationIntegrationTest(ConcentIntegrationTestCase):
         upload_report.save()
 
         with mock.patch('conductor.views.transaction.on_commit') as mock_transaction_on_commit:
-            response = self.client.post(
-                reverse(
-                    'conductor:report-upload',
-                    kwargs={
-                        'file_path': self.source_package_path
-                    }
-                ),
-                content_type='application/octet-stream',
-            )
+            with mock.patch('conductor.views.upload_finished.delay') as upload_finished:
+                response = self.client.post(
+                    reverse(
+                        'conductor:report-upload',
+                        kwargs={
+                            'file_path': self.source_package_path
+                        }
+                    ),
+                    content_type='application/octet-stream',
+                )
 
         self.assertEqual(response.status_code, 200)
         self.assertEqual(UploadReport.objects.count(), 2)
@@ -172,6 +173,7 @@ class ConductorVerificationIntegrationTest(ConcentIntegrationTestCase):
         self.assertEqual(upload_report.path, self.source_package_path)
 
         mock_transaction_on_commit.assert_called_once()
+        upload_finished.assert_not_called()
 
         verification_request.refresh_from_db()
         self.assertTrue(verification_request.upload_finished)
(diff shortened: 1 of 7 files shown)


concent_api/core/tasks.py
```diff
@@ -111,7 +111,10 @@ def call_upload_acknowledged() -> None:
                 result_package_hash=report_computed_task.package_hash,
             )
 
-        transaction.on_commit(call_upload_acknowledged, using='control')
+        transaction.on_commit(
+            call_upload_acknowledged,
+            using='control',
+        )
```
cameel: Why not keep it on a single line? It was short and readable.

cameel:
Would it be possible to cover these lines with some simple tests that would detect that the function is not being called when a view/task ends (i.e. at a point when the transaction is supposed to be committed)? I think that the fact that it was totally incorrect but the tests passed is a problem. Debugging code that simply does not run when you think it should is tricky on the cluster. We should probably put more effort into prevention here.

@dybi What do you think?

concent_api/conductor/tasks.py
```diff
@@ -178,7 +181,10 @@ def call_blender_verification_order() -> None:
             blender_crop_script=verification_request.blender_subtask_definition.blender_crop_script,
         )
 
-    transaction.on_commit(call_blender_verification_order, using='control')
+    transaction.on_commit(
+        call_blender_verification_order,
+        using='storage',
```
kbeker: I'm just curious why it should be `storage` database instead of `control`. Could you explain it to me? :)

concent_api/core/tasks.py
rwrzesien: I thought we wanted it that way to keep history changes of each parameters, didn't we?

rwrzesien:
@cameel 
>  I think that the fact that it was totally incorrect but the tests passed is a problem.

Tests was failing and fixed in previous pull requests related with that changes.

concent_api/conductor/tasks.py
rwrzesien: The whole task is wrapped in transaction using `storage` database.

cameel:
> > I think that the fact that it was totally incorrect but the tests passed is a problem.

> Tests was failing and fixed in previous pull requests related with that changes.

So you mean that someone merged it even though the tests were failing? If it was fixed then we wouldn't be working on this pull request now.

rwrzesien:
> So you mean that someone merged it even though the tests were failing? If it was fixed then we wouldn't be working on this pull request now.

No, what I have meant is that when doing the original issue of which this pull request is a derivative (https://github.com/golemfactory/concent/pull/992) I have to fix several tests because in those tests the calls to tasks `delay` was no longer called, and as we mock it, the mock check `was_called_*` was no longer right. But now as I think about it, the tests which started failing and which I have fixed was actually those where the problem was resolved correctly (function passed to `transaction.on_commit` was not being run instantly). And I wasn't able to notice the problem because incorrect tests were still passing. So yes, we should add check that tasks `delay` called inside function passed to `transaction.on_commit` is not called. I will try to add them in this pull request.

rwrzesien:
@cameel Added, are you planning on reviewing those?

PaweuB:
It has just appeared in Sentry:
https://talkback.golem.network/share/issue/2214d7f4b69d462aab1e6bf55bff9448/

rwrzesien:
@dybi 
> This is very misleading and cannot stay as it is.

What is your proposal to fix it then? I don't see a different solution for this.

Issue #1024, opened 2018-12-13
Bankster: Guard against passing the same claim twice to finalize_payment()


cameel:
Currently `finalize_payment()` does not check if the `DepositClaim` passed to it has `tx_hash` filled out or not. If it's filled out it will be ignored and overwritten.

This does not matter as long as Concent is using it correctly but I think we need at least a sanity check (`assert`) for that case.

Since Bankster is an internal API, it's also fine to use a normal `if` and raise an exception. But in that case you should replace other sanity checks with validatios too: for example the check that the use case is either UC3 or UC4 is now enforced with an assertion.

PaweuB:
Pivotal: 
https://www.pivotaltracker.com/story/show/162732447

PR #1027, opened 2018-12-14
Bankster: Guard against passing the same claim twice to finalize_paym…

rwrzesien:
…ent()

Resolves https://github.com/golemfactory/concent/issues/1024

### DIFF
--- a/concent_api/core/payments/bankster.py
+++ b/concent_api/core/payments/bankster.py
@@ -203,6 +203,7 @@ def finalize_payment(deposit_claim: DepositClaim) -> Optional[str]:
     """
 
     assert isinstance(deposit_claim, DepositClaim)
+    assert deposit_claim.tx_hash is None
 
     # Bankster asks SCI about the amount of funds available on the deposit account listed in the DepositClaim.
     available_funds = service.get_deposit_value(  # pylint: disable=no-value-for-parameter


rwrzesien:
@kbeker 
> Please fix branch name ....bankster-guard-against... And good for me

It is in Spanish.

concent_api/core/payments/bankster.py
```diff
@@ -203,6 +203,7 @@ def finalize_payment(deposit_claim: DepositClaim) -> Optional[str]:
     """
 
     assert isinstance(deposit_claim, DepositClaim)
+    assert deposit_claim.tx_hash is None
```
dybi: do we really want a crash here? What about logging an abnormal situation (I assume, this shouldn't normally happen)? Or even returning without a crash?

rwrzesien: > do we really want a crash here?

According to issue description, yes, this is abnormal situation. That also fits our coding style.

cameel: We absolutely do not want to handle it (unless you consider returning HTTP 500 handling it). It's never supposed to happen. And the check is here mainly to protect programmers from making this mistake in the first place. As I said in the description of the issue - since this is an API it would make sense to raise an exeception too. But we should not handle that exception anyway.

If this happens at runtime - we want to bail out as soon as possible and get a report in Sentry. There's nothing we can do that does not involve fixing the code - and the server won't patch itself at runtime :)

PR #1028, opened 2018-12-14
Add validation for force_payment message

kbeker:
This PR doesn't have issue. This is a sentry exception: https://talkback.golem.network/share/issue/b19e564a4e074f4890a314c69170f6e2/

### DIFF
--- a/concent_api/core/tests/test_unit_message_handlers.py
+++ b/concent_api/core/tests/test_unit_message_handlers.py
@@ -1,4 +1,5 @@
 from unittest import TestCase
+import pytest
 
 from django.conf import settings
 from django.test import override_settings
@@ -8,6 +9,7 @@ from golem_messages import message
 from common.helpers import parse_timestamp_to_utc_datetime
 from core.exceptions import GolemMessageValidationError
 from core.message_handlers import are_items_unique
+from core.message_handlers import handle_send_force_payment
 from core.message_handlers import update_and_return_updated_subtask
 from core.message_handlers import store_subtask
 from core.models import Subtask
@@ -156,3 +158,13 @@ class TestValidateRejectReportComputedTask(ConcentIntegrationTestCase):
         )
         with self.assertRaises(GolemMessageValidationError):
             validate_reject_report_computed_task(incorrect_reject_reported_computed_task)
+
+
+class TestCorrectMessageToForcePaymentHandler:
+
+    @pytest.mark.parametrize('subtask_results_accepted_list', (None, [], [message.Ping()], ()))
+    def test_that_force_payment_with_malformed_subtask_results_accepted_list_raises_assertion(self, subtask_results_accepted_list):  # pylint: disable=no-self-use
+        force_payment = message.concents.ForcePayment(subtask_results_accepted_list=subtask_results_accepted_list)
+
+        with pytest.raises(AssertionError):
+            handle_send_force_payment(force_payment)
--- a/concent_api/core/tests/test_unit_validation.py
+++ b/concent_api/core/tests/test_unit_validation.py
@@ -3,9 +3,10 @@ from assertpy import assert_that
 import mock
 import pytest
 
-from golem_messages.factories.tasks import ComputeTaskDefFactory, ReportComputedTaskFactory
-from golem_messages.factories.tasks import TaskToComputeFactory
+from golem_messages.factories.tasks import ComputeTaskDefFactory
+from golem_messages.factories.tasks import ReportComputedTaskFactory
 from golem_messages.factories.tasks import SubtaskResultsAcceptedFactory
+from golem_messages.factories.tasks import TaskToComputeFactory
 from golem_messages.factories.tasks import WantToComputeTaskFactory
 from golem_messages.message.tasks import ReportComputedTask
 from golem_messages.message.tasks import SubtaskResultsRejected
--- a/concent_api/core/message_handlers.py
+++ b/concent_api/core/message_handlers.py
@@ -26,6 +26,7 @@ from golem_messages.message.concents import ForcePaymentRejected
 from golem_messages.message.concents import ForceReportComputedTask
 from golem_messages.message.concents import ForceReportComputedTaskResponse
 from golem_messages.message.concents import ServiceRefused
+from golem_messages.message.tasks import SubtaskResultsAccepted
 from golem_messages.message.tasks import SubtaskResultsRejected
 from golem_messages.register import library
 
@@ -68,13 +69,13 @@ from core.utils import calculate_subtask_verification_time
 from core.utils import is_protocol_version_compatible
 from core.validation import is_golem_message_signed_with_key
 from core.validation import substitute_new_report_computed_task_if_needed
-from core.validation import validate_that_golem_messages_are_signed_with_key
-from core.validation import validate_reject_report_computed_task
 from core.validation import validate_all_messages_identical
 from core.validation import validate_ethereum_addresses
 from core.validation import validate_golem_message_subtask_results_rejected
+from core.validation import validate_reject_report_computed_task
 from core.validation import validate_report_computed_task_time_window
 from core.validation import validate_task_to_compute
+from core.validation import validate_that_golem_messages_are_signed_with_key
 
 from .utils import hex_to_bytes_convert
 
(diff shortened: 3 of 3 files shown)


kbeker:
> I appreciate the work put into this pull request, I am just not sure if some part of it is not already being done by golem messages. For example, can `ForcePayment` be deserialized if `subtask_results_accepted_list` is something different than list? And can elements on this list be something different than `SubtaskResultsAccepted`. I am not saying it is wrong, it is rather a matter of do we want to double check it? @cameel @dybi ?

Yea, it strange that this happened, that `ReportComputedTask`  hadn't nested `TaskToCompute` so I decided that the best will be to add validation to this.

concent_api/core/tests/test_unit_validation.py
```diff
@@ -403,3 +406,42 @@ def test_that_other_messages_than_task_to_compute_causes_message_validation_erro
         with pytest.raises(ConcentValidationError) as exception_wrapper:
             validate_task_to_compute(wrong_message)
         assert_that(exception_wrapper.value.error_code).is_equal_to(ErrorCode.MESSAGE_INVALID)
+
+
+class TestValidateForcePaymentContainCorrectMessages():
```
rwrzesien: I think there is missing test that element on `subtask_results_accepted_list` is something different than `SubtaskResultsAccepted`.

concent_api/core/validation.py
```diff
@@ -495,3 +496,42 @@ def validate_list_of_transaction_timestamp(
...
+    for subtask_results_accepted in force_payment_message.subtask_results_accepted_list:
+        if not isinstance(subtask_results_accepted, SubtaskResultsAccepted):
+            raise ConcentValidationError(
+                error_message=f'Expected SubtaskResultsAccepted in SubtaskResultsAccepted list, '
+                              f'but got {type(subtask_results_accepted).__name__}',
+                error_code=ErrorCode.MESSAGE_SUBTASK_RESULTS_ACCEPTED_MISSING,
+            )
+        if (
+            not hasattr(subtask_results_accepted, 'report_computed_task') or
+                subtask_results_accepted.report_computed_task is None
```
rwrzesien: This line should start on the same indent as above.

concent_api/core/validation.py
```diff
@@ -495,3 +496,42 @@ def validate_list_of_transaction_timestamp(
...
+        ):
+            raise ConcentValidationError(
+                error_message='Expected TaskToCompute nested in ReportComputedTask',
+                error_code=ErrorCode.MESSAGE_REPORT_COMPUTED_TASK_MISSING,
+            )
+        if (
+            hasattr(subtask_results_accepted.report_computed_task, 'task_to_compute') and
+                subtask_results_accepted.report_computed_task.task_to_compute is not None
```
rwrzesien: Same here.

rwrzesien:
@kbeker 
> Yea, it strange that this happened, that ReportComputedTask hadn't nested TaskToCompute so I decided that the best will be to add validation to this.

Yes, if it is possible from golem-messages point of view then we should definitely validate against it. I was just curious if those two cases I have mentioned in previous comment are possible and if they are not, if we should validate against them.

cameel:
> I appreciate the work put into this pull request, I am just not sure if some part of it is not already being done by golem messages. For example, can ForcePayment be deserialized if subtask_results_accepted_list is something different than list? And can elements on this list be something different than SubtaskResultsAccepted. I am not saying it is wrong, it is rather a matter of do we want to double check it?

I think we should have assertions in our code for that. This is something that should be checked in golem-messages but as this situation shows it's hard to rely on it because that code is far away. We need sanity checks to ensure that golem-messages is doing its job.

rwrzesien:
@kbeker According to @cameel comment please check if the two situations I have mentioned in my first commit are possible due to golem-messages validation during deserialization, and if they are not just change those two checks for asserts.

dybi:
My opinion is that we should understand the nature of this issue, i.e. when and why it might happen. 
Validation should be unnecessary, because there is validation in golem messages that kicks in during deserialization. For whatever reason it didn't work as expected. If there's a bug in golem messages, we should fix it there.
As to putting an assert - we shouldn't do it. We shouldn't let a malformed message crash concent. So again - we should understand what's going on and fix it.

kbeker:
OK, issue will be resolved in newest GM version. PR -> https://github.com/golemfactory/golem-messages/pull/293. @cameel @rwrzesien @dybi should we leave this validation or just delete it?

rwrzesien:
@kbeker Please just make asserts in cases covered by golem-messages.

cameel:
>  should we leave this validation or just delete it?

Validation should be deleted. I agree with @dybi that double validations are bad because of code duplication.

We should convert them to assertions instead. Not necessarily all of them. Only those without which our code is really broken.

It's like having two parts: the first is a protector for the other and the other has a label saying "do not use without a protector". Since they're separate parts, made by different factories and might even be sometimes sold separately, the label is the minimum you can do to make sure the user is aware that the protector must be there. Except it's even better in our case because we don't have to use a label (a comment) that could easily go out of date. We have sanity checks that are checked automatically for us.

We should have at least an assert for the case that has just failed. Then we'll easily spot the problem if it happens to anyone in development.

dybi:
I am against assert - I would prefer a simple validation that would result in responding with `ForcePaymentRejected` in case of broken/absent list of `SubtaskResultAccepted` messages

concent_api/core/tests/test_unit_message_handlers.py
```diff
@@ -156,3 +157,30 @@ def test_that_validation_raises_error_if_reason_is_none(self):
...
+        force_payment = message.concents.ForcePayment(subtask_results_accepted_list=[message.Ping()])
+
+        with self.assertRaises(AssertionError):
+            handle_send_force_payment(force_payment)
+
+    def test_that_force_payment_handler_will_raise_assertion_when_subtask_results_accepted_list_is_not_list_instance(self):
+        force_payment = message.concents.ForcePayment(subtask_results_accepted_list=())
+
+        with self.assertRaises(AssertionError):
+            handle_send_force_payment(force_payment)
```
pawelkisielewicz: this can be simplified using pytest.parametrize. You can parametrize `subtask_results_accepted_list` parameter and write only one test. Now it is lot of code duplication

cameel:
> I am against `assert`s - I would prefer a simple validation that would result in responding with `ForcePaymentRejected` in case of broken/absent list of `SubtaskResultAccepted` messages

Let's ignore `assert` for a moment. The question is - do we add any validations or not? The potential issue here is not that something will get through validations in golem-messages in production. It's that someone may remove or break these validations in golem-messages and not notice that it breaks Concent.

This is something that should simply not pass our tests. For each validation there should be a test that fails if it's removed or broken. If we have that, there's no chance that no one will notice broken validations.

So I don't think that adding this validation is wrong but it's a bit superfluous and defensive. I personally would not add it because I think it's like adding code to handle bugs instead of writing tests to detect those bugs but I'm not strongly against it if you really prefer it that way.

Now, the `assert` is a separate matter. I think assertions should be there not as a replacement for validations. Quite the opposite. They should be there precisely because we're certain they cannot fail. It's like a safety net that might catch you in development even if you're wrong and fail to detect the problem via other means like tests or review. If you think our test coverage is too small and there's real risk that the validations in golem-messages are missing or broken and that there's even a slight chance that we'll actually hit the code path that returns this `ForcePaymentRejected` then yeah - we can't assume that those assumptions are true and we cannot put `assert` there.

> Assertions or ServiceRefused. I personally agree with @dybi, even if it is theoretically impossible it is better to validate, log error and return ServiceRefused instead of crashing code. I'm not "so against" assertions as @dybi, I think they can be useful but for checking our code, preventing for ours mistakes. For external library I would prefer validation and ServiceRefused. I think also it is good subject to talk about at next meeting- validations vs assertions in general

Yeah, I want to have a tech talk about it at some point. I hope I'll have more time for things like this after we release Concent on the mainnet. In the meantime I'm trying to explain specific situations in comments whenever I spot them but looks like not everyone is convinced.

Anyway, I think that "validations vs assertions" is a really bad way to think about it. It makes you think that you can replace one with the other. Or that there are situations in which we validate with `assert`. It's really "validations vs no validations". It just so happens that if you do not need validations, that's because you have some hidden assumption (e.g. someone else validates that for me; or it's impossible). You could make a comment about this assumption that explains why we don't have to check it. You could write a test case that checks that indeed this assumption is met because even though you did not add a validation yourself, Concent rejects the request. And you could add an `assert` to check that assumption directly. You should probably do all of these things.

A real question would be "tests vs assertions" or "comments vs assertions". And I'd argue that they are not mutually exclusive. Each is a tool that's good at what it was designed to do. They all suck if you use them the wrong way. You absolutely can have too many assertions just like you can have too many tests or too many comments. They complement each other. Tests are more effective with sanity checks that detect inconsistencies you're not explicitly testing for. Comments can be accompanied by assertions that prove they are true. And tests should have comments/docstrings explaining what you're testing.

For me the killer feature of `assert` is simply how extremely cheap it is compared to the value it adds. It's not as heavy as a test. It does not go out of sync like a comment. It's something you already know and you just have to write it down. They cost so little that if I don't have enough time to make the code good, of the three things (tests, assertions, comments/docs) I always drop assertions last. Fortunately we don't have to choose here.

concent_api/core/tests/test_unit_message_handlers.py
kbeker: Changed to `@pytest.mark.parametrize`

Issue #1023, opened 2018-12-13
[Sentry] 'Ensure this value is less than or equal to 2147483647.'


Jakub89:
https://talkback.golem.network/sentry/concent/issues/44675/

Default for django `IntegerField()` is 4-bytes value. We can replace it with django `BigIntegerField()` where we can store value up to "9223372036854775807", but it still might not be enough if someone decide to make a deposit for millions of GNT. @cameel ? @rwrzesien ? @dybi ?

rwrzesien:
Can be `DecimalField(max_digits=78, decimal_places=0)` .

cameel:
Yeah, it should be the same type we use in our transaction storage. I now don't 100% remember if we ended up using `Decimal` but if so, we can use it here too.

This applies to any other places where we're storing GNT or ETH amounts. If you're aware of any (please look through our models to be sure), they need to be fixed too.

PaweuB:
Yes, it is in Pivotal since yesterday and @Jakub89 is assigned for this since yesterday (-; 

https://www.pivotaltracker.com/story/show/16264040

PR #1030, opened 2018-12-14
[M]Switch "big" IntegerField() to DecimalField in core models

Jakub89:
Resolves: #1023


concent_api/core/models.py
```diff
@@ -637,14 +637,14 @@ def clean(self) -> None:
                 'provider_eth_account': 'Provider ethereum account address must be diffrent than task owner key'
             })
 
-        if not isinstance(self.amount_pending, int) or self.amount_pending < 0:
+        if self.amount_pending < 0:
```
rwrzesien: I know that this is not related with that changes but as we are here, what about condition when it is equal 0? Neither the condition nor error message cover that case (so on of those should be fixed).
Apart from that, I think it should work still if you check that value type is either `int` or `Decimal`.

concent_api/core/models.py
```diff
@@ -637,14 +637,14 @@ def clean(self) -> None:
...
             raise ValidationError({
-                'amount_pending': 'Amount pending must be an integer and bigger than 0'
+                'amount_pending': 'Amount pending must be bigger than 0'
             })
 
-        if not isinstance(self.amount_paid, int) or self.amount_paid < 0:
+        if self.amount_paid < 0:
```
rwrzesien: Same here about the type, the message here is ok.

concent_api/core/models.py
```diff
@@ -765,9 +765,9 @@ def clean(self) -> None:
                                           f'(requestor, provider or Concent) must be string and must be exactly '
                                           f'{ETHEREUM_ADDRESS_LENGTH} characters long.'
             })
-        if not isinstance(self.amount, int) or self.amount <= 0:
```
rwrzesien: Same here about the type, the message is ok here.
Btw. please just add dots at the end of the sentences, I see that sometimes they are and sometimes not, lets keep it consistent (and correct). Could be separate commit.

concent_api/core/tests/test_unit_model.py
```diff
@@ -308,11 +308,6 @@ def test_that_exception_is_raised_when_amount_is_less_then_zero(self):
             self.deposit_claim.clean()
         self.assertIn('amount', exception_info.value.error_dict)
 
-    def test_that_exception_is_raised_when_amount_is_of_wrong_type(self):
-        self.deposit_claim.amount = 5.0
```
rwrzesien: I would suggest keeping this test but changing the value to a non accepted type. This way we still test one branch of the `if` condition (after doing the comments I have added). @dybi do you have opinion on that? Is it better to remove it?

concent_api/core/models.py
```diff
@@ -739,7 +739,7 @@ class DepositClaim(Model):
     payer_deposit_account = ForeignKey(DepositAccount)
     payee_ethereum_address = CharField(max_length=ETHEREUM_ADDRESS_LENGTH)
     concent_use_case = IntegerField()
-    amount = IntegerField()
+    amount = DecimalField(max_digits=78, decimal_places=0)
```
kbeker: We should have tests for all this model changes. Please add tests which will test maximum possible value for this changed fields

concent_api/core/models.py
dybi: I think that `amount_pending` can be `0` - if all claims have been satisfied.

concent_api/core/tests/test_unit_model.py
dybi: If `DecimalField` validations check that for us automatically, then we don't have to add tests for that. If not, then it would be good to have one.

Jakub89: Done

concent_api/core/models.py
Jakub89: Done

concent_api/core/models.py
Jakub89: Type not needed, dot added.

concent_api/core/models.py
rwrzesien: Each of those validations is not actually needed to be performed there, if you want to validate if value is grater than 0 you can set `MinValueValidator(0)` directly on the model. I thought we are doing this for custom error messages - if not, what is the reason for keeping this custom validation?

Jakub89: @cameel said the same thing on Friday, I can change it to django Validator then.

rwrzesien: Ok, please do that then.

concent_api/core/tests/test_unit_model.py
```diff
@@ -308,10 +308,15 @@ def test_that_exception_is_raised_when_amount_is_less_then_zero(self):
...
+    def test_that_exception_is_raised_when_amount_is_to_large(self):
```
kbeker: `is_to_large` -> `is_too_large`

Jakub89: Done

concent_api/core/models.py
Jakub89: Done

PR #1032, opened 2018-12-17
Fix for databases used in conductor in `transaction.on_commit` calls

rwrzesien:
This is part extracted from https://github.com/golemfactory/concent/pull/1019/ that covers the main issue.


kbeker:
Merged

Issue #1039, opened 2018-12-19
Revert changes from #982 so that the list of forced payments in UC5 is the list settlement payments again


PaweuB:
In #982 we have changed the implementation of `settle_overdue_acceptances()` to construct the list of forced subtask payments and use that list in the calculation. Turns it it was wrong. The previous code was correct. The list should contain settlement payments after all.

Revert the implementation changes but keep all the other stuff done in that pull request like tests or refactoring.

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/162755341

Issue #1009, opened 2018-12-11
Preventing multiple settlement payments for the same subtasks


cameel:
This task depends on #1017.

In #984 we have implemented a mechanism that checks if Concent has already issued a settlement payment and prevents the acceptances covered in that settlement from being submitted again. This mechanism has a critical flaw: the settlement payment may not be visible on the blockchain for several minutes or more because blockchain transactions are not instant. This gives the provider plenty of time to call Concent again and force another payment for the same subtasks.

(EDIT: Even though we're going to revert #984 the point still stands: if the provider runs UC5 multiple times, he should not be paid multiple times for the same thing. If the first run covered only 50% of what he was owed (due to not enough deposit), the next run should see the forced payment and cover no more than the other 50%.)

Fortunately Concent can know that it has issued a payment because this information is stored inside `DatabaseClaim`s created by Bankster.

Modify the implementation of UC5 so that:
1. It finds all `DatabaseClaims` that represent settlement payments from requestor's Ethereum account to provider's Ethereum account and for which a transaction has already been sent to the Ethereum client. It creates a list of those transactions.
2. It gets the list of settlement payments from the blockchain
3. It finds transactions that are on both lists and removes them from the list of transactions found in claims.
    - It compares the duplicates and if the information in the database (e.g. the amount or `closure_time`) differs from the information on the blockchain, aborts the settlement and responds with HTTP 500. This should never happen in practice and if it does, we want to get a report and fix the problem.
4. It uses both lists when checking existing settlement payments.

When the task is done, please drop me a note in #996 because the specification will need to be updated.

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/162732368

cameel:
I started wondering if we should maybe take into account claims that are not yet finalized (i.e. do not have a transaction). But I think ignoring them is better. Below is my rationale in case anyone is interested. I think it would be a good idea to create a test case for this.

Let's assume that requestor's deposit is 100 GNT, there's an unfinalized claim from UC3 for 50 GNT and the provider starts UC5 because he's not been paid 100 GNT for a different subtask.
1) If we take unfinalized claims into account
    1) If the claim from UC3 eventually gets finalized, provider gets 50 GNT in UC5 and 50 GNT in UC3. 100 GNT total.
    2) If the claim from UC3 eventually gets removed, provider gets 50 GNT in UC5 and 0 GNT in UC3. 50 GNT total.
[...]

Issue #1018, opened 2018-12-12
Confusing names in sci_backend


cameel:
Names in our `core.payments.backends.sci_backend` module may be confusing for someone who has read the #996 blueprint. Please change them to make the code easier to understand:
1) `make_force_payment_to_provider()` should be called `make_settlement_payment()`.
    - Also, if you see any comments or variables referring to "forced payments" (but watch out for "forced subtask payments" - it's not the same thing), it would be a good idea to change them too. It's too easy to confuse "forced payment" and "forced subtask payment". "Settlement payment" is a less ambiguous name for the former. Don't spend too much time on it though. Our main goal is changing the public names visible in `sci_backend.py`.
2) `make_force_payment_to_provider()` has a parameter called `payment_ts`. It should be called `closure_time` because it's used as the value for `closure_time` in the payment. Calling it `payment_ts` makes it too easy to confuse it with `SubtaskResultsAccepted.payment_ts` because they're often used in the same context.
3) `get_list_of_payments()` has a parameter called `payment_ts`. It would be better to call it `min_block_timestamp` because it has absolutely nothing to do with `payment_ts` in `SubtaskResultsAccepted` or `ForcePaymentCommitted`. They cannot even be compared directly.
4) `call_on_confirmed_transaction()` is a weird name for a callback. It's telling the programmer what he is supposed to do with it instead of telling it what it does. I'd call it `confirmed_transaction_callback()` or `confirmed_transaction_handler()`.
    - EDIT: After looking at it again, I see that I was wrong about what this function does (which only confirms that the current name is not good). It's the other function it gets as a parameter that gets called, not this function itself. In that case I think that `on_confirmed_transaction()` or `register_confirmed_transaction_handler()` would be better.

These names should be adjusted to avoid potential bugs caused by wrong interpretation of the code.

Please consider renaming badly named functions, parameters and variables also in other parts of our payment system  if you notice any (see for example https://github.com/golemfactory/concent/pull/1043#discussion_r243700491). Misleading or out of date names make the code harder to understand.

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/162632122

Issue #983, opened 2018-11-27
Concent responds with ForcePaymentRejected if there was a payment after one of the acceptances


cameel:
Originally reported in https://github.com/golemfactory/concent/pull/975#discussion_r236654251.

In the current implementation Concent responds with `ForcePaymentRejected` in use case 5 if not all payments found on the blockchain were made before the first submitted acceptance.

In other words, a situation like this provider's request will be rejected because there was a payment made after the first SRA:
```
     -100 GNT    1 GNT         -100 GNT
       |         |               |
------SRA-------PAY-------------SRA------> [time]
       |         |
       T0        T1
```
_T0 is the timestamp of the oldest acceptance._
_T1 is the timestamp of the youngest payment._

The expected outcome in this situation is for Concent to pay the provider 199 GNT from requestor's deposit.

The code that does this is in `settle_overdue_acceptances()`:
https://github.com/golemfactory/concent/blob/e90c59c971a0e5c48e0529c586a4e8db1cae557d/concent_api/core/payments/bankster.py#L365-L389

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/162265003

Jakub89:
@cameel from what I understand, this bug comes down to delete one of our validations:
``` python
# Concent checks if all passed SubtaskResultAccepted messages from subtask_results_accepted_list
# have payment_ts < T1
if any(
    youngest_transaction_timestamp > subtask_results_accepted.payment_ts
    for subtask_results_accepted in acceptances
):
    raise BanksterTimestampError
```
Should it be replace with something else?

cameel:
I think you can just remove it as long as all the unit tests pass. Please be sure to add a test for the situation from the description and any others you can think of.

PR #1043, opened 2018-12-19
Feature preventing multiple settlement payments for the same subtasks

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/1009
Resolves https://github.com/golemfactory/concent/issues/1039

### DIFF
--- a/concent_api/core/payments/bankster.py
+++ b/concent_api/core/payments/bankster.py
@@ -5,6 +5,7 @@ from typing import Union
 import logging
 
 from django.conf import settings
+from django.db.models import QuerySet
 from django.db.models import Sum
 from django.db.models.functions import Coalesce
 
@@ -19,6 +20,7 @@ from common.helpers import ethereum_public_key_to_address
 from common.helpers import parse_timestamp_to_utc_datetime
 from common.logging import log
 from core.constants import ETHEREUM_ADDRESS_LENGTH
+from core.exceptions import BanksterTransactionMismatchError
 from core.exceptions import TooSmallProviderDeposit
 from core.models import Client
 from core.models import DepositAccount
@@ -266,6 +268,8 @@ def settle_overdue_acceptances(
     assert len(provider_ethereum_address) == ETHEREUM_ADDRESS_LENGTH
     assert provider_ethereum_address != requestor_ethereum_address
 
+    validate_list_of_transaction_timestamp(acceptances)
+
     with non_nesting_atomic(using='control'):
         requestor_client: Client = get_or_create_safely(Client, public_key=requestor_public_key)
 
@@ -292,16 +296,24 @@ def settle_overdue_acceptances(
             sum_of_existing_claims=Coalesce(Sum('amount'), 0)
         )
 
-        assert sum_of_existing_requestor_claims['sum_of_existing_claims'] >= 0
-
-        # If the existing claims against requestor's deposit are greater or equal to his current deposit,
-        # we can't add a new claim.
-        if requestor_deposit_value <= sum_of_existing_requestor_claims['sum_of_existing_claims']:
-            return None
-
         # Concent defines time T0 equal to oldest payment_ts from passed SubtaskResultAccepted messages from
         # subtask_results_accepted_list.
         oldest_payments_ts = min(subtask_results_accepted.payment_ts for subtask_results_accepted in acceptances)
+
+        # Concent gets list of forced payments from payment API where T0 <= payment_ts + PAYMENT_DUE_TIME.
+        list_of_settlement_payments = service.get_list_of_payments(  # pylint: disable=no-value-for-parameter
+            requestor_eth_address=requestor_ethereum_address,
+            provider_eth_address=provider_ethereum_address,
+            payment_ts=oldest_payments_ts,
+            transaction_type=TransactionType.FORCE,
+        )
+
+        already_satisfied_claims_without_duplicates = find_unconfirmed_settlement_payments(
+            list_of_settlement_payments,
+            requestor_deposit_account,
+            provider_ethereum_address,
+            oldest_payments_ts,
+        )
 
         # Concent gets list of transactions from payment API where timestamp >= T0.
         list_of_transactions = service.get_list_of_payments(  # pylint: disable=no-value-for-parameter
@@ -311,19 +323,10 @@ def settle_overdue_acceptances(
             transaction_type=TransactionType.BATCH,
         )
 
-        validate_list_of_transaction_timestamp(list_of_transactions, acceptances)
-
-        # Concent gets list of forced payments from payment API where T0 <= payment_ts + PAYMENT_DUE_TIME.
-        list_of_forced_payments = service.get_list_of_payments(  # pylint: disable=no-value-for-parameter
-            requestor_eth_address=requestor_ethereum_address,
-            provider_eth_address=provider_ethereum_address,
-            payment_ts=oldest_payments_ts,
-            transaction_type=TransactionType.FORCE,
-        )
-
         (_amount_paid, amount_pending) = get_provider_payment_info(
-            list_of_forced_payments=list_of_forced_payments,
-            list_of_payments=list_of_transactions,
+            list_of_settlement_payments=list_of_settlement_payments,
+            list_of_transactions=list_of_transactions,
+            settlement_payment_claims=already_satisfied_claims_without_duplicates,
             subtask_results_accepted_list=acceptances,
         )
 
@@ -383,22 +386,85 @@ def sum_subtask_price(subtask_results_accepted_list: List[SubtaskResultsAccepted
 
 
 def get_provider_payment_info(
-    list_of_forced_payments: List[ForcedPaymentEvent],
-    list_of_payments: List[BatchTransferEvent],
+    list_of_settlement_payments: List[ForcedPaymentEvent],
+    list_of_transactions: List[BatchTransferEvent],
+    settlement_payment_claims: QuerySet,
     subtask_results_accepted_list: List[SubtaskResultsAccepted],
 ) -> Tuple[int, int]:
-    assert isinstance(list_of_payments, list)
-    assert isinstance(list_of_forced_payments, list)
+    assert isinstance(settlement_payment_claims, QuerySet)
+    assert isinstance(list_of_settlement_payments, list)
+    assert isinstance(list_of_transactions, list)
     assert isinstance(subtask_results_accepted_list, list)
 
-    force_payments_price = sum_payments(list_of_forced_payments)
-    payments_price = sum_payments(list_of_payments)
+    settlement_payment_total_price = sum_payments(list_of_settlement_payments)
+    payments_price = sum_payments(list_of_transactions)
+    satisfied_payments_price = settlement_payment_claims.aggregate(
+        sum_of_already_satisfied_claims=Coalesce(Sum('amount'), 0)
+    )['sum_of_already_satisfied_claims']
     subtasks_price = sum_subtask_price(subtask_results_accepted_list)
 
-    amount_paid = payments_price + force_payments_price
+    amount_paid = payments_price + satisfied_payments_price + settlement_payment_total_price
     amount_pending = subtasks_price - amount_paid
 
     return (amount_paid, amount_pending)
+
+
+def find_unconfirmed_settlement_payments(
+    list_of_settlement_payments: List[ForcedPaymentEvent],
+    requestor_deposit_account: DepositAccount,
+    provider_ethereum_address: str,
+    oldest_payments_ts: int,
+) -> QuerySet:
+    """
+    Retrieves claims of settlement payments with pending transaction
+    and removes duplicates which are on list of settlement payments.
+    """
+    already_satisfied_claims = get_already_satisfied_claims(
+        requestor_deposit_account,
+        provider_ethereum_address,
+        oldest_payments_ts,
+    )
+    settlement_payment_tx_hashes = [force_payment.tx_hash for force_payment in list_of_settlement_payments]
+
+    for matching_claim in already_satisfied_claims.filter(tx_hash__in=settlement_payment_tx_hashes):
+        assert matching_claim.concent_use_case == ConcentUseCase.FORCED_PAYMENT
+
+        matching_force_payment = next(iter([
+            force_payment for force_payment in list_of_settlement_payments
+            if force_payment.tx_hash == matching_claim.tx_hash
+        ]))
+
+        # If the information in the database differs from the information on the blockchain,
+        # aborts the settlement and responds with HTTP 500.
+        if (
+            matching_claim.closure_time != matching_force_payment.closure_time or
+            matching_claim.amount != matching_force_payment.amount
+        ):
+            raise BanksterTransactionMismatchError(
+                f'There is a data mismatch between DepositClaim and ForcedPaymentEvent with transaction hash: '
+                f'{matching_force_payment.tx_hash}'
+            )
+
+    return already_satisfied_claims.exclude(tx_hash__in=settlement_payment_tx_hashes)
+
+
+def get_already_satisfied_claims(
+    requestor_deposit_account: DepositAccount,
+    provider_ethereum_address: str,
+    payments_ts: int,
+) -> QuerySet:
+    """
+    Returns all DatabaseClaims that represent settlement payments from requestor's Ethereum account to
+    provider's Ethereum account and for which a transaction has already been sent to the Ethereum client.
+    """
+    return DepositClaim.objects.filter(
+        payer_deposit_account=requestor_deposit_account,
+        payee_ethereum_address=provider_ethereum_address,
+        concent_use_case=ConcentUseCase.FORCED_PAYMENT.value,
+        closure_time__gte=parse_timestamp_to_utc_datetime(payments_ts),
+    ).exclude(
+        tx_hash=None,
+    )
 
 
 def discard_claim(deposit_claim: DepositClaim) -> bool:
--- a/concent_api/core/tests/test_payments_bankster.py
+++ b/concent_api/core/tests/test_payments_bankster.py
@@ -257,540 +257,340 @@ class FinalizePaymentBanksterTest(ConcentIntegrationTestCase):
 
 class SettleOverdueAcceptancesBanksterTest(ConcentIntegrationTestCase):
 
-    def test_that_settle_overdue_acceptances_should_return_none_if_subtask_costs_where_already_paid(self):
-        task_to_compute = self._get_deserialized_task_to_compute(
-            price=3000,
-        )
+    def setUp(self):
+        super().setUp()
+        self.get_deposit_value_return_value_default = 15000
+        self.task_to_compute = None
+        self.subtask_results_accepted_list = None
+        self.requestor_client = None
+        self.requestor_deposit_account = None
+        self.validate_list_of_transaction_mock = None
+        self.get_deposit_value_mock = None
+        self.get_list_of_payments_mock = None
 
-        subtask_results_accepted_list = [
+    def create_subtask_results_accepted_list(self, price, number_of_items=1):
+        self.task_to_compute = self._get_deserialized_task_to_compute(price=price)
+        self.subtask_results_accepted_list = [
             self._get_deserialized_subtask_results_accepted(
                 report_computed_task=self._get_deserialized_report_computed_task(
                     timestamp="2018-02-04 10:00:05",
-                    task_to_compute=task_to_compute
+                    task_to_compute=self.task_to_compute
                 )
-            )
+            ) for _ in range(number_of_items)
         ]
 
-        with mock.patch('core.payments.bankster.validate_list_of_transaction_timestamp') as validate_list_of_transaction_mock:
-            with mock.patch(
-                'core.payments.bankster.service.get_list_of_payments',
-                return_value=self._get_list_of_force_transactions(),
-            ) as get_list_of_payments_mock:
-                claim_against_requestor = settle_overdue_acceptances(
-                    requestor_ethereum_address=task_to_compute.requestor_ethereum_address,
-                    provider_ethereum_address=task_to_compute.provider_ethereum_address,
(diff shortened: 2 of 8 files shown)


concent_api/core/payments/bankster.py
```diff
@@ -456,6 +460,35 @@ def get_provider_payment_info(
...
+        except DepositClaim.DoesNotExist:
+            continue
+
+        # If the information in the database differs from the information on the blockchain,
+        # aborts the settlement and responds with HTTP 500.
+        if (
+            matching_claim.closure_time != force_payment.closure_time or
+            matching_claim.amount != force_payment.amount
+        ):
+            raise BanksterTransactionMismatchError(
```
pawelkisielewicz: consider log this situation before raising `BanksterTransactionMismatchError`

concent_api/core/tests/test_payments_bankster.py
```diff
@@ -256,208 +257,327 @@ def test_that_when_there_are_other_deposit_claims_finalize_payment_substract_the
...
+                        claim_against_requestor = settle_overdue_acceptances(
+                            requestor_ethereum_address=self.task_to_compute.requestor_ethereum_address,
+                            provider_ethereum_address=self.task_to_compute.provider_ethereum_address,
+                            acceptances=self.subtask_results_accepted_list,
+                            requestor_public_key=hex_to_bytes_convert(self.task_to_compute.requestor_public_key),
+                        )
+
+        return claim_against_requestor
+
+    def assert_mocked_function_were_called(self):
```
dybi: please rename to `assert_mocked_sci_functions_were_called`. 

BTW: :+1: for code extraction ;)

concent_api/core/tests/test_payments_bankster.py
```diff
@@ -256,208 +257,327 @@ def test_that_when_there_are_other_deposit_claims_finalize_payment_substract_the
...
+        self.get_list_of_payments_mock.assert_called_once()
+        self.validate_list_of_transaction_mock.assert_called_once()
+
+    def test_that_settle_overdue_acceptances_should_return_none_if_subtask_costs_where_already_paid(self):
+        """
+        In this test we have following calculations:
+
+        TaskToCompute price:                                    3000
+        Sum of amounts from list of forced transactions:        3000
+        Requestor deposit value:                                does not matter in this case
```
dybi: I would suggest stating the value anyway, ie:
```suggestion
         Requestor deposit value:                                15000 (does not matter in this case)
```

concent_api/core/tests/test_payments_bankster.py
```diff
@@ -256,208 +257,327 @@ def test_that_when_there_are_other_deposit_claims_finalize_payment_substract_the
 
 class SettleOverdueAcceptancesBanksterTest(ConcentIntegrationTestCase):
```
dybi: Big :+1: for the tests

concent_api/core/tests/test_payments_bankster.py

concent_api/core/tests/test_payments_bankster.py

concent_api/core/payments/bankster.py
```diff
@@ -439,15 +441,17 @@ def sum_subtask_price(subtask_results_accepted_list: List[SubtaskResultsAccepted
...
-    list_of_payments: List[BatchTransferEvent],
+    claims_of_settlement_payments: QuerySet,
     subtask_results_accepted_list: List[SubtaskResultsAccepted],
 ) -> Tuple[int, int]:
-    assert isinstance(list_of_payments, list)
+    assert isinstance(claims_of_settlement_payments, QuerySet)
     assert isinstance(list_of_forced_payments, list)
     assert isinstance(subtask_results_accepted_list, list)
 
     force_payments_price = sum_payments(list_of_forced_payments)
```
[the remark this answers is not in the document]

concent_api/core/payments/bankster.py
```diff
@@ -439,15 +441,17 @@ def sum_subtask_price(subtask_results_accepted_list: List[SubtaskResultsAccepted
 
 def get_provider_payment_info(
     list_of_forced_payments: List[ForcedPaymentEvent],
-    list_of_payments: List[BatchTransferEvent],
+    claims_of_settlement_payments: QuerySet,
```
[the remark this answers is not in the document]

concent_api/core/payments/bankster.py
```diff
@@ -439,15 +441,17 @@ def sum_subtask_price(subtask_results_accepted_list: List[SubtaskResultsAccepted
...
     subtask_results_accepted_list: List[SubtaskResultsAccepted],
 ) -> Tuple[int, int]:
-    assert isinstance(list_of_payments, list)
+    assert isinstance(claims_of_settlement_payments, QuerySet)
     assert isinstance(list_of_forced_payments, list)
     assert isinstance(subtask_results_accepted_list, list)
 
     force_payments_price = sum_payments(list_of_forced_payments)
-    payments_price = sum_payments(list_of_payments)
+    payments_price = claims_of_settlement_payments.aggregate(
```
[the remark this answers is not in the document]

concent_api/core/payments/bankster.py
```diff
@@ -347,27 +349,22 @@ def settle_overdue_acceptances(
             sum_of_existing_claims=Coalesce(Sum('amount'), 0)
         )
 
-        assert sum_of_existing_requestor_claims['sum_of_existing_claims'] >= 0
+        # Bankster finds all DatabaseClaims that represent settlement payments from requestor's Ethereum account to
+        # provider's Ethereum account and for which a transaction has already been sent to the Ethereum client.
+        claims_of_settlement_payments = DepositClaim.objects.filter(
```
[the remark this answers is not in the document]

concent_api/core/payments/bankster.py
```diff
@@ -347,27 +349,22 @@ def settle_overdue_acceptances(
...
+            concent_use_case=ConcentUseCase.FORCED_PAYMENT.value,
+        ).exclude(
+            tx_hash=None,
+        )
 
-        # If the existing claims against requestor's deposit are greater or equal to his current deposit,
-        # we can't add a new claim.
-        if requestor_deposit_value <= sum_of_existing_requestor_claims['sum_of_existing_claims']:
-            return None
+        validate_list_of_transaction_timestamp(acceptances)
```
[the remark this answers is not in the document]

concent_api/core/payments/bankster.py
```diff
@@ -347,27 +349,22 @@ def settle_overdue_acceptances(
             sum_of_existing_claims=Coalesce(Sum('amount'), 0)
         )
 
-        assert sum_of_existing_requestor_claims['sum_of_existing_claims'] >= 0
+        # Bankster finds all DatabaseClaims that represent settlement payments from requestor's Ethereum account to
```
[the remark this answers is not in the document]
rwrzesien: I think that is a little bit of overkill, it is a single call to a method chain, used once, with a comment above it.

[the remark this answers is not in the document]
dybi: this is why I suggest to extract it - to get rid of that comment ;)

concent_api/core/payments/bankster.py
dybi: ok, unhandled exception are handled by the middleware... if so, the question is if we even need a crash logger...

concent_api/core/payments/bankster.py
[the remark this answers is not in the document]
cameel: I agree with @dybi here. The clean code was exactly about things like this where you could replace the comment with a well named function.

I'm not sure if this style of having tons of tiny function will really be the most readable one but we agreed to try so let's start doing it.

concent_api/core/payments/bankster.py
rwrzesien: As discussed, not necessary in this case.

concent_api/core/payments/bankster.py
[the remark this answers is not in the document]

concent_api/core/payments/bankster.py
```diff
@@ -384,23 +381,68 @@ def sum_subtask_price(subtask_results_accepted_list: List[SubtaskResultsAccepted
...
-    force_payments_price = sum_payments(list_of_forced_payments)
-    payments_price = sum_payments(list_of_payments)
+    forced_satisfied_payments_price = sum_payments(list_of_forced_payments)
+    satisfied_payments_price = already_satisfied_claims.aggregate(
+        sum_of_already_satisfied_claims=Coalesce(Sum('amount'), 0)
+    )['sum_of_already_satisfied_claims']
     subtasks_price = sum_subtask_price(subtask_results_accepted_list)
 
-    amount_paid = payments_price + force_payments_price
+    amount_paid = satisfied_payments_price + forced_satisfied_payments_price
```
cameel: `payments_price` cannot be replaced with `satisfied_payments_price`. They're completely different things.

`payments_price` is `sum(LT)` - i.e. the amount the requestor has paid the normal way. If you remove it, you're ignoring all the non-forced payments.
`satisfied_payments_price + forced_satisfied_payments_price` is `sum(LF)`

After this change the code is computing `V` as `sum(LAR) - sum(LF)` instead of `sum(LAR) - sum(LT) - sum(LF)` as it should.

concent_api/core/validation.py
```diff
@@ -468,28 +467,17 @@ def substitute_new_report_computed_task_if_needed(
...
 
-    if (
-        (
-            any(youngest_payment_ts > closure_time for closure_time in list_of_transactions) or
-            len(list_of_transactions) == 0
-        ) and (
-            cut_off_time < youngest_payment_ts + settings.PAYMENT_DUE_TIME
-        )
-    ):
+    if cut_off_time < youngest_payment_ts + settings.PAYMENT_DUE_TIME:
```
cameel: This condition would be easier to understand for me as:

``` python
if youngest_payment_ts > cut_off_time - settings.PAYMENT_DUE_TIME:
```
because the focus here is on `youngest_payment_ts`, not `cut_off_time` (which is just the current time).

concent_api/core/validation.py
```diff
@@ -468,28 +467,17 @@ def substitute_new_report_computed_task_if_needed(
...
     """
 
     cut_off_time = get_current_utc_timestamp()
-
     youngest_payment_ts = max(subtask_results_accepted.payment_ts for subtask_results_accepted in acceptances)
 
-    if (
-        (
-            any(youngest_payment_ts > closure_time for closure_time in list_of_transactions) or
-            len(list_of_transactions) == 0
```
cameel: This condition ensured that a timestamp error was returned if there were any payments before the first acceptance? Or am I missing something?

This seems wrong and I don't remember it at the time we were doing #983, so why is it here?

rwrzesien:
@cameel So I have a good feeling that you should take a look on this. This way we can maybe avoid going back to this soon. I will try to fix those in the later part of next week.

cameel:
I did take a look on this. You have my review above.

rwrzesien:
@cameel Yes, I know, that was my point, thanks to that you found some things that can be fixed within the same pull request, so it is good.

cameel:
Oh, OK then :)
Yeah, let's fix them. And I really think that we should add the test cases from the blueprint as unit tests because these bugs should not have slipped through. It means that our current tests are too weak.

concent_api/core/payments/bankster.py
```diff
@@ -321,9 +313,14 @@ def settle_overdue_acceptances(
             transaction_type=TransactionType.FORCE,
         )
 
+        already_satisfied_claims = remove_duplicated_transactions_from_deposit_claims(
+            already_satisfied_claims,
```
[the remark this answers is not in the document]

concent_api/core/payments/bankster.py
```diff
@@ -384,23 +381,68 @@ def sum_subtask_price(subtask_results_accepted_list: List[SubtaskResultsAccepted
...
+        try:
+            matching_claim = already_satisfied_claims.get(tx_hash=force_payment.tx_hash)
+        except DepositClaim.DoesNotExist:
+            continue
+
+        # If the information in the database differs from the information on the blockchain,
+        # aborts the settlement and responds with HTTP 500.
+        if (
+            matching_claim.closure_time != force_payment.closure_time or
+            matching_claim.amount != force_payment.amount
```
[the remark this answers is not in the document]

concent_api/core/payments/bankster.py
```diff
@@ -384,23 +381,68 @@ def sum_subtask_price(subtask_results_accepted_list: List[SubtaskResultsAccepted
...
+    already_satisfied_claims: QuerySet,
+    list_of_forced_payments: List[ForcedPaymentEvent],
+) -> QuerySet:
+    """
+    Removes duplicates from claims of settlement payments with pending transaction
+    which are on list of forced payments by adding exclusions to DepositClaim queryset.
+    """
+    for force_payment in list_of_forced_payments:
+        try:
+            matching_claim = already_satisfied_claims.get(tx_hash=force_payment.tx_hash)
```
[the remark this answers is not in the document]
rwrzesien: OK but please take into consideration that a trade off for this is that I have to re-search for the matching `force_payment` from `list_of_forced_payments` for each of the matching claims to compare . Still, should be faster than making many queries until ie. `list_of_forced_payments` is extremely large and it will have only matching element which will result in one query.

concent_api/core/payments/bankster.py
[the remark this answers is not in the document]
rwrzesien: :+1: done, I have also updated referrences in whole Concent code, luckily there were not many of them.

concent_api/core/payments/bankster.py
```diff
@@ -321,9 +313,14 @@ def settle_overdue_acceptances(
...
 
+        already_satisfied_claims = remove_duplicated_transactions_from_deposit_claims(
+            already_satisfied_claims,
+            list_of_forced_payments,
+        )
+
         (_amount_paid, amount_pending) = get_provider_payment_info(
             list_of_forced_payments=list_of_forced_payments,
-            list_of_payments=list_of_transactions,
+            already_satisfied_claims=already_satisfied_claims,
```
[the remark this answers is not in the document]
rwrzesien: :+1: done, about second part maybe you can update https://github.com/golemfactory/concent/issues/1018 which is not started yet.

concent_api/core/payments/bankster.py
```diff
@@ -384,23 +381,68 @@ def sum_subtask_price(subtask_results_accepted_list: List[SubtaskResultsAccepted
...
+
+def get_already_satisfied_claims(requestor_deposit_account: DepositAccount, provider_ethereum_address: str) -> QuerySet:
+    """
+    Returns all DatabaseClaims that represent settlement payments from requestor's Ethereum account to
+    provider's Ethereum account and for which a transaction has already been sent to the Ethereum client.
+    """
+    return DepositClaim.objects.filter(
+        payer_deposit_account=requestor_deposit_account,
+        payee_ethereum_address=provider_ethereum_address,
+        concent_use_case=ConcentUseCase.FORCED_PAYMENT.value,
```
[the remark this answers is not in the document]

concent_api/core/payments/bankster.py
rwrzesien: :+1: done, I thought that (from issue):

> It uses both lists when checking existing settlement payments.

means that it should use *only* those two lists.

concent_api/core/validation.py
rwrzesien: :+1: done

concent_api/core/validation.py
rwrzesien: Good question, my only guess is that something went wrong during rebase in other pull requests, as I remember that there were some other changes in this part of the code.

concent_api/core/payments/bankster.py
```diff
@@ -383,24 +390,82 @@ def sum_subtask_price(subtask_results_accepted_list: List[SubtaskResultsAccepted
...
+    """
+    settlement_payment_tx_hashes = [force_payment.tx_hash for force_payment in list_of_settlement_payments]
+
+    for matching_claim in already_satisfied_claims.filter(tx_hash__in=settlement_payment_tx_hashes):
+        assert matching_claim.concent_use_case == ConcentUseCase.FORCED_PAYMENT
+
+        matching_force_payment = [
+            force_payment for force_payment in list_of_settlement_payments
+            if force_payment.tx_hash == matching_claim.tx_hash
+        ][0]
```
cameel: You should use `next()` here. You give it an iterator and it either returns the first element or raises `StopIteration` (if there are no elements). It's a more readable and also a tiny bit more efficient because it does not need to construct the whole list just to discard everything but the first element.

concent_api/core/payments/bankster.py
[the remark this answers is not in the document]
cameel: I know. I think the queries are slowest operation here, not the Python loop.

Searching for each element by going over a list makes it O(n^2) but you could make it O(n log n) but having both lists sorted by `tx_hash` and then going over them in parallel in a single `while` loop. This would complicate the code a bit though so I think it's not worth it. Having a lot of forced payments between a single requestor-provider pair will be very rare. Each one requires a situation in which the requestor did not pay and the provider started UC5.

concent_api/core/payments/bankster.py
cameel: Then I'd specifically write that it uses two lists :) I wrote that it uses **both** - i.e. both lists mentioned in the text above it, as opposed to the old code that was using just one of them.

Anyway, it should be clear from context anyway. Removing `LT` makes no sense in the context of goal stated in the issue - preventing multiple settlements by taking into account payments that are not yet on the blockchain. How would ignoring `LT` help us accomplish that?

concent_api/core/payments/bankster.py
```diff
@@ -292,38 +296,41 @@ def settle_overdue_acceptances(
...
         # Concent defines time T0 equal to oldest payment_ts from passed SubtaskResultAccepted messages from
         # subtask_results_accepted_list.
         oldest_payments_ts = min(subtask_results_accepted.payment_ts for subtask_results_accepted in acceptances)
 
-        # Concent gets list of transactions from payment API where timestamp >= T0.
-        list_of_transactions = service.get_list_of_payments(  # pylint: disable=no-value-for-parameter
+        already_satisfied_claims = get_already_satisfied_claims(
+            requestor_deposit_account,
+            provider_ethereum_address,
+            oldest_payments_ts,
```
cameel: The only thing you do with `already_satisfied_claims` is pass it into `remove_duplicated_transactions_from_deposit_claims()`. You don't really need it here. Why not move it inside that function and rename the function to something like `find_unconfirmed_settlement_payments()`?

concent_api/core/payments/bankster.py
cameel: What if there's no claim corresponding to the payment in our database? We remove claims when their transactions are confirmed so it's a very common situation.

concent_api/core/payments/bankster.py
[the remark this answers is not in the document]
rwrzesien: Right, I have just suggested that maybe before starting https://github.com/golemfactory/concent/issues/1018 (we marked this one as blocked until other issues related with UC5 are resolved) it might be good for you to review the issue description.

concent_api/core/payments/bankster.py
rwrzesien: Then we won't reach this place in code, because in this part we iterate only over claims that have `tx_hash` matched with one of the settlement payments from `list_of_settlement_payments`. So if there is `matching_claim`, it means that there must be object with the same `tx_hash` in `list_of_settlement_payments`.

concent_api/core/payments/bankster.py

concent_api/core/payments/bankster.py
```diff
@@ -383,24 +390,82 @@ def sum_subtask_price(subtask_results_accepted_list: List[SubtaskResultsAccepted
...
+        if (
+            matching_claim.closure_time != matching_force_payment.closure_time or
+            matching_claim.amount != matching_force_payment.amount
+        ):
+            raise BanksterTransactionMismatchError(
+                f'There is a data mismatch between DepositClaim and ForcedPaymentEvent with transaction hash: '
+                f'{matching_force_payment.tx_hash}'
+            )
+
+        already_satisfied_claims = already_satisfied_claims.exclude(pk=matching_claim.pk)
```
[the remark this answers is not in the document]

concent_api/core/payments/bankster.py

concent_api/core/payments/bankster.py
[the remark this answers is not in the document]
cameel: I don't think updating the issue to exactly match the code is necessary. The gist of the issue is that the names are bad or went out of sync with specs and need to be improved.

There are low level details such as specific names that I found problematic and justification for why they should be renamed but that's only to make it clear what I found particularly bad about those names. that' specifically to make it easy for the implementer to get what I meant and apply it to the new names if they change.

concent_api/core/payments/bankster.py
cameel: Is it done? I still see the old version here.

concent_api/core/payments/bankster.py
cameel: OK. You're right. It seems fine after all.

### Example
Just for reference, an example I created to analyze this case:

required confirmations: 6

payments:
- A (100 confirmations)
- B (80 confirmations)
- C (30 confirmations)
- D (6 confirmations)
- E (6 confirmations)
- F (5 confirmations)
- G (0 confirmations)
- H (transaction not submitted to the blockchain client yet)

claims in the database:
- 1 (hash=E)
- 2 (hash=F)
- 3 (hash=G)
- 4 (hash=None)

Then:
- `list_of_settlement_payments`: A, B, C, D, E
- `settlement_payment_tx_hashes`: A, B, C, D, E
- `already_satisfied_claims`: 1, 2, 3
- `already_satisfied_claims.filter(tx_hash__in=[A, B, C, D, E])`: 1

The case I was worried about was claim F or G (has claim with a hash but is not visible on the list obtained from the blockchain). But the filter takes care of it.

concent_api/core/payments/bankster.py
rwrzesien: Yes, it is done, you see old version because commented line doesn't get updated here. It is problematic sometimes because you have to go to file view to see if it was really fixed, but it is good that the code is kept here for reference of what was originally commented.

cameel: No, I mean in the file view. I know that the snippets do not get updated. It was still visible in the code when I wrote that comment and this thread was not marked as "outdated'. I see that it has disappeared now. Maybe you did not push all the commits?

Issue #1033, opened 2018-12-17
Review model custom validation and replace it with Django validators.


rwrzesien:
Review model custom validation and replace it with Django validators.

Discovered in https://github.com/golemfactory/concent/pull/1030.

PR #1045, opened 2018-12-20
Chore review model custom validation and replace it with django validators

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/1033


concent_api/core/models.py
```diff
@@ -625,41 +627,37 @@ class RecipientType(ChoiceEnum):
...
+        validators=[
+            MinLengthValidator(TASK_OWNER_KEY_LENGTH),
+            MaxLengthValidator(TASK_OWNER_KEY_LENGTH),
+        ]
+    )
+    provider_eth_account = CharField(
+        max_length=ETHEREUM_ADDRESS_LENGTH,
+        validators=[
+            MinLengthValidator(ETHEREUM_ADDRESS_LENGTH),
+            MaxLengthValidator(ETHEREUM_ADDRESS_LENGTH),
```
dybi: I think this is redundant to `max_length=ETHEREUM_ADDRESS_LENGTH`. Correct me if I'm wrong, but my understanding was that using `max_length=ETHEREUM_ADDRESS_LENGTH` makes Django use `MaxLengthValidator` under the hood.

concent_api/core/models.py
```diff
@@ -723,24 +721,39 @@ class DepositAccount(Model):
     objects = DepositAccountManager()
 
     client = ForeignKey(Client)
-    ethereum_address = CharField(max_length=ETHEREUM_ADDRESS_LENGTH, unique=True)
+    ethereum_address = CharField(
+        max_length=ETHEREUM_ADDRESS_LENGTH,
+        unique=True,
+        validators=[
+            MinLengthValidator(ETHEREUM_ADDRESS_LENGTH),
+            MaxLengthValidator(ETHEREUM_ADDRESS_LENGTH),
```
dybi: same here

concent_api/core/models.py
```diff
@@ -723,24 +721,39 @@ class DepositAccount(Model):
...
 
 class DepositClaim(Model):
     subtask_id = CharField(max_length=MESSAGE_TASK_ID_MAX_LENGTH, blank=True, null=True)
     payer_deposit_account = ForeignKey(DepositAccount)
-    payee_ethereum_address = CharField(max_length=ETHEREUM_ADDRESS_LENGTH)
+    payee_ethereum_address = CharField(
+        max_length=ETHEREUM_ADDRESS_LENGTH,
+        validators=[
+            MinLengthValidator(ETHEREUM_ADDRESS_LENGTH),
+            MaxLengthValidator(ETHEREUM_ADDRESS_LENGTH),
```
dybi: and here

concent_api/core/models.py
```diff
@@ -723,24 +721,39 @@ class DepositAccount(Model):
...
     amount = DecimalField(max_digits=BIG_ENDIAN_INT_MAX_DIGITS, decimal_places=0)
-    tx_hash = CharField(max_length=64, blank=True, null=True, unique=True)
+    tx_hash = CharField(
+        max_length=ETHEREUM_TRANSACTION_HASH_LENGTH,
+        blank=True,
+        null=True,
+        unique=True,
+        validators=[
+            MinLengthValidator(ETHEREUM_TRANSACTION_HASH_LENGTH),
+            MaxLengthValidator(ETHEREUM_TRANSACTION_HASH_LENGTH),
```
dybi: and here

concent_api/core/tests/test_unit_model.py
```diff
@@ -214,24 +212,6 @@ def setUp(self):
         self.client.clean()
         self.client.save()
 
-    def test_that_exception_is_raised_when_ethereum_address_has_wrong_length(self):
```
dybi: Hm... why do we want to remove these?

concent_api/core/models.py
rwrzesien: Yes, I was thinking about this and just wanted to make it more visible, but you are right, after thinking it through again I agree it should be removed.

concent_api/core/tests/test_unit_model.py
rwrzesien: Because this is no longer our custom validation, but we are using framework functionality, and we don't want to test framework functionality, as it has its own tests.

concent_api/core/models.py
pawelkisielewicz: Would it be good place for one line comment with this information? @dybi ?

rwrzesien: :+1: done

concent_api/core/models.py
rwrzesien: :+1: done

concent_api/core/models.py
rwrzesien: :+1: done

concent_api/core/models.py
rwrzesien: :+1: done

rwrzesien:
> I haven't use django validators before, so I ask. Before we had ours ValidationError with error message. Now in case of validation error we get ValidationError, but from BaseValidator. It's constructor takes also optional parameter message. What do You think about keeping error messages and adding previous messages as parameter to new validators?

I think that if it is not required then we shouldn't care, and I don't think there is a specified content of those messages anywhere. I also think that default messages have enough information to identify the problem if it happens.

concent_api/core/tests/test_unit_model.py
dybi: that was a tricky question :)

Issue #1046, opened 2018-12-22
Concent doesn't handle "JPEG" format correctly in ComputeTaskDef in Verification Use Case


dybi:
this is connected with sentry report: https://talkback.golem.network/sentry/concent/issues/44745/?query=is:unresolved

PR #1047, opened 2018-12-22
Bugfix wrong jpeg handling in verification use case

dybi:
resolves:  https://github.com/golemfactory/concent/issues/1046


signing_service/signing_service/tests/daily_tresholds/2018-11-06
```diff
@@ -0,0 +1 @@
+0
```
rwrzesien: I guess this file appeared here accidentally?

concent_api/common/tests/test_decorators.py
```diff
@@ -301,8 +301,8 @@ def test_that_non_nesting_atomic_decorator_will_not_raise_exception_if_transacti
 
     @override_settings(
         DETECT_NESTED_TRANSACTIONS=False,
-    )
-    def test_that_non_nesting_atomic_decorator_will_not_raise_exception_if_transaction_is_nested_using_same_databases_and_detect_nested_transaction_setting_is_false(self):  # pylint: disable=no-self-use
+    )  # pylint: disable=no-self-use
```
rwrzesien: :+1: Thanks for fixing this.

concent_api/conductor/models.py
```diff
@@ -76,7 +76,7 @@ class BlenderSubtaskDefinition(Model):
     """
 
     class OutputFormat(ChoiceEnum):
-        JPG = 'jpg'
+        JPEG = 'jpeg'
```
rwrzesien: So what about `jpg` format? We should not support it anymore?

concent_api/core/queue_operations.py
```diff
@@ -3,7 +3,8 @@
 from conductor.tasks import blender_verification_request
 from common.helpers import get_storage_result_file_path
 from common.helpers import get_storage_source_file_path
-from core.utils import extract_name_from_scene_file_path
+from core.utils import extract_name_from_scene_file_path, adjust_format_name
```
rwrzesien: Please review imports in this pull request.

concent_api/conductor/tasks.py
```diff
@@ -54,11 +53,9 @@ def blender_verification_request(
         f'With blender_crop_script: {bool(blender_crop_script)}',
         subtask_id=subtask_id
     )
-    validate_frames(frames)
```
rwrzesien: I understand the purpose of this change, I want just to point out that doing this in core instead of here will not start whole verification process. Wouldn't this result in a different outcome for requestor/provider when this validation fails? @cameel ?

signing_service/signing_service/tests/daily_tresholds/2018-11-06
dybi: true, removing

concent_api/common/tests/test_decorators.py
dybi: :+1:

concent_api/core/queue_operations.py
dybi: fixed

concent_api/conductor/tasks.py
cameel: I think it's a change for the better. If the list of frames is invalid, `core`, should immediately reject the message as invalid without even starting the verification.

Actually, even after this change the validation happens too late. It's now in `send_blender_verification_request()` but this function is called after we have already accepted the request, updated the subtask and committed the changes to the database. It should be performed along with all the other validations for `TaskToCompute`.

Also, conductor is a separate component so I think it should validate its input separately from `core`. So it would be fine to have this check here too. Or at least `assert` that the frames are valid - just like it already does for example for `output_format`.

dybi: ok, I will move it insisde `validate_task_to_compute`

Issue #804, opened 2018-09-12
SigingService: Exclude `if __name__ == .__main__. ` from coverage report


rwrzesien:
There is a part of SigingService which is used to run main script. There is no way to unit test it (at least I couldn't find any which doesn't modify the current code). This part doesn't contain any logic, just instantiates components, so it could be excluded from report.

```
[report]
exclude_lines =
    if __name__ == .__main__.:
```

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/160478329

PR #1048, opened 2018-12-27
Chore increase coverage in signing service

dybi:
resolves: https://github.com/golemfactory/concent/issues/804

### DIFF
--- a/signing_service/signing_service/main.py
+++ b/signing_service/signing_service/main.py
@@ -1,0 +1,39 @@
+import os
+
+import logging.config
+from raven import Client
+
+from signing_service.signing_service import _parse_arguments
+from signing_service.signing_service import SigningService
+from signing_service.utils import get_notifier
+
+crash_logger = logging.getLogger('crash')
+
+if __name__ == '__main__':
+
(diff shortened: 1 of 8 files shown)


signing_service/signing_service/main.py
```diff
@@ -0,0 +1,39 @@
...
+
+from signing_service.signing_service import _parse_arguments
+from signing_service.signing_service import SigningService
+from signing_service.utils import get_notifier
+
+crash_logger = logging.getLogger('crash')
+
+if __name__ == '__main__':
+
+    logging.config.fileConfig(os.path.join(os.path.dirname(os.path.abspath(__file__)), '..', 'logging.ini'))
```
rwrzesien: How about utility function for creating this path?

signing_service/signing_service/tests/daily_tresholds/2018-11-06
```diff
@@ -0,0 +1 @@
+0
```
rwrzesien: Same here, should this file be here?

dybi: my bad, will delete it ;)

PR #1050, opened 2019-01-07
Chore update golem messages from v2.17.0 to v2.22.0

kbeker:
There is no issue for that

### DIFF
--- a/concent_api/concent_api/tests/test_shutdown_mode.py
+++ b/concent_api/concent_api/tests/test_shutdown_mode.py
@@ -158,7 +158,7 @@ class SoftShutdownModeTest(ConcentIntegrationTestCase):
         subtask_results_accepted_list = [
             self._get_deserialized_subtask_results_accepted(
                 timestamp="2018-02-05 10:00:15",
-                payment_ts="2018-02-05 12:00:16",
+                payment_ts="2018-02-05 9:00:16",
                 report_computed_task=self._get_deserialized_report_computed_task(
                     timestamp="2018-02-05 10:00:05",
                     task_to_compute=self._get_deserialized_task_to_compute(
--- a/concent_api/core/message_handlers.py
+++ b/concent_api/core/message_handlers.py
@@ -832,7 +832,7 @@ def handle_send_force_payment(
             payment_ts              = parse_datetime_to_timestamp(claim_against_requestor.closure_time),
             task_owner_key          = requestor_ethereum_public_key,
             provider_eth_account    = provider_eth_address,
-            amount_paid             = claim_against_requestor.amount,
+            amount_paid             = claim_against_requestor.amount_as_int,
             amount_pending          = 0,
             recipient_type          = message.concents.ForcePaymentCommitted.Actor.Provider,
         )
@@ -841,7 +841,7 @@ def handle_send_force_payment(
             payment_ts              = parse_datetime_to_timestamp(claim_against_requestor.closure_time),
             task_owner_key          = requestor_ethereum_public_key,
             provider_eth_account    = provider_eth_address,
-            amount_paid             = claim_against_requestor.amount,
+            amount_paid             = claim_against_requestor.amount_as_int,
             amount_pending          = 0,
             recipient_type          = message.concents.ForcePaymentCommitted.Actor.Requestor,
         )
@@ -1142,8 +1142,8 @@ def handle_messages_from_database(client_public_key: bytes) -> Union[message.Mes
                 payment_ts=parse_datetime_to_timestamp(payment_message.payment_ts),
                 task_owner_key=payment_message.task_owner_key.tobytes(),
                 provider_eth_account=payment_message.provider_eth_account,
-                amount_paid=payment_message.amount_paid,
-                amount_pending=payment_message.amount_pending,
+                amount_paid=payment_message.amount_paid_as_int,
+                amount_pending=payment_message.amount_pending_as_int,
             )
             if payment_message.recipient_type == PaymentInfo.RecipientType.Requestor.name:  # pylint: disable=no-member
                 response_to_client.recipient_type = message.concents.ForcePaymentCommitted.Actor.Requestor
--- a/concent_api/core/models.py
+++ b/concent_api/core/models.py
@@ -421,6 +421,13 @@ class Subtask(Model):
         new._current_state_name = new.state  # pylint: disable=no-member
         return new
 
+    def _deserialize_database_message(self, serialized_message: StoredMessage) -> message.Message:  # pylint: disable=no-self-use
+        if isinstance(serialized_message.data, bytes):
+            deserialized_message = deserialize_message(serialized_message.data)
+        else:
+            deserialized_message = deserialize_message(serialized_message.data.tobytes())  # pylint: disable=no-member
+        return deserialized_message
+
     def clean(self) -> None:
         super().clean()
 
@@ -491,29 +498,22 @@ class Subtask(Model):
                     )
                 })
 
-        if isinstance(self.report_computed_task.data, bytes):
-            deserialized_report_computed_task = deserialize_message(self.report_computed_task.data)
-        else:
-            deserialized_report_computed_task = deserialize_message(self.report_computed_task.data.tobytes())  # pylint: disable=no-member
+        deserialized_report_computed_task = self._deserialize_database_message(self.report_computed_task)
 
         # If available, the report_computed_task nested in force_get_task_result must match report_computed_task.
-        if (
-            self.force_get_task_result is not None and
-            deserialize_message(self.force_get_task_result.data).report_computed_task != deserialized_report_computed_task
-        ):
-            raise ValidationError({
-                'force_get_task_result': "ReportComputedTask nested in ForceGetTaskResult must match Subtask's ReportComputedTask."
-            })
+        if self.force_get_task_result is not None:
+            deserialized_force_get_task_result = self._deserialize_database_message(self.force_get_task_result)
+            if deserialized_force_get_task_result.report_computed_task != deserialized_report_computed_task:
+                raise ValidationError({
+                    'force_get_task_result': "ReportComputedTask nested in ForceGetTaskResult must match Subtask's ReportComputedTask."
+                })
 
         if not self.result_package_size == deserialized_report_computed_task.size:
             raise ValidationError({
                 'result_package_size': "ReportComputedTask size mismatch"
             })
 
-        if isinstance(self.task_to_compute.data, bytes):
-            deserialized_task_to_compute = deserialize_message(self.task_to_compute.data)
-        else:
-            deserialized_task_to_compute = deserialize_message(self.task_to_compute.data.tobytes())  # pylint: disable=no-member
+        deserialized_task_to_compute = self._deserialize_database_message(self.task_to_compute)
 
         if not parse_datetime_to_timestamp(self.computation_deadline) == deserialized_task_to_compute.compute_task_def['deadline']:
             raise ValidationError({
--- a/concent_api/core/models.py
+++ b/concent_api/core/models.py
@@ -651,6 +651,14 @@ class PaymentInfo(Model):
     )
     pending_response = ForeignKey(PendingResponse, related_name='payments')
 
+    @property
+    def amount_paid_as_int(self) -> int:
+        return int(self.amount_paid)
+
+    @property
+    def amount_pending_as_int(self) -> int:
+        return int(self.amount_pending)
+
     def clean(self) -> None:
         if self.task_owner_key == self.provider_eth_account:
             raise ValidationError({
@@ -748,6 +756,10 @@ class DepositClaim(Model):
     created_at = DateTimeField(auto_now_add=True)
     modified_at = DateTimeField(auto_now=True)
     closure_time = DateTimeField(blank=True, null=True)
+
+    @property
+    def amount_as_int(self) -> int:
+        return int(self.amount)
 
     class Meta:
         unique_together = ('subtask_id', 'concent_use_case', 'payee_ethereum_address')
--- a/concent_api/core/tests/test_auth_accept_or_reject_results.py
+++ b/concent_api/core/tests/test_auth_accept_or_reject_results.py
@@ -443,7 +443,7 @@ class AuthAcceptOrRejectIntegrationTest(ConcentIntegrationTestCase):
             timestamp="2018-02-05 10:00:43",
             subtask_results_accepted=self._get_deserialized_subtask_results_accepted(
                 timestamp="2018-02-05 10:00:43",
-                payment_ts="2018-02-05 10:00:44",
+                payment_ts="2018-02-05 9:00:44",
                 report_computed_task=report_computed_task,
             )
         )
@@ -477,7 +477,7 @@ class AuthAcceptOrRejectIntegrationTest(ConcentIntegrationTestCase):
             timestamp="2018-02-05 10:00:43",
             subtask_results_accepted=self._get_deserialized_subtask_results_accepted(
                 timestamp="2018-02-05 10:00:43",
-                payment_ts="2018-02-05 10:00:44",
+                payment_ts="2018-02-05 9:00:44",
                 report_computed_task=report_computed_task,
             )
         )
@@ -553,7 +553,7 @@ class AuthAcceptOrRejectIntegrationTest(ConcentIntegrationTestCase):
                 'timestamp': parse_iso_date_to_timestamp("2018-02-05 11:00:02"),
                 'subtask_results_accepted.timestamp': parse_iso_date_to_timestamp("2018-02-05 10:00:43"),
                 'subtask_results_accepted.task_to_compute.compute_task_def': task_to_compute.compute_task_def,
-                'subtask_results_accepted.payment_ts': parse_iso_date_to_timestamp("2018-02-05 10:00:44")
+                'subtask_results_accepted.payment_ts': parse_iso_date_to_timestamp("2018-02-05 9:00:44")
             }
         )
         self._assert_stored_message_counter_not_increased()
@@ -702,7 +702,7 @@ class AuthAcceptOrRejectIntegrationTest(ConcentIntegrationTestCase):
             timestamp="2018-02-05 10:00:43",
             subtask_results_accepted=self._get_deserialized_subtask_results_accepted(
                 timestamp="2018-02-05 10:00:43",
-                payment_ts="2018-02-05 10:00:44",
+                payment_ts="2018-02-05 9:00:44",
                 report_computed_task=report_computed_task,
             )
         )
--- a/concent_api/core/tests/test_core_views.py
+++ b/concent_api/core/tests/test_core_views.py
@@ -19,6 +19,7 @@ from common.constants import ERROR_IN_GOLEM_MESSAGE
 from common.helpers import get_current_utc_timestamp
 from common.helpers import parse_timestamp_to_utc_datetime
 from common.testing_helpers import generate_ecc_key_pair
+from core.tests.constants_for_tests import ZERO_SIGNATURE
 from core.message_handlers import store_subtask
 from core.models import Client
 from core.models import StoredMessage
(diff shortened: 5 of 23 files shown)


concent_api/core/constants.py
```diff
@@ -90,3 +90,6 @@ class VerificationResult(IntEnum):
 
 # Defines a string that is returned from Bankster to Core views to indicate timestamp error case.
 BANKSTER_TIMESTAMP_ERROR = 'timestamp_error'
+
+
+ZERO_SIGNATURE = b'\x00' * 65
```
rwrzesien: Isn't this available in golem-messages? If yes, maybe it would be better to import it from there.

concent_api/core/message_handlers.py
```diff
@@ -832,7 +832,7 @@ def handle_send_force_payment(
             payment_ts              = parse_datetime_to_timestamp(claim_against_requestor.closure_time),
             task_owner_key          = requestor_ethereum_public_key,
             provider_eth_account    = provider_eth_address,
-            amount_paid             = claim_against_requestor.amount,
+            amount_paid             = int(claim_against_requestor.amount),
```
rwrzesien: Please add property to `DepositClaim` like `amount_as_int` which will do the converting and use it.

concent_api/concent_api/tests/test_shutdown_mode.py
```diff
@@ -158,7 +158,7 @@ def test_in_soft_shutdown_mode_concent_should_not_accept_payment_requests(self):
         subtask_results_accepted_list = [
             self._get_deserialized_subtask_results_accepted(
                 timestamp="2018-02-05 10:00:15",
-                payment_ts="2018-02-05 12:00:16",
+                payment_ts="2018-02-05 9:00:16",
```
rwrzesien: Why this change?

concent_api/core/message_handlers.py
```diff
@@ -1142,8 +1142,8 @@ def handle_messages_from_database(client_public_key: bytes) -> Union[message.Mes
                 payment_ts=parse_datetime_to_timestamp(payment_message.payment_ts),
                 task_owner_key=payment_message.task_owner_key.tobytes(),
                 provider_eth_account=payment_message.provider_eth_account,
-                amount_paid=payment_message.amount_paid,
-                amount_pending=payment_message.amount_pending,
+                amount_paid=int(payment_message.amount_paid),
```
rwrzesien: Same here, add properties to `PaymentInfo` model.

middleman_protocol/middleman_protocol/concent_golem_messages/message.py
```diff
@@ -21,9 +23,16 @@ class NonceAbstractMessage(Message):
         'nonce',
     ] + Message.__slots__
 
+    def slots(self) -> List[List[Union[str, bytes, int]]]:
+        slots = super(NonceAbstractMessage, self).slots()
```
rwrzesien: You can call `super()` without parameters in Python3.

middleman_protocol/middleman_protocol/concent_golem_messages/message.py
```diff
@@ -21,9 +23,16 @@ class NonceAbstractMessage(Message):
...
+        if ['nonce', self.nonce] in slots:  # pylint: disable=no-member
+            slots[slots.index(['nonce', self.nonce])][1] = str(self.nonce)  # pylint: disable=no-member
```
rwrzesien: Is this really necessary? Maybe it is OK to keep internally `nonce` as int, and add here property too for getting is as string? But if there is a case that it is required for this conversion to happen in `slots()` please let me know.

middleman_protocol/middleman_protocol/concent_golem_messages/message.py
```diff
@@ -21,9 +23,16 @@ class NonceAbstractMessage(Message):
...
+        return slots
+
     def deserialize_slot(self, key: str, value: Any) -> Any:
         value = super().deserialize_slot(key, value)
         if key == 'nonce':
+            value = int(value)
```
rwrzesien: If you apply comment above you can remove this.

middleman_protocol/middleman_protocol/concent_golem_messages/message.py
```diff
@@ -49,6 +58,16 @@ class TransactionAbstractMessage(NonceAbstractMessage):
         'data',
     ] + NonceAbstractMessage.__slots__
 
+    def slots(self) -> List[List[Union[str, bytes, int]]]:
+        slots = super(TransactionAbstractMessage, self).slots()
```
rwrzesien: Same here, why is it required to convert those values?

concent_api/core/models.py
```diff
@@ -497,13 +497,15 @@ def clean(self) -> None:
...
 
         # If available, the report_computed_task nested in force_get_task_result must match report_computed_task.
-        if (
-            self.force_get_task_result is not None and
-            deserialize_message(self.force_get_task_result.data).report_computed_task != deserialized_report_computed_task
-        ):
-            raise ValidationError({
-                'force_get_task_result': "ReportComputedTask nested in ForceGetTaskResult must match Subtask's ReportComputedTask."
-            })
+        if self.force_get_task_result is not None:
```
pawelkisielewicz: Please look at the comment above. In your version it does not match to new block of code in my opinion. 
Second thing- I know that `clean` is very long method now and you did it similar to the rest, but please extract your new block of code to one inner method (e.g. `_deserialize_force_get_task_result`) and do the whole logic in this method.

concent_api/core/tests/test_auth_accept_or_reject_results.py
```diff
@@ -443,7 +443,7 @@ def test_requestor_sends_subtask_results_accepted_and_concent_should_return_it_t
             timestamp="2018-02-05 10:00:43",
             subtask_results_accepted=self._get_deserialized_subtask_results_accepted(
                 timestamp="2018-02-05 10:00:43",
-                payment_ts="2018-02-05 10:00:44",
+                payment_ts="2018-02-05 9:00:44",
```
pawelkisielewicz: In tests there is lot of changes like this - why is it changed and why exactly one hour?

concent_api/core/constants.py
pawelkisielewicz: What is this signature?

concent_api/core/tests/test_integration_force_payment.py
```diff
@@ -49,7 +49,7 @@ def test_provider_send_force_payment_with_subtask_results_accepted_signed_by_dif
             ),
             self._get_deserialized_subtask_results_accepted(
                 timestamp="2018-02-05 9:00:15",
-                payment_ts="2018-02-05 11:00:16",
+                payment_ts="2018-02-05 8:00:16",
```
pawelkisielewicz: why changed and why 3 hours? In other place I've seen also 2 hours

concent_api/core/tests/test_payments_bankster.py
```diff
@@ -417,14 +417,14 @@ def test_that_settle_overdue_acceptances_should_return_claim_deposit_with_amount
...
-                    payment_ts="2018-02-04 10:00:05",
+                    payment_ts="2018-02-04 9:00:05",
                 ),
                 self._get_deserialized_subtask_results_accepted(
                     report_computed_task=self._get_deserialized_report_computed_task(
                         timestamp="2018-02-04 23:00:05",
                         task_to_compute=task_to_compute
                     ),
-                    payment_ts="2018-02-05 10:00:05",
+                    payment_ts="2018-02-04 10:00:05",
```
pawelkisielewicz: Is this change necessary and why? BTW, timestamp of created above `report_computed_task` is 2018-02-04 23:00:05, so I think it is wrong to set `payment_ts` as You did.. It should be later

concent_api/core/models.py
Jakub89: Also, in previous version, if `self.force_get_task_result` was in fact `None`, we've raised `ValidationError`. Now it's just pass, is that a bug or we now want to allow `None` value?

concent_api/core/message_handlers.py

concent_api/core/message_handlers.py

middleman_protocol/middleman_protocol/concent_golem_messages/message.py
kbeker: Done

middleman_protocol/middleman_protocol/concent_golem_messages/message.py
kbeker: In one of few last updates golem_messages Golem Team changed `cbor2` library to `cbor`, because it's faster than `cbor2`. This library has issues with serialization big integers and that's why we need to convert it during serialization and opposite thing during deserialization. We in our code don't need `nonce`, `v`, `r` or `s` as a string, but this serialization needs it.

middleman_protocol/middleman_protocol/concent_golem_messages/message.py
kbeker: Answered above https://github.com/golemfactory/concent/pull/1050#discussion_r245930285

middleman_protocol/middleman_protocol/concent_golem_messages/message.py
kbeker: Added comment in code

concent_api/core/tests/test_payments_bankster.py
kbeker: Newest `golem_messages` does not allow to create `SubtaskResultsAccepted` with `payment_ts` from future. That's why here and in all others `payment_ts` changes it is changed that way.

concent_api/core/tests/test_integration_force_payment.py
kbeker: Answered in https://github.com/golemfactory/concent/pull/1050#discussion_r245933221

concent_api/core/tests/test_auth_accept_or_reject_results.py
kbeker: Answered in https://github.com/golemfactory/concent/pull/1050#discussion_r245933221

concent_api/core/models.py
kbeker: @pawelkisielewicz 
> Please look at the comment above. In your version it does not match to new block of code in my opinion.

This block of code has same rules of working like before. Only one difference in this is that we can deserialize it if `self.force_get_task_result.data` is database object instead of string of bytes.

>Second thing- I know that clean is very long method now and you did it similar to the rest, but please extract your new block of code to one inner method (e.g. _deserialize_force_get_task_result) and do the whole logic in this method.

I think that this issue and this branch is not good place for refactor like this. It could be added new issue for refactor `clean()` method in `Subtask` model. If we want and we have time for this we can refactor whole `Subtask` not just a part of it.

@Jakub89 
>Also, in previous version, if self.force_get_task_result was in fact None, we've raised ValidationError. Now it's just pass, is that a bug or we now want to allow None value?

Previously we raised `ValidationError` only when `self.force_get_task_result is not None` **and** nested `report_computed_task of correctly deserialized message was different than `report_computed_task` stored separately. This block of code still works the same:
- we check if `self.force_get_task_result` exists
- try to deserialize it
- compare `self.force_get_task_results.report_computed_task` with `report_computed_task` of this same `Subtask` but stored separately and if it's not the same we raise `ValidationError`

concent_api/core/constants.py
pawelkisielewicz: seems strange for me... but if is is so - :+1:

concent_api/core/tests/test_payments_bankster.py
pawelkisielewicz: OK.. But still we have freeze_time("**2018-02-04 10:00:15**") and timestamp="**2018-02-04 23:00:05**" in one of `report_computed_task`. This should fail for me, how can someone send at **10:00:15** `report_computed_task` with timestamp "**23:00:05**" (both in the same day). 
@kbeker You didn't change this times, so thanks for explanation. :+1: 
Should we validate such situations or leave it as it is now? @cameel ?

concent_api/core/models.py
pawelkisielewicz: >I think that this issue and this branch is not good place for refactor like this. It could be added new issue for refactor clean() method in Subtask model. If we want and we have time for this we can refactor whole Subtask not just a part of it.
I disagree, extraction some code (which is changed by You) to one method is not refactor. If You are touching some code and there is possibility to leave it better I think You should do it.

concent_api/core/tests/test_payments_bankster.py
cameel: `RCT.timestamp > SRA.payment_ts` would be weird but it should be very rare because it would require one of them to be in the future or in the past when sent. Clients validate that received messages are not too new or too old so only a modified client could do that. And with a modified client our only concern is whether it can break something when it's malicious.

I don't think it breaks anything. We have safeguards against `payments_ts` too far in the past where it matters (e.g. #1036). RCT cannot be too far in the future either because it must be earlier than the deadline.

Such weird situations with timestamps are very often possible with nested messages. My approach here is not to try to validate all the relationships (that would be best but it's too complex because we can never treat timestamps as exact and there are so many of them). Just validate the things that can cause problems.

concent_api/core/constants.py
```diff
@@ -90,3 +91,6 @@ class VerificationResult(IntEnum):
 
 # Defines a string that is returned from Bankster to Core views to indicate timestamp error case.
 BANKSTER_TIMESTAMP_ERROR = 'timestamp_error'
+
+
+ZERO_SIGNATURE = b'\0' * Message.SIG_LEN
```
dybi: it is used only in the tests - it shouldn't be here... Please move it to a new modul `tests_constants` or `constants_for_tests`

concent_api/core/models.py
```diff
@@ -497,13 +497,15 @@ def clean(self) -> None:
...
+            if isinstance(self.force_get_task_result.data, bytes):
```
dybi: why this new `if ... else`? When it would be possible for `self.force_get_task_result.data` not to be `bytes`?

concent_api/core/tests/test_core_views.py
```diff
@@ -190,17 +190,22 @@ def test_send_should_return_http_400_if_data_is_incorrect(self):
...
+        )
+        malformed_dumped_message = dump(
                 data,
                 self.PROVIDER_PRIVATE_KEY,
-                CONCENT_PUBLIC_KEY),
+                CONCENT_PUBLIC_KEY,
+        )
+        response = self.send_request(
+            url='core:send',
+            data=malformed_dumped_message[:-1],
```
dybi: please move this slicing (`[-:1]` to the variable definition - otherwise the name is wrong ;)

concent_api/core/tests/test_core_views.py
```diff
@@ -477,22 +482,30 @@ def test_send_task_to_compute_without_public_key_should_return_http_400(self):
...
+                report_computed_task=message.tasks.ReportComputedTask(
+                    task_to_compute=self.task_to_compute
+                ),
+            ),
+            privkey=self.PROVIDER_PRIVATE_KEY,
+            pubkey=settings.CONCENT_PUBLIC_KEY,
+        )
         with mock.patch('core.decorators.log') as log:
-            response = self._send_force_report_computed_task()
+            response = self._send_force_report_computed_task(serialized_force_report_computed_task=dumped_message[:-1])
```
dybi: please move this slicing (`[-:1]` to the variable definition - otherwise the name is wrong ;)

concent_api/core/tests/test_integration_force_payment.py
```diff
@@ -231,65 +231,6 @@ def test_that_if_provider_sends_force_payment_with_two_different_provider_ethere
             }
         )
 
-    def test_provider_send_force_payment_beyond_payment_time_concent_should_reject(self):
```
dybi: this was removed because now GM don't allow to create such messages with wrong `payment_ts` or what was the reason?

concent_api/core/tests/test_integration_report_computed_task.py
```diff
@@ -1835,54 +1834,6 @@ def test_provider_forces_computed_task_report_empty_key_returns_400_error(self):
 
         self._assert_client_count_is_equal(0)
 
-    def test_requestor_sends_ack_report_computed_task_with_message_cannot_compute_task(self):
```
dybi: and this? another validation in GM?

concent_api/core/tests/test_payments_bankster.py
dybi: I agree with @pawelkisielewicz - such a timestamp of the 2nd `report_computed_task` doesn't make sense... please modify it to be more reasonable

concent_api/core/tests/test_task_verification_result.py
```diff
@@ -201,7 +203,7 @@ def setUp(self):
         (self.PROVIDER_PRIVATE_KEY, self.PROVIDER_PUBLIC_KEY) = generate_ecc_key_pair()
         (self.REQUESTOR_PRIVATE_KEY, self.REQUESTOR_PUBLIC_KEY) = generate_ecc_key_pair()
 
-        task_to_compute = TaskToComputeFactory()
+        task_to_compute = TaskToComputeFactory(sig=ZERO_SIGNATURE)
```
dybi: why do we need this?

middleman_protocol/middleman_protocol/concent_golem_messages/message.py
```diff
@@ -49,6 +61,18 @@ class TransactionAbstractMessage(NonceAbstractMessage):
         'data',
     ] + NonceAbstractMessage.__slots__
 
+    def slots(self) -> List[List[Union[str, bytes, int]]]:
+        # cbor library has issuses with big integers serialization
+        # before serialization every big integer need to be convert into string
+        slots = super().slots()
+        if ['gasprice', self.gasprice] in slots:  # pylint: disable=no-member
```
dybi: I would be for some kind of conversion map, instead of all this `if`s ;)

middleman_protocol/middleman_protocol/concent_golem_messages/message.py
```diff
@@ -114,9 +140,23 @@ class SignedTransaction(TransactionAbstractMessage):
         's',
     ] + TransactionAbstractMessage.__slots__
 
+    def slots(self) -> List[List[Union[str, bytes, int]]]:
+        # cbor library has issuses with big integers serialization
+        # before serialization every big integer need to be convert into string
+        slots = super().slots()
+        if ['v', self.v] in slots:  # pylint: disable=no-member
```
dybi: same here - map would be much better in my opinion

concent_api/core/constants.py
kbeker: Done

concent_api/core/models.py
kbeker: Because there are some cases when `self.force_get_task_result.data` could be `StoredMessage` object instead of `bytes`. We have same situation with `self.report_computed_task.data`

concent_api/core/tests/test_core_views.py
kbeker: Changed :)

concent_api/core/tests/test_core_views.py
kbeker: Changed :)

concent_api/core/tests/test_integration_force_payment.py
kbeker: That's right, GM now does not allow to create messages with `payment_ts` from future

concent_api/core/tests/test_integration_report_computed_task.py
kbeker: Yup, GM now contains validation for wrong nested message.

concent_api/core/tests/test_task_verification_result.py
kbeker: @dybi 
When we store to database single message then during serialization its gets own signature - if we serialialize it without any keys then to message is assign `fake_sign'. Let me show you how it's work. I'm gonna ignore parameters which we don't need here e.g: 
```python
store_subtask(
                task_to_compute=task_to_compute,
                report_computed_task=report_computed_task,
            )
```

We are here storing some `Subtask` with `task_to_compute` and `report_compute_task` which has as a nested message `task_to_compute`. When we serialize `task_to_compute` it will get `sig=fake_sig`. Same situation will be with `report_computed_task` beside `task_to_compute` which is nested in RCT. Nested message won't get its own `fake_sig` and that's why it will be different then `task_to_compute` stored separately. When `Subtask` model runs `clean()` method it will get `ValidationError` because `task_to_compute != report_computed_task.task_to_cumpute`. This is the only reason why we have to add `ZERO_SIGNATURE` here. It won't happen in normal cases because `TaskToCompute` is always signed.

Issue #1053, opened 2019-01-07
Fix `non_nesting_atomic` to verify transaction nesting at runtime instead of on startup


rwrzesien:
Original sentry issues:
https://talkback.golem.network/sentry/concent/issues/44786/
https://talkback.golem.network/sentry/concent/issues/44785/

Currently `non_nesting_atomic` is implemented wrongly that it calls `get_connection(using)` during server startup. This cause it to try to connect to databases that are not available for certain Concent features (ie. `verifier`). It should call it only during runtime.

PR #1054, opened 2019-01-08
Bugfix fix non nesting atomic to verify transaction nesting in runtime instead of on startup

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/1053

### DIFF
--- a/concent_api/common/decorators.py
+++ b/concent_api/common/decorators.py
@@ -6,9 +6,9 @@ from typing import Union
 import traceback
 
 from django.conf import settings
-from django.db import DEFAULT_DB_ALIAS
 from django.db.transaction import Atomic
 from django.db.transaction import get_connection
+from django.utils.decorators import ContextDecorator
 
 from common.exceptions import ConcentFeatureIsNotAvailable
 from common.exceptions import ConcentPendingTransactionError
@@ -54,13 +54,34 @@ def log_task_errors(task: Callable) -> Callable:
     return wrapper
 
 
-def non_nesting_atomic(using: Union[str, Callable], savepoint: bool=True) -> Atomic:
-    if (
-        settings.DETECT_NESTED_TRANSACTIONS and
-        get_connection(using).in_atomic_block
-    ):
-        raise ConcentPendingTransactionError
-    if callable(using):
-        return Atomic(DEFAULT_DB_ALIAS, savepoint)(using)
-    else:
-        return Atomic(using, savepoint)
+class non_nesting_atomic(ContextDecorator):
+    """
+    This class acts as either decorator or context manager.
+    It discovers when the callable or code is being run in nested database transaction and raises exception.
+    Otherwise it wraps callable or code in database transaction.
+    """
+
+    def __init__(self, using: Union[str, Callable], savepoint: bool=True):
+        self.using = using
+        self.atomic_context_decorator = Atomic(self.using, savepoint)
+
+    def __enter__(self) -> None:
+        if (
+            settings.DETECT_NESTED_TRANSACTIONS and
+            get_connection(self.using).in_atomic_block
+        ):
+            raise ConcentPendingTransactionError(
+                'Creation of new transaction in the same database detected inside pending transaction.'
+            )
+        return self.atomic_context_decorator.__enter__()
+
+    def __exit__(self, *args: Any, **kwargs: Any) -> None:
+        return self.atomic_context_decorator.__exit__(*args, **kwargs)
+
+    def __call__(self, func: Callable) -> Callable:
+        @wraps(func)
+        def decorated(*args: Any, **kwargs: Any):
+            with self:
+                return func(*args, **kwargs)
+
+        return decorated
--- a/concent_api/common/tests/test_decorators.py
+++ b/concent_api/common/tests/test_decorators.py
@@ -1,6 +1,7 @@
 import json
 import mock
 from django.conf import settings
+from django.db.transaction import get_connection
 from django.http import HttpResponse
 from django.http import HttpResponseNotAllowed
 from django.http import JsonResponse
@@ -297,7 +298,8 @@ class NonNestingAtomicDecoratorTestCase(TransactionTestCase):
     def test_that_non_nesting_atomic_decorator_will_not_raise_exception_if_transaction_is_nested_using_different_databases(self):  # pylint: disable=no-self-use
         with non_nesting_atomic(using='control'):
             with non_nesting_atomic(using='storage'):
-                pass
+                self.assertTrue(get_connection('control').in_atomic_block)
+                self.assertTrue(get_connection('storage').in_atomic_block)
 
     @override_settings(
         DETECT_NESTED_TRANSACTIONS=False,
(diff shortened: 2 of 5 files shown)


concent_api/common/decorators.py
```diff
@@ -54,13 +54,38 @@ def wrapper(*args: Any, **kwargs: Any) -> None:
...
-    if (
-        settings.DETECT_NESTED_TRANSACTIONS and
-        get_connection(using).in_atomic_block
-    ):
-        raise ConcentPendingTransactionError
-    if callable(using):
-        return Atomic(DEFAULT_DB_ALIAS, savepoint)(using)
-    else:
-        return Atomic(using, savepoint)
+class non_nesting_atomic(ContextDecorator):
```
pawelkisielewicz: If it is class shouldn't name be camel-case?

concent_api/common/tests/test_decorators.py
```diff
@@ -297,12 +298,20 @@ def test_that_non_nesting_atomic_decorator_will_raise_exception_if_transaction_i
...
             with non_nesting_atomic(using='control'):
-                pass
+                self.assertTrue(get_connection('control').in_atomic_block)
+
+    def test_that_non_nesting_atomic_decorator_wraps_in_transaction_when_used_as_decorator(self):
+        @non_nesting_atomic(using='control')
+        def wrapped_function():
+            self.assertTrue(get_connection('control').in_atomic_block)
+
+        wrapped_function()
```
dybi: one test case is missing:
```python
    def test_that_decorated_function_called_in_transaction_should_rasie_exception(self):
        @non_nesting_atomic(using='control')
        def wrapped_function():
            pass

            with self.assertRaises(ConcentPendingTransactionError):
                wrapped_function()
```

concent_api/common/decorators.py
```diff
@@ -54,13 +54,38 @@ def wrapper(*args: Any, **kwargs: Any) -> None:
...
+            raise ConcentPendingTransactionError
+        return self.atomic_context_decorator.__enter__()
+
+    def __exit__(self, *args: Any, **kwargs: Any) -> None:
+        return self.atomic_context_decorator.__exit__(*args, **kwargs)
+
+    def __call__(self, func: Callable) -> Callable:
+        @wraps(func)
+        def decorated(*args: Any, **kwargs: Any):
+            if (
```
dybi: this `if` is redundant and unnecessary... please remove it.

concent_api/common/decorators.py
dybi: :+1: for `ContextDecorator`

concent_api/common/decorators.py
```diff
@@ -54,13 +54,38 @@ def wrapper(*args: Any, **kwargs: Any) -> None:
...
+class non_nesting_atomic(ContextDecorator):
+    """
+    This class acts as either decorator or context manager.
+    It discovers when the callable or code is being run in nested database transaction and raises exception.
+    Otherwise it wraps callable or code in database transaction.
+    """
+
+    def __init__(self, using: Union[str, Callable], savepoint: bool=True):
+        self.using = using
+        self.savepoint = savepoint
```
dybi: this field is unnecessary, please remove it

concent_api/common/decorators.py
rwrzesien: It could be but actually none of our tools is complaining (lint, flake, pycharm). I also think that when using this is more natural to have a decorator in this form than in camel case. It would look weirder if I do 

```
@NonNestingAtomic(using='control')
def something():
```

so I can sacrifice class name in one place in favor of using it as decorator in many places.

concent_api/common/decorators.py
rwrzesien: Why? It is called when you use this class as decorator.

concent_api/common/tests/test_decorators.py
rwrzesien: :+1: done

concent_api/common/decorators.py
rwrzesien: :+1: good catch

concent_api/common/decorators.py
rwrzesien: I would keep it, without it we won't have possibility to call `Atomic()` with this parameter if we need to. And there is no extra logic behind that.

dybi: but you can still call `Atomic` in this way:
```python
self.atomic_context_decorator = Atomic(self.using, savepoint)
```

rwrzesien: :+1: done

concent_api/conductor/views.py
```diff
@@ -24,6 +24,7 @@
 @provides_concent_feature('conductor-urls')
 @require_POST
 @csrf_exempt
+@transaction.non_atomic_requests(using='storage')
```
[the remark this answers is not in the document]
rwrzesien: It makes this view not being wrapped in transaction, as we have it configured by default (see: `ATOMIC_REQUESTS` are set to `True` in settings for each database). We also have this in `core.views`. Without it, whole view is wrapped in transaction, and there are calls inside which are creating nested transactions.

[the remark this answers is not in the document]
dybi: just as I thought ;) thanks!

concent_api/core/payments/storage.py
```diff
@@ -86,7 +84,6 @@ def get_all_tx(self) -> List[Transaction]:
             for ethereum_transaction in PendingEthereumTransaction.objects.all()
         ]
 
-    @non_nesting_atomic(using='control')
```
cameel: What was wrong here? You removed the decorator from the function but everything is still wrapped in `@atomic` so this should not change behavior in any way.

concent_api/core/payments/bankster.py
```diff
@@ -241,7 +241,6 @@ def finalize_payment(deposit_claim: DepositClaim) -> Optional[str]:
         # The code below is executed in another transaction, so - in theory - deposit_claim object could be modified in
         # the meantime. Here we are working under assumption that it's not the case and it is coder's responsibility to
         # ensure that.
-        # Bankster puts transaction ID in DepositClaim.tx_hash.
```
cameel: This commit says that you "Removed nested transactions in Bankster" but it only removes this comment.

concent_api/conductor/views.py
```diff
@@ -24,6 +24,7 @@
...
 @non_nesting_atomic(using='storage')
 def report_upload(_request: HttpRequest, file_path: str) -> HttpResponse:
```
cameel: You're turning atomic requests off and then back on immediately:
``` python
@transaction.non_atomic_requests(using='storage')
@non_nesting_atomic(using='storage')
```

This can only detect something if someone calls our view from within a transaction. But we never call views explicitly. We let Django router do this. So in practice it's not necessary.

It seems redundant but it's not wrong. But if you prefer it that way, you should do the same with `send/`, `receive/` and any other views for consistency.

concent_api/core/payments/bankster.py
rwrzesien: Originally it was, but after merging https://github.com/golemfactory/concent/pull/1041/ which has same changes this is only what is left from this commit.

concent_api/core/payments/storage.py
rwrzesien: Not everything, `self._get_nonce()` is not ;)

concent_api/conductor/views.py
rwrzesien: > But we never call views explicitly. We let Django router do this. So in practice it's not necessary.

I agree, but I would prefer to keep it for consistency, in case if it would be used differently and someone would miss this. And views in `send/` and `receive` are also using `@transaction.non_atomic_requests()` already.

Issue #1062, opened 2019-01-14
api-e2e-force-accept-or-reject-results-test.py case 2a is not complete


Jakub89:
Issue happen  when trying to deploy concent on cluster. CI block deployment because script that runs all e2e tests fails, because `api-e2e-force-accept-or-reject-results-test.py` creates unresolved  `DepositClaim` that effect the outcome of `api-e2e-force-payment.py`.

rwrzesien:
@Jakub89 Could you please provide more info about the effect? Does it block some amount of the deposit?

Jakub89:
@rwrzesien Yes, when we have unresolved DepositClaim(), it's being add to sum of payments in `force-payment` test. So the bug is just wrong(other then we expect) value that's being claimed from requestor.

PR #1063, opened 2019-01-14
Bugfix e2e tests for acceptance use case

Jakub89:
Resolves: #1062

### DIFF
--- a/concent_api/api-e2e-force-accept-or-reject-results-test.py
+++ b/concent_api/api-e2e-force-accept-or-reject-results-test.py
@@ -356,7 +356,7 @@ def test_case_2b_not_enough_funds(cluster_consts: ProtocolConstants, cluster_url
 
 @count_fails
 def test_case_2a_send_duplicated_force_subtask_results(cluster_consts: ProtocolConstants, cluster_url: str) -> None:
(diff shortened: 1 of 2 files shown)


concent_api/api-e2e-force-accept-or-reject-results-test.py
```diff
@@ -356,7 +356,7 @@ def test_case_2b_not_enough_funds(cluster_consts: ProtocolConstants, cluster_url
 
 @count_fails
 def test_case_2a_send_duplicated_force_subtask_results(cluster_consts: ProtocolConstants, cluster_url: str) -> None:
-    #  Test CASE 2A + 2D + 3 - Send ForceSubtaskResults with same task_id as stored by Concent before
+    #  Test CASE 2A + 2D + 3 + 4B + 5 - Send ForceSubtaskResults with same task_id as stored by Concent before
```
dybi: shouldn't this me modified in `test_case_2a_send_duplicated_force_subtask_results`? ;)

Jakub89: I'm not sure, that part "4B + 5" is repeated in most of this e2e tests, so the main purpose of this particular test is to check duplication. I would leave it as it is.

Issue #1056, opened 2019-01-09
get_or_create_with_retry() will not be able to handle all IntegrityErrors it's supposed to


dybi:
The issue originates from comment: https://github.com/golemfactory/concent/pull/1041#pullrequestreview-187567211

EDIT by @cameel:
"Get or create" operations implemented at the application level always have an inherent race condition. They need to check if the object exists and then either fetch or create it - but there's a short time gap between the check and the creation. Another process running in parallel can insert the row in that time. This is not merely a theoretical problem: see for example #486.

When this happens, the create operation in the first process fails with `IntegrityError` (`pgcode`: `UNIQUE_VIOLATION`) due to non-unique primary key. It's safe to deal with it by simply retrying the operation. #1041/#1042 introduced `get_or_create_with_retry()` which is a wrapper over Django's `get_or_create()` which does just that.

The problem with the wrapper is that it does not commit its changes to the database. This means that when it's called within an `@atomic` block, the `IntegrityError` may be raised at the end of that block when the transaction is committed - i.e. outside of the `try`/`except` block. This is because at the default transaction isolation level used by Django (`READ COMMITTED`) the processes do not see each other's changes until commit and the database won't detect a conflict until that point. As a result the `IntegrityError` will not be caught by the wrapper.

The simplest way to fix the wrapper is to put the body of the `try` block in an `@atomic` block.

This caused some controversy in https://github.com/golemfactory/concent/pull/1041#pullrequestreview-187567211 so please try to reproduce such a situation and report your findings in comments. It would actually be best to have a test instead but unit tests that test stuff related to transactions are often tricky - so it's enough to modify code locally to cause a situation in which there's an `IntegrityError` on commit and make sure that it can happen and that the fix works.

Also, while at it, please replace `get_or_create()` in the `except` block with just a simple `get()`. If the first create failed, it's pointless to try to create it again. This will save us one database query.

PaweuB:
Please provide more information about Issue. 
Is it critical?

cameel:
@PaweuB I have added a proper description.

PR #1070, opened 2019-01-16
Use non_nestion_atomic in get_or_create_with_retry

pawelkisielewicz:
Resolves https://github.com/golemfactory/concent/issues/1056

### DIFF
--- a/concent_api/api-e2e-force-accept-or-reject-results-test.py
+++ b/concent_api/api-e2e-force-accept-or-reject-results-test.py
@@ -3,57 +3,32 @@
 import os
 import sys
 import time
-from freezegun import freeze_time
 from typing import Optional
+
+import requests
+from freezegun import freeze_time
+from golem_messages import message
 from mock import Mock
 
-from golem_messages import message
-
-from common.helpers import get_current_utc_timestamp
-from common.helpers import sign_message
-
 from api_testing_common import api_request
+from api_testing_common import calculate_deadline
+from api_testing_common import calculate_deadline_too_far_in_the_future
+from api_testing_common import calculate_timestamp
 from api_testing_common import count_fails
+from api_testing_common import create_ack_report_computed_task
 from api_testing_common import create_client_auth_message
+from api_testing_common import create_force_subtask_results
 from api_testing_common import create_signed_report_computed_task
 from api_testing_common import create_signed_subtask_results_accepted
 from api_testing_common import create_signed_task_to_compute
 from api_testing_common import receive_pending_messages_for_requestor_and_provider
-from api_testing_common import REPORT_COMPUTED_TASK_SIZE
 from api_testing_common import run_tests
 from api_testing_common import timestamp_to_isoformat
+from common.helpers import get_current_utc_timestamp
+from common.helpers import sign_message
 from protocol_constants import ProtocolConstants
 
-import requests
-
-from core.utils import calculate_maximum_download_time
-
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "concent_api.settings")
-
-
-def create_force_subtask_results(
-    timestamp: Optional[str]=None,
-    ack_report_computed_task: Optional[message.tasks.AckReportComputedTask]=None,
-) -> message.concents.ForceSubtaskResults:
-    with freeze_time(timestamp):
-        return message.concents.ForceSubtaskResults(
-            ack_report_computed_task=ack_report_computed_task,
-        )
-
-
-def create_ack_report_computed_task(
-    requestor_private_key: bytes,
-    timestamp: Optional[str]=None,
-    report_computed_task: Optional[message.tasks.ReportComputedTask]=None,
-) -> message.tasks.AckReportComputedTask:
-    with freeze_time(timestamp):
-        signed_message: message.tasks.AckReportComputedTask = sign_message(
-            message.tasks.AckReportComputedTask(
-                report_computed_task=report_computed_task,
-            ),
-            requestor_private_key,
-        )
-        return signed_message
 
 
 def create_force_subtask_results_response(
@@ -83,31 +58,6 @@ def create_subtask_results_rejected(
             requestor_private_key,
         )
         return signed_message
-
-
-def calculate_timestamp(current_time: int, concent_messaging_time: int, minimum_upload_rate: int) -> str:
-    return timestamp_to_isoformat(
-        current_time - (2 * concent_messaging_time + _precalculate_subtask_verification_time(minimum_upload_rate, concent_messaging_time))
-    )
-
-
-def calculate_deadline(current_time: int, concent_messaging_time: int, minimum_upload_rate: int) -> int:
-    return current_time - (concent_messaging_time + _precalculate_subtask_verification_time(minimum_upload_rate, concent_messaging_time))
-
-
-def calculate_deadline_too_far_in_the_future(current_time: int, minimum_upload_rate: int, concent_messaging_time: int) -> int:
-    return current_time - (1 + (20 * _precalculate_subtask_verification_time(minimum_upload_rate, concent_messaging_time)))
-
-
-def _precalculate_subtask_verification_time(minimum_upload_rate: int, concent_messaging_time: int) -> int:
-    maxiumum_download_time = calculate_maximum_download_time(
-        size=REPORT_COMPUTED_TASK_SIZE,
-        rate=minimum_upload_rate,
-    )
-    return (
-        (4 * concent_messaging_time) +
-        (3 * maxiumum_download_time)
-    )
 
 
 @count_fails
--- a/concent_api/api-e2e-multi-threads-test.py
+++ b/concent_api/api-e2e-multi-threads-test.py
@@ -3,8 +3,8 @@ import os
 import sys
 import time
 from typing import Optional
-import requests
 
+import requests
 from golem_messages import message
 from golem_messages.factories.tasks import ReportComputedTaskFactory
 from golem_messages.message.concents import ForceGetTaskResult
@@ -12,23 +12,31 @@ from golem_messages.message.concents import ForceReportComputedTask
 from golem_messages.message.tasks import AckReportComputedTask
 from golem_messages.message.tasks import ReportComputedTask
 
-from api_testing_common import call_function_in_threads
-from api_testing_common import count_fails, assert_content_equal
 from api_testing_common import PROVIDER_PRIVATE_KEY
 from api_testing_common import REQUESTOR_PRIVATE_KEY
 from api_testing_common import api_request
+from api_testing_common import assert_content_equal
+from api_testing_common import calculate_deadline
+from api_testing_common import calculate_timestamp
+from api_testing_common import call_function_in_threads
+from api_testing_common import count_fails
+from api_testing_common import create_ack_report_computed_task
+from api_testing_common import create_force_subtask_results
+from api_testing_common import create_signed_report_computed_task
 from api_testing_common import create_signed_task_to_compute
 from api_testing_common import run_tests
 from api_testing_common import timestamp_to_isoformat
 from common.helpers import get_current_utc_timestamp
 from common.helpers import sign_message
+from common.testing_helpers import generate_ecc_key_pair
+from protocol_constants import ProtocolConstants
 
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "concent_api.settings")
 
 REPORT_COMPUTED_TASK_SIZE = 10
 
 NUMBER_OF_TESTING_THREADS = 3  # changing this variable remember to correct number of expected responses
-MAXIMUM_WAITING_TIME_FOR_ALL_RESPONSES = 3  # seconds
+MAXIMUM_WAITING_TIME_FOR_ALL_RESPONSES = 6  # seconds
 
 responses_global = []  # type: list
 
@@ -119,6 +127,32 @@ def send_correct_force_get_task_result(
     responses_global.append(response.__class__.__name__)
 
 
+def send_correct_force_subtask_results(
+    cluster_url: str,
+    report_computed_task: ReportComputedTask,
+    current_time: int,
+    provider_private_key: bytes,
+    requestor_private_key: bytes,
+    expected_status: Optional[int] = None,
+) -> None:
+    response = api_request(
+        cluster_url,
+        'send',
+        provider_private_key,
+        CONCENT_PUBLIC_KEY,
+        create_force_subtask_results(
+            timestamp=timestamp_to_isoformat(current_time),
+            ack_report_computed_task=create_ack_report_computed_task(
+                timestamp=timestamp_to_isoformat(current_time),
+                report_computed_task=report_computed_task,
+                requestor_private_key=requestor_private_key
+            )
+        ),
+        expected_status=expected_status,
+    )
+    responses_global.append(response.__class__.__name__)
+
+
 @count_fails
 def test_case_multiple_requests_concerning_one_subtask_will_be_processed_one_by_one_if_subtask_exists_in_database(
     cluster_consts: str,
@@ -185,6 +219,53 @@ def test_case_multiple_force_get_task_result_concerning_one_subtask_will_be_proc
             break
 
     expected_responses = ['AckForceGetTaskResult', 'ServiceRefused', 'ServiceRefused']
+    assert len(expected_responses) == NUMBER_OF_TESTING_THREADS, 'Did you change number of testing threads and forget to change expected_responses?'
+
+    print('Responses = ' + str(responses_global))
+    assert_content_equal(actual=responses_global, expected=expected_responses)
+    print('Single test passed successfully')
+
+
+@count_fails
+def test_case_multiple_force_subtask_results_does_not_cause_integrity_errors(
+    cluster_consts: ProtocolConstants,
+    cluster_url: str
+) -> None:
+    clear_responses()
+    provider_private_key, provider_public_key = generate_ecc_key_pair()
+    requestor_private_key, requestor_public_key = generate_ecc_key_pair()
+    current_time = get_current_utc_timestamp()
+    signed_task_to_compute = create_signed_task_to_compute(
+        timestamp=calculate_timestamp(current_time, cluster_consts.concent_messaging_time, cluster_consts.minimum_upload_rate),
+        deadline=calculate_deadline(current_time, cluster_consts.concent_messaging_time, cluster_consts.minimum_upload_rate),
+        price=1000,
+        provider_public_key=provider_public_key,
+        provider_private_key=provider_private_key,
+        requestor_public_key=requestor_public_key,
+        requestor_private_key=requestor_private_key
+    )
+    report_computed_task = create_signed_report_computed_task(
+        task_to_compute=signed_task_to_compute,
+        provider_private_key=provider_private_key,
+    )
+
+    call_function_in_threads(
+        func=send_correct_force_subtask_results,
+        number_of_threads=NUMBER_OF_TESTING_THREADS,
+        cluster_url=cluster_url,
+        report_computed_task=report_computed_task,
+        current_time=current_time,
+        provider_private_key=provider_private_key,
+        requestor_private_key=requestor_private_key,
+    )
+
+    end_time = time.time() + MAXIMUM_WAITING_TIME_FOR_ALL_RESPONSES
+    while len(responses_global) != NUMBER_OF_TESTING_THREADS:
+        time.sleep(0.1)
+        if time.time() >= end_time:
+            break
+
+    expected_responses = ['ServiceRefused', 'ServiceRefused', 'ServiceRefused']
     assert len(expected_responses) == NUMBER_OF_TESTING_THREADS, 'Did you change number of testing threads and forget to change expected_responses?'
 
     print('Responses = ' + str(responses_global))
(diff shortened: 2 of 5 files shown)


concent_api/api-e2e-multi-threads-test.py
```diff
@@ -119,6 +128,32 @@ def send_correct_force_get_task_result(
     responses_global.append(response.__class__.__name__)
 
 
+def send_correct_force_subtask_results(
+    cluster_url: str,
+    signed_report_computed_task: ReportComputedTask,
```
dybi: why is `signed_report_computed_task` needed?

concent_api/common/decorators.py
```diff
@@ -85,3 +89,35 @@ def decorated(*args: Any, **kwargs: Any) -> None:
...
+        while True:
+            try:
+                result = self.func(*args, **kwargs)
+                break
+            except IntegrityError:
+                self._validate_if_max_number_of_tries_exceeded()
+        return result
+
+    def _initialize_retry_parameters(self) -> None:
+        self.max_number_of_retries = settings.MAXIMUM_NUMBER_OF_RETRIES  # pylint: disable=attribute-defined-outside-init
```
dybi: this should be the part of the `__init__`, as it needs to be defined only once

concent_api/common/decorators.py
```diff
@@ -85,3 +89,35 @@ def decorated(*args: Any, **kwargs: Any) -> None:
...
+        self.func = func
+
+    def __call__(self, *args: Any, **kwargs: Any) -> Any:
+        self._initialize_retry_parameters()
+        while True:
+            try:
+                result = self.func(*args, **kwargs)
+                break
+            except IntegrityError:
+                self._validate_if_max_number_of_tries_exceeded()
```
dybi: please rename to `validate_number_of_retries`

concent_api/core/exceptions.py
```diff
@@ -83,3 +83,7 @@ class BanksterTimestampError(Exception):
 
 class BanksterTransactionMismatchError(Exception):
     pass
+
+
+class MaxRetriesExceededError(ConcentBaseException):
```
dybi: It inherits from `ConcentBaseException` which is handled as `HTTP400` is really a bad request when `MaxRetriesExceededError`? I wouldn't say so.

concent_api/api-e2e-multi-threads-test.py
pawelkisielewicz: In all tests `report_computed_task` is signed. The naming is confusing, because in all other places there is just `report_computed_task` and here `signed_report_computed_task`. All of them are signed, my mistake for confusing names. Renamed

concent_api/common/decorators.py
pawelkisielewicz: :+1: good catch

concent_api/common/decorators.py
pawelkisielewicz: I renamed to `_validate_number_of_retries`, because this is "private" method and should not be called from outside. Is it OK or it should be just `validate_number_of_retries`? What do You think?

concent_api/core/exceptions.py
pawelkisielewicz: OK, changed. I did it to provide return message, but there is no need, Exception type is pretty obvious.

concent_api/common/decorators.py
```diff
@@ -85,3 +89,32 @@ def decorated(*args: Any, **kwargs: Any) -> None:
                 return func(*args, **kwargs)
 
         return decorated
+
+
+class retry_on_database_error(object):
```
rwrzesien: There is no need to inherit from `object` in Python3.

concent_api/common/decorators.py
```diff
@@ -1,3 +1,4 @@
+import time
```
rwrzesien: Should be directly above to `import time`.

concent_api/common/decorators.py
```diff
@@ -85,3 +89,32 @@ def decorated(*args: Any, **kwargs: Any) -> None:
                 return func(*args, **kwargs)
 
         return decorated
+
+
+class retry_on_database_error(object):
+
+    def __init__(self, func: Callable) -> None:
+        self.func = func
+        self.max_number_of_retries = settings.MAXIMUM_NUMBER_OF_RETRIES  # pylint: disable=attribute-defined-outside-init
```
rwrzesien: There is no need to make it a class attribute. You never modify it, so you can use settings directly in `_validate_number_of_retries`.

concent_api/common/decorators.py
```diff
@@ -85,3 +89,32 @@ def decorated(*args: Any, **kwargs: Any) -> None:
...
+
+    def __call__(self, *args: Any, **kwargs: Any) -> Any:
+        self._initialize_retry_parameters()
+        while True:
+            try:
+                result = self.func(*args, **kwargs)
+                break
+            except IntegrityError:
+                self._validate_number_of_retries()
```
rwrzesien: Shouldn't you also check here if the exception type is `pg_errorcodes.UNIQUE_VIOLATION`?

concent_api/common/decorators.py
```diff
@@ -85,3 +89,32 @@ def decorated(*args: Any, **kwargs: Any) -> None:
...
+        return result
+
+    def _initialize_retry_parameters(self) -> None:
+        self.wait_time_in_sec = settings.STARTING_WAIT_TIME_FOR_RETRY_IN_SECONDS  # pylint: disable=attribute-defined-outside-init
+        self.current_number_of_retries = 0  # pylint: disable=attribute-defined-outside-init
+
+    def _validate_number_of_retries(self) -> None:
```
rwrzesien: Please rename to `_increase_retry_parameters`. There is no real validation here (data is not from client).

concent_api/common/decorators.py
```diff
@@ -85,3 +89,32 @@ def decorated(*args: Any, **kwargs: Any) -> None:
...
+        if self.current_number_of_retries < self.max_number_of_retries:
+            self.wait_time_in_sec *= 2
+            self.current_number_of_retries += 1
+            time.sleep(self.wait_time_in_sec)
+        else:
+            raise MaxRetriesExceededError
```
rwrzesien: Please add error message in exception.

concent_api/core/exceptions.py
```diff
@@ -83,3 +83,7 @@ class BanksterTimestampError(Exception):
 
 class BanksterTransactionMismatchError(Exception):
     pass
+
+
+class MaxRetriesExceededError(Exception):
```
rwrzesien: Isn't this exception that we want to handle somehow?

concent_api/common/decorators.py
```diff
@@ -85,3 +89,32 @@ def decorated(*args: Any, **kwargs: Any) -> None:
...
         return decorated
+
+
+class retry_on_database_error(object):
+
+    def __init__(self, func: Callable) -> None:
+        self.func = func
+        self.max_number_of_retries = settings.MAXIMUM_NUMBER_OF_RETRIES  # pylint: disable=attribute-defined-outside-init
+
+    def __call__(self, *args: Any, **kwargs: Any) -> Any:
```
rwrzesien: Are you sure this will be called each time in runtime, not only on server initialization?

concent_api/core/payments/bankster.py
```diff
@@ -75,19 +76,17 @@ def claim_deposit(
...
         )
 
         if is_claim_against_provider:
-            provider_client: Client = get_or_create_with_retry(Client, public_key=provider_public_key)
+            provider_client = Client.objects.get_or_create_full_clean(public_key=provider_public_key)
 
-            provider_deposit_account: DepositAccount = get_or_create_with_retry(
-                DepositAccount,
+            provider_deposit_account = DepositAccount.objects.get_or_create_full_clean(
                 client=provider_client,
```
cameel: I see that it's already been broken before the pull request but anyway, there's a problem here: we're not locking `DepositAccount` before doing operations on `DepositClaims`. The atomic block below **must** fetch `provider_deposit_account` and `requestor_deposit_account` again with `select_for_update()`. Otherwise any other process would be able to run `claim_deposit()` in parallel and not see the funds required by the current process as already reserved.

concent_api/common/decorators.py
cameel: I think this is an overkill. It's too generic. We don't need any generic code for retrying on database errors because there is only one, verify specific edge case where we might want to do it - a race condition between two processes trying to create an object. And even then, the retry is only a workaround for the fact that Django can't do this properly at the database level with `INSERT ... ON CONFLICT ...` yet. Basically, if we could use the full power of SQL, this case would not need to result in an exception.

Actually, I've just checked and it seems that we might be able to use `INSERT ... ON CONFLICT ...` now:
- [django-postgres-extra > Using the manager > Conflict handling](https://django-postgres-extra.readthedocs.io/manager/) - this library was created in 2017 and seems to support `on_conflict`.
- [Add ON CONFLICT IGNORE support to QuerySet.bulk_create()](https://code.djangoproject.com/ticket/28668) - Django is adding `on_conflict` to `bulk_create()`. Not sure about ordinary `create()`. But we may have this feature in some form when we decide to update to a new release.

So I think something like this is not necessary and not even well adapted to our use case:
- Name does not say we're only retrying on uniqueness violation.
- We don't need complex retry conditions. We only ever want to retry once and we'll never use a timeout.
- This looks like a decorator but we barely ever want to put it on the whole function. It would be much more useful to be able to use it as a context manager to put it only around the operations we want to retry.

concent_api/core/payments/bankster.py
```diff
@@ -254,6 +253,7 @@ def finalize_payment(deposit_claim: DepositClaim) -> Optional[str]:
     return deposit_claim.tx_hash
 
 
+@retry_on_database_error
 def settle_overdue_acceptances(
```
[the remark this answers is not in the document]
pawelkisielewicz: ok... decorator removed

concent_api/concent_api/settings/base.py
```diff
@@ -360,3 +360,7 @@
 
 # Define if exception should be raised when database transactions are nested.
 DETECT_NESTED_TRANSACTIONS = True
+
+MAXIMUM_NUMBER_OF_RETRIES = 5
+
+STARTING_WAIT_TIME_FOR_RETRY_IN_SECONDS = 0.1
```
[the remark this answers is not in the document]
pawelkisielewicz: After removing decorator this two settings are removed as well

concent_api/core/model_helpers.py
```diff
@@ -22,18 +18,3 @@ def get_one_or_none(
         instances = model_or_query_set.filter(**conditions)
         assert len(instances) <= 1
         return None if len(instances) == 0 else instances[0]
-
-
-def get_or_create_with_retry(model: Type[Model], **kwargs: Any) -> Model:
-    try:
-        model_object = model.objects.get_or_create_full_clean(
-            **kwargs,
```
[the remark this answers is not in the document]
pawelkisielewicz: `@atomic` added

concent_api/core/payments/bankster.py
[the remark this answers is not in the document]
pawelkisielewicz: Once if there are two requests at the same time. What if it will be more of them? Just curious, in new implementation it will be no up to date any more :)

concent_api/core/payments/bankster.py
pawelkisielewicz: Added in separate commit

concent_api/common/decorators.py
pawelkisielewicz: ok.. decorator removed.. And thanks for explanation,  i'm interested in this subject so I will read this carefully for sure :)

concent_api/api-e2e-multi-threads-test.py
```diff
@@ -1,34 +1,43 @@
 #!/usr/bin/env python3
+
 import os
 import sys
 import time
 from typing import Optional
-import requests
 
+import requests
```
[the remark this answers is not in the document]
rwrzesien: Requests is 3rd party library.

concent_api/core/payments/bankster.py
```diff
@@ -103,6 +103,13 @@ def claim_deposit(
     # Bankster puts database locks on DepositAccount objects
     # that will be used as payers in newly created DepositClaims.
     with non_nesting_atomic(using='control'):
+        DepositAccount.objects.select_for_update().filter(
+            client=requestor_client,
+            ethereum_address=requestor_ethereum_address,
+        ).filter(
+            client=provider_client if is_claim_against_provider else None,
```
rwrzesien: `client` is defined as `client = ForeignKey(Client)` so it can't be `None`. This way you won't ever match any object. You have to add this filter condition conditionally.

concent_api/api-e2e-multi-threads-test.py
```diff
@@ -1,34 +1,43 @@
 #!/usr/bin/env python3
+
```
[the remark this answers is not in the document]
pawelkisielewicz: It was added automatically during optimizing imports. Removed

concent_api/api-e2e-force-accept-or-reject-results-test.py
```diff
@@ -85,31 +60,6 @@ def create_subtask_results_rejected(
...
-    maxiumum_download_time = calculate_maximum_download_time(
-        size=REPORT_COMPUTED_TASK_SIZE,
-        rate=minimum_upload_rate,
-    )
-    return (
-        (4 * concent_messaging_time) +
-        (3 * maxiumum_download_time)
-    )
-
-
```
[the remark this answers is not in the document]
pawelkisielewicz: I knew it.. done..

concent_api/api-e2e-multi-threads-test.py
```diff
@@ -119,6 +128,32 @@ def send_correct_force_get_task_result(
...
+            timestamp=timestamp_to_isoformat(current_time),
+            ack_report_computed_task=create_ack_report_computed_task(
+                timestamp=timestamp_to_isoformat(current_time),
+                report_computed_task=report_computed_task,
+                requestor_private_key=requestor_private_key
+            )
+        ),
+        expected_status=expected_status,
+    )
+    responses_global.append(response.__class__.__name__)
```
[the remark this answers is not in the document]
pawelkisielewicz: I disagree or don't understand Your intent. I moved functions using in 2 files to avoid code duplication. This is used only in this test file. Do You agree or what and why should be moved?

concent_api/common/decorators.py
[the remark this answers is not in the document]
pawelkisielewicz: In lot of places it is imported like this. But after other changes it is removed anyway

concent_api/common/decorators.py
```diff
@@ -85,3 +89,32 @@ def decorated(*args: Any, **kwargs: Any) -> None:
...
+            try:
+                result = self.func(*args, **kwargs)
+                break
+            except IntegrityError:
+                self._validate_number_of_retries()
+        return result
+
+    def _initialize_retry_parameters(self) -> None:
+        self.wait_time_in_sec = settings.STARTING_WAIT_TIME_FOR_RETRY_IN_SECONDS  # pylint: disable=attribute-defined-outside-init
+        self.current_number_of_retries = 0  # pylint: disable=attribute-defined-outside-init
```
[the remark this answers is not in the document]
pawelkisielewicz: It was to spare "boiler plate code". But now this whole class is removed

concent_api/core/model_helpers.py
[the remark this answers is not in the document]
pawelkisielewicz: Because it is already in transaction. So for this very small part of code I think it should be like this to avoid breaking the outside atomic block.

concent_api/core/model_helpers.py
```diff
@@ -3,7 +3,7 @@
 from typing import Type
 from typing import Union
 
-from django.db import IntegrityError
+from django.db import IntegrityError, transaction
```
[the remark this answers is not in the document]
pawelkisielewicz: yes..

concent_api/core/model_helpers.py
```diff
@@ -26,11 +26,12 @@ def get_one_or_none(
...
-        model_object = model.objects.get_or_create_full_clean(
-            **kwargs,
-        )
+        with transaction.atomic(using='control'):
+            model_object = model.objects.get_or_create_full_clean(
+                **kwargs,
+            )
     except IntegrityError as exception:
-        if exception.pgcode == pg_errorcodes.UNIQUE_VIOLATION:
+        if exception.__cause__.pgcode == pg_errorcodes.UNIQUE_VIOLATION:
```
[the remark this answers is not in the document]
pawelkisielewicz: I have downgraded to 9.6 and with `__cause__` everything is OK

But without:
``` python
 File "/home/pawel/PycharmProjects/concent/concent/concent_api/core/model_helpers.py", line 35, in get_or_create_with_retry
    if exception.pgcode == pg_errorcodes.UNIQUE_VIOLATION:
AttributeError: 'IntegrityError' object has no attribute 'pgcode'
```

So I think it should stay

concent_api/core/model_helpers.py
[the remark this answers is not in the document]
rwrzesien: If it is already in transaction then what is the point of wrapping it in another transaction? @cameel ?

concent_api/core/payments/bankster.py
rwrzesien: It is not a matter of clean code and it will not work correctly. `filter` in Django works as SQL `AND` operator, so if you have ie.
```
In [6]: Client.objects.filter(public_key='ZlXidtPmabypX/uTKl8e7drZcGrwyrEcfsHvxMNc6RAiWWfpJX9/FgPLdEW3HueIqK6gUKgyYuQ9KSnvL4McUg==')
Out[6]: <QuerySet [<Client: Client object>]>
```

This returns object, but when you add 

```
In [7]: Client.objects.filter(public_key='ZlXidtPmabypX/uTKl8e7drZcGrwyrEcfsHvxMNc6RAiWWfpJX9/FgPLdEW3HueIqK6gUKgyYuQ9KSnvL4McUg==').filter(public_key=None)
Out[7]: <QuerySet []>
```

This returns empty result, so even the first matching result will not be locked. So you have split it into two queries, or add second filter conditionally.

cameel: I agree with @rwrzesien. It works like an AND. It's possible to achieve OR with `Q()` expressions but I think that doing two separate queries makes it more apparent that two separate objects may be locked.

Locking them in a single operation does have one advantage though - it makes it harder to cause a deadlock. When two processes (X and Y) need to lock two separate objects (A and B) there's a risk that X will lock A, Y will lock B and then X will be waiting for B while Y will be waiting for A - indefinitely. That's a deadlock. To avoid this you have to always lock objects in the same order - i.e. both X and Y should try to lock A first and only when that succeeds try to lock B. Locking both in a single operation is another solution.

Please look at all the other functions that may lock two objects and make sure that they either try to acquire locks in the same order or lock both in a single operation.

concent_api/core/model_helpers.py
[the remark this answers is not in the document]
cameel: @rwrzesien is right. `get_or_create_with_retry()` won't work correctly (i.e. it won't catch the `IntegrityError`) if you don't commit the transaction. And if the transactions are nested, there will be no commit here. You should use `@non_nesting_atomic` to detect such situations.

If some other function calls `get_or_create_with_retry()` inside its transaction then that function must be fixed.

concent_api/core/payments/bankster.py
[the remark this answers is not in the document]
cameel: I does not depend on the number of requests.

It's like this:
- N requests come at the same time.
- N processes try to create an object at the same time.
- 1 process succeeds. N-1 processes fail.
- Each of the N-1 processes tries to get the object.
    - Get should succeed because the object now exists. If it fails again, it must be for some other, unknown reason and we don't want to silence it and there's no point in retrying.

concent_api/core/payments/bankster.py
pawelkisielewicz: OK, i will do it in two separate queries. My doubt is only that we have two database queries this way and it could be achieved with only one (OR with `Q()` as you wrote). I do it as you wrote, 2 queries and if you would decide to make it in one please let me know.

rwrzesien: Should be OK with `OR` and will be good exercise to get familiar how to construct such query.

cameel: One more query per HTTP request is insignificant. Now worth worrying about. You should start worrying when the number of queries depends on the size of the input or data stored in the database. I.e. when you have a list and have to make a separate query for each item.

If you want to learn more about database performance problems in Django, here are some articles:
- [All you need to know about prefetching in Django](https://hackernoon.com/all-you-need-to-know-about-prefetching-in-django-f9068ebe1e60)
- [Solving Performance Problems in the Django ORM](https://medium.com/@hansonkd/performance-problems-in-the-django-orm-1f62b3d04785)

In this case either way of doing it - one query or two - is fine for me. Each one has its pros and cons. The important thing was the bug pointed out by @rwrzesien (doing an AND operation). Now that it's fixed, the rest is just a matter of preference and readability.

pawelkisielewicz: Thanks, I will read this. I like this part of programming :)

Issue #850, opened 2018-09-25
SCI Callback to Middleman connection retry


PaweuB:
First, investigate what happens when gunicorn terminates the request due to a timeout. I.e. does it kill the Django process or does it let it finish but just terminates the connection and does not send the response to the client. After doing this, please add comment in this issue with the result.

Then add timeout to SCI callback when sending request to Middleman defined as new Django setting.

In case of timeout, the operation should be retried `SCI_CALLBACK_RETRIES` times, where `SCI_CALLBACK_RETRIES` is a Django setting. Setting it to zero means there should be no retry. Setting it to one or more means that whole operation should be retried that amount of times in case of exception. Off course, the retry should also include timeout.

If the timeout is still reached in last retry, raise custom exception.

Please remember to log info that the call is retried.
Please remember that only sending request to Middleman is counted in timeout and only this should be retried.

cameel:
Please don't create "placeholder" issues. We have Pivotal for that. Github is for when the issue needs a longer description and only when that description can already be provided. If the only thing we know is that that a task needs to be done, but we're yet to decide when or how, adding it in Pivotal is all we need.

rwrzesien:
@cameel Should reaching SigningService and increasing its daily threshold info be counted as "irreversible operation"?

rwrzesien:
@cameel 
2.
> If I remember correctly, when we talked about it, we agreed that you'll investigate what happens when gunicorn terminates the request due to a timeout. I.e. does it kill the Django process or does it let it finish but just terminates the connection and does not send the response to the client?

I have added to issue description that this should be investigated first, and then depending on the outcome of this investigation the rest of the issue might be redefined. Is that ok?

3.
> In this particular situation I'm more worried about middleman's internal state or extensions we might add in SS than the counter. Increasing the counter and then not actually making the payment is not a big deal. The limit is not an exact value anyway. But I think communication with SS should be treated as irreversible because it might start doing other things.

But we are not able to determine if request reached SigningService or not. This means that every request sent by Callback should be treated as irreversible, so we should never make retry.

cameel:
> I have added to issue description that this should be investigated first, and then depending on the outcome of this investigation the rest of the issue might be redefined. Is that ok?

I don't think we'll need to redefine it based on the outcome. We should retry either way. If we discover behavior that could break stuff we were not aware of until now, that's a separate concern. Create separate issues for that.

[...]

PR #1071, opened 2019-01-17
Feature sci callback to middleman connection retry

Jakub89:
Resolves: #850

### DIFF
--- a/concent_api/concent_api/settings/base.py
+++ b/concent_api/concent_api/settings/base.py
@@ -360,3 +360,6 @@ GOLEM_MESSAGES_VERSION = __version__
 
 # Define if exception should be raised when database transactions are nested.
 DETECT_NESTED_TRANSACTIONS = True
+
+# Define how long concent wait for middleman response before raising exception.
+SCI_CALLBACK_RETRIES_TIME = 60
--- a/concent_api/concent_api/system_check.py
+++ b/concent_api/concent_api/system_check.py
@@ -14,6 +14,7 @@ from golem_messages import constants
 from common.exceptions import ConcentValidationError
 from concent_api.constants import AVAILABLE_CONCENT_FEATURES
 from core.constants import ETHEREUM_PUBLIC_KEY_LENGTH
+from core.constants import SCI_CALLBACK_MAXIMUM_TIMEOUT
 from core.validation import validate_bytes_public_key
 
 
@@ -445,6 +446,22 @@ def create_error_62_additional_verification_call_time_has_wrong_value() -> Error
     )
 
 
+def create_error_63_sci_callback_retries_is_not_set() -> Error:
+    return Error(
+        "SCI_CALLBACK_RETRIES_TIME is not set.",
+        hint="SCI_CALLBACK_RETRIES_TIME must be set to non-negative integer.",
+        id="concent.E063",
+    )
+
+
+def create_error_64_sci_callback_retries_has_wrong_value() -> Error:
+    return Error(
+        "SCI_CALLBACK_RETRIES_TIME has wrong value.",
+        hint=f"SCI_CALLBACK_RETRIES_TIME must be set to non-negative integer, grater then SCI_CALLBACK_MAXIMUM_TIMEOUT({SCI_CALLBACK_MAXIMUM_TIMEOUT}).",
+        id="concent.E064",
+    )
+
+
 @register()
 def check_settings_concent_features(app_configs: None=None, **kwargs: Any) -> list:  # pylint: disable=unused-argument
 
@@ -812,3 +829,17 @@ def check_additional_verification_call_time(app_configs: None=None, **kwargs: An
     if not isinstance(settings.ADDITIONAL_VERIFICATION_CALL_TIME, int) or settings.ADDITIONAL_VERIFICATION_CALL_TIME < 0:
         return [create_error_62_additional_verification_call_time_has_wrong_value()]
     return errors
+
+
+@register()
+def check_sci_callback_retries(app_configs: None=None, **kwargs: Any) -> list:  # pylint: disable=unused-argument
+    if 'concent-api' in settings.CONCENT_FEATURES:
+        if not hasattr(settings, 'SCI_CALLBACK_RETRIES_TIME'):
+            return [create_error_63_sci_callback_retries_is_not_set()]
+        if (
+            not isinstance(settings.SCI_CALLBACK_RETRIES_TIME, int) or settings.SCI_CALLBACK_RETRIES_TIME < 0 or
+            settings.SCI_CALLBACK_RETRIES_TIME <= SCI_CALLBACK_MAXIMUM_TIMEOUT
+        ):
+            return [create_error_64_sci_callback_retries_has_wrong_value()]
+
+    return []
--- a/concent_api/core/decorators.py
+++ b/concent_api/core/decorators.py
@@ -1,5 +1,6 @@
 from functools import wraps
 from logging import getLogger
+from time import time
 from typing import Any
 from typing import Callable
 from typing import Union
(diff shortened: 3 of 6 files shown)


concent_api/core/decorators.py
```diff
@@ -1,3 +1,4 @@
+import time
```
kbeker: Please change it into `from time import time`

concent_api/core/decorators.py
```diff
@@ -316,3 +319,20 @@ def wrapper(request: HttpRequest, golem_message: message.Message, client_public_
         response_from_view = view(request,  golem_message, client_public_key)
         return response_from_view
     return wrapper
+
+
+def retry_middleman_connection_if_not_pass_timeout(send_request_to_middleman: Callable) -> Callable:
```
kbeker: I can't see any test for this decorator. Please add some

Jakub89: Did You check coverage? :D Most tests from `test_payments_sci_callback.py` go through that decorator, and the `test_that_sci_callback_should_raise_exception_on_timeout()` check every line. I can add more tests of course, but I don't know what else would I check in them?

concent_api/core/decorators.py
Jakub89: Done

concent_api/core/decorators.py
kbeker: Yea, but these tests just go through this decorator and their are not checking the way this function works. I think that you should check it's timeouts etc. :)

Jakub89: `test_that_sci_callback_should_raise_exception_on_timeout()` does. Try it with debuger. That's why I add `SCI_CALLBACK_RETRIES_TIME=1` to `override_settings`. Loop in that decorator will go on for as much seconds as You put in that setting.

concent_api/concent_api/settings/base.py
```diff
@@ -360,3 +360,6 @@
 
 # Define if exception should be raised when database transactions are nested.
 DETECT_NESTED_TRANSACTIONS = True
+
+# Define how long concent wait for middleman response before raising exception.
+SCI_CALLBACK_RETRIES_TIME = 60
```
rwrzesien: Should be `SCI_CALLBACK_RETRY_TIME`, because this is the time of single retry, not many.

concent_api/concent_api/system_check.py
```diff
@@ -812,3 +829,15 @@ def check_additional_verification_call_time(app_configs: None=None, **kwargs: An
...
         return [create_error_62_additional_verification_call_time_has_wrong_value()]
     return errors
+
+
+@register()
+def check_sci_callback_retries(app_configs: None=None, **kwargs: Any) -> list:  # pylint: disable=unused-argument
+    if 'concent-api' in settings.CONCENT_FEATURES:
+        if not hasattr(settings, 'SCI_CALLBACK_RETRIES_TIME'):
+            return [create_error_63_sci_callback_retries_is_not_set()]
+        if not isinstance(settings.SCI_CALLBACK_RETRIES_TIME, int) or settings.SCI_CALLBACK_RETRIES_TIME < 0 or \
```
rwrzesien: Please fix styling in breaking this if into lines.

concent_api/core/decorators.py
```diff
@@ -316,3 +319,20 @@ def wrapper(request: HttpRequest, golem_message: message.Message, client_public_
...
+
+    @wraps(send_request_to_middleman)
+    def wrapper(middleman_message: GolemMessageFrame) -> bytes:
```
rwrzesien: Please provide the mentioned description on why implementation differs from description.

concent_api/core/decorators.py
```diff
@@ -316,3 +319,20 @@ def wrapper(request: HttpRequest, golem_message: message.Message, client_public_
...
+        connection_start = time()
+        retry_counter = 0
+        while time() - connection_start < settings.SCI_CALLBACK_RETRIES_TIME:
+            try:
+                return send_request_to_middleman(middleman_message)
+            except (SCICallbackTimeoutError, SystemExit):
+                log(logger, f"Concent didn't get any response from middleman. Retrying. Retry amount: {retry_counter}.")
+                retry_counter += 1
+        log(logger, 'error: Concent failed to get response from middleman.')
```
rwrzesien: You can add `in X retries`.

concent_api/core/decorators.py
rwrzesien: Yes, some 2-3 dedicated tests would be good.

concent_api/concent_api/settings/base.py
Jakub89: Actually it is time of many retries. It's going to be hard to explain without looking at the code. The way @dybi explain it to me, is that we already have a timeout in `core/constants.py` in form of:
```
# Defines how many seconds should SCI callback wait for response from MiddleMan.
SCI_CALLBACK_MAXIMUM_TIMEOUT = 30
```
And now, depending on how long `SCI_CALLBACK_RETRIES_TIME` will be, that how many times retry will happen, or in other words how many times  `SCI_CALLBACK_MAXIMUM_TIMEOUT` will fit inside `SCI_CALLBACK_RETRIES_TIME`.

rwrzesien: Right, I have checked implementation but forgot to go back to this comment, thanks for explanation.

concent_api/core/decorators.py
kbeker: Oki, I tested it also and you are right. Im giving approvie, but I still think, that it should has it own test :P

Jakub89: Yep, I'm trying to write that test @rwrzesien explained.

Jakub89: @rwrzesien I've added 2 simple tests, check it out :)

concent_api/concent_api/system_check.py
rwrzesien: Could be each condition in separate line, more readable, but this is still OK ;)

concent_api/core/decorators.py
rwrzesien: @Jakub89 This is not exactly what I have meant. You test if exception is being raised and raise it inside function. This does not change the main mechanism of the decorator.

Instead, you should override the settings and constant (using mock) and set it to some low values (2-3 seconds). Write three tests:
1. In first, define the dummy function which does sleep which will exceed `SCI_CALLBACK_RETRIES_TIME`. Check if exception is raised.
2. In seconds, define the dummy function which does sleep which will  **not** exceed `SCI_CALLBACK_RETRIES_TIME` and is just below this value. Return value from it. Check if value is returned.
3. In third, define the dummy function which calls a different function. That different function should make sleep and raise `SCICallbackTimeoutError` in first run, then return value which will be then returned by dummy function. The mentioned sleep does sleep should not will  **not** exceed `SCI_CALLBACK_RETRIES_TIME`. Check if value is returned, even if exception was raised and caught and function was retried.

Jakub89: @rwrzesien settings are override for all tests `SCI_CALLBACK_RETRIES_TIME=1`, and I don't see why would we need to override constant?

Ad1. `test_retry_middleman_connection_returns_error_after_timout()` does exactly that, and there's no need for sleep. Dummy function raise exception all the time until one second pass and then decorator raise the same exception, which I check for.

Ad2. Doesn't `test_retry_middleman_connection_returns_correct_value()` do exactly that? It returns `None` value. And again, what's the point in `sleep`? It will only make the test last longer, with no added value. You can't manipulate that first time check with sleep, because function runs after it.

Ad3. That's how `test_that_sci_callback_should_sign_transaction_after_two_retries()` works, I tried to rewrite it similar to the decorator tests, but run in to some problems. I'll try again if it's necessary.

rwrzesien: Right, I thought it works a little bit different, but after examining the code I think those tests are OK.

Jakub89: tenor

Issue #1044, opened 2018-12-20
Payment report


cameel:
Currently checking if Concent has really paid out the correct amount in all use cases is very tedious and we never do this when testing. It's however crucial if we want to verify that it's working correctly. The goal of this task is to add a custom page in the admin panel that compares Concent's responses (stored in the database) with payments available on the blockchain and reports any discrepancies.

The page should:
1) List all the known pairs of Ethereum accounts where one belongs to a provider and the other to the requestor.
    - The list should be based on known subtasks.
    - The order matters. (A, B) and (B, A) are a different pair.
    - Each client can have multiple accounts. (A, B) and (C, D) are different pairs even if A and C belong to the same provider and B and D belong to the same provider.
2) For each pair list the following:
    - All `SubtaskResultsSettled`, `SubtaskResultsRejected` and `ForcePaymentCommitted` messages stored in the database.
        - If an identical message was sent to both requestor and provider, it's enough to list one of them.
    - All payments between that account pair (including payments from deposit):
        - regular payments
        - settlement payments
        - forced subtask payments
        - additional verification payments
3) Try to match payments with responses:
    - For each `SubtaskResultsSettled` message find forced subtask payments for the same subtask.
    - For each additional verification payment find the corresponding `SubtaskResultsSettled` or `SubtaskResultsRejected`.
    - For each `ForcePaymentCommitted` message find the corresponding settlement payment.
4) Print an error if:
    - You can't match a message with a payment.
        - Exception: if verification cost is zero, there should be no payment.
        - Regular payments do not have to match. They're displayed just as additional information.
    - There's more than one message corresponding to a payment.
    - There's more than one payment corresponding to a message.
    - The value in the message is different than the value of the payment.
        - In case of `SubtaskResultsRejected`: the value of the payment is different than verification cost defined in settings.
    - There's more than one `SubtaskResultsSettled` or `SubtaskResultsRejected` for the same subtask.
    - The value to be paid in `SubtaskResultsSettled` plus the value that will not be paid because deposit is too low is different than the cost of the subtask.
    - `ForcePaymentCommitted.payment_ts` differs from `closure_timestamp` of the settlement payment.
    - There deposit claim corresponding to a forced payment has not been removed from the database.
5) Print a warning if:
    - `ForcePaymentCommitted` or `SubtaskResultsSettled` indicates that the deposit was not high enough to cover the whole cost.
    - You cannot match a message with a payment but there's a `DepositClaim` that indicates that the transaction is pending.
[...]

Issue #960, opened 2018-11-22
get_list_of_payments() does not search through the right range of blocks


PaweuB:
EDIT: Description rewritten by @cameel 

Currently `get_list_of_payments()` does not use the correct range of block when searching for payments.

The range of blocks should be as follows:
1) When searching for regular payments or settlement payments:
    - start block: first block with timestamp >= `payment_ts`. Then filter out payments for which `closure_time` < `payment_ts`.
    - end block: last block - `SCIImplementation.REQUIRED_CONFS`
        - If there are less than `SCIImplementation.REQUIRED_CONFS` block overall, then no block qualify
2) When searching for forced subtask payments or additional verification payments:
    - start block: a few blocks before first block with timestamp >= `payment_ts`. How much exactly is "a few" should be determined by a constant. 10 should be enough in nearly all situations.
    - end block: last block - `SCIImplementation.REQUIRED_CONFS`

Actually, after recent talks with Łukasz on Rocket, turns out that 2. is not necessary at all. We'll never need to search for forced subtask payments or additional verification payments and we need to revert #982. The rules are given just in case it's needed after all but we should really just to remove support for searching for forced subtask payments.

dybi:
@PaweuB I know - just do it

PaweuB:
Pivotal: 
https://www.pivotaltracker.com/story/show/162197293

cameel:
I have updated the description with stuff from my comment above. I've also added the information that this function does not really need to return forced subtask payments.

Below the old description in case someone is confused by the comments here:

> @dybi @cameel 
>
> it needs description
>
>
> EDIT:
> `BlocksHelper(payment_interface).get_first_block_after(payment_ts).number` raises exception when `payment_ts` is after publication of the newest block

dybi:
@cameel, how does this relate to current implementation of `settle_overdue_acceptances()`, where:
```python
        # Concent gets list of forced payments from payment API where T0 <= payment_ts + PAYMENT_DUE_TIME.
        list_of_forced_payments = service.get_list_of_payments(  # pylint: disable=no-value-for-parameter
            requestor_eth_address=requestor_ethereum_address,
            provider_eth_address=provider_ethereum_address,
            payment_ts=oldest_payments_ts,
            transaction_type=TransactionType.FORCE,
        )
```
?

cameel:
- `get_list_of_payments()` currently does not take `REQUIRED_CONFS` into account. It should.
[...]

PR #1073, opened 2019-01-19
Feature payment report

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/1044

### DIFF
--- a/concent_api/core/payments/report.py
+++ b/concent_api/core/payments/report.py
@@ -1,0 +1,766 @@
+from enum import Enum
+from enum import unique
+from typing import Any
+from typing import List
+from typing import NamedTuple
+from typing import Tuple
+
+from django.conf import settings
+from django.db.models import QuerySet
+
+from golem_sci.events import CoverAdditionalVerificationEvent
+from golem_sci.events import ForcedPaymentEvent
+from golem_sci.events import ForcedSubtaskPaymentEvent
+
+from common.constants import ConcentUseCase
+from common.helpers import deserialize_message
+from common.helpers import get_current_utc_timestamp
+from common.helpers import parse_datetime_to_timestamp
+from core.models import Client
+from core.models import DepositClaim
+from core.models import PendingResponse
+from core.models import Subtask
+from core.payments import service
+from core.payments.backends.sci_backend import TransactionType
+
+
+@unique
+class MatchingError(Enum):
+    MESSAGE_NOT_MATCHED_WITH_ANY_PAYMENT = 'message_not_matched_with_any_payment'
+    PAYMENT_NOT_MATCHED_WITH_ANY_MESSAGE = 'payment_not_matched_with_any_message'
+    MORE_THAN_EXPECTED_MATCHING_MESSAGES_FOR_PAYMENT = 'more_than_expected_matching_messages_for_payment'
+    MORE_THAN_ONE_MATCHING_PAYMENT_FOR_MESSAGE = 'more_than_one_matching_payment_for_message'
+    MESSAGE_VALUE_DIFFERS_FROM_PAYMENT_VALUE = 'message_value_differs_from_payment_value'
+    MORE_THAN_ONE_SUBTASK_RESULT_FOR_SUBTASK = 'more_than_one_subtask_result_for_subtask'
+    FORCE_PAYMENT_COMMITED_PAYMENT_TS_DIFFERS_FROM_SETTLEMENT_PAYMENT_CLOSURE_TIMESTAMP = 'force_payment_committed_payment_ts_differs_from_settlement_payment_closure_timestamp'
+    DEPOSIT_CLAIM_RELATED_TO_FORCED_PAYMENT_EXISTS = 'deposit_claim_related_to_forced_payment_exists'
+    DEPOSIT_CLAIM_RELATED_TO_SETTLEMENT_PAYMENT_EXISTS = 'deposit_claim_related_to_settlement_payment_exists'
+
+
+@unique
+class MatchingWarning(Enum):
+    NOT_ENOUGH_DEPOSIT_TO_COVER_WHOLE_COST = 'not_enough_deposit_to_cover_whole_cost'
+    NO_MATCHING_MESSAGE_FOR_PAYMENT_BUT_DEPOSIT_CLAIM_EXISTS = 'no_matching_message_for_payment_but_deposit_claim_exists'
+    PAYMENT_VALUE_DIFFERS_FROM_VERIFICATION_COST = 'payment_value_differs_from_verification_cost'
+    VERIFICATION_PAYMENT_DO_NOT_HAVE_RELATED_SUBTASK = 'verification_payment_do_not_have_related_subtask'
+
+
+PendingResponses = NamedTuple(
+    'PendingResponses',
+    [
+        ('subtask_results_settled', list),
+        ('subtask_results_rejected', list),
+        ('force_payment_committed', list),
+    ]
+)
+
+
+Payments = NamedTuple(
+    'Payments',
+    [
+        ('regular_payments', list),
+        ('settlement_payments', list),
+        ('forced_subtask_payments', list),
+        ('additional_verification_payments_for_requestor', list),
+        ('additional_verification_payments_for_provider', list),
+    ]
+)
+
+
+Summary = NamedTuple(
+    'Summary',
+    [
+        ('number_of_pairs', int),
+        ('number_of_subtasks', int),
+        ('number_of_regular_payments', int),
+        ('number_of_forced_payments', int),
+        ('total_amount_to_be_paid', int),
+        ('total_amount_actually_paid', int),
+        ('total_amount_unrecoverable', int),
+    ]
+)
+
+
+PairReport = NamedTuple(
+    'PairReport',
+    [
+        ('requestor', Client),
+        ('provider', Client),
+        ('pending_responses', PendingResponses),
+        ('payments', Payments),
+        ('errors', list),
+        ('warnings', list),
+        ('summary', Summary),
+    ]
+)
+
+
+Report = NamedTuple(
+    'Report',
+    [
+        ('pair_reports', list),
+        ('summary', Summary),
+    ]
+)
+
+
+Error = NamedTuple(
+    'Error',
+    [
+        ('matching_error', MatchingError),
+        ('message', str),
+    ]
+)
+
+
+Warnings = NamedTuple(
+    'Warnings',
+    [
+        ('matching_warning', MatchingWarning),
+        ('message', str),
+    ]
+)
+
+
+def create_report() -> Report:
+    """ Create payments report for Concent. """
+    pair_reports = []
+
+    for subtask in get_subtasks_with_distinct_clients_pairs():  # pylint: disable=not-an-iterable
+        (subtasks_ids, provider_ethereum_keys, requestor_ethereum_key) = get_subtask_ids_and_ethereum_keys_for_clients_pair(
+            subtask.requestor,
+            subtask.provider,
+        )
+
+        related_pending_responses_for_pair = get_related_pending_responses_for_subtasks_ids(subtasks_ids)
+
+        related_payments_for_pair = get_related_payments_for_pair(provider_ethereum_keys, requestor_ethereum_key)
+
+        (errors, warnings) = get_errors_and_warnings(
+            related_payments_for_pair,
+            related_pending_responses_for_pair,
+        )
+
+        pair_report = PairReport(
+            requestor=subtask.requestor,
+            provider=subtask.provider,
+            pending_responses=related_pending_responses_for_pair,
+            payments=related_payments_for_pair,
+            errors=errors,
+            warnings=warnings,
+            summary=create_pair_summary(
+                requestor=subtask.requestor,
+                provider=subtask.provider,
+                pending_responses=related_pending_responses_for_pair,
+                payments=related_payments_for_pair,
+            ),
+        )
+        pair_reports.append(pair_report)
+
+    return Report(
+        pair_reports,
+        summary=create_report_summary(pair_reports),
+    )
+
+
+def get_subtasks_with_distinct_clients_pairs() -> QuerySet:
+    """ Get Subtasks with distinct pairs of requestor and provider Clients. """
+    return Subtask.objects.order_by('provider', 'requestor').distinct('provider', 'requestor')
+
+
+def get_subtask_ids_and_ethereum_keys_for_clients_pair(requestor: Client, provider: Client) -> Tuple:
+    """ Get Subtasks Ids and all Ethereum keys for Clients pair. """
+    subtask_ids = []
+    provider_ethereum_keys = []
+    requestor_ethereum_keys = []
+
+    for subtask in Subtask.objects.filter(requestor=requestor, provider=provider):  # pylint: disable=no-member
+        subtask_ids.append(subtask.subtask_id)
+        task_to_compute = deserialize_message(subtask.task_to_compute.data.tobytes())
+        provider_ethereum_keys.append(task_to_compute.provider_ethereum_address)
+        requestor_ethereum_keys.append(task_to_compute.requestor_ethereum_address)
+
+    return subtask_ids, set(provider_ethereum_keys), set(requestor_ethereum_keys)
+
+
+def get_related_pending_responses_for_subtasks_ids(subtasks_ids: List[str]) -> PendingResponses:
+    """ Group and return pending responses from Subtask queryset. """
+    return PendingResponses(
+        subtask_results_settled=PendingResponse.objects.filter(
+            response_type=PendingResponse.ResponseType.SubtaskResultsSettled.name,  # pylint: disable=no-member
+            subtask__subtask_id__in=subtasks_ids,
+        ),
+        subtask_results_rejected=PendingResponse.objects.filter(
+            response_type=PendingResponse.ResponseType.SubtaskResultsRejected.name,  # pylint: disable=no-member
+            subtask__subtask_id__in=subtasks_ids,
+        ),
+        force_payment_committed=prefetch_payment_info(
+            PendingResponse.objects.filter(
+                response_type=PendingResponse.ResponseType.ForcePaymentCommitted.name,  # pylint: disable=no-member
+                subtask__subtask_id__in=subtasks_ids,
+            )
+        ),
+    )
+
+
+def prefetch_payment_info(queryset: QuerySet) -> QuerySet:
+    """ Prefetches related PaymentInfo to PendingResponse model queryset. """
+    return queryset.select_related('payment_info')
+
+
+def get_related_payments_for_pair(requestor_ethereum_address: str, provider_ethereum_address: str) -> Payments:
+    """ Get all payments between given requestor/provider pair. """
+    regular_payments = service.get_list_of_payments(  # pylint: disable=no-value-for-parameter
+        requestor_eth_address=requestor_ethereum_address,
+        provider_eth_address=provider_ethereum_address,
+        min_block_timestamp=get_oldest_subtask_creation_date(),
+        transaction_type=TransactionType.BATCH,
+    )
+
+    settlement_payments = service.get_list_of_payments(  # pylint: disable=no-value-for-parameter
+        requestor_eth_address=requestor_ethereum_address,
+        provider_eth_address=provider_ethereum_address,
+        min_block_timestamp=get_oldest_subtask_creation_date(),
+        transaction_type=TransactionType.SETTLEMENT,
+    )
+
+    forced_subtask_payments = service.get_list_of_payments(  # pylint: disable=no-value-for-parameter
+        requestor_eth_address=requestor_ethereum_address,
+        provider_eth_address=provider_ethereum_address,
+        min_block_timestamp=get_oldest_subtask_creation_date(),
+        transaction_type=TransactionType.FORCED_SUBTASK_PAYMENT,
+    )
+
+    additional_verification_payments_for_requestor = service.get_covered_additional_verification_costs(  # pylint: disable=no-value-for-parameter
+        client_eth_address=requestor_ethereum_address,
+        payment_ts=get_oldest_subtask_creation_date(),
+    )
+    additional_verification_payments_for_provider = service.get_covered_additional_verification_costs(  # pylint: disable=no-value-for-parameter
+        client_eth_address=provider_ethereum_address,
+        payment_ts=get_oldest_subtask_creation_date(),
+    )
+
+    return Payments(
+        regular_payments=regular_payments,
+        settlement_payments=settlement_payments,
+        forced_subtask_payments=forced_subtask_payments,
+        additional_verification_payments_for_requestor=additional_verification_payments_for_requestor,
+        additional_verification_payments_for_provider=additional_verification_payments_for_provider,
+    )
+
+
+def find_forced_subtask_payments_for_subtask_results_settled(
+    subtask_results_settled: PendingResponse,
+    forced_subtask_payments: list,
+) -> List[ForcedSubtaskPaymentEvent]:
+    """ Find forced subtask payments for the same subtask for given SubtaskResultsSettled message. """
+    return [
+        forced_subtask_payment for forced_subtask_payment in forced_subtask_payments
+        if forced_subtask_payment.subtask_id == subtask_results_settled.subtask.subtask_id
+    ]
+
+
+def find_subtask_result_for_additional_verification_payment(
+    additional_verification_payment: CoverAdditionalVerificationEvent,
+    subtask_results_settled: List[PendingResponses],
+    subtask_results_rejected: List[PendingResponses],
+) -> List[PendingResponse]:
+    """ Find the SubtaskResultsSettled or SubtaskResultsRejected for given additional verification payment. """
+    matched_subtask_results_settled = [
+        results_settled for results_settled in subtask_results_settled
+        if results_settled.subtask.subtask_id == additional_verification_payment.subtask_id
+    ]
+
+    matched_subtask_results_rejected = [
+        results_rejected for results_rejected in subtask_results_rejected
+        if results_rejected.subtask.subtask_id == additional_verification_payment.subtask_id
+    ]
+
+    return matched_subtask_results_settled + matched_subtask_results_rejected
+
+
+def find_settlement_payment_for_force_payment_committed(
+    force_payment_committed_pending_response: PendingResponse,
+    settlement_payments: list,
+) -> List[ForcedPaymentEvent]:
+    """ Find settlement payment for given ForcePaymentCommitted message. """
+    # assert len(force_payment_committed_pending_response.payment_info) == 1
+    if force_payment_committed_pending_response.payment_info is None:
+        return []
+    return [
+        settlement_payment for settlement_payment in settlement_payments
+        if (
+            int(parse_datetime_to_timestamp(force_payment_committed_pending_response.payment_info.payment_ts)) - 10 <=
+            settlement_payment.closure_time <=
+            int(parse_datetime_to_timestamp(force_payment_committed_pending_response.payment_info.payment_ts)) + 10
+        )
+    ]
+
+
+def get_errors_and_warnings(payments: Payments, responses: PendingResponses) -> Tuple[List[Error], List[Warnings]]:
+    """ Get errors and warnings for messages and related payments. """
+    errors: List[Error] = []
+    warnings: List[Warnings] = []
+
+    for subtask_results_settled_pending_response in responses.subtask_results_settled:
+        matched_forced_subtask_payments_with_subtask_results_settled = find_forced_subtask_payments_for_subtask_results_settled(
+            subtask_results_settled_pending_response,
+            payments.forced_subtask_payments,
+        )
+
+        error_was_added = check_error_cannot_match_message_with_any_payment(
+            errors,
+            matched_forced_subtask_payments_with_subtask_results_settled,
+            subtask_results_settled_pending_response,
+        )
+
+        if error_was_added:
+            check_warning_deposit_claim_exists_for_missing_match(
+                warnings,
+                subtask_results_settled_pending_response.subtask.subtask_id,
+                ConcentUseCase.FORCED_ACCEPTANCE,
+            )
+
+        check_error_more_matching_payment_for_message_than_given_amount(
+            errors,
+            matched_forced_subtask_payments_with_subtask_results_settled,
+            subtask_results_settled_pending_response,
+        )
+
+    for additional_verification_payment in payments.additional_verification_payments_for_requestor:
+        matched_subtask_result_with_additional_verification_payment_for_requestor = find_subtask_result_for_additional_verification_payment(
+            additional_verification_payment,
+            responses.subtask_results_settled,
+            responses.subtask_results_rejected,
+        )
+
+        check_additional_verification_payment_errors_and_warning(
+            errors,
+            warnings,
+            matched_subtask_result_with_additional_verification_payment_for_requestor,
+            additional_verification_payment,
+        )
+
+    for additional_verification_payment in payments.additional_verification_payments_for_provider:
+        matched_subtask_result_with_additional_verification_payment_for_provider = find_subtask_result_for_additional_verification_payment(
+            additional_verification_payment,
+            responses.subtask_results_settled,
+            responses.subtask_results_rejected,
+        )
+
+        check_additional_verification_payment_errors_and_warning(
+            errors,
+            warnings,
+            matched_subtask_result_with_additional_verification_payment_for_provider,
+            additional_verification_payment,
+        )
+
+    for force_payment_committed_pending_response in responses.force_payment_committed:
+        matched_settlement_payment_for_force_payment_committed = find_settlement_payment_for_force_payment_committed(
+            force_payment_committed_pending_response,
+            payments.settlement_payments,
+        )
+
+        error_was_added = check_error_cannot_match_message_with_any_payment(
+            errors,
+            matched_settlement_payment_for_force_payment_committed,
+            force_payment_committed_pending_response,
+        )
+
+        if error_was_added:
+            check_warning_deposit_claim_exists_for_missing_match(
+                warnings,
+                force_payment_committed_pending_response.subtask.subtask_id,
+                ConcentUseCase.FORCED_PAYMENT,
+            )
+
+        check_error_more_matching_payment_for_message_than_given_amount(
+            errors,
+            matched_settlement_payment_for_force_payment_committed,
+            force_payment_committed_pending_response,
+        )
+        check_error_message_value_is_different_than_payment_value(
+            errors,
+            matched_settlement_payment_for_force_payment_committed,
+            force_payment_committed_pending_response,
+        )
+        check_error_message_payment_ts_is_different_than_payment_closure_time(
+            errors,
+            matched_settlement_payment_for_force_payment_committed,
+            force_payment_committed_pending_response,
+        )
+        check_warning_force_payment_committed_indicates_that_deposit_was_too_low_to_cover_whole_cost(
+            warnings,
+            force_payment_committed_pending_response,
+        )
+
+    for forced_subtask_payment in payments.forced_subtask_payments:
+        check_error_deposit_claim_exists_for_forced_payment(
+            errors,
+            forced_subtask_payment.tx_hash,
+        )
+
+    for settlement_payment in payments.settlement_payments:
+        check_error_deposit_claim_exists_for_settlement_payment(
+            errors,
+            settlement_payment.tx_hash,
+        )
+
+    return errors, warnings
+
+
+def check_additional_verification_payment_errors_and_warning(
+    errors: List[Error],
+    warnings: List[Warnings],
+    matched_objects: List[PendingResponse],
+    additional_verification_payment: CoverAdditionalVerificationEvent,
+) -> None:
+    """ Check errors and warning related to CoverAdditionalVerificationEvent object. """
+    error_was_added = check_error_cannot_match_payment_with_any_message(
+        errors,
+        matched_objects,
+        additional_verification_payment
+    )
+
+    if error_was_added:
+        check_warning_deposit_claim_exists_for_missing_match(
+            warnings,
+            additional_verification_payment.subtask_id,
+            ConcentUseCase.ADDITIONAL_VERIFICATION,
+        )
+
+    check_error_more_matching_message_for_payment_than_given_amount(
+        errors,
+        matched_objects,
+        additional_verification_payment,
+        amount=2,
+    )
+    check_error_more_than_one_subtask_result_of_the_same_type_for_subtask(
+        errors,
+        matched_objects,
+        additional_verification_payment,
+    )
+    check_warning_no_subtask_related_with_additional_verification_payment(
+        warnings,
+        additional_verification_payment.subtask_id,
+    )
+    check_warning_payment_value_differs_from_verification_cost_in_settings(
+        warnings,
+        matched_objects,
+        additional_verification_payment,
+    )
+
+
+def check_error_cannot_match_message_with_any_payment(
+    errors: List[Error],
+    matched_objects: List,
+    pending_response: PendingResponse,
+) -> bool:
+    """ Add error if message cannot be matched with any payment."""
+    if len(matched_objects) == 0:
+        errors.append(
+            Error(
+                matching_error=MatchingError.MESSAGE_NOT_MATCHED_WITH_ANY_PAYMENT,
+                message=f'PendingResponse of type {pending_response.response_type} and pk {pending_response.pk} does not have any matching payment.',
+            )
+        )
+        return True
+    return False
+
+
+def check_error_cannot_match_payment_with_any_message(
+    errors: List[Error],
+    matched_objects: List,
+    payment: Any,
+) -> bool:
+    """ Add error if payment cannot be matched with any message."""
+    if len(matched_objects) == 0:
+        errors.append(
+            Error(
+                matching_error=MatchingError.PAYMENT_NOT_MATCHED_WITH_ANY_MESSAGE,
+                message=f'Payment of type {type(payment)} and tx_hash {payment.tx_hash} does not have any matching messages.',
+            )
+        )
+        return True
+    return False
+
+
+def check_error_more_matching_message_for_payment_than_given_amount(
+    errors: List[Error],
+    matched_objects: List,
+    payment: Any,
+    amount: int = 1,
+) -> None:
+    """ Add error if there are more messages matched with a payment than given amount. """
+    if len(matched_objects) > amount:
+        errors.append(
+            Error(
+                matching_error=MatchingError.MORE_THAN_EXPECTED_MATCHING_MESSAGES_FOR_PAYMENT,
+                message=f'Payment of type {type(payment)} and tx_hash {payment.tx_hash} has {len(matched_objects)} matching messages.',
+            )
+        )
+
+
+def check_error_more_matching_payment_for_message_than_given_amount(
+    errors: List[Error],
+    matched_objects: List,
+    pending_response: PendingResponse,
+    amount: int = 1,
+) -> None:
+    """ Add error if there are more payments matched with a message than given amount. """
+    if len(matched_objects) > amount:
+        errors.append(
+            Error(
+                matching_error=MatchingError.MORE_THAN_ONE_MATCHING_PAYMENT_FOR_MESSAGE,
+                message=f'PendingResponse of type {pending_response.response_type} and pk {pending_response.pk} has {len(matched_objects)} matching payments.',
+            )
+        )
+
+
+def check_error_message_value_is_different_than_payment_value(
+    errors: List[Error],
+    matched_objects: List,
+    force_payment_committed_pending_response: PendingResponse,
+) -> None:
+    """
+    Add error if PaymentInfo.amount_paid related to stored PendingResponse of type ForcePaymentCommitted
+    differs from the settlement_payment.amount value.
+    """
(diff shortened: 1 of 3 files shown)


concent_api/core/tests/test_payments_report.py
```diff
@@ -0,0 +1,786 @@
...
+                task_to_compute=task_to_compute,
+                report_computed_task=report_computed_task,
+            )
+
+            self.subtasks.append(subtask)
+
+        assert Client.objects.count() == self.number_of_clients_per_type * 2
+
+    def _prepare_related_payments_mocks(self):
+        for subtask_pair in get_client_pairs_from_subtasks():
```
Jakub89:
```
E:189,28: Non-iterable value get_client_pairs_from_subtasks() is used in an iterating context (not-an-iterable)
```

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,759 @@
+from datetime import datetime
```
Jakub89:
```
W:  1, 0: Unused datetime imported from datetime (unused-import)
```

concent_api/core/payments/backends/sci_backend.py
```diff
@@ -145,6 +153,21 @@ def call_on_confirmed_transaction(
     )
 
 
+def get_covered_additional_verification_costs(client_eth_address: str, payment_ts: int) -> list:
+    assert isinstance(client_eth_address, str) and len(client_eth_address) == ETHEREUM_ADDRESS_LENGTH
```
pawelkisielewicz: I saw in couple of Your pull requests such checks. Consider create function e.g. `validate_ethereum_address` and use this instead of this checks. There are different approaches to asserts, for me some of them are OK, but you could extract low level logic to separate, even one line function and use it instead. I don't say that is must be done now, please consider and decide.

Advantage of this is as well, when something will change (e.g. `payment_ts >= 0 -> payment_ts >= ??` it will be necessary to change it in only one place

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,759 @@
...
+            forced_subtask_payment.tx_hash,
+        )
+
+    for settlement_payment in payments.settlement_payments:
+        check_error_deposit_claim_exists_for_settlement_payment(
+            errors,
+            settlement_payment.tx_hash,
+        )
+
+    return errors, warnings
```
kbeker: I would change into `NamedTuple`

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,759 @@
...
+        )
+
+        if error_was_added:
+            check_warning_deposit_claim_exists_for_missing_match(
+                warnings,
+                subtask_results_settled_pending_response.subtask.subtask_id,
+                ConcentUseCase.FORCED_ACCEPTANCE,
+            )
+
+        check_error_more_matching_payment_for_message_than_given_amount(
```
kbeker: I am not a fan of changing parameter inside function and not returning it, but maybe here you had nire unoirtabt goal to achieve :)

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,759 @@
...
+        settlement_payment for settlement_payment in settlement_payments
+        if (
+            int(parse_datetime_to_timestamp(payment_info.payment_ts)) - 10 <=
+            settlement_payment.closure_time <=
+            int(parse_datetime_to_timestamp(payment_info.payment_ts)) + 10
+        )
+    ]
+
+
+def get_errors_and_warnings(payments: Payments, responses: PendingResponses) -> Tuple[List[Error], List[Warnings]]:
```
kbeker: I think that it might be good if you would divide this function into few small ones. Then names of smaller functions would tell everything what is going inside :)

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,759 @@
...
+def check_error_more_than_one_subtask_result_of_the_same_type_for_subtask(
+    errors: List[Error],
+    matched_objects: List,
+    additional_verification_payment: CoverAdditionalVerificationEvent,
+) -> None:
+    """ Add error if there is more than two SubtaskResultsSettled or SubtaskResultsRejected for the same subtask. """
+    if (
+        any(
+            [subtask_result.response_type == PendingResponse.ResponseType.SubtaskResultsSettled.name  # pylint: disable=no-member
+             for subtask_result in matched_objects]
```
kbeker: Here you created `[]` a little bit different then in our coding style. You should move comparison and for loop to new line

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,759 @@
...
+def find_settlement_payment_for_force_payment_committed(
+    force_payment_committed_pending_response: PendingResponse,
+    settlement_payments: list,
+) -> List[ForcedPaymentEvent]:
+    """ Find settlement payment for given ForcePaymentCommitted message. """
+    payment_info = force_payment_committed_pending_response.payments.order_by('id').last()
+    return [
+        settlement_payment for settlement_payment in settlement_payments
+        if (
+            int(parse_datetime_to_timestamp(payment_info.payment_ts)) - 10 <=
```
pawelkisielewicz: `payment_ts` is parsed from datetime to timestamp? It should be timestamp  without parsing.

concent_api/core/payments/report.py
pawelkisielewicz: There is very, very lot of low level logic in this function and it is for me really hard to see what is actually happening here. Consider extracting each `for` to separate inner, helper function responsible for low level logic and the names of this helpers will explain high level, business logic.

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,759 @@
...
+
+
+def check_additional_verification_payment_errors_and_warning(
+    errors: List[Error],
+    warnings: List[Warnings],
+    matched_objects: List[PendingResponse],
+    additional_verification_payment: CoverAdditionalVerificationEvent,
+) -> None:
+    """ Check errors and warning related to CoverAdditionalVerificationEvent object. """
+    error_was_added = check_error_cannot_match_payment_with_any_message(
```
pawelkisielewicz: :+1: for constructing- see above. Each function has name explaining what is going on and low level logic is hidden

concent_api/core/tests/test_payments_sci_backend.py
```diff
@@ -135,6 +135,45 @@ def test_that_sci_backend_get_list_of_payments_should_return_list_of_batch_trans
...
+            payee_address=Web3.toChecksumAddress(self.task_to_compute.provider_ethereum_address),
+            from_block=self.last_block,
+            to_block=self.block_number,
+        )
+
+        new_sci_rpc.return_value.get_block_number.assert_called()
+
+        get_first_block_after.assert_called_with(
+            self.current_time - 1
+        )
```
kbeker: It should be inlined

concent_api/core/tests/test_payments_sci_backend.py
```diff
@@ -280,3 +319,38 @@ def test_that_sci_backend_cover_additional_verification_cost_should_return_trans
...
+            address=Web3.toChecksumAddress(self.task_to_compute.provider_ethereum_address),
+            from_block=self.last_block,
+            to_block=self.block_number,
+        )
+
+        new_sci_rpc.return_value.get_block_number.assert_called()
+
+        get_first_block_after.assert_called_with(
+            self.current_time
+        )
```
kbeker: Also here

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,759 @@
...
+    if len(matched_objects) > amount:
+        errors.append(
+            Error(
+                matching_error=MatchingError.MORE_THAN_ONE_MATCHING_PAYMENT_FOR_MESSAGE,
+                message=f'PendingResponse of type {pending_response.response_type} and pk {pending_response.pk} has {len(matched_objects)} matching payments.',
+            )
+        )
+
+
+def check_error_message_value_is_different_than_payment_value(
```
pawelkisielewicz: This function is typical inner helper for other function in this file. In java it should be 100% private function, how is it python? Shouldn't the name start with `_` to mark this as private, inner helper instead of public api?

concent_api/core/tests/test_payments_report.py
```diff
@@ -0,0 +1,786 @@
...
+        self._prepare_requestors()
+        self._prepare_subtasks()
+        self._prepare_related_payments_mocks()
+        self._prepare_related_pending_responses()
+
+        assert self.default_payment_amount > 0
+        assert self.number_of_clients_per_type > 1
+        assert self.number_of_subtasks > 0
+
+    def test_create_report(self):
```
kbeker: Please move this test after all private function where are all preparation for tests.

concent_api/core/tests/test_payments_report.py
```diff
@@ -0,0 +1,786 @@
...
+            )
+            client.full_clean()
+            client.save()
+            self.requestors.append(client)
+
+    def _prepare_subtasks(self):
+        for i in range(self.number_of_subtasks):
+            task_to_compute = self._get_deserialized_task_to_compute(
+                provider_public_key=self.providers[i // 3].public_key,
+                requestor_public_key=self.requestors[(i + 1) // 3].public_key if i != (self.number_of_subtasks - 1) else self.requestors[0].public_key,
```
kbeker: Please split it to more lines.

concent_api/core/tests/test_payments_report.py
```diff
@@ -0,0 +1,786 @@
...
+            )
+            report_computed_task = self._get_deserialized_report_computed_task(
+                task_to_compute=task_to_compute
+            )
```
kbeker: This actually could be in one line

concent_api/core/tests/test_payments_report.py
```diff
@@ -0,0 +1,786 @@
...
+
+            self.subtasks.append(subtask)
+
+        assert Client.objects.count() == self.number_of_clients_per_type * 2
+
+    def _prepare_related_payments_mocks(self):
+        for subtask_pair in get_client_pairs_from_subtasks():
+            # The SCI methods are called in the same order in `core.payments.report.get_related_payments_for_pair`.
+            self.get_list_of_payments_mocks.append(
+                [self._create_batch_payment_object(
```
kbeker: Please move opening bracket for a list to the line above

concent_api/core/tests/test_payments_report.py
```diff
@@ -0,0 +1,786 @@
...
+                    self.default_payment_amount,
+                    self.closure_time,
+                )]
+            )
+            self.get_list_of_payments_mocks.append(
+                [self._create_settlement_payment_object(
```
kbeker: Same here

concent_api/core/tests/test_payments_report.py
```diff
@@ -0,0 +1,786 @@
...
+                    self.default_payment_amount,
+                    self.closure_time,
+                )]
+            )
+            self.get_list_of_payments_mocks.append(
+                [self._create_forced_subtask_payment_object(self.default_payment_amount, subtask.subtask_id)
+                 for subtask in Subtask.objects.filter(requestor=subtask_pair.requestor, provider=subtask_pair.provider)]
```
kbeker: I'm curious which way would be best:
- divide it and make it statement more sloppy
- leave as it is, but it does not fit in one screen e.g. in github

concent_api/core/tests/test_payments_report.py
```diff
@@ -0,0 +1,786 @@
...
+)
+class GetErrorsAndWarningsPaymentsReportTest(ConcentIntegrationTestCase):
+
+    def setUp(self):
+        super().setUp()
+
+        task_to_compute = self._get_deserialized_task_to_compute()
+        report_computed_task = self._get_deserialized_report_computed_task(
+            task_to_compute=task_to_compute
+        )
```
kbeker: As some comment above. Please make it inline statement

concent_api/core/payments/report.py
Jakub89: Had what goal? "nire unoirtabt"

concent_api/core/payments/backends/sci_backend.py
cameel: `validate_ethereum_address()` suggests that it would be a function that checks the condition and raises an exception. For assertions we need simple boolean conditions instead.

Also these are two separate checks. I would do them with separate assertions:
``` python
assert isinstance(client_eth_address, str)
assert len(client_eth_address) == ETHEREUM_ADDRESS_LENGTH
```

And these checks are so simple that wrapping them in functions could only make things less clear.

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,759 @@
...
+from core.models import PendingResponse
+from core.models import Subtask
+from core.payments import service
+from core.payments.backends.sci_backend import TransactionType
+
+
+@unique
+class MatchingError(Enum):
+    MESSAGE_NOT_MATCHED_WITH_ANY_PAYMENT = \
+        'message_not_matched_with_any_payment'
```
cameel: We don't have a line length limit. I'd suggest not breaking lines like this.

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,759 @@
...
+    concent_use_case: ConcentUseCase,
+) -> None:
+    """
+    Add warning if message cannot be matched with a payment
+    but there's a DepositClaim that indicates that the transaction is pending.
+    """
+    if DepositClaim.objects.filter(
+        subtask_id=subtask_id,
+        concent_use_case=concent_use_case,
+    ).exists():
```
cameel: If `tx_hash` is `None`, this could be an error because in that situation it's certain that something failed - Concent tried to make a payment and even claimed the funds but did not finalize it.

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,759 @@
...
+                matching_error=MatchingError.DEPOSIT_CLAIM_RELATED_TO_FORCED_PAYMENT_EXISTS,
+                message=f'DepositClaim still exists for transaction with tx_hash {tx_hash}',
+            )
+        )
+
+
+def check_warning_deposit_claim_exists_for_missing_match(
+    warnings: List[Warnings],
+    subtask_id: str,
+    concent_use_case: ConcentUseCase,
```
cameel: This functions seems incomplete to me. It does not really do what the name says - it just checks if the subtask has any claims. There's nothing about "missing match" inside it.

Either the name is misleading, the implementation is incomplete or the function is only meant to be used when some other conditions are met and is missing `assert`s that would ensure that these conditions are really true.

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,759 @@
...
+            Error(
+                matching_error=MatchingError.DEPOSIT_CLAIM_RELATED_TO_SETTLEMENT_PAYMENT_EXISTS,
+                message=f'DepositClaim still exists for transaction with tx_hash {tx_hash}',
+            )
+        )
+
+
+def check_error_deposit_claim_exists_for_forced_payment(errors: List[Error], tx_hash: str) -> None:
+    """ Add error if there is DepositClaim related with forced payment. """
+    if DepositClaim.objects.filter(tx_hash=tx_hash).exists():
```
cameel: This function and the function above check the same thing and only have different messages. Just like in my previous comment, this seems incomplete. The check is not self-contained so it's hard to even say if your check here is right or wrong. It sometimes is and sometimes is not, depending on how you use the function. It makes it harder to understand and review this.

concent_api/core/payments/report.py
cameel: I'd suggest to putting the result of `any()` in a variable with a good name. It will make the `if` more concise and easier to understand what you're checking.

Also, `[]` is not really necessary. `any()` can work with a generator expression too.

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,759 @@
...
+                )
+            )
+
+
+def check_error_more_than_one_subtask_result_of_the_same_type_for_subtask(
+    errors: List[Error],
+    matched_objects: List,
+    additional_verification_payment: CoverAdditionalVerificationEvent,
+) -> None:
+    """ Add error if there is more than two SubtaskResultsSettled or SubtaskResultsRejected for the same subtask. """
```
cameel: Your `if` is not checking for more than two `SubtaskResultsSettled` or `SubtaskResultsRejected`. Only for more than one.

Maybe you mean `if there exist both SubtaskResultsSettled and SubtaskResultsRejected for the same subtask`?

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,759 @@
...
+    matched_objects: List,
+    force_payment_committed_pending_response: PendingResponse,
+) -> None:
+    """
+    Add error if PaymentInfo.payment_ts related stored PendingResponse of type ForcePaymentCommitted
+    differs from closure_time of the settlement payment.
+    """
+    payment_info = force_payment_committed_pending_response.payments.order_by('id').last()
+    for settlement_payment in matched_objects:
+        if settlement_payment.closure_time != parse_datetime_to_timestamp(payment_info.payment_ts):
```
cameel: I think you should assert that `parse_datetime_to_timestamp(payment_info.payment_ts)` does not have fractional part. In some places you convert timestamps to `int`, in others you don't. If they do have fractional parts, the results will be inconsistent.

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,759 @@
...
+            errors.append(
+                Error(
+                    matching_error=MatchingError.FORCE_PAYMENT_COMMITED_PAYMENT_TS_DIFFERS_FROM_SETTLEMENT_PAYMENT_CLOSURE_TIMESTAMP,
+                    message=f'PaymentInfo with pk {payment_info.pk} payment ts value {payment_info.payment_ts} '
```
cameel: Either say `payment_ts` (if you want to refer to the field in the code or `payment timestamp` (if you want to refer to what the field means). `payment ts` is a weird hybrid.

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,759 @@
...
+                            f'related to stored PendingResponse of type ForcePaymentCommitted '
+                            f'with pk {force_payment_committed_pending_response.pk} '
+                            f'differs from the settlement payment with tx_hash {settlement_payment.tx_hash} '
+                            f'closure time value {settlement_payment.closure_time}.',
```
cameel: Don't you need a comma here? `with closure time` or `, closure time`. I think this text might be hard to understand otherwise.

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,759 @@
...
+    for settlement_payment in matched_objects:
+        if settlement_payment.amount != payment_info.amount_paid:
+            errors.append(
+                Error(
+                    matching_error=MatchingError.MESSAGE_VALUE_DIFFERS_FROM_PAYMENT_VALUE,
+                    message=f'PaymentInfo with pk {payment_info.pk} amount_paid value {payment_info.amount_paid} '
+                            f'related to stored PendingResponse of type ForcePaymentCommitted '
+                            f'with pk {force_payment_committed_pending_response.pk} '
+                            f'differs from the settlement payment with tx_hash {settlement_payment.tx_hash} '
+                            f'amount value {settlement_payment.amount}.',
```
cameel: `amount value`? Sounds really weird. Why not `amount`? Or `"amount" value`, though I think that `value` in these messages is redundant. They're a little to verbose overall.

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,759 @@
...
+    if len(matched_objects) > amount:
+        errors.append(
+            Error(
+                matching_error=MatchingError.MORE_THAN_EXPECTED_MATCHING_MESSAGES_FOR_PAYMENT,
+                message=f'Payment of type {type(payment)} and tx_hash {payment.tx_hash} has {len(matched_objects)} matching messages.',
+            )
+        )
+
+
+def check_error_more_matching_payment_for_message_than_given_amount(
```
cameel: `amount` -> `number`
"amount" is only for uncountable things.

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,759 @@
...
+    """ Add error if payment cannot be matched with any message."""
+    if len(matched_objects) == 0:
+        errors.append(
+            Error(
+                matching_error=MatchingError.PAYMENT_NOT_MATCHED_WITH_ANY_MESSAGE,
+                message=f'Payment of type {type(payment)} and tx_hash {payment.tx_hash} does not have any matching messages.',
+            )
+        )
+        return True
+    return False
```
cameel: Why some of these `check_` functions return booleans and others don't?

I think `check_` is not a good name because it's vague and does not really tell you what the function does. Does it return a value? Raise an exception? Just check something and keep quiet about it?

If I were designing it, I'd probably make those functions return the error and warning objects instead of giving it a list that it's supposed to modify.

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,759 @@
...
+    error_was_added = check_error_cannot_match_payment_with_any_message(
+        errors,
+        matched_objects,
+        additional_verification_payment
+    )
+
+    if error_was_added:
+        check_warning_deposit_claim_exists_for_missing_match(
+            warnings,
+            additional_verification_payment.subtask_id,
```
cameel: I think `error_was_added` should be checked inside the function. And of course it should not be checking if there was an error but the condition that caused the error. Otherwise the check is incomplete.

concent_api/core/payments/report.py
cameel: I agree. I think the function should return the errors/warning objects.

concent_api/core/tests/test_payments_report.py
rwrzesien: :+1: done

concent_api/core/payments/report.py
rwrzesien: :+1: done

concent_api/core/payments/backends/sci_backend.py
```diff
@@ -56,6 +57,13 @@ def get_list_of_payments(
             from_block=first_block_after_payment_number,
             to_block=latest_block_number - payment_interface.REQUIRED_CONFS,  # pylint: disable=no-member
         )
+    elif transaction_type == TransactionType.FORCE:
+        payments_list = payment_interface.get_forced_subtask_payments(  # pylint: disable=no-member
+            payer_address=Web3.toChecksumAddress(requestor_eth_address),
+            payee_address=Web3.toChecksumAddress(provider_eth_address),
+            from_block=first_block_after_payment_number,
+            to_block=payment_interface.get_block_number(),  # pylint: disable=no-member
```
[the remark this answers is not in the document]
rwrzesien: :+1: done

concent_api/core/payments/backends/sci_backend.py
```diff
@@ -56,6 +57,13 @@ def get_list_of_payments(
...
+        )
```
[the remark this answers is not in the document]
rwrzesien: I think `assert False` should be good, we don't use this code outside Concent, and `transaction_type` is never coming from outside. :+1: done

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,759 @@
...
+        ('message', str),
+    ]
+)
+
+
+def create_report() -> Report:
+    """ Create payments report for Concent. """
+    pair_reports = []
+
+    for subtask_pair in get_client_pairs_from_subtasks():  # pylint: disable=not-an-iterable
```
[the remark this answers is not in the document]
rwrzesien: :+1: done

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,759 @@
...
+        )
+
+        task_to_compute = deserialize_message(subtask_pair.task_to_compute.data.tobytes())
+
+        related_payments_for_pair = get_related_payments_for_pair(
+            task_to_compute.requestor_ethereum_address,
+            task_to_compute.provider_ethereum_address,
+        )
+
+        (errors, warnings) = get_errors_and_warnings(
```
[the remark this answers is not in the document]
rwrzesien: At first I have it this way, but then during the implementation of the logic it only gave unnecessary complexity. It is better to keep them in a flat structure.

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,759 @@
...
+    )
+
+
+def get_client_pairs_from_subtasks() -> QuerySet:
+    """
+    Get all unique pairs of Ethereum accounts where one belongs to a provider and the other to the requestor.
+    """
+    return Subtask.objects.distinct(
+        'provider', 'requestor'
+    )
```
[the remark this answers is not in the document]
rwrzesien: :+1: done

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,759 @@
...
+def get_related_payments_for_pair(requestor_ethereum_address: str, provider_ethereum_address: str) -> Payments:
+    """ Get all payments between given requestor/provider pair. """
+    regular_payments = service.get_list_of_payments(  # pylint: disable=no-value-for-parameter
+        requestor_eth_address=requestor_ethereum_address,
+        provider_eth_address=provider_ethereum_address,
+        min_block_timestamp=get_oldest_subtask_creation_date(),
+        transaction_type=TransactionType.BATCH,
+    )
+
+    # Settlement payments.
```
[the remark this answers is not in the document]
rwrzesien: :+1: done

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,759 @@
...
+    settlement_payments = service.get_list_of_payments(  # pylint: disable=no-value-for-parameter
+        requestor_eth_address=requestor_ethereum_address,
+        provider_eth_address=provider_ethereum_address,
+        min_block_timestamp=get_oldest_subtask_creation_date(),
+        transaction_type=TransactionType.SETTLEMENT,
+    )
+
+    # Forced subtask payments.
```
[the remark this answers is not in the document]
rwrzesien: :+1: done

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,759 @@
...
+    forced_subtask_payments = service.get_list_of_payments(  # pylint: disable=no-value-for-parameter
+        requestor_eth_address=requestor_ethereum_address,
+        provider_eth_address=provider_ethereum_address,
+        min_block_timestamp=get_oldest_subtask_creation_date(),
+        transaction_type=TransactionType.FORCE,
+    )
+
+    # Additional verification payments.
```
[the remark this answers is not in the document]
rwrzesien: :+1: done

concent_api/core/payments/report.py
rwrzesien: Answered in https://github.com/golemfactory/concent/pull/1073#discussion_r250418559

concent_api/core/payments/report.py
rwrzesien: How would like to make the call then? All functions are adding the error conditionally.

```
error = check_x()
if error:
    errors.append(error)
```

This is much less readable.

```
errors.append(check_x())
```

This results in error looking like `[None]`.

concent_api/core/payments/report.py
rwrzesien: I think it already is. I think that another level of abstraction would complicate it. Now you have a loop for every list that does two things:
1. Match the related object.
2. Check for errors and warnings.

All what is done in this function are points 3,4 and 5 from issue description.

But I am open for discussion on this, so if you have some proposal how to split without overcomplicating, lets discuss.

concent_api/core/tests/test_payments_report.py
rwrzesien: Please provide example of both, it would be easier to chose.

concent_api/core/payments/backends/sci_backend.py
rwrzesien: > Advantage of this is as well, when something will change (e.g. payment_ts >= 0 -> payment_ts >= ?? it will be necessary to change it in only one place

This can be also partially achieved by using constants.

concent_api/core/payments/report.py
rwrzesien: Answered in https://github.com/golemfactory/concent/pull/1073#discussion_r250423558, please keep this discussion in single tread.

concent_api/core/payments/report.py
rwrzesien: The reason for that is that it is used twice.

concent_api/core/payments/report.py
rwrzesien: Yes, exactly as @cameel wrote, it is a database column which casted to `datetime` object.

concent_api/core/payments/report.py
rwrzesien: In python this is just a convention. You can call/import private and protected function however you like. Some tools (I think Pycharm does it) might complain about it but its only that. I think making things private or public makes sense in object oriented approach - which is AFAIK how Java works - so there are actually no functions in Java. And when implementing a class in Python I would follow this convention. But in functional programming I can't see a logical division into public/protected/private functions, as any of those can be reused outside of current module.

rwrzesien:
@cameel 

> You're making database queries as if they were just Python calls. You should always minimize the number of queries. And I mean mostly the obvious stuff - i.e. avoiding n queries to do something that you can do with a constant number of them.
> In this particular feature the performance impact of those queries will probably be dwarfed by the blockchain calls - though you can never be sure until you profile. You're doing tons of queries and simply asking Geth for information might actually turn out to be pretty fast. There are also possible race conditions because you're not getting your information from a consistent snapshot. Both of these are acceptable for now since this is just an initial version of a development tool - but I think we should create an issue for fixing it. Please be more careful about queries in future pull requests.

Time was the factor here. The code of this report was constantly evolving so I have decided not to pay any attention to this in the first version and focus on delivering functionality. Then, when the functionality is accepted/modified, focus on optimization. 
My very first idea was to create a full report with single SQL query, until I have realized it can't be done because of SCI calls.

> Also, I did not manage to fully review the errors and warnings. I've seen all the check_ functions but then I realized that they're pretty useless on their own and part of their logic is hidden in the function that calls them. It makes it much more time consuming to analyze. In the parent function you only see the context and in the functions you only have a partial condition. If you're splitting it into functions, these functions should be more self-contained.

I think this is only for single case (which happens in three different loops), but it might give that impression as it is first case of this function. But if it is something else, please let me know.

concent_api/core/payments/report.py
pawelkisielewicz: When I wrote this comment of @kbeker was pending. I haven't seen it, this is the reason of duplication. But i still think it would be better to split... Please continue in  https://github.com/golemfactory/concent/pull/1073#discussion_r250423558

concent_api/core/tests/test_payments_report.py
```diff
@@ -0,0 +1,789 @@
...
+        self._prepare_requestors()
+        self._prepare_subtasks()
+        self._prepare_related_payments_mocks()
+        self._prepare_related_pending_responses()
+
+        assert self.default_payment_amount > 0
+        assert self.number_of_clients_per_type > 1
+        assert self.number_of_subtasks > 0
+
+    def test_create_report(self):
```
dybi: this test is very complex and contains too much logic. Quite frankly, I cannot understand what is being done here...

concent_api/core/tests/test_payments_report.py
```diff
@@ -0,0 +1,789 @@
...
+        # Test SCI calls.
+        self.assertEqual(get_list_of_payments_mock.call_count, self.number_of_pairs * 3)
+        self.assertEqual(get_covered_additional_verification_costs_mock.call_count, self.number_of_pairs * 2)
+
+        # Test Report Summary.
+        self.assertEqual(report.summary.number_of_pairs, self.number_of_pairs)
+        self.assertEqual(report.summary.number_of_subtasks, self.number_of_subtasks)
+        # 1 payment per pair.
+        self.assertEqual(report.summary.number_of_regular_payments, self.number_of_pairs)
+        # (2 regular payments for odds * 10 odds) + (2 regular payments for even * 10 evens)
```
dybi: odds and evens what?

concent_api/core/tests/test_payments_report.py
```diff
@@ -0,0 +1,789 @@
...
+            )
+            client.full_clean()
+            client.save()
+            self.requestors.append(client)
+
+    def _prepare_subtasks(self):
+        for i in range(self.number_of_subtasks):
+            task_to_compute = self._get_deserialized_task_to_compute(
+                provider_public_key=self.providers[i // 3].public_key,
+                requestor_public_key=(
```
dybi: it should extracted to a well named function to describe the logic here

concent_api/core/tests/test_payments_report.py
```diff
@@ -0,0 +1,789 @@
...
+        self.pending_response_subtask_results_rejected = PendingResponse.objects.last()
+
+        store_pending_message(
+            response_type=PendingResponse.ResponseType.ForcePaymentCommitted,
+            client_public_key=self.subtask.requestor.public_key_bytes,
+            queue=PendingResponse.Queue.Receive,
+            subtask=self.subtask,
+        )
+
+        task_to_compute = deserialize_message(self.subtask.task_to_compute.data)
```
dybi: why this? you have already defined `task_to_compute` in line `266`

concent_api/core/tests/test_payments_sci_backend.py
```diff
@@ -57,7 +57,7 @@ def test_that_if_number_of_blocks_from_timestamp_is_smaller_than_required_confs_
 
         self.assertEqual(list_of_payments, [])
 
-    def test_that_sci_backend_get_list_of_payments_should_return_list_of_forced_subtask_payments(self):
+    def test_that_sci_backend_get_list_of_payments_should_return_list_of_forced_payments(self):
```
dybi: "forced" in the test name, `SETTLEMENT` in the test's body...

concent_api/core/tests/test_payments_sci_backend.py
```diff
@@ -280,3 +317,36 @@ def test_that_sci_backend_cover_additional_verification_cost_should_return_trans
             value=self.transaction_value,
             subtask_id=sci_backend._hexencode_uuid(self.task_to_compute.subtask_id),
         )
+
+    def test_that_sci_backend_get_covered_additional_verification_costs_should_return_transaction_hash(self):
```
dybi: transaction has is not checked in this Test Case...

concent_api/core/payments/report.py
dybi: @cameel, the problem is that `payment.closure_time` is int and all timestamps could be float... we shouldn't assert because timestamps **could** be floats...

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,759 @@
...
+    NO_MATCHING_MESSAGE_FOR_PAYMENT_BUT_DEPOSIT_CLAIM_EXISTS = \
+        'no_matching_message_for_payment_but_deposit_claim_exists'
+    PAYMENT_VALUE_DIFFERS_FROM_VERIFICATION_COST = \
+        'payment_value_differs_from_verification_cost'
+    VERIFICATION_PAYMENT_DO_NOT_HAVE_RELATED_SUBTASK = \
+        'verification_payment_do_not_have_related_subtask'
+
+
+PendingResponses = NamedTuple(
+    'PendingResponses',
```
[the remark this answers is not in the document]
rwrzesien: It is already being used.

concent_api/core/payments/report.py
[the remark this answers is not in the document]
rwrzesien: Right, I though it works differently and pure `distinct(*fields)` is enough. It works correctly for me in both versions, so maybe it have impact when there is greater number of subtasks or they are created in different order.

> Does this code really work? If it does then it returns a random subtask for each pair. This is wrong because the report should process all the available subtasks.

Here I am getting only Clients pairs and TTC. All available subtasks for each pair are retrieved in different place.

:+1: done

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,759 @@
...
+    return [
+        forced_subtask_payment for forced_subtask_payment in forced_subtask_payments
+        if forced_subtask_payment.subtask_id == subtask_results_settled.subtask.subtask_id
+    ]
+
+
+def find_subtask_result_for_additional_verification_payment(
+    additional_verification_payment: CoverAdditionalVerificationEvent,
+    subtask_results_settled: list,
+    subtask_results_rejected: list,
```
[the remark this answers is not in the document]
rwrzesien: :+1: done

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,759 @@
...
+
+    return matched_subtask_results_settled + matched_subtask_results_rejected
+
+
+def find_settlement_payment_for_force_payment_committed(
+    force_payment_committed_pending_response: PendingResponse,
+    settlement_payments: list,
+) -> List[ForcedPaymentEvent]:
+    """ Find settlement payment for given ForcePaymentCommitted message. """
+    payment_info = force_payment_committed_pending_response.payments.order_by('id').last()
```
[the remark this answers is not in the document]
rwrzesien: I think its better to use `prefetch_related` to have it already on available `force_payment_committed_pending_response`, otherwise I will have to pass this dict around from top to bottom.

But no matter which version is chosen, that means that those functions cannot be used as a standalone functions any more.

:+1: done

concent_api/core/payments/report.py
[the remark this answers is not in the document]
rwrzesien: I think that it was returning `int` in first implementation but then there was an issue with it related to golem-messages.

I have checked and it changed in https://github.com/golemfactory/concent/pull/883.

@dybi Do you remember maybe what was the case behind that float return from helper?

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,759 @@
...
+        number_of_regular_payments=len(payments.regular_payments),
+        number_of_forced_payments=(
+            len(payments.settlement_payments) +
+            len(payments.forced_subtask_payments) +
+            len(payments.additional_verification_payments_for_requestor) +
+            len(payments.additional_verification_payments_for_provider)
+        ),
+        total_amount_to_be_paid=(
+            sum([int(force_payment_committed.payments.order_by('id').last().amount_paid)
+                 for force_payment_committed in pending_responses.force_payment_committed])
```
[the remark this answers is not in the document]
rwrzesien: :+1: done

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,759 @@
...
+        ).count(),
+        number_of_regular_payments=len(payments.regular_payments),
+        number_of_forced_payments=(
+            len(payments.settlement_payments) +
+            len(payments.forced_subtask_payments) +
+            len(payments.additional_verification_payments_for_requestor) +
+            len(payments.additional_verification_payments_for_provider)
+        ),
+        total_amount_to_be_paid=(
+            sum([int(force_payment_committed.payments.order_by('id').last().amount_paid)
```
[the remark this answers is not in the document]
rwrzesien: Resolved in https://github.com/golemfactory/concent/pull/1073#discussion_r250678041

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,759 @@
...
+        total_amount_actually_paid=(
+            sum([payment.amount for payment in payments.settlement_payments]) +
+            sum([payment.amount for payment in payments.forced_subtask_payments]) +
+            sum([payment.amount for payment in payments.additional_verification_payments_for_requestor]) +
+            sum([payment.amount for payment in payments.additional_verification_payments_for_provider])
+        ),
+        total_amount_not_paid=(
+            sum([int(force_payment_committed.payments.order_by('id').last().amount_pending)
+                 for force_payment_committed in pending_responses.force_payment_committed])
+        ),
```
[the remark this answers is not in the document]
rwrzesien: :+1: done

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,759 @@
...
+        ('matching_error', MatchingError),
+        ('message', str),
+    ]
+)
+
+
+Warnings = NamedTuple(
+    'Warnings',
+    [
+        ('matching_warning', MatchingWarning),
```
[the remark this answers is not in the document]
rwrzesien: `Warning` is a built-in in Pycharm and it is complaining about overriding it. I think it has no impact on the code itself, but I am not sure if it won't have impact on ie. running tests related to this feature in Pycharm.

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,759 @@
...
+    for subtask_pair in get_client_pairs_from_subtasks():  # pylint: disable=not-an-iterable
+        related_pending_responses_for_pair = get_related_pending_responses_for_pair(
+            subtask_pair.requestor,
+            subtask_pair.provider,
+        )
+
+        task_to_compute = deserialize_message(subtask_pair.task_to_compute.data.tobytes())
+
+        related_payments_for_pair = get_related_payments_for_pair(
+            task_to_compute.requestor_ethereum_address,
```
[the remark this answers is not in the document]
rwrzesien: :+1: done

concent_api/core/payments/report.py
rwrzesien: :+1: done

concent_api/core/payments/bankster.py
```diff
@@ -310,7 +310,7 @@ def settle_overdue_acceptances(
             requestor_eth_address=requestor_ethereum_address,
             provider_eth_address=provider_ethereum_address,
             min_block_timestamp=oldest_payments_ts,
-            transaction_type=TransactionType.FORCE,
+            transaction_type=TransactionType.SETTLEMENT,
```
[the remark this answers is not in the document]
cameel: @dybi
`FORCE` used to refer to **forced payments** (i.e. the settlement payments). I suggested renaming it because it's confusing but it has not been done yet. Now @rwrzesien is renaming `FORCE` to `SETTLEMENT` and making `FORCE` return **forced subtask payments** instead just like I suggested back then.

cameel:
@rwrzesien @dybi Here's my proposal for splitting it into smaller pull requests.

First, get things that are not the focus of the pull request out of the way:
1) Cleanup:
    - Removing unnecessary `filter()` parameters in handlers (https://github.com/golemfactory/concent/pull/1073/files#r250510064)
    - Removing defaults from `make_force_payment_to_provider()`.
2) Adding support for `TransactionType.FORCE` in `get_list_of_payments()`.
    - BTW: if you're adding it there, my remarks from #960 still apply.
3) Adding `get_covered_additional_verification_costs()`

Then, the report itself:

4) The simplest report that just lists messages and payments for each pair. Does not report errors or warnings. Does not try to match them together.
5) Matching and finding errors separately for each situation:
    - `SubtaskResultsSettled` vs forced subtask payements
    - `PaymentCommitted` vs settlement payments
    - Verification payments
    - Anything else that's not included in the above.
6) Summary for the whole report.

concent_api/core/payments/report.py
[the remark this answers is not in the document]
cameel: OK. [`Warning`](https://docs.python.org/3.7/library/exceptions.html#Warning) is one of the exception types available in the global scope.

But `Warnings` is definitely misleading. It does not contain multiple warnings. Maybe `Warning_`? `WarningDetails`? `WarningInfo`?

concent_api/core/payments/report.py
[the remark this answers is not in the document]
cameel: Yes, but that can be achieved by having it set to `None` in other situations.

concent_api/core/payments/report.py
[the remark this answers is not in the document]
cameel: OK. But you're defining it just like an old-style `namedtuple`. Why not:

``` python
class PendingResponses(NamedTuple):
    subtask_results_settled: list
    subtask_results_rejected: list
    force_payment_committed: list
```

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,753 @@
...
+        ),
+    )
+
+
+def get_related_payments_for_pair(requestor_ethereum_address: str, provider_ethereum_address: str) -> Payments:
+    """ Get all payments between given requestor/provider pair. """
+    regular_payments = service.get_list_of_payments(  # pylint: disable=no-value-for-parameter
+        requestor_eth_address=requestor_ethereum_address,
+        provider_eth_address=provider_ethereum_address,
+        min_block_timestamp=get_oldest_subtask_creation_date(),
```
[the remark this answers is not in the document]
cameel: Using creation time of the oldest subtask is just a heuristic. It's not foolproof as you have just noted. The idea is to prevent stuff from before the last database reset from getting in the way - so we don't want to go too far in the past.

But yeah, using the timestamp from RCT seems like a better heuristic. It's still not that far in the past but would eliminate the problems your're talking about.

concent_api/core/tests/test_payments_sci_backend.py
cameel: Let's settle on calling them settlement payments :)

concent_api/core/payments/report.py
cameel: I'm assuming it's being validated elsewhere (e.g. in golem-messages) If not I think it should be. Or are golem-messages allowed to contain fractional times? The rule we agreed on early in the project for validation is that anything that does not fail whan passed to `int()` is a valid timestamp (see https://github.com/golemfactory/concent/issues/54#issuecomment-358923252). Has GM dropped this rule?

concent_api/core/payments/report.py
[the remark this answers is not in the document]
cameel: If you can eliminate the extra query with `prefetch_related()` then it's fine. But seeing how many queries are being made, I'm not sure if you can achieve this in every case.

Jakub89:
We have split this PR into 3 part's:
https://github.com/golemfactory/concent/tree/feature-payment-report-part-1
https://github.com/golemfactory/concent/tree/feature-payment-report-part-2
https://github.com/golemfactory/concent/tree/feature-payment-report-part-3

So that they can be developed separately in the future.

Issue #803, opened 2018-09-12
SigningService: Make _set_was_sigterm_caught_true a method and test it


rwrzesien:
In early stages of `SigningService` implementation `_set_was_sigterm_caught_true` was defined as function inside a method. This way it cannot be unit tested.

I have made a quick prove of concept that it can be a method too, which will allow to unit test it.

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/160478329

dybi:
still valid, priority `low`

PR #1074, opened 2019-01-21
SigningService: Make _set_was_sigterm_caught_true a method and add te…

rwrzesien:
…st for it

Resolves https://github.com/golemfactory/concent/issues/803

### DIFF
--- a/signing_service/signing_service/signing_service.py
+++ b/signing_service/signing_service/signing_service.py
@@ -115,12 +115,8 @@ class SigningService:
 
         self._validate_arguments()
 
-        def _set_was_sigterm_caught_true(signum: int, frame: Optional[FrameType]) -> None:  # pylint: disable=unused-argument
-            logger.info('Closing connection and exiting on SIGTERM.')
-            self.was_sigterm_caught = True
-
         # Handle shutdown signal.
-        signal.signal(signal.SIGTERM, _set_was_sigterm_caught_true)
+        signal.signal(signal.SIGTERM, self._set_was_sigterm_caught_true)
 
     def run(self) -> None:
         """
@@ -343,6 +339,10 @@ class SigningService:
     def _was_sigterm_caught(self) -> bool:
         """ Helper function which checks if SIGTERM signal was caught. """
         return self.was_sigterm_caught
+
+    def _set_was_sigterm_caught_true(self, signum: int, frame: Optional[FrameType]) -> None:  # pylint: disable=unused-argument
+        logger.info('Closing connection and exiting on SIGTERM.')
+        self.was_sigterm_caught = True
 
     def _validate_arguments(self) -> None:
         if not 0 < self.port < 65535:
--- a/signing_service/signing_service/tests/test_signing_service_methods.py
+++ b/signing_service/signing_service/tests/test_signing_service_methods.py
@@ -417,3 +417,27 @@ class SigningServiceDailyThresholdTestCase(TestCase):
         self.signing_service._update_daily_transactions_limit_file_name()
         self.assertEqual(self.signing_service.daily_transactions_limit_file_name, datetime.datetime.now().strftime('%Y-%m-%d'))
         self.assertEqual(self.signing_service.signing_service_daily_transaction_sum_so_far, 0)
+
+
+class SigningServiceSetSigtermTestCase(TestCase):
+
+    def setUp(self):
+        self.host = '127.0.0.1'
+        self.port = 8000
+        self.initial_reconnect_delay = 2
+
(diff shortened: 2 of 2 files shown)


signing_service/signing_service/tests/test_signing_service_methods.py
```diff
@@ -417,3 +417,27 @@ def test_that_signing_service_update_daily_transactions_limit_file_name_overwrit
...
+                self.port,
+                self.initial_reconnect_delay,
+                CONCENT_PUBLIC_KEY,
+                SIGNING_SERVICE_PRIVATE_KEY,
+                TEST_ETHEREUM_PRIVATE_KEY,
+                SIGNING_SERVICE_DEFAULT_RECONNECT_ATTEMPTS,
+                ConsoleNotifier(),
+            )
+
+    def test_that_set_was_sigterm_caught(self):
```
dybi: Something's wrong with the name.

signing_service/signing_service/tests/test_signing_service_methods.py
```diff
@@ -417,3 +417,27 @@ def test_that_signing_service_update_daily_transactions_limit_file_name_overwrit
...
+        self.signing_service._set_was_sigterm_caught_true(None, None)
```
dybi: this method's signature is `(int, Optional[Frame])` - why is it called with `(None, None)`?

signing_service/signing_service/tests/test_signing_service_methods.py
rwrzesien: :+1: fixed

signing_service/signing_service/tests/test_signing_service_methods.py
rwrzesien: Because they are not used inside method in any way, and the test is testing something different, they don't need to be any real values. In runtime this method is called by standard lib code, which pass those values.

dybi: I would be for using real values - it helps to understand what is really being tested and documents the code better.

signing_service/signing_service/signing_service.py
```diff
@@ -341,6 +337,10 @@ def _was_sigterm_caught(self) -> bool:
         """ Helper function which checks if SIGTERM signal was caught. """
         return self.was_sigterm_caught
 
+    def _set_was_sigterm_caught_true(self, signum: int, frame: Optional[FrameType]) -> None:  # pylint: disable=unused-argument
```
pawelkisielewicz: Why `signum: int` and `frame: Optional[FrameType]` are in function declaration?

rwrzesien: This function is connected to signal. It means that when signal is being send, this function will be called with arguments send by signal. And this function is called with arguments send by `signal.SIGTERM`. They come from standard library code. We don't do anything with them, but we still have to accept them.

signing_service/signing_service/tests/test_signing_service_methods.py
rwrzesien: What are real values here? We don't define them in our code, they come from standard library code. I am not even sure how to create proper `FrameType` object, because I don't have to do it - standard library does it.

The documentation for SIGTERM is not something we created, it is a part of standard lib https://docs.python.org/3/library/signal.html.

rwrzesien:
@dybi 
> I would expect that test will ensure that field was_sigterm_caught is changed to True, upon receiving SIGTERM. Please change and/or add such test.

This is something that if tested automatically needs some extra unnecessary work, because you have to involve `subprocess` into it. However, it can be very easily tested manually, which I did.

signing_service/signing_service/signing_service.py
pawelkisielewicz: thanks for explanation

signing_service/signing_service/tests/test_signing_service_methods.py
dybi: In such cases - please use `Mock()` instances or `sentinel` values - these are used to indicate that non real data is being used...

dybi:
@rwrzesien 
> This is something that if tested automatically needs some extra unnecessary work, because you have to involve subprocess into it. However, it can be very easily tested manually, which I did.

We would like to have automatic tests ;) Please see link below:
https://stackoverflow.com/a/26159031

signing_service/signing_service/tests/test_signing_service_methods.py
rwrzesien: :+1: done, please let me know if it should look like this.

Issue #1075, opened 2019-01-21
Replace raising SigningServiceMaximumReconnectionAttemptsExceeded in SigningService with logging


rwrzesien:
We have recurring `SigningServiceMaximumReconnectionAttemptsExceeded` exception being raised and reported to Sentry, which was good when we wanted to synchronize connection, but now it might happen for various reasons and we could replace it with logging.

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/163361953

cameel:
Yeah. Detecting connection problems is the job of the monitoring system (meant for cluster admins), rather than error reporting system (meant for developers). This task is about freeing developers from this nagging report they can't really do anything about. Back when the setting for max reconnections in SS was being implemented I discussed the feature with @dybi. I'm not sure if we explicitly agreed about whether this should be reported or not (we probably did) but now I think it should not. Sentry because Sentry is being flooded by these messages while it's not something developers can fix or are responsible for. Having them there only desensitizes everyone to real problems.

Instead we need to redirect it to monitoring. Monitoring of Signing Service itself is up to Golem team but we can still notice the problem on Concent side. I think that Middleman should log an error after it has not seen SS for a significant period of time (e.g. 2 minutes; should be configurable of course). After that we should add filters to detect events like this and enable notifications for them on the clusters where it really matters (`testnet`, `mainnet`). These notifications should only go to the people managing these clusters (so me and @bartoszbetka for now).

@rwrzesien @dybi @bartoszbetka 
What do you think about it?

rwrzesien:
@cameel Sounds good. Apart from that, if we wanted to move monitoring of `SigningService` exceptions out of our Sentry we might just set `sentry_dsn` to a different one.

PR #1079, opened 2019-01-21
SigningService: Replace raising SigningServiceMaximumReconnectionAtte…

rwrzesien:
…mptsExceeded with logging

Resolves https://github.com/golemfactory/concent/issues/1075.

### DIFF
--- a/signing_service/signing_service/signing_service.py
+++ b/signing_service/signing_service/signing_service.py
@@ -138,7 +138,11 @@ class SigningService:
                 except (socket.error, socket.timeout) as exception:
                     logger.error(f'Socket error occurred: {exception}')
 
-                    self._attempt_reconnection()
+                    try:
+                        self._attempt_reconnection()
+                    except SigningServiceMaximumReconnectionAttemptsExceeded:
+                        logger.error('Maximum number of reconnections exceeded.')
+                        break
                 except KeyboardInterrupt:
                     # Handle keyboard interrupt.
                     logger.info('Closing connection and exiting on KeyboardInterrupt.')
@@ -153,7 +157,6 @@ class SigningService:
         self._increase_delay()
         self.reconnection_counter += 1
         if self.reconnection_counter > self.maximum_reconnection_attempts:
-            crash_logger.error("Maximum reconnection attempts exceeded.")
             raise SigningServiceMaximumReconnectionAttemptsExceeded
         logger.info(f'Reconnecting... (attempt: {self.reconnection_counter}/{self.maximum_reconnection_attempts})')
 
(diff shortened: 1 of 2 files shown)


signing_service/signing_service/signing_service.py
```diff
@@ -138,7 +138,11 @@ def run(self) -> None:
                 except (socket.error, socket.timeout) as exception:
                     logger.error(f'Socket error occurred: {exception}')
 
-                    self._attempt_reconnection()
+                    try:
+                        self._attempt_reconnection()
+                    except SigningServiceMaximumReconnectionAttemptsExceeded:
+                        logger.error('Maximum number of reconnection exceeded.')
```
cameel: `reconnection` -> `reconnections`

signing_service/signing_service/signing_service.py
```diff
@@ -153,7 +157,6 @@ def _attempt_reconnection(self) -> None:
         self._increase_delay()
         self.reconnection_counter += 1
         if self.reconnection_counter > self.maximum_reconnection_attempts:
-            crash_logger.error("Maximum reconnection attempts exceeded.")
             raise SigningServiceMaximumReconnectionAttemptsExceeded
```
cameel: Why not log it here?

rwrzesien: It could be either here or in the place where it is currently (in the mail loop). I have decided to put it there because every `except` block there has a logger usage and for somebody reading the code logically from top to bottom (from main loop down to the more specific functions) it might be more describing. But I have not any better argument for that. Definitely it should not be logged twice.

signing_service/signing_service/signing_service.py
rwrzesien: I have this in my first version, but for some reason Pycharm is marking me this form as a typo. Well. I guess Pycharm is not a vocabulary software after all.
:+1: fixed

signing_service/signing_service/signing_service.py
```diff
@@ -138,7 +138,11 @@ def run(self) -> None:
                 except (socket.error, socket.timeout) as exception:
                     logger.error(f'Socket error occurred: {exception}')
 
-                    self._attempt_reconnection()
+                    try:
+                        self._attempt_reconnection()
+                    except SigningServiceMaximumReconnectionAttemptsExceeded:
+                        logger.error('Maximum number of reconnections exceeded.')
+                        break
```
dybi: This break will close the `SigningService` - am I right? Is it our intention?

rwrzesien: Yes. Also, it was working exactly like this before this change.

signing_service/signing_service/signing_service.py
cameel: OK

signing_service/signing_service/signing_service.py
cameel: In singular this sounds very Japanese :)

Issue #1078, opened 2019-01-21
Remove a constraint in Subtask model that says that the requestor must be unique together with task_id


rwrzesien:
Original Sentry issue: https://talkback.golem.network/sentry/concent/issues/57029

There is a constrain in Subtask model that requestor must be unique together with task_id. It should be removed, test for this case added, and any related failing tests fixed.

This also require blueprint uppdate https://github.com/golemfactory/concent/issues/139.

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/163386405

PR #1080, opened 2019-01-22
Bugfix remove constraint in subtask model that says that the requestor must be unique together with task

rwrzesien:
Resolves https://github.com/golemfactory/concent/issues/1078


concent_api/core/tests/test_core_views.py
```diff
@@ -240,6 +240,49 @@ def test_send_should_return_http_400_if_task_id_already_use(self):
...
+                force_report_computed_task,
+                self.PROVIDER_PRIVATE_KEY,
+                CONCENT_PUBLIC_KEY,
+            ),
+        )
+
+        self.assertIsInstance(response_202, HttpResponse)
+        self.assertEqual(response_202_again.status_code, 202)
+
+        subtask_1 = Subtask.objects.all().order_by('created_at')[0]
```
Jakub89:
```
E:280,20: Instance of 'list' has no 'order_by' member (no-member)
```

concent_api/core/tests/test_core_views.py
```diff
@@ -240,6 +240,49 @@ def test_send_should_return_http_400_if_task_id_already_use(self):
...
+        subtask_2 = Subtask.objects.all().order_by('created_at')[1]
```
Jakub89:
```
E:281,20: Instance of 'list' has no 'order_by' member (no-member)
```

concent_api/core/tests/utils.py
```diff
@@ -283,7 +283,11 @@ def _get_deserialized_task_to_compute(
                 want_to_compute_task=want_to_compute_task if want_to_compute_task is not None else self._get_deserialized_want_to_compute_task(
                     provider_public_key=provider_public_key,
                     provider_ethereum_public_key=provider_ethereum_public_key,
-                    task_id=task_id,
+                    task_id=(
+                        compute_task_def['task_id']
```
dybi: this part seems unnecessary here - if `compute_task_def` is not passed as argument, it is created with the same `task_id`

concent_api/core/models.py
```diff
@@ -365,7 +365,6 @@ class SubtaskState(ChoiceEnum):
 
     class Meta:
         unique_together = (
-            ('requestor', 'task_id'),
             ('requestor', 'subtask_id'),
```
dybi: shouldn't we remove this as well? and have 
```python
unique_together = (
    ('task_id', 'subtask_id')
)
```
instead?

concent_api/core/tests/test_core_views.py
```diff
@@ -240,6 +240,49 @@ def test_send_should_return_http_400_if_task_id_already_use(self):
             error_code=ErrorCode.SUBTASK_DUPLICATE_REQUEST
         )
 
+    @freeze_time("2017-11-17 10:00:00")
+    def test_send_should_return_http_202_if_task_id_already_in_use_but_subtask_id_differs_for_the_same_requestor(self):
+
+        response_202 = self.send_request(
+            url='core:send',
+            data=dump(
+                self.correct_golem_data,
```
dybi: what's `correct_golem_data`?

concent_api/core/tests/utils.py
rwrzesien: It was not working this way, that's why I have changed it.

concent_api/core/tests/test_core_views.py
rwrzesien: It is defined in `setUp` of this function and used in most tests in this class.

concent_api/core/tests/test_core_views.py

concent_api/core/tests/test_core_views.py

concent_api/core/tests/utils.py
dybi: what exactly is not working?

concent_api/core/models.py
dybi: ok, so it would seem we need `('requestor', 'task_id')` **and** `('task_id', 'subtask_id')`?

concent_api/core/tests/test_core_views.py
dybi: then I would suggest renaming it, because it doesn't say anything about what kind of message is being sent/what USE CASE is being started here...

concent_api/core/tests/utils.py
kbeker: You mean it wasn't working when CTD was already created and we create TTC with new task_id?

concent_api/core/models.py
rwrzesien: It might work like this, but I think this change will result in fixing some decent amount of tests, so if you want to do this add a separate issue for this. This issue is only about removing one of the constrains.

concent_api/core/tests/test_core_views.py
rwrzesien: Because it doesn't matter which use case it is. Those tests are not supposed to test any particular use case (even if there are leftovers from testing `force report`), but some general rules among all use cases.

concent_api/core/tests/utils.py
pawelkisielewicz: I see, but when I looked at this first time it looked strange for me as well. `compute_task_def` is passed to this method or created, so in this place it exists. Is it possible that `compute_task_def` does not have field `['task_id']`? Could it be changed simple to 
`task_id=compute_task_def['task_id']` ?

BTW. :+1: for discovering it

concent_api/core/models.py
dybi: @cameel, your thoughts?

rwrzesien: @cameel @dybi I think it requires very small amount of work, it would be only adding constraint, migration and few extra tests. I doubt that more than several existing tests will fail, as they already use unique `subtask_id`.

rwrzesien: Created https://github.com/golemfactory/concent/issues/1096.

Issue #1084, opened 2019-01-24
Move relation between PendingResponse and PaymentInfo on the former


rwrzesien:
Original source comment:
https://github.com/golemfactory/concent/pull/1073#discussion_r250290947

Move relation between `PendingResponse` and `PaymentInfo` on the `PendingResponse` and make it it a nullable `OneToOne` field.

Fix related usages (`message_handlers` and `payment_report`) and failing tests.

PR #1089, opened 2019-01-28
Chore move relation between pending response and payment info on the former

pawelkisielewicz:
Resolves https://github.com/golemfactory/concent/issues/1084


concent_api/core/models.py
```diff
@@ -618,54 +667,6 @@ def clean(self) -> None:
             })
 
 
-class PaymentInfo(Model):
```
rwrzesien: Why have you moved the whole model?

concent_api/core/transfer_operations.py
```diff
@@ -82,10 +75,19 @@ def store_pending_message(
                 amount_paid=payment_message.amount_paid,
                 recipient_type=payment_message.recipient_type.name,  # pylint: disable=no-member
                 amount_pending=payment_message.amount_pending,
-                pending_response=receive_queue
             )
             payment_committed_message.full_clean()
             payment_committed_message.save()
+        receive_queue = PendingResponse(
```
rwrzesien: This can be achieved in more readable way:
1. Create `PendingResponse` object.
2. `if payment_message is not None:` create `PaymentInfo` and assign to created `PendingResponse` object.
3. `full_clean` and `save` on `PendingResponse` object.

concent_api/core/models.py
```diff
@@ -604,9 +651,11 @@ def response_type_enum(self) -> 'ResponseType':
     created_at = DateTimeField(auto_now_add=True)
     modified_at = DateTimeField(auto_now=True)
 
+    payment_info = OneToOneField(PaymentInfo, related_name='payments', blank=True, null=True)
```
rwrzesien: Related name should be `payment_info`, `payments` suggests a collection.

concent_api/core/models.py
```diff
@@ -566,6 +566,53 @@ def _current_state_enum(self) -> 'SubtaskState':
         return Subtask.SubtaskState[self._current_state_name]  # type: ignore
 
 
+class PaymentInfo(Model):
+    """
+    Stores the information needed to construt a PaymentCommitted message
+    """
```
cameel: Please put changes like moving this class to a new place in a separate commit. Otherwise it's hard to see what really changed.

concent_api/core/models.py
cameel: Is it even necessary to explicitly set `related_name` here?

cameel: BTW, don't forget to regenerate the migration if you change it.

concent_api/core/models.py
rwrzesien: I think the whole move is not necessary, the same effect can be achieved when wrapping class name in a string in relation field.

pawelkisielewicz: Yes, I wrapped the name in string and everything is OK- thanks, didn't know about it. @cameel I dont't move the classes, so separate commit is not needed any more.

concent_api/core/models.py
pawelkisielewicz: It is answered in https://github.com/golemfactory/concent/pull/1089#discussion_r251551319

concent_api/core/models.py
pawelkisielewicz: `related_name` is not necessary, removed. I remembered about migration, but thanks :)

concent_api/core/transfer_operations.py
pawelkisielewicz: done with helper function

PR #1097, opened 2019-01-31
Bugfix message not identical exception

kbeker:
No issue for that. Problem appeared after fix of node_integration_tests

### DIFF
--- a/concent_api/core/tests/test_unit_authentication_utilities.py
+++ b/concent_api/core/tests/test_unit_authentication_utilities.py
@@ -1,8 +1,11 @@
+from copy import deepcopy
+
 from django.test                    import override_settings
 from django.test                    import TestCase
 
 from golem_messages                 import message
 from golem_messages.factories       import tasks
+from golem_messages.factories.tasks import TaskToComputeFactory
 from golem_messages.shortcuts       import dump
 from golem_messages.shortcuts       import load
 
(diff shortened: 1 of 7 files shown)


concent_api/core/validation.py
```diff
@@ -127,20 +127,21 @@ def validate_all_messages_identical(golem_messages_list: List[message.Message])
...
-                    'The difference is on field {}: {} is not equal {}'.format(
-                        type(base_golem_message).__name__,
-                        i,
-                        slot,
-                        getattr(base_golem_message, slot),
-                        getattr(golem_message, slot),
-                    ),
-                    error_code=ErrorCode.MESSAGES_NOT_IDENTICAL,
-                )
+        if not base_golem_message.__eq__(golem_message):
```
rwrzesien: Was the problem here related with nested messages in TTC?

concent_api/core/tests/test_unit_authentication_utilities.py
```diff
@@ -152,16 +149,11 @@ def test_that_validate_all_messages_identical_should_raise_http400_when_not_all_
...
-            message.TaskToCompute(
-                requestor_id = 1
-            ),
-            message.TaskToCompute(
-                requestor_id = 1
-            ),
-            message.TaskToCompute(
-                requestor_id = 1
-            ),
+            task_to_compute,
```
dybi: please explicitly change `encrypted` param in one of the messages to be different

concent_api/core/validation.py
kbeker: It was z problem with `encrypted` field in golem_messages. This field can be different and this method allows us to use method from GM and we will always know for sure is it equaled or not :)

dybi: 1) 
```suggestion
        if  base_golem_message != golem_message:
```
2) I would get rid of that "iteration" part that seeks which slot is different. But this is only my PoV - you decide ;)

concent_api/core/tests/test_unit_authentication_utilities.py
kbeker: Done

concent_api/core/validation.py
kbeker: Fixed :)

PR #1102, opened 2019-02-12
Chore downgrade gm from 2.24.2 to 2.15.0

kbeker:
Needs testing on cluster. All tests pass.


kbeker:
> Seems legit, but there is one question: what about `ForcePayment` msg? Can it have an empty `SubtaskResultsAccepted`list in this version of `GM`? If so, we need a test and additional check in the handler.
@dybi Added in newest commit: https://github.com/golemfactory/concent/pull/1102/commits/6df2bf538f156a0b19f88e2a3d4664aabd653c47

Issue #1106, opened 2019-02-13
Pending messages in older versions block receiving any other messages from `receive/`


cameel:
Blueprint #857 states the following with regard to a client communicating with the `receive/` endpoint:

> If the response needs to contain a nested message taken from the database and that message is not compatible with Concent's version of golem-messages, Concent responds with `ServiceRefused`. The client is always required to stay on the same protocol version while communicating about the same subtask.

This is unfortunately not viable in practice. It makes it impossible to receive any messages from Concent in the situation where one updates his Golem client to a newer, incompatible version while there are still pending messages. The pending older messages stand in the way of newer ones but they can't be skipped or discarded.

![concent-receive-version-conflict-no-filtering](https://user-images.githubusercontent.com/137030/52723514-f521ee00-2fad-11e9-9090-7a127b87b23e.png)

The best solution for this problem is to simply ignore messages that can't be processed using the available version of golem-messages. `receive/` should simply filter messages by version and return only those compatible.

![concent-receive-version-conflict-with-filtering](https://user-images.githubusercontent.com/137030/52723516-f521ee00-2fad-11e9-9a39-efcc8dc5e468.png)

### Implementation notes
1) Keep in mind that _compatible versions_ does not mean _identical versions_. For example 2.24.1 is compatible with 2.24.2.
2) To filter the `PendingMessage` message table you'll need to make a `JOIN` query `Subtask` and also add `protocol_version` to `Subtask` table. Currenty only `StoredMessage` contains the version and there can be many `StoredMessage`s associated with one `Subtask`.
    - There should be a validation that all `StoredMessage` objects associated with a `Subtask` have a version compatible with subtask's `protocol_version`.
    - It's best to store only the two first version components with a `Subtask`. For example `2.24` rather than `2.24.1` or `2.24.5rc3`.
        - Now we'll have two fields with the same name but different content: `Subtask.protocol_version` with the first two components and `StoredMessage.protocol_version` with the full version. I suggest renaming `StoredMessage.protocol_version` to `StoredMessage.golem_messages_version`.
3) Telling the client that there are no pending messages while there are some is misleading. Consider adding a header that would indicate that there are pending messages in a different version. E.g. `Concent-Incompatible-Messages-Pending: 2.14`. It could contain the version of the last pending message. It would be missing or empty if there are really no pending messages.
    - This is optional since it's doubtful that a Golem client will ever use in practice it but I think a good, complete and friendly API should have a way to convey that information.

[...]

PR #1107, opened 2019-02-14
Bugfix pending messages in older versions block receiving any other messages

kbeker:
Resolves https://github.com/golemfactory/concent/issues/1106

Missing:
Add point 3 from issue https://github.com/golemfactory/concent/issues/1106

Locally api-tests pass


concent_api/core/message_handlers.py
```diff
@@ -1,5 +1,6 @@
 from base64 import b64encode
 from collections import Counter
+
```
rwrzesien: Not needed I think.

concent_api/core/message_handlers.py
```diff
@@ -956,34 +957,33 @@ def store_subtask(
 def handle_messages_from_database(client_public_key: bytes) -> Union[message.Message, None]:
     assert client_public_key not in ['', None]
     encoded_client_public_key = b64encode(client_public_key)
-
     with non_nesting_atomic(using='control'):
-        pending_response = PendingResponse.objects.select_for_update().filter(
+        payments_info_responses = PendingResponse.objects.select_for_update().filter(
```
rwrzesien: It is actually one `response` because you do `first()` at the end of the query.

concent_api/core/message_handlers.py
```diff
@@ -956,34 +957,33 @@ def store_subtask(
...
+            payment_info__isnull=False,
             client__public_key=encoded_client_public_key,
             delivered=False,
         ).order_by('created_at').first()
 
-        if pending_response is None:
-            return None
+        subtask_responses = PendingResponse.objects.select_for_update().filter(
```
rwrzesien: Same here, only one `response`.

concent_api/core/message_handlers.py
```diff
@@ -956,34 +957,33 @@ def store_subtask(
...
+            subtask__protocol_version=get_major_and_minor_golem_messages_version(settings.GOLEM_MESSAGES_VERSION),
+            client__public_key=encoded_client_public_key,
+            delivered=False,
+        ).order_by('created_at').first()
+        if payments_info_responses is not None and subtask_responses is not None:
```
rwrzesien: Could you please explain why we differentiate logic for `PendingResponse` with `payment_info` and without?

concent_api/core/message_handlers.py
```diff
@@ -956,34 +957,33 @@ def store_subtask(
...
+                pending_response = payments_info_responses
+            else:
+                pending_response = subtask_responses
+        else:
+            if payments_info_responses is not None:
+                pending_response = payments_info_responses
+            elif subtask_responses is not None:
+                pending_response = subtask_responses
+            else:
+                return None
```
rwrzesien: This line will be never reached.

concent_api/core/message_handlers.py
```diff
@@ -956,34 +957,33 @@ def store_subtask(
...
+            client__public_key=encoded_client_public_key,
+            delivered=False,
+        ).order_by('created_at').first()
+        if payments_info_responses is not None and subtask_responses is not None:
+            if payments_info_responses.created_at > subtask_responses.created_at:
+                pending_response = payments_info_responses
+            else:
+                pending_response = subtask_responses
+        else:
+            if payments_info_responses is not None:
```
rwrzesien: Can be simplified to 
`pending_response = payments_info_responses or subtask_responses`

concent_api/core/subtask_helpers.py
```diff
@@ -162,7 +163,7 @@ def check_compatibility(subtask: Subtask, client_public_key: bytes) -> None:
     if not is_protocol_version_compatible(subtask.task_to_compute.protocol_version):
         error_message = f'Unsupported version of golem messages in stored messages. ' \
             f'Version stored in database is {subtask.task_to_compute.protocol_version}, ' \
-            f'Concent version is {settings.GOLEM_MESSAGES_VERSION}.'
+            f'Concent version is {get_major_and_minor_golem_messages_version(settings.GOLEM_MESSAGES_VERSION)}.'
```
rwrzesien: Why not making this a constant? The version should not change without a server restart. If you make it a constant it will be calculated only once.

concent_api/core/tests/test_integration_version_compatibility.py
```diff
@@ -0,0 +1,197 @@
...
+        delivered=False,
+        queue=PendingResponse.Queue.Receive.name,  # pylint: disable=no-member
+        subtask__protocol_version=golem_message_version
+    )
+
+
+@override_settings(
+    CONCENT_PRIVATE_KEY=CONCENT_PRIVATE_KEY,
+    CONCENT_PUBLIC_KEY=CONCENT_PUBLIC_KEY,
+
```
rwrzesien: Please remove blank line.

concent_api/core/utils.py
```diff
@@ -121,14 +121,19 @@ def extract_name_from_scene_file_path(absoulte_scene_file_path_in_docker: str) -
     return relative_scene_file_path_in_archive
 
 
+def get_major_and_minor_golem_messages_version(protocol_version: str) -> str:
+    divided_version = protocol_version.split('.')
```
rwrzesien: I don't get this part - you are splitting the string at `.` just to join it in the next line.

concent_api/core/message_handlers.py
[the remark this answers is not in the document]
dybi: okay, forget it: using @rwrzesien's proposal is good solution here: 
```python
else:
    pending_response = payments_info_responses or subtask_responses
```

concent_api/core/message_handlers.py

concent_api/core/message_handlers.py
kbeker: Changed to your soulution

concent_api/core/message_handlers.py

concent_api/core/message_handlers.py

concent_api/core/message_handlers.py

concent_api/core/tests/test_integration_version_compatibility.py
kbeker: Done

concent_api/core/subtask_helpers.py
kbeker: @rwrzesien Changed to setting: `settings.MAJOR_MINOR_GOLEM_MESSAGES_VERSION` If better name will come to your mind feel free to say it because I don't have any better idea for it and I still don't like this name :P

concent_api/core/utils.py
kbeker: @rwrzesien Becuase sometimes we can get here 2 elements out of 3 and we want to store this version as a string in this form: `major_number.minor_number`. Golem Messages version has also patch which is unnecessary for us so I decided to do this convertion this way :)
@dybi I had a lot of thinking about good name for this function(what you wrote was one of it) and I still didn't like any of my thought about name :P If you want me to change name of this function I will. Also maybe today you got any new idea about the name? :D

concent_api/core/message_handlers.py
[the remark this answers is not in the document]

concent_api/core/utils.py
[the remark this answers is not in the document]
kbeker: But we don't need tuple of strings. We need simply string which will contain major and minor from golem messages version separated by a dot. Example: `2.24`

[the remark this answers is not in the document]
kbeker: @dybi But we don't need tuple of strings. We need simply string which will contain major and minor from golem messages version separated by a dot. Example: `2.24`

concent_api/core/message_handlers.py
kbeker: @rwrzesien We need to differentiate because `PaymentInfo` does not contain `subtask_id` and we can't sort it by GolemMessages version which is stored in database. Also `PaymentInfo` is not dependent on the GolemMessages version becuase all information are stored in separate columns in database so we can create and serialize message in `2.24` and '3.2' GM version. Unless construction of `ForcePaymentComitted` will change.

concent_api/core/subtask_helpers.py
rwrzesien: I think at some point we wanted to have pure values in settings only, and keep calculations like this in constants. I am OK with how you have done it currently, because I am used to make calculations and function calls in settings, but please make sure with @cameel .

cameel: @kbeker
> If better name will come to your mind feel free to say it because I don't have any better idea for it and I still don't like this name :P

In deployment I'm calling it simply _protocol version_ because the first two components of GM version determine whether the protocol is the same or not (see https://github.com/golemfactory/concent-deployment/issues/346).

@kbeker @rwrzesien 
> I think at some point we wanted to have pure values in settings only, and keep calculations like this in constants. I am OK with how you have done it currently, because I am used to make calculations and function calls in settings, but please make sure with @cameel .

Good point @rwrzesien. Thanks for pinging me. You're right, I think these should not be in settings but in this case it's not just about calculations. The fact that just importing `golem_messages.__version__` may read/create files is bad too but even if that was not the case, storing versions in a setting does not make any sense to me:
- Do we even support changing it? Having a setting for it suggests that we do.
    - If not, it should be a constant. Settings are more "public" than constants. They can even be considered a kind of an API and that comes with all the drawbacks of having to keep an API backwards-compatible.
- Will the code even still work when it's changed? Do we have proper checks/validations against values that could break the app?
    - Constants do not need such extensive checks because we have complete control over them. You can even make assertions about them. No so with settings.
    - A setting requires a bigger upfront investment from us to make sure our code works with all supported values. With a constant it's always a single value and you know that if anyone wants to change it, the code must be modified and go through all the usual testing and Q&A to make sure it still works.
- Does changing it even make sense? Will anyone realistically need to set it to anything other than `golem_messages.__version__`?
    - Sometimes it's useful to expose something as a setting even if it's just a temporary or internal thing because it's still useful in our limited circumstances. It's not the case here. I can't think of a situation where we'd want to manually set GM version in our deployment.

It's not even more convenient or readable than simply using `golem_messages.__version__`. In the code. So why do it?

I think these settings should be removed.

rwrzesien: Issue created https://github.com/golemfactory/concent/issues/1111.

PR #1113, opened 2019-02-19
Fix ValueError with 'no block after {timestamp}'

kbeker:
Add missing check for block number. Problem occurred when with @Jakub89 we started to test PaymentReport

### DIFF
--- a/concent_api/core/constants.py
+++ b/concent_api/core/constants.py
@@ -51,6 +51,9 @@ SCI_CALLBACK_MAXIMUM_TIMEOUT = 30
 # should be listed.
 PAYMENTS_FROM_BLOCK_SAFETY_MARGIN = 10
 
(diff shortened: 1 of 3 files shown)


concent_api/core/constants.py
```diff
@@ -51,6 +51,9 @@
 # should be listed.
 PAYMENTS_FROM_BLOCK_SAFETY_MARGIN = 10
 
+# Defines minimum block timestamp margin for fetching newest block
+MINIMUM_BLOCK_TIMESTAMP_MARGIN = 20
```
dybi: why `20`?

kbeker:
PR is outdated after [this](https://github.com/golemfactory/golem-smart-contracts-interface/pull/50) pull request is merged into golem-smart-contract-interface repository.

PR #1116, opened 2019-02-20
Feature payment report part 1

kbeker:
Part of https://github.com/golemfactory/concent/issues/1044

### DIFF
--- a/concent_api/core/payments/report.py
+++ b/concent_api/core/payments/report.py
@@ -1,0 +1,212 @@
+from typing import List
+from typing import NamedTuple
+from typing import Tuple
+
+from django.db.models import QuerySet
+from golem_messages.exceptions import MessageError
+
+from common.helpers import deserialize_message
+from common.helpers import get_current_utc_timestamp
+from core.exceptions import SCINotSynchronized
+from core.models import Client
+from core.models import PendingResponse
+from core.models import Subtask
+from core.payments import service
+from core.payments.backends.sci_backend import TransactionType
+
+
+class PendingResponses(NamedTuple):
+    subtask_results_settled: QuerySet
+    subtask_results_rejected: QuerySet
+    force_payment_committed: QuerySet
+
+    def is_empty(self) -> bool:
+        is_subtask_results_settled_empty = self.subtask_results_settled.count() == 0
+        is_subtask_results_rejected_empty = self.subtask_results_rejected.count() == 0
+        is_force_payment_committed_empty = self.force_payment_committed.count() == 0
+
+        return is_subtask_results_settled_empty and is_subtask_results_rejected_empty and is_force_payment_committed_empty
+
+
+class Payments(NamedTuple):
+    regular_payments: list
+    settlement_payments: list
+    forced_subtask_payments: list
+    requestor_additional_verification_payments: list
+    provider_additional_verification_payments: list
+
+    def is_empty(self) -> bool:
+        is_regular_payments_empty = len(self.regular_payments) == 0
+        is_settlement_payments_empty = len(self.settlement_payments) == 0
+        is_forced_subtask_payments_empty = len(self.forced_subtask_payments) == 0
+        is_requestor_additional_verification_payments_empty = len(self.requestor_additional_verification_payments) == 0
+        is_provider_additional_verification_payments_empty = len(self.provider_additional_verification_payments) == 0
+
+        return is_regular_payments_empty and is_settlement_payments_empty and is_forced_subtask_payments_empty and is_requestor_additional_verification_payments_empty and is_provider_additional_verification_payments_empty
+
+
+class PairReport(NamedTuple):
+    requestor: Client
+    provider: Client
+    requestor_ethereum_addresses: list
+    provider_ethereum_addresses: list
+    pending_responses: PendingResponses
+    payments: Payments
+
+
+class ClientEthereumAddresses(NamedTuple):
+    provider: str
+    requestor: str
+
+
+class Report(NamedTuple):
+    pair_reports: list
+
+
+def create_report() -> Report:
+    """ Create payments report for Concent. """
+    pair_reports = []
+
+    for subtask in get_subtasks_with_distinct_clients_pairs():  # pylint: disable=not-an-iterable
+        (subtasks_ids, clients_ethereum_addresses) = get_subtask_ids_and_ethereum_addresses_for_clients_pair(
+            subtask.requestor,
+            subtask.provider,
+        )
+
+        related_pending_responses_for_pair = get_related_pending_responses_for_subtasks_ids(
+            subtasks_ids,
+            subtask.requestor,
+            clients_ethereum_addresses,
+        )
+        try:
+            related_payments_for_pair = get_related_payments_for_pair(clients_ethereum_addresses)
+        except SCINotSynchronized:
+            related_payments_for_pair = Payments([], [], [], [], [])
+
+        if not(related_payments_for_pair.is_empty() and related_pending_responses_for_pair.is_empty()):  # pylint: disable=no-member
+            pair_report = PairReport(
+                requestor=subtask.requestor,
+                provider=subtask.provider,
+                requestor_ethereum_addresses=[client_ethereum_address.requestor for client_ethereum_address in clients_ethereum_addresses],
+                provider_ethereum_addresses=[client_ethereum_address.provider for client_ethereum_address in clients_ethereum_addresses],
+                pending_responses=related_pending_responses_for_pair,
+                payments=related_payments_for_pair,
+            )
+            pair_reports.append(pair_report)
+
+    return Report(pair_reports)
+
+
+def get_subtasks_with_distinct_clients_pairs() -> QuerySet:
+    """ Get Subtasks with distinct pairs of requestor and provider Clients. """
+    return Subtask.objects.order_by('provider', 'requestor').distinct('provider', 'requestor')
+
+
+def get_subtask_ids_and_ethereum_addresses_for_clients_pair(requestor: Client, provider: Client) -> Tuple:
+    """ Get Subtasks Ids and all Ethereum keys for Clients pair. """
+    subtask_ids = []
+    clients_ethereum_addresses = []
+
+    for subtask in Subtask.objects.filter(requestor=requestor, provider=provider):  # pylint: disable=no-member
+        subtask_ids.append(subtask.subtask_id)
+        try:
+            task_to_compute = deserialize_message(subtask.task_to_compute.data.tobytes())
+            clients_ethereum_addresses.append(
+                ClientEthereumAddresses(
+                    task_to_compute.provider_ethereum_address,
+                    task_to_compute.requestor_ethereum_address,
+                )
+            )
+        except MessageError:
+            pass
+
+    return subtask_ids, list(set(clients_ethereum_addresses))
+
+
+def get_related_pending_responses_for_subtasks_ids(subtasks_ids: List[str], requestor: Client, clients_ethereum_addresses: list) -> PendingResponses:
+    """ Group and return pending responses from Subtask queryset. """
+    return PendingResponses(
+        subtask_results_settled=PendingResponse.objects.filter(
+            response_type=PendingResponse.ResponseType.SubtaskResultsSettled.name,  # pylint: disable=no-member
+            subtask__subtask_id__in=subtasks_ids,
+        ),
+        subtask_results_rejected=PendingResponse.objects.filter(
+            response_type=PendingResponse.ResponseType.SubtaskResultsRejected.name,  # pylint: disable=no-member
+            subtask__subtask_id__in=subtasks_ids,
+        ),
+        force_payment_committed=prefetch_payment_info(
+            PendingResponse.objects.filter(
+                response_type=PendingResponse.ResponseType.ForcePaymentCommitted.name,  # pylint: disable=no-member
+                client=requestor,
+                payment_info__provider_eth_account__in=[client_ethereum_address.provider for client_ethereum_address in clients_ethereum_addresses]
+            )
+        ),
+    )
+
+
+def prefetch_payment_info(queryset: QuerySet) -> QuerySet:
+    """ Prefetches related PaymentInfo to PendingResponse model queryset. """
+    return queryset.select_related('payment_info')
+
+
+def get_related_payments_for_pair(clients_ethereum_addresses: list) -> Payments:
+    """ Get all payments between given requestor/provider pair. """
+    regular_payments = []  # type: List[str]
+    settlement_payments = []  # type: List[str]
+    forced_subtask_payments = []  # type: List[str]
+    requestor_additional_verification_payments = []  # type: List[str]
+    provider_additional_verification_payments = []  # type: List[str]
+    oldest_subtask_creation_date = get_report_computed_task_timestamp_of_oldest_subtask()
+    for clients_pair in clients_ethereum_addresses:
+        regular_payments = service.get_list_of_payments(  # pylint: disable=no-value-for-parameter
+            requestor_eth_address=clients_pair.requestor,
+            provider_eth_address=clients_pair.provider,
+            min_block_timestamp=oldest_subtask_creation_date,
+            transaction_type=TransactionType.BATCH,
+        )
+
+        settlement_payments = service.get_list_of_payments(  # pylint: disable=no-value-for-parameter
+            requestor_eth_address=clients_pair.requestor,
+            provider_eth_address=clients_pair.provider,
+            min_block_timestamp=oldest_subtask_creation_date,
+            transaction_type=TransactionType.SETTLEMENT,
+        )
+
+        forced_subtask_payments = service.get_list_of_payments(  # pylint: disable=no-value-for-parameter
+            requestor_eth_address=clients_pair.requestor,
+            provider_eth_address=clients_pair.provider,
+            min_block_timestamp=oldest_subtask_creation_date,
+            transaction_type=TransactionType.FORCED_SUBTASK_PAYMENT,
+        )
+
+        requestor_additional_verification_payments = service.get_covered_additional_verification_costs(  # pylint: disable=no-value-for-parameter
+            client_eth_address=clients_pair.requestor,
+            payment_ts=oldest_subtask_creation_date,
+        )
+        provider_additional_verification_payments = service.get_covered_additional_verification_costs(  # pylint: disable=no-value-for-parameter
+            client_eth_address=clients_pair.provider,
+            payment_ts=oldest_subtask_creation_date,
+        )
+
+    return Payments(
+        regular_payments=regular_payments,
+        settlement_payments=settlement_payments,
+        forced_subtask_payments=forced_subtask_payments,
+        requestor_additional_verification_payments=requestor_additional_verification_payments,
+        provider_additional_verification_payments=provider_additional_verification_payments,
+    )
+
+
+def get_report_computed_task_timestamp_of_oldest_subtask() -> int:
+    """
+    Return ReportComputedTask timestamp of oldest Subtask stored in database.
+    If there are no Subtasks in database, return current timestamp.
+    """
+    subtasks_set = Subtask.objects.order_by('created_at')
+    for subtask in subtasks_set:  # pylint: disable=not-an-iterable
+        try:
+            return deserialize_message(subtask.report_computed_task.data.tobytes()).timestamp
+        except MessageError:
+            pass
+
+    return get_current_utc_timestamp()
--- a/concent_api/core/tests/test_payments_report.py
+++ b/concent_api/core/tests/test_payments_report.py
@@ -1,0 +1,213 @@
+import base64
+import mock
+
+from django.test import override_settings
+from golem_messages import message
+from golem_messages.cryptography import ECCx
+
+from common.helpers import deserialize_message
+from core.message_handlers import store_subtask
+from core.models import Client
+from core.models import PaymentInfo
+from core.models import PendingResponse
+from core.models import Subtask
+from core.payments.report import create_report
+from core.payments.report import get_subtasks_with_distinct_clients_pairs
+from core.tests.utils import ConcentIntegrationTestCase
+from core.transfer_operations import store_pending_message
+from core.utils import get_current_utc_timestamp
+from core.utils import hex_to_bytes_convert
+from core.utils import parse_timestamp_to_utc_datetime
+
+
+@override_settings(
+    ADDITIONAL_VERIFICATION_COST=1,
+)
+class PaymentsReportTest(ConcentIntegrationTestCase):
+
+    def setUp(self):
+        super().setUp()
+        self.providers = []
+        self.requestors = []
+        self.subtasks = []
+        self.get_list_of_payments_mocks = []
+        self.get_covered_additional_verification_costs_mocks = []
+        self.closure_time = get_current_utc_timestamp()
+        self.payment_ts = parse_timestamp_to_utc_datetime(self.closure_time)
+        self.default_payment_amount = 10
+        self.number_of_clients_per_type = 10
+        self.number_of_subtasks = 30
+        # The value below is determined by approach of mixing requestors and providers in _prepare_subtasks
+        self.number_of_pairs = self.number_of_clients_per_type * 2
+
+        self._prepare_client("providers")
+        self._prepare_client("requestors")
+        self._prepare_subtasks()
+        self._prepare_related_payments_mocks()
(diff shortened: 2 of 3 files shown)


kbeker:
> ```
> concent_api/core/payments/report.py:23: error: Function is missing a type annotation
> concent_api/core/payments/report.py:38: error: Function is missing a type annotation
> concent_api/core/payments/report.py:80: error: Call to untyped function "is_empty" in typed context
> concent_api/core/payments/report.py:145: error: Need type annotation for 'regular_payments'
> concent_api/core/payments/report.py:146: error: Need type annotation for 'settlement_payments'
> concent_api/core/payments/report.py:147: error: Need type annotation for 'forced_subtask_payments'
> concent_api/core/payments/report.py:148: error: Need type annotation for 'additional_verification_payments_for_requestor'
> concent_api/core/payments/report.py:149: error: Need type annotation for 'additional_verification_payments_for_provider'
> ```

tenor

fixed

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,202 @@
...
+from core.payments import service
+from core.payments.backends.sci_backend import TransactionType
+
+
+class PendingResponses(NamedTuple):
+    subtask_results_settled: QuerySet
+    subtask_results_rejected: QuerySet
+    force_payment_committed: QuerySet
+
+    def is_empty(self) -> bool:
```
rwrzesien: If you will keep the version with `len()` this and next class `empty()` method can be generalized to master class, in which in `empty()` method you iterate over attributes, but ...

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,202 @@
...
+        is_subtask_results_settled_empty = len(self.subtask_results_settled) == 0
```
rwrzesien: ... for querysets you should use `.count()` instead of `len()`. If you use `count()`, you 1. calculate the result on database side, which is faster. 2. You return only count from database instead of whole list of data, which is faster.

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,202 @@
...
+class Report(NamedTuple):
+    pair_reports: list
+
+
+def create_report() -> Report:
+    """ Create payments report for Concent. """
+    pair_reports = []
+
+    for subtask in get_subtasks_with_distinct_clients_pairs():  # pylint: disable=not-an-iterable
+        (subtasks_ids, clients_ethereum_addresses) = get_subtask_ids_and_ethereum_addresses_for_clients_pair(
```
rwrzesien: I think the code is too compact in this function, and it would be good to separate logical blocks by empty lines.

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,202 @@
...
+            subtask.requestor,
+            subtask.provider,
+        )
+        related_pending_responses_for_pair = get_related_pending_responses_for_subtasks_ids(subtasks_ids, subtask.requestor, clients_ethereum_addresses)
```
rwrzesien: This line could be split.

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,202 @@
...
+        try:
+            related_payments_for_pair = get_related_payments_for_pair(clients_ethereum_addresses)
+        except SCINotSynchronized:
+            related_payments_for_pair = Payments([], [], [], [], [])
+        is_related_payments_for_pair_empty = related_payments_for_pair.is_empty()  # pylint: disable=no-member
```
rwrzesien: I think those calls can be used directly in `if` statement.

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+def get_subtask_ids_and_ethereum_addresses_for_clients_pair(requestor: Client, provider: Client) -> Tuple:
+    """ Get Subtasks Ids and all Ethereum keys for Clients pair. """
+    subtask_ids = []
+    clients_ethereum_addresses = []
+
+    for subtask in Subtask.objects.filter(requestor=requestor, provider=provider):  # pylint: disable=no-member
+        subtask_ids.append(subtask.subtask_id)
+        task_to_compute = deserialize_message(subtask.task_to_compute.data.tobytes())
```
rwrzesien: What if GM version will not match here and deserialization will fail?

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,202 @@
...
+        clients_ethereum_addresses.append(
+            ClientEthereumAddresses(
+                task_to_compute.provider_ethereum_address,
+                task_to_compute.requestor_ethereum_address,
+            )
+        )
+
+    return subtask_ids, list(set(clients_ethereum_addresses))
```
rwrzesien: If there will be two different `ClientEthereumAddresses` objects but with the same keys will `set()` consider them a duplicate?

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+def get_related_pending_responses_for_subtasks_ids(subtasks_ids: List[str], requestor: Client, clients_ethereum_addresses: list) -> PendingResponses:
+    """ Group and return pending responses from Subtask queryset. """
+    return PendingResponses(
+        subtask_results_settled=PendingResponse.objects.filter(
+            response_type=PendingResponse.ResponseType.SubtaskResultsSettled.name,  # pylint: disable=no-member
+            subtask__subtask_id__in=subtasks_ids,
+        ).distinct('subtask'),
```
rwrzesien: Why do you need `distinct('subtask')` here?

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,202 @@
...
+        additional_verification_payments_for_provider=additional_verification_payments_for_provider,
+    )
+
+
+def get_oldest_subtask_creation_date() -> int:
+    """
+    Return creation date as timestamp of oldest Subtask stored in database.
+    If there are no Subtasks in database, return current timestamp.
+    """
+    oldest_subtask = Subtask.objects.order_by('created_at').first()
```
rwrzesien: I think that @dybi has some important remarks regarding this function in https://github.com/golemfactory/concent/pull/1073.

concent_api/core/payments/report.py
kbeker: Ok, as discussed I changed it to `count()` method where it is possible :)

concent_api/core/payments/report.py
kbeker: As we discussed, done :)

concent_api/core/payments/report.py
kbeker: Done

concent_api/core/payments/report.py
kbeker: Done

concent_api/core/payments/report.py
kbeker: Done. At first I did it this way and `MYPY` told me that this is wrong, so I tried again and now it is working. MIRACLE :D

concent_api/core/payments/report.py
kbeker: Good point, we should add filtering by protocol_version

rwrzesien: This is actually a little bit more complicated. We ignored this problem in the first version of report on purpose, and we might continue doing that here too, depending on @cameel decision.

The problem is that if you will ignore it, it might simply fail with the first ever run.
The problem is that if you will not ignore it, and add filtering by procol version, the report output can be misleading, because later payments could not be matched with responses, and the calculated values will be different than real.

I think the simplest solution is to create report only for time period since latest GM version is being used. But that is bigger architecture decision.

kbeker: I agree with you. Let's wait for @cameel and then I will change or leave it

concent_api/core/payments/report.py
kbeker: Yes, if addresses will be duplicated `set()` will do its job :)

concent_api/core/payments/report.py
kbeker: Because we don't want duplicated `PendingResponses` which are stored in database for requestor and provider. Thanks to this we got only one message

concent_api/core/payments/report.py
kbeker: I can't find any comment attached to this line of code. But we can wait for him :)

concent_api/core/payments/report.py
rwrzesien: According to https://docs.djangoproject.com/pl/2.1/ref/models/querysets/#distinct you should use `order_by('subtask')` too, but also change it to `subtask_id`.

But I am also worried about something else. If you for example have 4 PendingResponses of type `SubtaskResultsSettled` for single Subtask, then you would return only one of them. Is that ok?

concent_api/core/payments/report.py
rwrzesien: https://github.com/golemfactory/concent/pull/1073#discussion_r250523208

kbeker: I added that what @cameel and @dybi discussed about this function in previous PR.

concent_api/core/tests/test_payments_report.py
```diff
@@ -0,0 +1,226 @@
...
+                    )
+                ]
+            )
+            self.get_list_of_payments_mocks.append(
+                [self._create_forced_subtask_payment_object(self.default_payment_amount, subtask.subtask_id)
+                 for subtask in Subtask.objects.filter(requestor=subtask.requestor, provider=subtask.provider)]
+            )
+            self.get_covered_additional_verification_costs_mocks.append(
+                [self._create_cover_additional_verification_costs_object(self.default_payment_amount, subtask.subtask_id)
+                 for subtask in Subtask.objects.filter(requestor=subtask.requestor, provider=subtask.provider)]
```
dybi: this `for` loop is executed three times per iteration - list of related `subtask_ids` could be extracted to a variable and reused to produce:
`get_list_of_payments_mocks`,
`get_covered_additional_verification_costs_mocks`

concent_api/core/tests/test_payments_report.py
```diff
@@ -0,0 +1,226 @@
...
+            )
+            self.get_covered_additional_verification_costs_mocks.append(
```
dybi: why is it called second time with exactly the same params?

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,205 @@
...
+        forced_subtask_payments = service.get_list_of_payments(  # pylint: disable=no-value-for-parameter
+            requestor_eth_address=clients_pair.requestor,
+            provider_eth_address=clients_pair.provider,
+            min_block_timestamp=get_oldest_subtask_creation_date(),
+            transaction_type=TransactionType.FORCED_SUBTASK_PAYMENT,
+        )
+
+        additional_verification_payments_for_requestor = service.get_covered_additional_verification_costs(  # pylint: disable=no-value-for-parameter
+            client_eth_address=clients_pair.requestor,
+            payment_ts=get_oldest_subtask_creation_date(),
```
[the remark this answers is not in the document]
Jakub89: I see it's being used five times. Once for every payment.

concent_api/core/payments/report.py
rwrzesien: Then please change:
1. Function name.
2. Function comment.

Now the issue with different version of GM is also valid here.

concent_api/core/tests/test_payments_report.py
```diff
@@ -0,0 +1,226 @@
...
+        # The value below is determined by approach of mixing requestors and providers in _prepare_subtasks
+        self.number_of_pairs = self.number_of_clients_per_type * 2
+
+        self._prepare_providers()
+        self._prepare_requestors()
+        self._prepare_subtasks()
+        self._prepare_related_payments_mocks()
+        self._prepare_related_pending_responses()
+
+        assert self.default_payment_amount > 0
```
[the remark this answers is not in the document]
rwrzesien: There was some idea behind that in initial draft, but it is not valid anymore and those can be removed.

concent_api/core/tests/test_payments_report.py
```diff
@@ -0,0 +1,226 @@
...
+                requestor_public_key=base64.b64decode(task_to_compute.requestor_public_key),
+                state=Subtask.SubtaskState.ACCEPTED,
+                next_deadline=None,
+                task_to_compute=task_to_compute,
+                report_computed_task=report_computed_task,
+            )
+
+            self.subtasks.append(subtask)
+
+        assert Client.objects.count() == self.number_of_clients_per_type * 2
```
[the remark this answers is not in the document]
rwrzesien: It is sanity check that the right number of Clients is generated by code above.

concent_api/core/tests/test_payments_report.py
rwrzesien: Because we want two payments to exist and be mocked, because in report they are retrieved twice in a single run - once for provider and once for requestor.

concent_api/core/tests/test_payments_report.py
```diff
@@ -0,0 +1,226 @@
...
+            client.save()
+            self.requestors.append(client)
+
+    def _prepare_subtasks(self):
+        for i in range(self.number_of_subtasks):
+            task_to_compute = self._get_deserialized_task_to_compute(
+                provider_public_key=self.providers[i // 3].public_key,
+                requestor_public_key=(
+                    self.requestors[(i + 1) // 3].public_key
+                    if i != (self.number_of_subtasks - 1) else self.requestors[0].public_key
```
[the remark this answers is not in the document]
rwrzesien: We want to generate following list of Subtasks:
- Subtask 1 | Provider 1 | Requestor 1
- Subtask 2 | Provider 1 | Requestor 1
- Subtask 3 | Provider 1 | **Requestor 2**
- Subtask 4 | Provider 2 | Requestor 2
- Subtask 5 | Provider 2 | Requestor 3
- Subtask 6 | Provider 2 | **Requestor 3**
- Subtask 7 | Provider 3 | Requestor 3
- Subtask 8 | Provider 3 | Requestor 3
- Subtask 9 | Provider 3 | **Requestor 4**
.....
- Subtask 25 | Provider 9 | Requestor 9
- Subtask 26 | Provider 9 | Requestor 9
- Subtask 27 | Provider 9 | **Requestor 10**
- Subtask 28 | Provider 10 | Requestor 10
- Subtask 29 | Provider 10 | Requestor 10
- Subtask 30 | Provider 10 | **Requestor 11**

So this is the way to do this programmatically.

concent_api/core/payments/report.py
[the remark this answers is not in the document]
Jakub89: Done

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,205 @@
...
+        )
+        additional_verification_payments_for_provider = service.get_covered_additional_verification_costs(  # pylint: disable=no-value-for-parameter
```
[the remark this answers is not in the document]
Jakub89: Done

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,205 @@
...
+        )
+
+        forced_subtask_payments = service.get_list_of_payments(  # pylint: disable=no-value-for-parameter
+            requestor_eth_address=clients_pair.requestor,
+            provider_eth_address=clients_pair.provider,
+            min_block_timestamp=get_oldest_subtask_creation_date(),
+            transaction_type=TransactionType.FORCED_SUBTASK_PAYMENT,
+        )
+
+        additional_verification_payments_for_requestor = service.get_covered_additional_verification_costs(  # pylint: disable=no-value-for-parameter
```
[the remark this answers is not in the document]
Jakub89: Done

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,205 @@
...
+        additional_verification_payments_for_provider = service.get_covered_additional_verification_costs(  # pylint: disable=no-value-for-parameter
+            client_eth_address=clients_pair.provider,
+            payment_ts=get_oldest_subtask_creation_date(),
+        )
+
+    return Payments(
+        regular_payments=regular_payments,
+        settlement_payments=settlement_payments,
+        forced_subtask_payments=forced_subtask_payments,
+        additional_verification_payments_for_requestor=additional_verification_payments_for_requestor,
```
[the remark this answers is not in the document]
Jakub89: Done

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,205 @@
...
+        additional_verification_payments_for_provider=additional_verification_payments_for_provider,
```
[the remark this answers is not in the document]
Jakub89: Done

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,205 @@
...
+        is_force_payment_committed_empty = self.force_payment_committed.count() == 0
+
+        return is_subtask_results_settled_empty and is_subtask_results_rejected_empty and is_force_payment_committed_empty
+
+
+class Payments(NamedTuple):
+    regular_payments: list
+    settlement_payments: list
+    forced_subtask_payments: list
+    additional_verification_payments_for_requestor: list
```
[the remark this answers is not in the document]
Jakub89: Done

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,205 @@
...
+    additional_verification_payments_for_provider: list
```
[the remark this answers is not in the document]
Jakub89: Done

concent_api/core/tests/test_payments_report.py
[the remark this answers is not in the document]
Jakub89: Done, deleted all three.

concent_api/core/tests/test_payments_report.py
```diff
@@ -0,0 +1,226 @@
...
+            # 1 for each of 1 or 2 Subtasks.
+            self.assertIn(len(pair_report.payments.forced_subtask_payments), [1, 2])
+            # 1 for each of 1 or 2 Subtasks.
+            self.assertIn(len(pair_report.payments.additional_verification_payments_for_requestor), [1, 2])
+            # 1 for each of 1 or 2 Subtasks.
+            self.assertIn(len(pair_report.payments.additional_verification_payments_for_provider), [1, 2])
+
+        self.assertEqual(pairs_with_two_subtasks_count, self.number_of_pairs / 2)
+
+    def _prepare_providers(self):
```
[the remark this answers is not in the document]
Jakub89: Done

concent_api/core/payments/report.py
cameel: See https://github.com/golemfactory/concent/issues/1044#issuecomment-453200820

Add all the information you need to `Subtask` or create models corresponding to the messages.

Please also make sure that your changes are backwards compatible with the current database schema.

concent_api/core/tests/test_payments_report.py
Jakub89: I've just extract that database call to variable 
```
subtask_list = Subtask.objects.filter(requestor=subtask.requestor, provider=subtask.provider)
``` 
and use it in those three for loops. I don't know if this is what You mean.

concent_api/core/tests/test_payments_report.py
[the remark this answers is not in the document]
dybi: the solution could be simplified:
```python
    def _prepare_client(self, client_type):
        for _ in range(self.number_of_clients_per_type):
            client = Client(
                public_key=base64.b64encode(ECCx(None).raw_pubkey)
            )
            client.full_clean()
            client.save()
            getattr(self, client_type).append(client)

```
although it can stay as is.

concent_api/core/tests/test_payments_report.py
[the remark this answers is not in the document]
dybi: hm... how about:
```python
        for i in range(self.number_of_subtasks):
            requestor_index = ((i + 1) // 3) % self.number_of_clients_per_type 
            task_to_compute = self._get_deserialized_task_to_compute(
                provider_public_key=self.providers[i // 3].public_key,
                requestor_public_key=(
                    self.requestors[requestor_index].public_key
                ),
            )
```
? 
Wouldn't this be simpler/easier to understand? @Jakub89 @rwrzesien ?

concent_api/core/tests/test_payments_report.py
dybi: actually, you could make list of `subtask_ids`:
```python
subtask_ids = list(subtask.id for subtask in Subtask.objects.filter(requestor=subtask.requestor, provider=subtask.provider))
```
so you don't have to store whole subtask in the memory, just their IDs (you don't need more for filtering)

rwrzesien: https://docs.djangoproject.com/en/2.1/ref/models/querysets/#values-list

concent_api/core/tests/test_payments_report.py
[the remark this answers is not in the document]
rwrzesien: @dybi Works for me, but then maybe lets add both provider and request index getting above in separated lines.

concent_api/core/tests/test_payments_report.py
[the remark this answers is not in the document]
Jakub89: Done

concent_api/core/tests/test_payments_report.py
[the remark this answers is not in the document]
dybi: good idea

concent_api/core/tests/test_payments_report.py
dybi: oh, it seems I need to catch up with django ;)

concent_api/core/payments/report.py
Jakub89: @rwrzesien what other name and comment are You suggesting? There's nothing about that in [#1073(comment)](https://github.com/golemfactory/concent/pull/1073#discussion_r250523208).

concent_api/core/tests/test_payments_report.py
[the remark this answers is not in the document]
Jakub89: Done

concent_api/core/payments/report.py
rwrzesien: @kbeker @Jakub89 How was this case finally resolved? I see `.distinct('subtask')` is removed, so we are going to print all 1-4 messages?

concent_api/core/payments/report.py
rwrzesien: Let me ask this way - does function name reflects what the function is actually returning? The same for comment.

concent_api/core/payments/report.py
rwrzesien: As discussed, we want to show all extra messages and mark them as warning, which will be done in next pull request related to payment report.

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,206 @@
...
+
+def get_oldest_subtask_creation_date() -> int:
+    """
+    Return creation date as timestamp of oldest Subtask stored in database.
+    If there are no Subtasks in database, return current timestamp.
+    """
+    oldest_subtask = Subtask.objects.order_by('created_at').first()
+
+    if oldest_subtask is not None:
+        return deserialize_message(oldest_subtask.report_computed_task.data.tobytes()).timestamp
```
rwrzesien: We need a change here - it might happen that some old RCT on old Subtasks can't get deserialized, but from some point of time they can. I think there is no other way to determine that than iterating through Subtasks from oldest one and trying to deserialize. If there will be an exception, continue. If it will be deserialized correctly, break the loop and return the value.

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,214 @@
...
+    )
+
+
+def get_report_computed_task_timestamp_of_oldest_subtask() -> int:
+    """
+    Return ReportComputedTask timestamp of oldest Subtask stored in database.
+    If there are no Subtasks in database, return current timestamp.
+    """
+    subtasks_set = Subtask.objects.order_by('created_at')
+    if subtasks_set.exists():
```
rwrzesien: There is no need to check this, it will actually do one more database query, it is enough to iterate over subtasks, and remove `else` below.

concent_api/core/payments/report.py
```diff
@@ -0,0 +1,214 @@
...
+        subtasks_iterator = subtasks_set.iterator()
+        for subtask in subtasks_iterator:
```
rwrzesien: You can iterate directly over queryset, there is no need to cretea `iterator()`. Honestly I see this method for the first time :)

concent_api/core/payments/report.py
Jakub89: Done

concent_api/core/payments/report.py
Jakub89: Done

Issue #1069, opened 2019-01-16
Prepare django admin template for payment report


rwrzesien:
This is created as a part of https://github.com/golemfactory/concent/issues/1044

Once the backend part of the report is ready, it should be displayed in admin panel. As the data is not taken from queryset, I think it would be necessary to create also a separate admin view along with the html template.

The general guidelines were defined in related issue:
1. Use colors. Mark wrong amounts, wrong timestamps or unmatched messages in red.
2. Use tables. You could have a row for each message or payment. Message details would be in columns on the left and payment details in columns on the right. Common things like values and timestamps could be in the center.
3. If you have your own ideas that would make the report more concise and readable, use them but please drop a note in comment below the issue first.

Jakub89:
Just for the record about time it take, to generate whole report on `GETH_ADDRESS = 'https://rinkeby.golem.network:55555'`:
```
regular_payments in:  3.284846067428589
settlement_payments in:  2.7103161811828613
forced_subtask_payments in:  2.5918567180633545
additional_verification_payments_for_requestor in:  2.5945217609405518
additional_verification_payments_for_provider in:  2.641995668411255
Sum - related payments for subtask 1 in:  13.824137210845947

regular_payments in:  2.978281021118164
settlement_payments in:  2.505110502243042
forced_subtask_payments in:  2.698831796646118
additional_verification_payments_for_requestor in:  2.89158296585083
additional_verification_payments_for_provider in:  2.501680850982666
Sum - related payments for subtask 2 in:  13.57610297203064

total time:  27.51136088371277
16:55:32 INFO     | "GET /admin/payment_report/ HTTP/1.1" 200 2823
```
Every requestor-provider pair needs 5 sci calls, where each takes 2-4 second(on my potato laptop).
`total time` include also all the database calls in `create_report()` from both pairs , which takes around 0.1 second.

PR #1117, opened 2019-02-21
Feature payment report admin view and template part1

Jakub89:
First part of #1069

### DIFF
--- a/concent_api/core/admin.py
+++ b/concent_api/core/admin.py
@@ -1,11 +1,20 @@
-import datetime
 from decimal import Decimal
+from typing import Any
+from typing import Dict
+from typing import List
+
+import datetime
+
+from django.conf import settings
 from django.contrib import admin
+from django.db import transaction
 from django.db.models import Q
 from django.db.models import QuerySet
 from django.db.models import Sum
 from django.db.models.functions import Coalesce
 from django.http.request import HttpRequest
+from django.utils.decorators import method_decorator
+from django.views.generic import TemplateView
 from common.admin import ModelAdminReadOnlyMixin
 from common.helpers import get_current_utc_timestamp
 from common.helpers import parse_timestamp_to_utc_datetime
@@ -16,9 +25,46 @@ from .models import PendingResponse
 from .models import StoredMessage
 from .models import Subtask
 from .models import SubtaskWithTimingColumnsManager
+from .payments.report import create_report
 
 ACTIVE_STATE_NAMES = [x.name for x in Subtask.ACTIVE_STATES]
 PASSIVE_STATE_NAMES = [x.name for x in Subtask.PASSIVE_STATES]
+
+
+@method_decorator(transaction.non_atomic_requests(using='control'), name='dispatch')
+class PaymentReportAdminView(TemplateView):
+
+    template_name = 'admin/payment_report.html'
+
+    def get_context_data(self, **kwargs: Any) -> Dict:
+        context_data = super().get_context_data(**kwargs)
+        context_data['payment_report'] = create_report()
+        context_data['eth_network'] = self._get_eth_network()
+        return context_data
+
+    @staticmethod
+    def _get_eth_network() -> str:
+        if settings.ETHEREUM_CHAIN == 'rinkeby':
+            return 'rinkeby.'
+        else:
+            return ''
+
+
+old_get_urls = admin.site.get_urls
+
+
+def get_urls() -> List:
+    from django.conf.urls import url
+
+    urls = old_get_urls()
+    urls += [
+        url(r'^payment_report/$', admin.site.admin_view(PaymentReportAdminView.as_view()), name='payment_report')
+    ]
+    return urls
+
+
+# This is a way to add custom non-model related view to Django admin.
+admin.site.get_urls = get_urls
 
 
 class ActivePassiveDownloadsStateFilter(admin.SimpleListFilter):
--- a/concent_api/core/templates/admin/index.html
+++ b/concent_api/core/templates/admin/index.html
@@ -1,0 +1,25 @@
+{% extends "admin/index.html" %}
+{% block content %}
+{{ block.super }}
+{% endblock %}
+
+{% block sidebar %}
+{{ block.super }}
+<style>
+a.button {
+    background-color: #79aec8;
+    border-color: #247c94;
+    color: white;
+    padding: 10px 26px;
+    text-align: center;
+    text-decoration: none;
+    display: inline-block;
+    font-size: 18px;
+    margin: 4px 2px;
+    cursor: pointer;
+}
+</style>
+<div>
+    <a href="{% url 'admin:payment_report' %}" class="button">$ Generate Payment Report $</a>
+</div>
+{% endblock %}
--- a/concent_api/core/templates/admin/payment_report.html
+++ b/concent_api/core/templates/admin/payment_report.html
@@ -51,36 +51,25 @@ br {
             <th rowspan="3">PendingResponses</th>
             <th>SubtaskResultsSettled</th>
             <td>
-            {% for subtask_results_settled in pair_report.pending_responses.subtask_results_settled %}
-                <p><a href="{% url 'admin:core_pendingresponse_change' subtask_results_settled.id %}">
-                - subtask_id: {{ subtask_results_settled.subtask.subtask_id }},
-                delivered: {{ subtask_results_settled.delivered }},
-                created_at: {{ subtask_results_settled.created_at }}</a>
-                </p>
-            {% endfor %}
+                {% with pending_responses=pair_report.pending_responses.subtask_results_settled %}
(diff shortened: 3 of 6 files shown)


concent_api/core/admin.py
```diff
@@ -16,9 +27,53 @@
...
 from .models import Subtask
 from .models import SubtaskWithTimingColumnsManager
+from .payments.report import create_report
 
 ACTIVE_STATE_NAMES = [x.name for x in Subtask.ACTIVE_STATES]
 PASSIVE_STATE_NAMES = [x.name for x in Subtask.PASSIVE_STATES]
+DECORATORS = [transaction.non_atomic_requests(using='control')]
+
+
+@method_decorator(DECORATORS, name='dispatch')
```
rwrzesien: You can use decorator directly here, there is no need for `DECORATORS`.

concent_api/core/admin.py
```diff
@@ -16,9 +27,53 @@
...
+
+    @staticmethod
+    def _get_eth_network() -> str:
+        session = requests.Session()
+        data = json.dumps({'id': 67, 'jsonrpc': '2.0', 'method': 'net_version', 'params': []})
+        headers = {'Content-type': 'application/json'}
+        try:
+            response = session.post(settings.GETH_ADDRESS, data=data, headers=headers)  # type: ignore
+        except ConnectionError:
+            return ''
```
rwrzesien: Maybe return `couldn't connect` or something like this?

concent_api/core/admin.py
```diff
@@ -16,9 +27,53 @@
...
+        context_data = super().get_context_data(**kwargs)
+        context_data['payment_report'] = create_report()
+        context_data['eth_network'] = self._get_eth_network()
+        return context_data
+
+    @staticmethod
+    def _get_eth_network() -> str:
+        session = requests.Session()
+        data = json.dumps({'id': 67, 'jsonrpc': '2.0', 'method': 'net_version', 'params': []})
+        headers = {'Content-type': 'application/json'}
```
rwrzesien: Those two lines above are good candidates for a constants.

concent_api/core/admin.py
```diff
@@ -16,9 +27,53 @@
...
+        try:
+            response = session.post(settings.GETH_ADDRESS, data=data, headers=headers)  # type: ignore
```
rwrzesien: I am not sure if there is a need to do this via `Session` object, I think you can call `requests.post` directly here.

concent_api/core/templates/admin/index.html
```diff
@@ -0,0 +1,25 @@
...
+    padding: 10px 26px;
+    text-align: center;
+    text-decoration: none;
+    display: inline-block;
+    font-size: 18px;
+    margin: 4px 2px;
+    cursor: pointer;
+}
+</style>
+    <div>
```
rwrzesien: I think there is no need for indentation here.

concent_api/core/templates/admin/payment_report.html
```diff
@@ -0,0 +1,155 @@
...
+table, th, td {
+  border: 1px solid black;
+  border-collapse: collapse;
+}
+th, td {
+  padding: 5px;
+  text-align: left;
+}
+br {
+    padding: 5px;
```
rwrzesien: Please unify indentation, either 4 spaces or 2.

concent_api/core/templates/admin/payment_report.html
```diff
@@ -0,0 +1,155 @@
...
+}
+</style>
+
+    {% for pair_report in payment_report.pair_reports %}
+    <table style="width:100%">
```
rwrzesien: This should be done via CSS class too.

concent_api/core/templates/admin/payment_report.html
```diff
@@ -0,0 +1,155 @@
...
+            </td>
+        </tr>
+
+        {#  Pending Responses table  #}
+        <tr>
+            <th rowspan="3">PendingResponses</th>
+            <th>SubtaskResultsSettled</th>
+            <td>
+            {% for subtask_results_settled in pair_report.pending_responses.subtask_results_settled %}
+                <p><a href="{% url 'admin:core_pendingresponse_change' subtask_results_settled.id %}">
```
rwrzesien: You should create partial template for this and include it three times.

concent_api/core/templates/admin/payment_report.html
```diff
@@ -0,0 +1,155 @@
...
+        </tr>
+
+        {#  Payments table  #}
+        <tr>
+            <th rowspan="5">Payments</th>
+            <th>Regular payments</th>
+            <td>
+            {% for payment in pair_report.payments.regular_payments %}
+                <p>
+                    <a href="https://{{ eth_network }}etherscan.io/address/{{ payment.provider }}">- provider: {{ payment.provider }}</a>,
```
rwrzesien: Same here, partial template and include.

concent_api/core/templates/admin/payment_report.html
```diff
@@ -0,0 +1,155 @@
...
+                </p>
+            {% endfor %}
+            </td>
+        </tr>
+        <tr>
+            <th>additional_verification_payments_for_requestor</th>
+            <td>
+            {% for payment in pair_report.payments.additional_verification_payments_for_requestor %}
+                <p>
+                - subtask_id: {{ payment.subtask_id }},
```
rwrzesien: Same here, partial template and include.

concent_api/core/admin.py
Jakub89: I don't know if we want to crush concent just because GETH from settings is not working at that moment. Do we want to render payment report without payments? Or we could render custom error 5xx page, that shows exceptions?

concent_api/core/admin.py
Jakub89: I'll try

concent_api/core/admin.py
rwrzesien: I am not talking about crashing, I am talking about returning and showing a string with info that Ethereum network could not be retrieved.

PaweuB:
Has anyone got holy water?

concent_api/core/templates/admin/payment_report.html
Jakub89: I'm not sure how that suppose to look. Could You give me an example?

rwrzesien: https://docs.djangoproject.com/en/2.1/ref/templates/builtins/#include

rwrzesien: @Jakub89 You simply extract repeating part to separate html file, and use `include` tag to attach it.

concent_api/core/admin.py
Jakub89: As we spoke, this string is a part of link in template, so it need to stay like that. Or we should change handling that exception completely, @cameel ?

cameel: We do not need to contact Geth to get this information. We should already know if we're using mainnet or testnet. We need this information to initialize SCI anyway.

BTW, I have noticed that we always tell SCI to use Rinkeby, no matter whether we're using mainnet or testnet. That's a bug: #1120.

Jakub89: I didn't find any other way. There is no setting that would define that right now. `GETH_ADDRESS` is usually just some ip address. It's going to be much easier once #1120 is done.

concent_api/core/admin.py
Jakub89: Done

concent_api/core/admin.py
Jakub89: Done

concent_api/core/admin.py
Jakub89: works. Done

concent_api/core/templates/admin/index.html
Jakub89: Done

concent_api/core/templates/admin/payment_report.html
Jakub89: Done

concent_api/core/templates/admin/payment_report.html
Jakub89: Done

concent_api/core/templates/admin/payment_report.html
Jakub89: Done

concent_api/core/templates/admin/payment_report.html
Jakub89: Done

concent_api/core/templates/admin/payment_report.html
Jakub89: Done

concent_api/core/admin.py
cameel: You can just hard-code it to Rinkeby and add a note in #1120 to use the setting for both. Right now Concent is hard-coded for Rinkeby anyway.

PR #1129, opened 2019-03-25
Chore add script for downloading render tools for development

### LINKED ISSUE #1127: Downloading parts of blender_verifier image for use in development
In https://github.com/golemfactory/concent-deployment/pull/361 we have switched to Golem's new blender image (`blender_verifier:1.1`). The new image now accepts invidividual task parameters and renders a helper script for blender from a template instead of expecting to get the script from the outside (until now it was being sent in `TaskToCompute`). This is safer and allows Concent to start Blender the same way Golem does.

Since we're not running verified inside the blender image in development, we need to fetch those scripts from the image or from Golem's repository.
1. Create a bash script that accepts Golem version (tag/branch/commit) as a parameter and downloads files directly from `golem` repository. We need the following files:
    - `golem/apps/blender/resources/images/entrypoints/scripts/render_tools/scenefileeditor.py`
    - `golem/apps/blender/resources/images/entrypoints/scripts/render_tools/templates/blendercrop.py.template`
    - The scripts should be placed somewhere where they will not conflict with Concent's files and will be importable by verfier. It's probably best to put them in a subdirectory in verifier app.
    - Add the files to `.gitignore`.
    - Make the version parameter optional. Set it to `develop` by default.
2. `README` should say to run the script when configuring one's development environment.
    - Suggest to use tag with the latest version as parameter.
    - Say what is the default value for the parameter.
    - Say that you need to rerun the script regularly to update the files, especially when a new Golem version is released.
3. This should be merged together with https://github.com/golemfactory/concent-deployment/issues/374. Otherwise our integration tests will stop working with Jenkins.

kbeker:
Resolves #1127

Must be merged before #1128

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -118,3 +118,6 @@ ENV/
 
 # Signing service daily tresholds
 daily_thresholds/
+
+# Render tools downloaded from golem repository
+concent_api/verifier/golem_render_tools/
--- a/concent_api/concent_api/settings/base.py
+++ b/concent_api/concent_api/settings/base.py
@@ -370,3 +370,6 @@ SCI_CALLBACK_RETRIES_TIME = 60
 
 # Define chain where PaymentInterface has to connect to geth
 # ETHEREUM_CHAIN =
+
+# Define dir where render_tools from golem repository should be stored
+# BLENDER_RENDER_TOOLS_DIR = ''
--- a/concent_api/verifier/utils.py
+++ b/concent_api/verifier/utils.py
@@ -1,5 +1,3 @@
-import sys
-
 import hashlib
 from base64 import b64encode
 from typing import Dict
@@ -36,7 +34,6 @@ from verifier.exceptions import VerificationMismatch
 from .constants import UNPACK_CHUNK_SIZE
 
 
-sys.path.append('/golem/entrypoints/scripts/render_tools')
 logger = logging.getLogger(__name__)
 crash_logger = logging.getLogger('concent.crash')
 FramesToParsedFilePaths = Dict[int, List[str]]
--- a/download-render-tools-from-golem-repository.sh
+++ b/download-render-tools-from-golem-repository.sh
@@ -1,0 +1,53 @@
+#!/bin/bash -e
+
+default_download_path="concent_api/verifier/golem_render_tools"
+default_version="develop"
+
+
+function display_usage() {
+    printf "Usage:\n"
+    printf " download_render_tools_from_golem_repository
+           [-r || --repository-tag=<name of tag or branch from where files should be downloaded (default: ${default_version})>]
+           [-p || --full-path-to-store-files=<full path where downloaded files should be stored (default: ${default_download_path}>])\n"
+}
+
+
+organization_name="golemfactory"
+repository_name="golem"
+version=${default_version}
+path_to_store_files=${default_download_path}
+files_to_download=(
+    "scenefileeditor.py"
+    "templates/blendercrop.py.template"
+)
+
+download_url="https://raw.githubusercontent.com/${organization_name}/${repository_name}/${version}/apps/blender/resources/images/entrypoints/scripts/render_tools"
+
+
+for argument in "$@"
+do
+    if [[ $argument == "--help" ||  $argument == "-h" ]]; then
+        display_usage
+        exit 0
+    fi
+    case $argument in
+        -p=*|--full-path-to-store-files=*)
+        path_to_store_files="${argument#*=}"
+        ;;
+        -r=*|--repository-tag=*)
+        version="${argument#*=}"
+    esac
+done
+
+
+if [ ! -d "${path_to_store_files}" ]
+then
+    mkdir -p "${path_to_store_files}/templates"
+    echo "Created new directory in ${path_to_store_files}"
+fi
+
+for file in ${files_to_download[*]}
+do
+    curl "${download_url}/${file}" --output "${path_to_store_files}/${file}"
+    echo "Downloaded ${file} and stored in ${path_to_store_files}/${file}"
+done
--- a/install-development-requirements.sh
+++ b/install-development-requirements.sh
@@ -23,3 +23,7 @@ cd ..
 printf "=================== INSTALL REQUIREMENTS DEVELOPMENT ====================\n"
 pip install -r requirements-development.txt
 printf "\n"
+
+printf "=================== DOWNLOAD RENDER_TOOLS FROM GOLEM REPOSITORY ====================\n"
+./download-render-tools-from-golem-repository.sh
+printf "\n"
--- a/concent_api/concent_api/settings/development.py
+++ b/concent_api/concent_api/settings/development.py
@@ -1,5 +1,4 @@
 # pylint: disable=unused-wildcard-import
-import os
 
 from middleman.constants import DEFAULT_INTERNAL_PORT
 from .base import *  # NOQA  # pylint: disable=wildcard-import
@@ -17,8 +16,6 @@ DATABASES['storage']['PASSWORD'] = ''
 
 CONCENT_PUBLIC_KEY  = b'\xf3\x97\x19\xcdX\xda\x86tiP\x1c&\xd39M\x9e\xa4\xddb\x89\xb5,]O\xd5cR\x84\xb85\xed\xc9\xa17e,\xb2s\xeb\n1\xcaN.l\xba\xc3\xb7\xc2\xba\xff\xabN\xde\xb3\x0b\xa6l\xbf6o\x81\xe0;'
 CONCENT_PRIVATE_KEY = b'l\xcdh\x19\xeb$>\xbcG\xa1\xc7v\xe8\xd7o\x0c\xbf\x0e\x0fM\x89lw\x1e\xd7K\xd6Hnv$\xa2'
-
-SIGNING_SERVICE_PUBLIC_KEY = b'\xc0$\xa3\xab\x8d\x04[q\x0b\x1b\x00#\x8b\x18\xd9\xcf\x95\xf1\\`\x08w@\xf0\x9d\xf6B7;\xe56\xc8\xb1pp\x97\xac\xb3\xd3\xc1\xb0\x96a6\xcat\x8b\xea\xbb\x98\x04\xcfD28\x97$\xa9\xb9b8\x17\xe1\xfc'
 
 CONCENT_FEATURES = [
     "admin-panel",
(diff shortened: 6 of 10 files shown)


download_render_tools_from_golem_repository.sh
```diff
@@ -0,0 +1,47 @@
...
+REPOSITORY_TAG="develop"
+PATH_TO_STORE_FILES="/home/$USER/golem_render_tools"
+FILES_TO_DOWNLOAD=("scenefileeditor.py" "templates/blendercrop.py.template")
+
+DOWNLOAD_URL="https://raw.githubusercontent.com/${ORGANIZATION_NAME}/${REPOSITORY_NAME}/${REPOSITORY_TAG}/apps/blender/resources/images/entrypoints/scripts/render_tools"
+
+
+for argument in "$@"
+do
+    if [[ ( $argument == "--help") ||  $argument == "-h" ]]; then
```
dybi: for me it is a bit unusual to see `if` and `switch` mixed like that

download_render_tools_from_golem_repository.sh
```diff
@@ -0,0 +1,47 @@
...
+        display_usage
+        exit 0
+    fi
+    case $argument in
+        -p=*|--full_path_to_store_files=*)
+        PATH_TO_STORE_FILES="${argument#*=}"
+        ;;
+        -r=*|--repository_tag=*)
```
dybi: shouldn't this be a separate `case` entry?

install-development-requirements.sh
```diff
@@ -23,3 +23,7 @@ cd ..
 printf "=================== INSTALL REQUIREMENTS DEVELOPMENT ====================\n"
 pip install -r requirements-development.txt
 printf "\n"
+
+printf "=================== DOWNLOAD RENDER_TOOLS FROM GOLEM REPOSITORY ====================\n"
+./download_render_tools_from_golem_repo.sh
```
dybi: :heart:

download_render_tools_from_golem_repository.sh
kbeker: I did it exactly the same as we have it in `run_tests.sh`: https://github.com/golemfactory/concent/blob/master/run_tests.sh#L18

download_render_tools_from_golem_repository.sh
kbeker: I did it exactly the same as we have it in `run_tests.sh`: https://github.com/golemfactory/concent/blob/master/run_tests.sh#L31

download_render_tools_from_golem_repository.sh
```diff
@@ -0,0 +1,47 @@
...
+    printf " download_render_tools_from_golem_repository
+           [-r || --repository_tag=<name of tag or branch from where files should be downloaded (default: develop)>]
+           [-p || --full_path_to_store_files=<full path where downloaded files should be stored (default: /home/$USER/golem_render_tools>])\n"
+}
+
+
+ORGANIZATION_NAME="golemfactory"
+REPOSITORY_NAME="golem"
+REPOSITORY_TAG="develop"
+PATH_TO_STORE_FILES="/home/$USER/golem_render_tools"
```
cameel: The usual convention in Bash is to user lowercase names for your own variables local to the script. Only the system and configuration variables defined in the environment (like `PATH` or `PWD`) should be capitalized.

cameel: Why is this an absolute path? I think it would be better to assume that the script is being run from a working copy of Concent and put it in a location relative to that.

download_render_tools_from_golem_repository.sh
```diff
@@ -0,0 +1,47 @@
...
+FILES_TO_DOWNLOAD=("scenefileeditor.py" "templates/blendercrop.py.template")
```
cameel: You can break lines in bash lists like this:
``` bash
FILES_TO_DOWNLOAD=(
    "scenefileeditor.py"
    "templates/blendercrop.py.template"
)
```

download_render_tools_from_golem_repository.sh
```diff
@@ -0,0 +1,47 @@
...
+    printf "Usage:\n"
+    printf " download_render_tools_from_golem_repository
+           [-r || --repository_tag=<name of tag or branch from where files should be downloaded (default: develop)>]
+           [-p || --full_path_to_store_files=<full path where downloaded files should be stored (default: /home/$USER/golem_render_tools>])\n"
+}
+
+
+ORGANIZATION_NAME="golemfactory"
+REPOSITORY_NAME="golem"
+REPOSITORY_TAG="develop"
```
cameel: It does not have to be a tag so the name is a bit misleading. It can also be a commit or branch. We're usually referring to this as "version" in deployment.

cameel: Since `REPOSITORY_TAG` is just the default, I think you should have "default" in its name to make it clear.

download_render_tools_from_golem_repository.sh
cameel: The parenthesis is not necessary here.

download_render_tools_from_golem_repository.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash -e
+
+
+display_usage(){
```
cameel: To be consistent with our conventions in deployment please use the `function` keyword when defining a function and leave a space before the `{`.

download_render_tools_from_golem_repository.sh
```diff
@@ -0,0 +1,47 @@
...
+
+
+for argument in "$@"
+do
+    if [[ ( $argument == "--help") ||  $argument == "-h" ]]; then
+        display_usage
+        exit 0
+    fi
+    case $argument in
+        -p=*|--full_path_to_store_files=*)
```
cameel: Please use dashes in option names: `--full-path-to-store-files`.

concent_api/verifier/utils.py
```diff
@@ -36,7 +36,7 @@
 from .constants import UNPACK_CHUNK_SIZE
 
 
-sys.path.append('/golem/entrypoints/scripts/render_tools')
+sys.path.append(os.path.join(settings.BLENDER_RENDER_TOOLS_DIR, '/golem/entrypoints/scripts/render_tools'))
```
cameel: I don't think utils is a good location for this. If I was looking for something like this I would not think to look here. Why not put it in some more obvious location, e.g. in the file with verifier's views or even somewhere in `concent_api/concent_api/`? Actually the `manage.py` file might be the best location though I'm not sure if you'll be able to import config there without any annoying side-effects.

concent_api/concent_api/settings/base.py
```diff
@@ -372,4 +372,4 @@
 # ETHEREUM_CHAIN =
 
 # Define dir where render_tools from golem repository should be stored
-BLENDER_RENDER_TOOLS_DIR = ''
+# BLENDER_RENDER_TOOLS_DIR = ''
```
cameel: I think this should have a default in `base.py`. It should point at the directory where these files would if we were running in our Docker container.

install-development-requirements.sh
```diff
@@ -25,5 +25,5 @@ pip install -r requirements-development.txt
 printf "\n"
 
 printf "=================== DOWNLOAD RENDER_TOOLS FROM GOLEM REPOSITORY ====================\n"
-./download_render_tools_from_golem_repo.sh
+./download_render_tools_from_golem_repository.sh
```
cameel: Please use dashes in Bash script names: `download-render-tools-from-golem-repository.sh`.

kbeker: Done

concent_api/concent_api/settings/base.py
kbeker: Done

concent_api/verifier/utils.py
kbeker: Ok, moved to `manage.py`.

download_render_tools_from_golem_repository.sh
kbeker: Done

download_render_tools_from_golem_repository.sh
kbeker: Done

download_render_tools_from_golem_repository.sh
kbeker: Done

download_render_tools_from_golem_repository.sh
kbeker: Done

kbeker: Done, changed to `version`

download_render_tools_from_golem_repository.sh
kbeker: Done

download_render_tools_from_golem_repository.sh
kbeker: Changed to `concent_api/verifier/golem_render_tools`

kbeker: Changed to local variables to lowercase names

concent_api/verifier/utils.py
```diff
@@ -36,7 +36,7 @@
 from .constants import UNPACK_CHUNK_SIZE
 
 
-sys.path.append('/golem/entrypoints/scripts/render_tools')
+
```
cameel: Why is this change visible as removal (instead of just disappearing? Is this already in master?

download-render-tools-from-golem-repository.sh
```diff
@@ -0,0 +1,53 @@
...
+organization_name="golemfactory"
+repository_name="golem"
+version=${default_version}
+path_to_store_files=${default_download_path}
+files_to_download=(
+    "scenefileeditor.py"
+    "templates/blendercrop.py.template"
+)
+
+DOWNLOAD_URL="https://raw.githubusercontent.com/${organization_name}/${repository_name}/${version}/apps/blender/resources/images/entrypoints/scripts/render_tools"
```
cameel: This should be lower case too.

.gitignore
```diff
@@ -118,3 +118,6 @@ ENV/
 
 # Signing service daily tresholds
 daily_thresholds/
+
+# Render tools downloaded from golem repository
+/concent_api/verifier/golem_render_tools
```
cameel: Please add a trailing slash so that it's clear it's a directory.

concent_api/verifier/utils.py
kbeker: This change is visible because it was added in previous PR. In this PR I am changing it to work with custom directory

download-render-tools-from-golem-repository.sh
kbeker: Ups, I missed it. Done

.gitignore
kbeker: Done

kbeker:
Also, @cameel I changed where path is added. I decided that better place than `manage.py` and `wsgi.py` would be `__init__.pl` file in `verifier` directory. I have also added setting `CUSTOM_BLENDER_RENDER_TOOLS` which in production default is False.

PR #1138, opened 2019-05-21
Revert payment report part 1

Jakub89:
Payment report will be moved to separate branch since we will not develop/support it right now.


Jakub89:
Lint/tests all clear.

PR #1158, opened 2020-01-15
Bugfix float to string for Decimal input in blender_crop_script

Jakub89:
Resolves: https://talkback.golem.network/sentry/concent/issues/72830/
https://talkback.golem.network/sentry/concent/issues/72829

### DIFF
--- a/concent_api/conductor/service.py
+++ b/concent_api/conductor/service.py
@@ -120,15 +120,21 @@ def filter_frames_by_blender_subtask_definition(blender_subtask_definition: Blen
 
 
 def _store_blender_crop_script_parameters(blender_parameters: Any) -> BlenderCropScriptParameters:
+    """
+    Create and save BlenderCropScriptParameters model in database.
+    String conversion and splicing is done to avoid DecimalField exceptions on full_clean.
+    [:11] is needed for DecimalField(max_digits=10, decimal_places=9).
+    """
+
     blender_crop_script_parameters = BlenderCropScriptParameters(
         resolution_x=blender_parameters['resolution'][0],
         resolution_y=blender_parameters['resolution'][1],
         samples=blender_parameters['samples'],
         use_compositing=blender_parameters['use_compositing'],
-        borders_x_min=blender_parameters['borders_x'][0],
-        borders_x_max=blender_parameters['borders_x'][1],
-        borders_y_min=blender_parameters['borders_y'][0],
-        borders_y_max=blender_parameters['borders_y'][1],
+        borders_x_min=(str(blender_parameters['borders_x'][0]))[:11],
+        borders_x_max=(str(blender_parameters['borders_x'][1]))[:11],
+        borders_y_min=(str(blender_parameters['borders_y'][0]))[:11],
+        borders_y_max=(str(blender_parameters['borders_y'][1]))[:11],
     )
     blender_crop_script_parameters.full_clean()
     blender_crop_script_parameters.save()
--- a/concent_api/conductor/tests/test_service.py
+++ b/concent_api/conductor/tests/test_service.py
@@ -1,12 +1,18 @@
 import mock
+import pytest
+from assertpy import assert_that
+
+from golem_messages.factories.tasks import ComputeTaskDefFactory
 
 from common.helpers import get_current_utc_timestamp
 from common.helpers import parse_timestamp_to_utc_datetime
 from conductor.exceptions import VerificationRequestAlreadyInitiatedError
+from conductor.models import BlenderCropScriptParameters
 from conductor.models import ResultTransferRequest
 from conductor.models import UploadReport
 from conductor.models import VerificationRequest
 from conductor.service import update_upload_report
+from conductor.service import _store_blender_crop_script_parameters
 from core.tests.utils import ConcentIntegrationTestCase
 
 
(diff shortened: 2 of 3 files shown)


concent_api/conductor/service.py
```diff
@@ -125,10 +125,10 @@ def _store_blender_crop_script_parameters(blender_parameters: Any) -> BlenderCro
...
         samples=blender_parameters['samples'],
         use_compositing=blender_parameters['use_compositing'],
-        borders_x_min=blender_parameters['borders_x'][0],
-        borders_x_max=blender_parameters['borders_x'][1],
-        borders_y_min=blender_parameters['borders_y'][0],
-        borders_y_max=blender_parameters['borders_y'][1],
+        borders_x_min=str(blender_parameters['borders_x'][0]),
+        borders_x_max=str(blender_parameters['borders_x'][1]),
+        borders_y_min=str(blender_parameters['borders_y'][0]),
+        borders_y_max=str(blender_parameters['borders_y'][1]),
```
cameel: When fixing an error like this you should add a test for the failing case.  So `_store_blender_crop_script_parameters()` with the bad float values.

Also, a comment about why it's being converted to `str()` would be useful. It's not obvious.

concent_api/conductor/service.py
```diff
@@ -125,10 +125,10 @@ def _store_blender_crop_script_parameters(blender_parameters: Any) -> BlenderCro
         resolution_y=blender_parameters['resolution'][1],
         samples=blender_parameters['samples'],
         use_compositing=blender_parameters['use_compositing'],
-        borders_x_min=blender_parameters['borders_x'][0],
-        borders_x_max=blender_parameters['borders_x'][1],
-        borders_y_min=blender_parameters['borders_y'][0],
-        borders_y_max=blender_parameters['borders_y'][1],
+        borders_x_min=str(blender_parameters['borders_x'][0]),
```
cameel: What happens when `blender_parameters['borders_x'][0]` is `None`?

Jakub89: It won't get to that point, we have validations for types in `message_handlers.py`. But:
```
../service.py:133: in _store_blender_crop_script_parameters
    blender_crop_script_parameters.full_clean()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <BlenderCropScriptParameters: BlenderCropScriptParameters object>
exclude = ['borders_y_max'], validate_unique = True

    def full_clean(self, exclude=None, validate_unique=True):
        """
            Calls clean_fields, clean, and validate_unique, on the model,
            and raises a ``ValidationError`` for any errors that occurred.
            """
        errors = {}
        if exclude is None:
            exclude = []
        else:
            exclude = list(exclude)
    
        try:
            self.clean_fields(exclude=exclude)
        except ValidationError as e:
            errors = e.update_error_dict(errors)
    
        # Form.clean() is run even if other validation fails, so do the
        # same with Model.clean() for consistency.
        try:
            self.clean()
        except ValidationError as e:
            errors = e.update_error_dict(errors)
    
        # Run unique checks, but only for fields that passed validation.
        if validate_unique:
            for name in errors.keys():
                if name != NON_FIELD_ERRORS and name not in exclude:
                    exclude.append(name)
            try:
                self.validate_unique(exclude=exclude)
            except ValidationError as e:
                errors = e.update_error_dict(errors)
    
        if errors:
>           raise ValidationError(errors)
E           django.core.exceptions.ValidationError: {'borders_y_max': ["'None' value must be a decimal number."]}
```

concent_api/conductor/service.py
Jakub89: Yup, writing it right now. I've pushed it so @bartoszbetka could deploy it faster.

Jakub89: Done

concent_api/conductor/tests/test_service.py
```diff
@@ -1,12 +1,17 @@
...
+import pytest
+from assertpy import assert_that
+
+from golem_messages.factories.tasks import ComputeTaskDefFactory
 
 from common.helpers import get_current_utc_timestamp
 from common.helpers import parse_timestamp_to_utc_datetime
 from conductor.exceptions import VerificationRequestAlreadyInitiatedError
-from conductor.models import ResultTransferRequest
+from conductor.models import ResultTransferRequest, BlenderCropScriptParameters
```
kbeker: You should move it to new line but this is only style issue.

concent_api/conductor/service.py
```diff
@@ -120,15 +120,20 @@ def filter_frames_by_blender_subtask_definition(blender_subtask_definition: Blen
...
     blender_crop_script_parameters = BlenderCropScriptParameters(
         resolution_x=blender_parameters['resolution'][0],
         resolution_y=blender_parameters['resolution'][1],
         samples=blender_parameters['samples'],
         use_compositing=blender_parameters['use_compositing'],
-        borders_x_min=blender_parameters['borders_x'][0],
-        borders_x_max=blender_parameters['borders_x'][1],
-        borders_y_min=blender_parameters['borders_y'][0],
-        borders_y_max=blender_parameters['borders_y'][1],
+        borders_x_min=(str(blender_parameters['borders_x'][0]))[:11],
```
kbeker: I would also add note why slice is done exactly to 11th char because it might be not understandable in future.

concent_api/conductor/tests/test_service.py
Jakub89: Done

concent_api/conductor/service.py
Jakub89: Done

Issue #1161, opened 2020-01-28
ValidationError {'SubtaskResultsRejected': ['Nested ReportComputedTask message must be the same as ReportComputedTask stored separately']}


Jakub89:
https://talkback.golem.network/sentry/concent/issues/72930/

We were able to reproduce this bug on `dev` cluster, by running test where report_computed_task was different between messages.
PR: #1159 

This means that bug is somewhere on golem application side, but we could handle it differently. We are waiting for golem team to reply.

PR #1164, opened 2020-02-03
Bugfix remove nested rct messages validations

Jakub89:
Resolves: #1161


Jakub89:
@kbeker there are tests for this `test_unit_model.py`.

